PR #972, opened 2015-07-27
Basic support for Sensor Kid

sndpl:
Basic support for Sensor Kid (http://www.generation-msx.nl/software/sony/sensor-kid/release/1149/)
- It generates 3 random waves for sound, light and temperature
- When enable/disable port x in basic, it will mention it in a on screen message

Thanks to Wouter and Manuel for all the help!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@
 *~
 core.*
 core
+.cproject
+.project
--- a/share/extensions/SensorKid.xml
+++ b/share/extensions/SensorKid.xml
@@ -1,0 +1,29 @@
+<?xml version="1.0"?>
+<!DOCTYPE msxconfig SYSTEM "msxconfig2.dtd">
+<msxconfig>
+  <info>
+    <name>Sensor Kid Cartridge</name>
+    <manufacturer>Sony</manufacturer>
+    <code>HBJ-H081C</code>
+    <release_year>1988</release_year>
+    <description>Sony Sensor Kid cartridge</description>
+    <type>sensor cartridge</type>
+  </info>
+  <devices>
+    <SensorKid id="Sensor Kid">
+      <io base="0x00" num="2"/>
+    </SensorKid>
+  
+    <primary slot="any">
+      <secondary slot="any">
+        <ROM id="Sensor Kid ROM">
+          <rom>
+              <sha1>794aca3c2136933f1fe6405e56e0ce7fdbb10385</sha1>
+              <filename>sensor-kid.rom</filename>
+          </rom>
+          <mem base="0x4000" size="0x8000"/>
+        </ROM>
+      </secondary>
+    </primary>
+  </devices>
+</msxconfig>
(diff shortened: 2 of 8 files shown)


m9710797:
Hi,

This is the first ever pull request we get in openMSX (before we always got patches). So I'm not sure what's the best way to handle it (suggestions for a better workflow are welcome). I still want to make some changes before merging this pull request into master. So what I've done instead is pull your work in a 'sensorkid' branch and made some changes on top of that. Could you check those changes and test it again (I only made sure it still compiles).

Here's a summary of my changes:
- in SensorKid.xml, renamed id to sensor_kid  (see later why)
- umrcallback,tcl: slightly rewritten, no functional changes
- GlobalSettings.{hh,cc}: reverted your changes (see later why)
- SensorKid.{hh.cc}:
  - Added a reset() method.
  - Implemented the serialize() method. All member variables that need to be (re)stored from/in a savestate should be enumerated here.
  - (detail) Replaced rand() with c++11 random functions.
  - (detail) Various smaller cleanups.

But the most important change is about the 'sensor_kid_port_status_callback' setting. I moved that setting from GlobalSettings to SensorKid. So the settings is only created when SensorKid is in use. The name of the settings is also calculated from the device name (specified in the .xml file). So if you'd use two Sensorkid devices simultaneous they now each have their own setting.

Before merging your work to master I'd still like to make one more change: I'd like to move the triangle wave stuff from c++ to Tcl. So the getAnalog() method will call a Tcl proc. In umrcallback.tcl I'll add a default implementation for that proc that returns a triangle wave. But interested users can implement their own stuff. I'll try to make time for this later this week.

sndpl:
I just checked the new branch, compiled it (with make staticbindist) and tested the cartridge. All functions still work like before :+1:

m9710797:
I added 'sensor_kid_acquire_callback'. With this you can select the name of a Tcl proc. That proc gets called with a port value (0-3) as parameter, and it should return the value for that port (0-255). I also provided a default Tcl proc that returns a sine, sawtooth and triangle wave on the first 3 ports.

I committed this to the 'sensorkid' branch. I tested it, and it still seems to work fine.

One (last?) thing I'd like to change before merging to master is to integrate the rom and the extension. At the moment you need to specify both (e.g. openmsx roms/Sensor_Kid.rom -ext SensorKid). Might be more convenient if one of the two were enough.

m9710797:
Hmm, the extension does already contain a ROM.  But for some reason it didn't start correctly for me if I didn't also specify it on the command line. I'll investigate ...

m9710797:
I don't now what I did wrong before, but now it's working perfectly (openmsx -ext SensorKid).

@sndpl: If you can do some last testing and review/ack my changes, I'll merge it to master.

sndpl:
Hi @m9710797, tested the code and everything works perfect. I like the changes you made to move the values generator to tcl, looks a lot cleaner now.

m9710797:
Merged into master.
Thanks for working on this!

PR #1051, opened 2017-02-11
Master

toughkidcst:


mthuurne:
Eh, there doesn't seem to be any content in those commits. What changes were you trying to offer?

mthuurne:
Half a year passed and no clarification was given. I'm closing the request.

PR #1052, opened 2017-02-12
Added HB-4100 extension file (64Kb RAM expansion)

plainspooky:
The [HB-4100](https://www.msx.org/wiki/Sharp-Epcom_HB-4100) is a 64KiB memory expansion cartridge used by HB-MCP (a CP/M 2.2 version) as RAMDISK and by MT-BASE (a dutch database manager and sold in Brazil by Sharp as HOTDATA).

### DIFF
--- a/share/extensions/Sharp_HB-4100.xml
+++ b/share/extensions/Sharp_HB-4100.xml
@@ -1,0 +1,21 @@
+<?xml version="1.0" ?>
+<!DOCTYPE msxconfig SYSTEM 'msxconfig2.dtd'>
+<msxconfig>
+    <info>
+        <name>Expansão de Memória RAM 64Kb</name>
+        <manufacturer>Sharp</manufacturer>
+        <code>HB-4100</code>
+        <release_year>1987</release_year>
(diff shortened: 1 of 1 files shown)


MBilderbeek:
This is nice, but it's basically identical to the generic ram64k extension. I.e. there is no difference in emulation, it's just plain 64kB RAM. I'm not sure if we should add all variations of 64kB RAM cartridges in which only the meta-data is different. Identical would be the Sony HBM-64 and the Philips VU0034, and the Casio OR-264, just to name a few.

plainspooky:
Yes, I did something that already exists :-D
I checked in my openMSX installation (version 0.12.0) and had found only mappers (512K, 1M, 2M and 4M) on the extensions folder. I should have checked the main git repository first.
No problems, was a pretty nice exercise and (wow) I've created exactly the same file in the end!

MBilderbeek:
No problem, thanks anyway for your contribution!

Note that the latest version is 0.13.0, which had this extension already (I created it in April 2016).

PR #1128, opened 2018-12-01
Launch openMSX with retina support on macOS.

grauw:
This makes the window chrome appear sharp rather than blurred on a retina display.

Before:

schermafbeelding 2018-12-01 om 12 22 20

After:

schermafbeelding 2018-12-01 om 12 23 14

Note that if you symlink to openMSX and run it from the command line it currently already ignores the app’s Info.plist settings and opens the window with retina support, and I’ve been running openMSX like this for years without issues.

### DIFF
--- a/build/package-darwin/Info.plist
+++ b/build/package-darwin/Info.plist
@@ -2,6 +2,8 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>NSHighResolutionCapable</key>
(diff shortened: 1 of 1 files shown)


grauw:
More info about the Info.plist setting if desired:

https://developer.apple.com/documentation/foundation/nshighresolutioncapable
https://developer.apple.com/library/archive/documentation/GraphicsAnimation/Conceptual/HighResolutionOSX/Explained/Explained.html

Basically when Apple introduced retina displays, they introduced this setting, and apps without it will run in a pixelated x2 magnified mode as you can see on the screenshots above. The idea is that once apps have been tested and possibly adapted for retina displays, this flag is added to the Info.plist manifest.

Issue #1133, opened 2018-12-18
Using openMSX in build pipeline


immetoo2:
Hi, I'm tring to use openMSX as one or more build steps to package msx software.
For example to run the orginal file packers,custom utils and unittests.

Its working great already when manually closing openMSX constantly.
It would be nice to automate this, for this I'm requesting the following features;

* Add CommandDevice to send command from msx to emulator.
    Example usecases;
    * omsxctl exit
    * omsxctl set speed 999
    * omsxctl plug joyporta mouse
    * omsxctl set fail_after 100000

* Add optional exitCode argument to exit command.
    So msx can do shutdown with error to flag build process.

* Add failAfter command + command line option.
    When starting openMSX with this command line option it
    will always do exit(1) after the timeout.
    Expect for when the normal exit command is executed.
    It should be allowed to change this value runtime.

Is this something which is able to be added to openMSX ?

Have created a ~working commanddevice will add a PR for review.
The other two features I not sure how and where to impl it best.

MBilderbeek:
Just to understand better what you're attempting to do here, I have some questions to trigger some discussion.

Wouldn't this also be possible to implement with only a Tcl script? E.g. you could write a script that puts an I/O watchpoint on a certain unused Z80 port and the MSX program writes data to that port, which the scripts interprets.

For the third thing: there is already quite some timing possible in Tcl. A Tcl script could simply check some condition after a certain amount of MSX time and then perform the exit (with or without flag).

For the 2nd thing: so, you basically want some simple and easy way to communicate back to the program that started openMSX. I suppose the CliComm could be used for this as well, but it's not as easy as this. CliComm allows you to give commands to openMSX remotely. Then the remote program is running these Tcl commands to check certain conditions and establish a 'fail' or 'success' state of the running MSX program, see http://openmsx.org/manual/openmsx-control.html

Anyway, as I said, just asking some questions to trigger discussion, to see what the best way would be to solve the use cases you're proposing, which do sound quite interesting!

Also wondering what the other devs think.

m9710797:
I like this idea.

Let's first answer your last two points.

* Extend the openMSX 'exit' command with an optional exitcode. Shouldn't be
too hard to implement. I'll try to add it.
Manuel's suggestion (use CliComm for communication between openMSX and the
other process) would technically work, but it's too hard for a simple use
case like using openMSX as a tool in a build pipeline.

* fail_after can be implemented like this
      proc fail_after {time} { after $time {exit 1} }
[...]

PR #1134, opened 2018-12-18
MSX CommandDevice for build usage

immetoo2:
This Pull Request adds CommandDevice like seen in #1133


immetoo2:
Feature is now implemented in tcl.

PR #1141, opened 2019-01-01
Added msxbuild tcl scripts.

immetoo2:
Added improved tcl scripts like seen in #1133


immetoo2:
Not improved sufficient enough.

PR #1151, opened 2019-03-22
remove 'dead' code

Pencioner:
the condition:
    ```
            ((type == ROM_ASCII16) &&
                (typeGuess[ROM_GENERIC_8KB] == typeGuess[ROM_ASCII16]))
    ```
    is always `false` due to the fact that `typeGuess[ROM_GENERIC_8KB]`
    is always zero and `type` is initialized as `ROM_GENERIC_8KB` -
    would change to `ROM_ASCII16` only if `typeGuess[ROM_ASCII16]` > 0
    which makes second part of condition `false`

    PS i wanted to track down to commit which has changed some logic and created
    this leftover but it wasn't possible - some part of commit history was lost
    during move from svn to git and this code was already there back in 2004

### DIFF
--- a/src/memory/RomFactory.cc
+++ b/src/memory/RomFactory.cc
@@ -146,14 +146,6 @@ static RomType guessRomType(const Rom& rom)
 				type = static_cast<RomType>(i);
 			}
 		}
-		// in case of doubt we go for type ROM_GENERIC_8KB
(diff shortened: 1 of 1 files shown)


m9710797:
Thanks. I've applied your patch with some additions in the commit message.
See d34640f2b79d72f6.

PR #1156, opened 2019-04-20
Replace WavData with WavReader

D15C0DE:
This patch replaces the usage of SDL_LoadWAV() with a custom implementation. Instead of loading the entire WAV file into RAM, WavReader uses a precached file. Furthermore the difference between 8bit unsigned and 16bit signed PCM is abstracted away and the need for several different getSample() implementations is removed.

### DIFF
--- a/src/cassette/WavImage.cc
+++ b/src/cassette/WavImage.cc
@@ -1,6 +1,6 @@
 #include "WavImage.hh"
-#include "LocalFileReference.hh"
 #include "File.hh"
+#include "Filename.hh"
 #include "FilePool.hh"
 #include "Math.hh"
 #include "xrange.hh"
@@ -27,35 +27,22 @@ static void filter(float sampleFreq, int16_t* begin, int16_t* end)
 WavImage::WavImage(const Filename& filename, FilePool& filePool)
 	: clock(EmuTime::zero)
 {
-	LocalFileReference localFile;
 	{
-		// File object must be destroyed before localFile is actually
-		// used by an external API (see comments in LocalFileReference
-		// for details).
+		// Scoped to avoid the same file being opened twice.
 		File file(filename);
 		setSha1Sum(filePool.getSha1Sum(file));
-		localFile = LocalFileReference(file);
 	}
-	wav = WavData(localFile.getFilename(), 16, 0);
+	wav = WavData(filename.getResolved());
 	clock.setFreq(wav.getFreq());
 
-	auto* buf = static_cast<int16_t*>(wav.getData());
+	auto* buf = wav.getData();
 	auto* end = buf + wav.getSize();
 	filter(wav.getFreq(), buf, end);
 }
 
-int16_t WavImage::getSample(unsigned pos) const
-{
-	if (pos < wav.getSize()) {
-		auto* buf = static_cast<const int16_t*>(wav.getData());
-		return buf[pos];
-	}
-	return 0;
-}
-
 int16_t WavImage::getSampleAt(EmuTime::param time)
 {
-	return getSample(clock.getTicksTill(time));
+	return wav.getSample(clock.getTicksTill(time));
 }
 
 EmuTime WavImage::getEndTime() const
@@ -74,7 +61,7 @@ void WavImage::fillBuffer(unsigned pos, int** bufs, unsigned num) const
 {
 	if (pos < wav.getSize()) {
 		for (auto i : xrange(num)) {
-			bufs[0][i] = getSample(pos + i);
+			bufs[0][i] = wav.getSample(pos + i);
 		}
 	} else {
 		bufs[0] = nullptr;
--- a/src/cassette/WavImage.hh
+++ b/src/cassette/WavImage.hh
@@ -22,8 +22,6 @@ public:
 	void fillBuffer(unsigned pos, int** bufs, unsigned num) const override;
 
 private:
-	int16_t getSample(unsigned pos) const;
-
 	WavData wav;
 	DynamicClock clock;
 };
--- a/src/sound/SamplePlayer.cc
+++ b/src/sound/SamplePlayer.cc
@@ -73,12 +73,7 @@ void SamplePlayer::setWavParams()
 	if ((currentSampleNum < samples.size()) &&
 	    samples[currentSampleNum].getSize()) {
 		auto& wav = samples[currentSampleNum];
-		sampBuf = wav.getData();
 		bufferSize = wav.getSize();
-
-		unsigned bits = wav.getBits();
-		assert((bits == 8) || (bits == 16));
-		bits8 = (bits == 8);
 
 		unsigned freq = wav.getFreq();
 		if (freq != getInputRate()) {
@@ -102,13 +97,6 @@ void SamplePlayer::repeat(unsigned sampleNum)
 	}
 }
 
-inline int SamplePlayer::getSample(unsigned idx)
-{
-	return bits8
-	     ? (static_cast<const uint8_t*>(sampBuf)[idx] - 0x80) * 256
-	     :  static_cast<const int16_t*>(sampBuf)[idx];
-}
-
 void SamplePlayer::generateChannels(int** bufs, unsigned num)
 {
 	// Single channel device: replace content of bufs[0] (not add to it).
@@ -116,6 +104,8 @@ void SamplePlayer::generateChannels(int** bufs, unsigned num)
 		bufs[0] = nullptr;
 		return;
 	}
+
+	auto& wav = samples[currentSampleNum];
 	for (unsigned i = 0; i < num; ++i) {
 		if (index >= bufferSize) {
 			if (nextSampleNum != unsigned(-1)) {
@@ -129,7 +119,7 @@ void SamplePlayer::generateChannels(int** bufs, unsigned num)
 				break;
 			}
 		}
-		bufs[0][i] = 3 * getSample(index++);
+		bufs[0][i] = 3 * wav.getSample(index++);
 	}
 }
 
--- a/src/sound/SamplePlayer.hh
+++ b/src/sound/SamplePlayer.hh
@@ -45,7 +45,6 @@ public:
 	void serialize(Archive& ar, unsigned version);
 
 private:
-	int getSample(unsigned idx);
 	void setWavParams();
 	void doRepeat();
 
@@ -54,12 +53,10 @@ private:
 
 	std::vector<WavData> samples;
 
-	const void* sampBuf;
 	unsigned index;
 	unsigned bufferSize;
 	unsigned currentSampleNum;
 	unsigned nextSampleNum;
-	bool bits8;
 };
 
 } // namespace openmsx
--- a/src/sound/WavAudioInput.cc
+++ b/src/sound/WavAudioInput.cc
@@ -25,7 +25,7 @@ WavAudioInput::~WavAudioInput()
 
 void WavAudioInput::loadWave()
 {
-	wav = WavData(audioInputFilenameSetting.getString().str(), 16, 0);
+	wav = WavData(audioInputFilenameSetting.getString().str());
 }
 
 const string& WavAudioInput::getName() const
@@ -75,10 +75,7 @@ int16_t WavAudioInput::readSample(EmuTime::param time)
 {
 	if (wav.getSize()) {
 		unsigned pos = (time - reference).getTicksAt(wav.getFreq());
-		if (pos < wav.getSize()) {
-			auto buf = static_cast<const int16_t*>(wav.getData());
-			return buf[pos];
-		}
+		return wav.getSample(pos);
 	}
 	return 0;
 }
--- a/src/sound/WavData.cc
+++ b/src/sound/WavData.cc
@@ -1,48 +1,81 @@
 #include "WavData.hh"
+#include "File.hh"
 #include "MSXException.hh"
-#include "SDLSurfacePtr.hh"
-#include <cassert>
-
-using std::string;
+#include "endian.hh"
 
 namespace openmsx {
 
-static inline bool is8Bit(Uint16 format)
+WavData::WavData(const std::string& filename)
 {
-	return (format == AUDIO_U8) || (format == AUDIO_S8);
+	File file(filename);
+
+	struct WavHeader {
+		char riffID[4];
+		Endian::L32 riffSize;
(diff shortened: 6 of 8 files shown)


m9710797:
Hi, thanks for your patch.

I don't have much time right now, so I only took a quick look. At first sight your code looks pretty clean. But I hope to do a more thorough review 2 or 3 days from now.

You explained what your patch does, but I'm missing the motivation for the change (why is this patch useful). I do see the benefit of less memory usage. Is that the main (only?) motivation?

I also have one major concern: WavImage::getSample() and SamplePlayer::getSample() can get called from the audio thread. In that context latency is _very_ important. Before this patch, these functions read from an in-memory buffer, after they perform file I/O. This seems like a bad idea.

Thanks.

Wouter

D15C0DE:
Hi!

The main motivation is to reduce the dependency on SDL. To that end I also have a sound driver in the works which provides low-latency audio on Linux, Windows and MacOS.

If latency is an issue the patch can be modified to preload the WAV in a MemBuffer, but then you lose the memory usage advantage. I would assume the openMSX precached file with the caching provided by the OS will have enough speed to handle this.

In any case, performance quality during WavImage usage is not so important (just screeches and noise), and SamplePlayer usage is a very rare occurance. However, I didn't notice more audio issues than normal on my system.

Discode

m9710797:
Hi,

The pre-cache feature was added ~15 years ago when CDROM was still common. When using a disk image directly from CDROM, it was very annoying to have to wait for the CDROM to spin-up (and have the emulation blocked on this). In it's current form the pre-cache feature is limited to 1MB files (so won't work for many wav files). Even with that fixed, it relies on the OS to buffer the data. Thus reading the data still requires a system call (in the current implementation a call for every sample). That's 2-4(!) orders of magnitude slower than accessing memory. I'm sure that in practice on most systems it will mostly work fine anyway. But in case of audio it's better to optimize for the worst case. So summary: i think it's better to preload the wav in memory.

Reducing dependencies on 3rdparty libraries (SDL) is nice. But it also has a cost: I quickly compared your WAV parser with the version in SDL, and it seems the SDL version can handle more variations. I _guess_ that in this specific case (MSX cassette tapes, and SamplePlayer) your simpler parser covers 99% or more of all cases, but I'm not sure.

About the changes in WavImage.cc:
- Before the DC-filter was applied on the whole wav at once with an R-constant based on the wav's sample rate.
- After your patch the filter is applied on demand on each getSample() query. But the MSX samples the cassette port at a different rate (likely in a loop in the MSX-BIOS, but not always), this rate depends on whether the MSX is running on the normal 3.5MHz speed or on 5.3MHz or 7MHz 'turbo' speed. This means that the calculated R-value is no longer correct. Also this filter has a history (previous samples have an influence on future samples). Thus when 'jumping around' in the stream there will be some transition period where the filtered result is not the same compared to the reading the same samples in a timely-ordered sequence. This happens when using the 'reverse' feature in openMSX, and thus there is a (small?) chance that this leads to the MSX seeing a different cassette input signal in a normal-run compared to a replayed-reverse-run.

About SamplePlayer.cc:
- The current code can handle the case that the wav file for one or more samples is missing. In that case a warning is printed and we play silence when that sample is requested.
- I _think_ after your patch it will still print the warning, but then crash when the sample is played (dereference a null (unique_)ptr).

Wouter

D15C0DE:
> The pre-cache feature was added ~15 years ago when CDROM was still common. When using a disk image directly from CDROM, it was very annoying to have to wait for the CDROM to spin-up (and have the emulation blocked on this). In it's current form the pre-cache feature is limited to 1MB files (so won't work for many wav files).

Time to delete the cruft then? ;-)

> Reducing dependencies on 3rdparty libraries (SDL) is nice. But it also has a cost: I quickly compared your WAV parser with the version in SDL, and it seems the SDL version can handle more variations. I _guess_ that in this specific case (MSX cassette tapes, and SamplePlayer) your simpler parser covers 99% or more of all cases, but I'm not sure.

What "variations" are you referring to? My parser works according to the format specification as far as I know.

I will update the pull request in accordance with your other concerns.

m9710797:
> > The pre-cache feature was added ~15 years ago when CDROM was still common ...

> Time to delete the cruft then? ;-)

It still works when I (very rarely nowadays) use a disk image from my MCCM CDs. So I'll keep the code :-) But If I were to re-implement this feature today, I'd probably do it in a different way.

> > Reducing dependencies on 3rdparty libraries (SDL) is nice. But it also has a cost: I quickly compared your WAV parser with the version in SDL, and it seems the SDL version can handle more variations. I _guess_ that in this specific case (MSX cassette tapes, and SamplePlayer) your simpler parser covers 99% or more of all cases, but I'm not sure.

> What "variations" are you referring to? My parser works according to the format specification as far as I know.

See src/audio/SDL_wave.c around line 515 in the SDL source code. I think that switch corresponds with the 'header.wFormatTag' field in your patch. You only handle value 1 (=PCM), SDL handles several more. But as i said before: I _believe_ the simpler parser handles most (all??) practical cases. So if your patch has substantial other advantages it might be OK to drop support for these more exotic(?) formats.

> I will update the pull request in accordance with your other concerns.

Thanks. I'm looking forward to it.

D15C0DE:
> See src/audio/SDL_wave.c around line 515 in the SDL source code. I think that switch corresponds with the 'header.wFormatTag' field in your patch. You only handle value 1 (=PCM), SDL handles several more. But as i said before: I _believe_ the simpler parser handles most (all??) practical cases. So if your patch has substantial other advantages it might be OK to drop support for these more exotic(?) formats.

WAV is a container that can also hold for example MP3 data, but PCM is the usual use-case.

> > I will update the pull request in accordance with your other concerns.
> 
> Thanks. I'm looking forward to it.

I already did so 5 hours ago, directly after my comment :-)

m9710797:
> I already did so 5 hours ago, directly after my comment :-)

Ah, sorry i missed that. (i was expecting some kind of notification from github in case a pull-request changes).

I did a new review. Most of the remaining stuff are details, and I only found one bug. How do you prefer to continue? Will you make another iteration of your patch, or do you prefer that I handle it myself? Either is fine for me.

First detail is about the name of the class, WavReader vs WavData. I agree that in your first patch WavReader was a better name, but now that the behavior changed back to buffering the whole wav in memory, I personally think WavData is again a better name.

Cleanup the WavReader (WavData) interface. WavReader::getBits() is no longer needed (it's always 16 bit). Similarly the return type of getData() can now be int16_t (instead of void*, because before it could be either 8 bit or 16 bit).

The WavReader (WavData) class has a 'header' and a 'data' private member variable. Though that information is only needed during the constructor. So these should be a local variables in the constructor. The only exception is 'header.dwSamplesPerSec', that must be copied to a (new) member variable.

I think there is a bug when converting 8-bit samples to the internal 16-bit sample buffer. Inside the conversion loop, only the lower byte of the 'Endian::L16 sample' variable gets overwritten (by read()). That's only correct when you first initialize 'sample' to zero. But even then it took me a minute to verify that this indeed works (and _probably_ doesn't violate the c++ TBAA rules). Maybe it's cleaner to have a separate 'sample' variable for 8 bit and for 16 bit?

I know the WavReader constructor is not performance critical, but still it bothers me a bit that you do a read() call per sample instead of reading all data at once (the latter is easily 1000x faster). I know that the multiple-channels stuff complicates this.

Maybe a bug on windows, I'm not sure. In the WavImage constructor you removed the extra scope for the local 'file' variable. A few lines lower when the WavReader object gets constructed, the same file gets opened. I don't know if it applies here, but in some cases on windows you get an error when you try to open the same file multiple times at once. Maybe this can be solved by giving WavReader a second constructor that takes a File& parameter instead of a filename (and the existing constructor then delegates to this new constructor)?

(Detail: could already be simplified before your patch). The member variable 'SamplePlayer::currentWav' (or SamplePlayer::sampBuf before your patch) is redundant. It can be replaced (without performance cost) with 'auto& wav = samples[currentSampleNum];' on the single location where it's used.


Wouter

D15C0DE:
> I did a new review. Most of the remaining stuff are details, and I only found one bug. How do you prefer to continue? Will you make another iteration of your patch, or do you prefer that I handle it myself? Either is fine for me.

I'll iterate.

> First detail is about the name of the class, WavReader vs WavData. I agree that in your first patch WavReader was a better name, but now that the behavior changed back to buffering the whole wav in memory, I personally think WavData is again a better name.

done, although I personally think WavReader fit better with WavWriter.

> Cleanup the WavReader (WavData) interface. WavReader::getBits() is no longer needed (it's always 16 bit). Similarly the return type of getData() can now be int16_t (instead of void*, because before it could be either 8 bit or 16 bit).

done, although it became uint16_t instead.

> The WavReader (WavData) class has a 'header' and a 'data' private member variable. Though that information is only needed during the constructor. So these should be a local variables in the constructor. The only exception is 'header.dwSamplesPerSec', that must be copied to a (new) member variable.

done

> I think there is a bug when converting 8-bit samples to the internal 16-bit sample buffer. Inside the conversion loop, only the lower byte of the 'Endian::L16 sample' variable gets overwritten (by read()). That's only correct when you first initialize 'sample' to zero. But even then it took me a minute to verify that this indeed works (and _probably_ doesn't violate the c++ TBAA rules). Maybe it's cleaner to have a separate 'sample' variable for 8 bit and for 16 bit?

The & 0xFF takes care of the upper byte being wrong, there is no bug. Using more code paths feels like adding unnecessary technical debt.

> I know the WavReader constructor is not performance critical, but still it bothers me a bit that you do a read() call per sample instead of reading all data at once (the latter is easily 1000x faster). I know that the multiple-channels stuff complicates this.

The original WavData let SDL load the entire WAV into RAM, then iterate over it to filter out the unused data, and then did a memcpy of the resulting data. I personally feel this is a massive waste of RAM for WAV files that can easily be dozens of megabytes in size. How would you suggest to skip the unused data, without wasting a lot of RAM or adding more code paths?

> Maybe this can be solved by giving WavReader a second constructor that takes a File& parameter instead of a filename (and the existing constructor then delegates to this new constructor)?

Unfortunately you can't delegate constructors if a temporary is created. I solved it by scoping the setSha1Sum part.

D15C0DE:
> (Detail: could already be simplified before your patch). The member variable 'SamplePlayer::currentWav' (or SamplePlayer::sampBuf before your patch) is redundant. It can be replaced (without performance cost) with 'auto& wav = samples[currentSampleNum];' on the single location where it's used.

done

m9710797:
> I'll iterate.

We're getting close now.

> > ... Similarly the return type of getData() can now be int16_t (instead of void*, because before it could be either 8 bit or 16 bit).
> 
> done, although it became uint16_t instead.

I think it should be int16_t because now you've introduced a bug in the filter() function in WavImage.cc.
It should also be consistent with the getSample() method which must return signed values. Ah .. that's now also wrong i think currently it returns signed integers in the range 0..65536, but that must be -32768..32767  (and personally i'd make the return type of getSample() 'int16_t' instead of 'int' to better reflect this).

> > The WavReader (WavData) class has a 'header' and a 'data' private member variable.  ...

> done

Detail: there are some left over #includes in WavData.hh or ones that can now be moved to the .cc file. In other files as well (i think).
 
> The original WavData let SDL load the entire WAV into RAM, then iterate over it to filter out the unused data, and then did a memcpy of the resulting data. I personally feel this is a massive waste of RAM for WAV files that can easily be dozens of megabytes in size. How would you suggest to skip the unused data, without wasting a lot of RAM or adding more code paths?

(No need to change anything, current code is fine). The extra buffer is only temporary. And for the common case of mono 16-bit (or even mono/stereo 8-bit) you can convert in-place.

> Unfortunately you can't delegate constructors if a temporary is created. I solved it by scoping the setSha1Sum part.

(No need to change anything). Indeed the new constructor should take File by value, and WavImage should then std::move() the file object to the WavData constructor.

D15C0DE:
> > > ... Similarly the return type of getData() can now be int16_t (instead of void*, because before it could be either 8 bit or 16 bit).
> > 
> > 
> > done, although it became uint16_t instead.
> 
> I think it should be int16_t because now you've introduced a bug in the filter() function in WavImage.cc.
> It should also be consistent with the getSample() method which must return signed values. Ah .. that's now also wrong i think currently it returns signed integers in the range 0..65536, but that must be -32768..32767 (and personally i'd make the return type of getSample() 'int16_t' instead of 'int' to better reflect this).

That's a problem, because the sample data is unsigned 16bit, so it would not fit in a int16_t. Do you think it's better to store all the sample data signed then? (ie. do sign conversion on 16bit WAVs and not on 8bit WAVs, as currently.)

m9710797:
> > I think it should be int16_t because now you've introduced a bug in the filter() function in WavImage.cc.
> > It should also be consistent with the getSample() method which must return signed values. Ah .. that's now also wrong i think currently it returns signed integers in the range 0..65536, but that must be -32768..32767 (and personally i'd make the return type of getSample() 'int16_t' instead of 'int' to better reflect this).
> 
> That's a problem, because the sample data is unsigned 16bit, so it would not fit in a int16_t. Do you think it's better to store all the sample data signed then? (ie. do sign conversion on 16bit WAVs and not on 8bit WAVs, as currently.)

I'm not sure I fully understand your remark, so I'll give a more general reply.
16-bit wav files store their samples as signed values (-32768..32767). OTOH 8-bit wavs store unsigned values (0..255, with 128=center). The WavData class in your patch abstracts the difference between 8 and 16 bit wav files, so it's logical to convert the 8-bit variant to the 16-bit format (the other way won't fit). Most (all?) other sound related code in openMSX assumes samples are signed, this is much easier to work with than unsigned.

D15C0DE:
> I'm not sure I fully understand your remark, so I'll give a more general reply.

That's because I was confused and thought the 16bit WAVs were unsigned. I'll fix it after asperine and food.

D15C0DE:
> It should also be consistent with the getSample() method which must return signed values. Ah .. that's now also wrong i think currently it returns signed integers in the range 0..65536, but that must be -32768..32767 (and personally i'd make the return type of getSample() 'int16_t' instead of 'int' to better reflect this).

done, sorry for that.

> Detail: there are some left over #includes in WavData.hh or ones that can now be moved to the .cc file. In other files as well (i think).

done, I couldn't find any in other files though. Unless you mean the &lt;cstdint&gt; in WavImage.hh, since that's already included in WavData.hh, but I figure it's cleaner this way.

PR #1159, opened 2019-05-01
Add PortAudio sound driver

D15C0DE:
This sound driver provides low latency audio on Windows, Linux and Mac OS. I have no idea how to add PortAudio to the build system, so I tested it by adding "pkg-config portaudio-2.0" to the COMPILE_FLAGS and LINK_FLAGS in the platform-*.mk files.

This driver further reduces the dependence on SDL. The PortMedia project which PortAudio is part of also provides PortMIDI, which we can use to replace the current platform-specific MIDI implementations of openMSX in the (hopefully near) future.

### DIFF
--- a/build/3rdparty/portaudio-v190600_20161030.diff
+++ b/build/3rdparty/portaudio-v190600_20161030.diff
@@ -32,6 +32,17 @@ diff -ru portaudio-v190600_20161030.org/configure portaudio-v190600_20161030/con
          fi
          ;;
  
+diff -ru portaudio-v190600_20161030.org/src/hostapi/alsa/pa_linux_alsa.c portaudio-v190600_20161030/src/hostapi/alsa/pa_linux_alsa.c
+--- portaudio-v190600_20161030.org/src/hostapi/alsa/pa_linux_alsa.c	2016-10-30 02:23:04.000000000 +0100
++++ portaudio-v190600_20161030/src/hostapi/alsa/pa_linux_alsa.c	2019-09-23 19:22:02.033192258 +0200
+@@ -762,6 +762,7 @@
+         Terminate function.
+     */
+     /*ENSURE_( snd_lib_error_set_handler(AlsaErrorHandler), paUnanticipatedHostError );*/
++	snd_lib_error_set_handler(NULL);
+ 
+     PA_ENSURE( BuildDeviceList( alsaHostApi ) );
+ 
 diff -ru portaudio-v190600_20161030.org/src/hostapi/wasapi/pa_win_wasapi.c portaudio-v190600_20161030/src/hostapi/wasapi/pa_win_wasapi.c
 --- portaudio-v190600_20161030.org/src/hostapi/wasapi/pa_win_wasapi.c	2016-10-30 02:23:04.000000000 +0100
 +++ portaudio-v190600_20161030/src/hostapi/wasapi/pa_win_wasapi.c	2019-09-23 07:05:01.040185589 +0200


m9710797:
Hi, thanks for your patch.

I did a quick initial review, and in isolation the new sound driver looks fine (I'm not familiar with the PortAudio API, so I mostly skipped that part). The main difference between this and the current code seems to be that: PortAudio is a push-based API, SDL is a pull-based API.

But before doing a more detailed review, I'd like to focus on the motivation (why should we add this new code). Especially given past experience: in the past we had extra sound drivers (libao for linux, directx for windows). But these got dropped because they were not better (anymore) than the SDL driver (because SDL keeps on improving (especially after switch to SDL2)). And these extra drivers did increase the maintenance burden.

The commit message on your patch is relatively short, but you mention these advantages:
 a) low latency
 b) reduce dependency on SDL
 c) PortMIDI
Did I miss anything?

a) low latency. Did you measure this? If so, how? On what platform? How does it compare to the current code (and which SDL driver is selected on your system)? I tried googling for comparisons between PortAudio and SDL but didn't find much useful info.

b) reduce dependency on SDL. I think this is the major drawback of your patch. I does not eliminate the dependency on SDL (only, like you say, reduce it). But it does add an extra (possibly optional?) dependency on PortAudio. So netto it does complicate the build process. And like you said, the changes for the build system are completely missing from your patch. Personally I'm not too familiar (and i prefer not to be) with the details of openMSX' build system. We actually only have one developer who really knows the build system. So this is a development bottleneck. Therefor I prefer to only add an extra build dependency when it offers _significant_ advantages.

I don't believe we will be able to remove all SDL dependencies. Certainly not in the short/medium term (~5 years). And by the time we do (if ever), there's a good chance the "C++ audio proposal"
(http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1386r0.pdf) has been implemented. And if so, we may prefer that over PortAudio (because then it's no extra build dependency anymore).

c) PortMIDI. Could be interesting, but for now let's keep the discussion focussed on audio.


So if I have to be honest, for now, I'm inclined to not merge your patch. But please do answer my above questions. I'm especially interested in quantitative comparisons between SDL and PortAudio (performance / quality / latency / ...).

D15C0DE:
Hi,

> The main difference between this and the current code seems to be that: PortAudio is a push-based API, SDL is a pull-based API.

PortAudio also offers a pull-based API (in fact it's the only one mentioned on Wikipedia), but I decided against using that because it's more complicated to use (see SDLSoundDriver), and to avoid all the extra buffer copying. The push-based API is also better suited to emulation, which is in essence push-based itself.

> a) low latency. Did you measure this? If so, how? On what platform? How does it compare to the current code (and which SDL driver is selected on your system)? I tried googling for comparisons between PortAudio and SDL but didn't find much useful info.

I did not measure this, but it's a fact that PortAudio provides this. For example, on Windows it is the underlying library for JACK, which is dependent on low latency audio.

SDL since version v2.0.6 uses WASAPI on Windows (as does PortAudio), so that at least has the potential of offering low latency. However, older versions of SDL had huge latencies on Windows, which is why openMSX used to have the DirectX DirectSound driver. SDL underneath also used DirectSound, and yet there was still a big latency difference. So there must have been (and there still might be) other sources of latency inside SDL. Whether this is still the case I did not verify, as the goal of this patch is not to improve on SDL per se, but to provide an equal or better alternative.

> b) reduce dependency on SDL. I think this is the major drawback of your patch. I does not eliminate the dependency on SDL (only, like you say, reduce it).

This patch eliminates (optionally) the dependency on SDL for audio. Eliminating SDL is a step by step process. My WavData patch also didn't eliminate the dependency on the entire SDL library, but brought it a step closer. You wouldn't accept a single monolithic patch that gets rid of SDL in one go, I'm sure!

>  But it does add an extra (possibly optional?) dependency on PortAudio.

Sure, it could use the openMSX component system to optionally include it. In fact, I would like to use that same system to optionally exclude the SDL driver.

> We actually only have one developer who really knows the build system. So this is a development bottleneck.

Actually I bypass most of the openMSX build system and build openMSX with my own pkg-config based solution instead, on both Linux and Windows. Like yourself, I don't dare to get into the currently very complicated openMSX build system.

> I don't believe we will be able to remove all SDL dependencies. Certainly not in the short/medium term (~5 years).

I have faith in it. What I'm working towards is a build-time choice of "desktop openMSX" (NOT using SDL) and "handheld/embedded openMSX" (using SDL). When using the PortAudio driver, SDL is currently only necessary for keyboard/mouse/joystick input, and graphics. In all of these areas, there are plenty of other cross-platform solutions. I believe this can be achieved in under a year.

> And by the time we do (if ever), there's a good chance the "C++ audio proposal"
> (http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1386r0.pdf) has been implemented. And if so, we may prefer that over PortAudio (because then it's no extra build dependency anymore).

This is wishful thinking. Besides, the PASoundDriver is just a few lines of code, it is much simpler than the SDLSoundDriver. The build dependency could be as simple as "pkg-config portaudio-2.0"! Honestly, I don't see what could be objectionable.

> c) PortMIDI. Could be interesting, but for now let's keep the discussion focussed on audio.

You've involved the entirety of SDL into the discussion, so it's not focused on audio already. PortMIDI would be a huge reduction in cumbersome cross-platform code and is just as much part of the PortMedia project as graphics and input are part of the SDL project. Anticipating this reduction in platform-specific code which SDL cannot abstract away for you should be part of the equation.

> So if I have to be honest, for now, I'm inclined to not merge your patch. But please do answer my above questions. I'm especially interested in quantitative comparisons between SDL and PortAudio (performance / quality / latency / ...).

It is virtually guaranteed to be equal or better, but I'm not going to do these measurements. I mean no offense, but after discussions on your Discord channel and reading this response, your entire team seems to have Stockholm Syndrome with regards to SDL. Basically your only reason to deny this is having to ask your build system maintainer to integrate it in your unnecessarily complicated build system!

I believe denying this essentially really small patch would be a mistake. The platform-specific sound drivers you mentioned earlier should not be equated to this cross-platform alternative. If you're not willing to accept alternative sound drivers and your desire is to remain fully dependent on SDL, why would you keep the technical debt of an internal driver system?

Discode

mthuurne:
> > The main difference between this and the current code seems to be that: PortAudio is a push-based API, SDL is a pull-based API.
> 
> PortAudio also offers a pull-based API (in fact it's the only one mentioned on Wikipedia), but I decided against using that because it's more complicated to use (see SDLSoundDriver), and to avoid all the extra buffer copying. The push-based API is also better suited to emulation, which is in essence push-based itself.

I agree that a push API is a more natural fit.

> > a) low latency. Did you measure this? If so, how? On what platform? How does it compare to the current code (and which SDL driver is selected on your system)? I tried googling for comparisons between PortAudio and SDL but didn't find much useful info.
> 
> I did not measure this, but it's a fact that PortAudio provides this. For example, on Windows it is the underlying library for JACK, which is dependent on low latency audio.

PortAudio being created for tasks such as soft synthesis does make it likely that its authors cared about low latency. At the same time, one has to be careful with generic performance claims, since they can be outdated or only relevant in particular situations.

How much do we care for lower latency? Of course lower latency is better, just like faster code is better, but there is a maintenance cost. If SDL2's audio driver has low enough latency then we don't gain much by lowering it further.

> > b) reduce dependency on SDL. I think this is the major drawback of your patch. I does not eliminate the dependency on SDL (only, like you say, reduce it).
> 
> This patch eliminates (optionally) the dependency on SDL for audio. Eliminating SDL is a step by step process. My WavData patch also didn't eliminate the dependency on the entire SDL library, but brought it a step closer. You wouldn't accept a single monolithic patch that gets rid of SDL in one go, I'm sure!

For the record: I'm interested in eliminating the mandatory dependency on SDL2, but not in eliminating the option to use SDL2. From my experience, SDL2 is performing well, is widely ported and actively maintained, so I think it will remain the main build configuration for openMSX for the foreseeable future.

That said, I think the openMSX design benefits (*) from keeping the dependencies on SDL at its edges and to judge whether the interfaces to the SDL-specific code are in fact generic and not re-stated versions of SDL interfaces, we need alternative implementations.

(*) One benefit is to keep options open if we do want to switch away from SDL2 at some point. One of the reasons the switch from SDL 1.2 to SDL2 took a lot of effort is that we had SDL 1.2 dependencies in many places in the code (threading in particular).
Another benefit is that if we want to be able to use openMSX as a core for a multi-emulator front-end or a music player, we need small and clear interfaces at its edges.

> > But it does add an extra (possibly optional?) dependency on PortAudio.
> 
> Sure, it could use the openMSX component system to optionally include it. In fact, I would like to use that same system to optionally exclude the SDL driver.

That sounds reasonable. We have a dummy sound driver, so it would be feasible to build openMSX without the SDL audio driver even if we didn't add another audio driver.

> > We actually only have one developer who really knows the build system. So this is a development bottleneck.
> 
> Actually I bypass most of the openMSX build system and build openMSX with my own pkg-config based solution instead, on both Linux and Windows. Like yourself, I don't dare to get into the currently very complicated openMSX build system.

I'm not entirely happy with our build system either, and I wrote most of it. For its future, I'm considering two options:

- make pkg-config mandatory (it already sort-of is because of FreeType) to simplify the probe and port the Python scripts to Python 3
- port the entire build to Meson plus a much smaller collection of Python 3 helper scripts

So far Meson seems promising, but I haven't tried yet to build for the more complex platforms we support.

> > I don't believe we will be able to remove all SDL dependencies. Certainly not in the short/medium term (~5 years).
> 
> I have faith in it. What I'm working towards is a build-time choice of "desktop openMSX" (NOT using SDL) and "handheld/embedded openMSX" (using SDL). When using the PortAudio driver, SDL is currently only necessary for keyboard/mouse/joystick input, and graphics. In all of these areas, there are plenty of other cross-platform solutions. I believe this can be achieved in under a year.

Graphics will increasingly depend on OpenGL, so any framework that can provide a GL or GL ES context should work.

> > And by the time we do (if ever), there's a good chance the "C++ audio proposal"
> > (http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1386r0.pdf) has been implemented. And if so, we may prefer that over PortAudio (because then it's no extra build dependency anymore).
> 
> This is wishful thinking. Besides, the PASoundDriver is just a few lines of code, it is much simpler than the SDLSoundDriver. The build dependency could be as simple as "pkg-config portaudio-2.0"! Honestly, I don't see what could be objectionable.

If we want end users to be able to use the PortAudio driver, we need to include PortAudio in our builds, which means integrating it into the 3rdparty system and into the Visual Studio build. That's doable, but it's certainly far more effort than a one-line change.

Having the driver as a source option that's not enabled in the officially released binaries will likely mean that barely anyone uses it. We had this before with the JACK driver.

> > c) PortMIDI. Could be interesting, but for now let's keep the discussion focussed on audio.
> 
> You've involved the entirety of SDL into the discussion, so it's not focused on audio already. PortMIDI would be a huge reduction in cumbersome cross-platform code and is just as much part of the PortMedia project as graphics and input are part of the SDL project. Anticipating this reduction in platform-specific code which SDL cannot abstract away for you should be part of the equation.

I see MIDI as a more attractive part of PortAudio than the sample output, to be honest. The problem with the current MIDI code is that I have no easy way to test changes on anything other than Linux. And the current Windows MIDI code is a mess that does need rewriting. So having a cross-platform MIDI abstraction could simplify development.

> > So if I have to be honest, for now, I'm inclined to not merge your patch. But please do answer my above questions. I'm especially interested in quantitative comparisons between SDL and PortAudio (performance / quality / latency / ...).
> 
> It is virtually guaranteed to be equal or better, but I'm not going to do these measurements. I mean no offense, but after discussions on your Discord channel and reading this response, your entire team seems to have Stockholm Syndrome with regards to SDL. Basically your only reason to deny this is having to ask your build system maintainer to integrate it in your unnecessarily complicated build system!

We have a Discord channel?

Anyway, please refrain from speculating. The reason we're reluctant to include a new audio driver is that we've had 3 different alternative audio drivers in the past and all of them got removed because they weren't used. The DirectX driver was used at some point in time, while the JACK and libao drivers never saw much use.

One concern that I have with PortAudio specifically is whether the project is actively maintained. It seems they have one release every 3 years, which is a bit on the low side: even if the library is feature complete, just keeping it building on all platforms would likely require more updates than that.

D15C0DE:
> For the record: I'm interested in eliminating the mandatory dependency on SDL2, but not in eliminating the option to use SDL2. From my experience, SDL2 is performing well, is widely ported and actively maintained, so I think it will remain the main build configuration for openMSX for the foreseeable future.

Agreed, except for that last part, but we'll see!

> I'm not entirely happy with our build system either, and I wrote most of it. For its future, I'm considering two options:
> 
>     * make pkg-config mandatory (it already sort-of is because of FreeType) to simplify the probe and port the Python scripts to Python 3
> 
>     * port the entire build to Meson plus a much smaller collection of Python 3 helper scripts
> 
> 
> So far Meson seems promising, but I haven't tried yet to build for the more complex platforms we support.

For what it's worth, because I obviously have no say in this, I would greatly prefer the pkg-config approach. Much of the 3rdparty system should be obsolete for packages that support static building using the --static flag.
I don't know Meson, but I already see problems using that in the build environments I use. Having to manually keep a list of files to compile into the project is a big annoyance, which I can tell from the openMSX commit history is already problematic for Visual Studio project files.

> If we want end users to be able to use the PortAudio driver, we need to include PortAudio in our builds, which means integrating it into the 3rdparty system and into the Visual Studio build. That's doable, but it's certainly far more effort than a one-line change.

In the current system yes...

> We have a Discord channel?

Yes! lol

> Anyway, please refrain from speculating. The reason we're reluctant to include a new audio driver is that we've had 3 different alternative audio drivers in the past and all of them got removed because they weren't used.

Maybe that's your point of view, but I've talked to eg. turbor, who is as far as I know the openMSX founder, and he was quite adamant in his statements that finding alternatives to SDL was impossible. Seeing another openMSX core developer then write that it's unlikely even in 5 years time makes my head explode. Call it speculation if you will, but some amount of tunnel vision is evident.

> One concern that I have with PortAudio specifically is whether the project is actively maintained. It seems they have one release every 3 years, which is a bit on the low side: even if the library is feature complete, just keeping it building on all platforms would likely require more updates than that.

PortAudio also used by the Audacity project, as well as in MAME (which additionally uses PortMIDI). MAME also have a "3rdparty" system and could be used as a reference.

mthuurne:
> For what it's worth, because I obviously have no say in this, I would greatly prefer the pkg-config approach. Much of the 3rdparty system should be obsolete for packages that support static building using the --static flag.

Eh, the main job of the 3rdparty system is to actually build all the libraries, not just to get their link flags.

> I don't know Meson, but I already see problems using that in the build environments I use. Having to manually keep a list of files to compile into the project is a big annoyance, which I can tell from the openMSX commit history is already problematic for Visual Studio project files.

Currently only Visual Studio requires sources to be listed. Therefore a developer who adds a file on a different platform will not notice if he forgets to list it. With Meson, this requirement would apply to all platforms.

I think the trade-off Meson makes by requiring files to be listed files explicitly is only worth it for huge projects. For projects the size of openMSX, the overhead of scanning the source dir is tiny. I'll likely write a Python script to keep the file lists in sync.

> > If we want end users to be able to use the PortAudio driver, we need to include PortAudio in our builds, which means integrating it into the 3rdparty system and into the Visual Studio build. That's doable, but it's certainly far more effort than a one-line change.
> 
> In the current system yes...

Well, the current system is what we have to evaluate it in if you want it to get this merged any time soon ;)

> > Anyway, please refrain from speculating. The reason we're reluctant to include a new audio driver is that we've had 3 different alternative audio drivers in the past and all of them got removed because they weren't used.
> 
> Maybe that's your point of view, but I've talked to eg. turbor, who is as far as I know the openMSX founder, and he was quite adamant in his statements that finding alternatives to SDL was impossible. Seeing another openMSX core developer then write that it's unlikely even in 5 years time makes my head explode. Call it speculation if you will, but some amount of tunnel vision is evident.

The migration from SDL 1.2 to SDL2 took a long time and that's probably the closest comparable change we had in the past. However, a lot of what we did in that migration was getting rid of unnecessary SDL dependencies, so the code should be easier to adapt to new APIs now.

I think it would be good to test whether the lower latency is noticeable, so having a branch with PortAudio support could be useful. I don't think we should merge it into master though at this point in time.

I saw some things in the code that might be bugs; I'll add comments to the commit itself.

D15C0DE:
> > For what it's worth, because I obviously have no say in this, I would greatly prefer the pkg-config approach. Much of the 3rdparty system should be obsolete for packages that support static building using the --static flag.
> 
> Eh, the main job of the 3rdparty system is to actually build all the libraries, not just to get their link flags.

My point is that if you use package-based build environments like MSYS2 on Windows, Linux distributions, or the linux to windows cross-compiling environment MXE, static libraries are likely already built, so pkg-config suffices. Only packages that do not provide static libraries would need them to be built. And yes, I realize none of these environments support Visual Studio, which I would recommend to drop entirely. I understand that is a pretty radical change, but it works for me.

D15C0DE:
Updated the code with suggested changes. For the naming suggestions that are copied from SDL I suggest you make those changes first and then this patch can be modified according to those.

D15C0DE:
Anything happening here? I explained how to get Portaudio without having to add it to the build system by just using pkg-config, so that can't be the hold-up ;-)

Meanwhile I've been making progress towards my goals, but I would like to know if this will get accepted at all, or whether I'm looking at having to maintain a fork instead... That would really be unfortunate.

m9710797:
I'm still interested in experimenting with this patch.
Though last month has been (and following weeks will continue to be) pretty busy. Whenever I could spare a few hours to work on openMSX, because of the build stuff hassle, I choose to instead spend it on reviewing other patches or running other smaller experiments.
So for me personally, these extra build steps *are* among the main reasons for the hold-up.

m9710797:
I finally got around to actually building/testing this pull request. I've attached a patch (on top of this pull request) that integrates this new sound driver (as an optional component) into the openMSX build system. In addition it updates the code for recent openMSX code changes (switch from int16->float32 sample format). 
[portaudio.patch.txt](https://github.com/openMSX/openMSX/files/3268319/portaudio.patch.txt)

I did some comparison between the SDL and the new port-audio sound driver. I short: both seem to have equally good audio quality (I only tested on my system, other systems (or other OSes) may show different results):
- I couldn't measure any difference in latency:
  * Wrote a Tcl script that selects a random (SDL or pa) sound driver.
  * Wrote a BASIC program that measures the reaction time between playing a sound an seeing a key press.
  * Only after obtaining a result check which driver was selected.
- I tried lowering the value of the 'samples' setting. Both drivers seem to react the same:
  * samples=64: very noticeable audio stutter
  * samples=128: occasional stutter when system is loaded
  * samples=256: only audible stutter with very high load
- I tried replicating the "underflow-scenario" D15C0DE  talked about (Pointless Fighting with MoonSound and FM-PAC). With sound_driver=pa (and samples<512), I do see messages like these from time to time on stderr:
`ALSA lib pcm.c:8424:(snd_pcm_recover) underrun occurred`
With sound_driver=SDL I see similar underflows (when I instrument our SDLSoundDriver to print on underflow) but these are less frequent. However I personally still couldn't tell the drivers apart (again using the randomly selected driver approach). Both sounded equally good to me.

(On my system) there are two annoying things with the new pa driver:
- When the pa driver is (first) initialized it prints the following messages on stderr
`ALSA lib pcm_dmix.c:1108:(snd_pcm_dmix_open) unable to open slave
ALSA lib pcm.c:2565:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.rear
ALSA lib pcm.c:2565:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.center_lfe
ALSA lib pcm.c:2565:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.side
ALSA lib pcm_route.c:869:(find_matching_chmap) Found no matching channel map
ALSA lib pcm_oss.c:377:(_snd_pcm_oss_open) Unknown field port
ALSA lib pcm_oss.c:377:(_snd_pcm_oss_open) Unknown field port
ALSA lib pcm_usb_stream.c:486:(_snd_pcm_usb_stream_open) Invalid type for card
ALSA lib pcm_usb_stream.c:486:(_snd_pcm_usb_stream_open) Invalid type for card
ALSA lib pcm_dmix.c:1108:(snd_pcm_dmix_open) unable to open slave`
This does not happen with the SDL driver, also not when I select the ALSA backend in the SDL driver. (This is on a standard Kubuntu 19.04 installation with Intel 200 Series PCH HD Audio).
- Maybe or maybe not related to the first point. There's noticeable more delay when initializing the pa driver compared to the SDL driver (measured with a command like `time openmsx exit.tcl` where `exit.tcl` only contains an `exit` command). Usually this extra delay is around 0.2s, but with occasional spikes up to 1.0s!

I quickly read the PortAudio documentation, but I couldn't find anything about handling (or suppressing) errors of the underlying (ALSA) sound driver. I quickly checked the implementation of the SDL ALSA backend, but also there I couldn't find any specific code for handling (or suppressing) errors. So in short: no clue what's going on, but it is annoying.

D15C0DE:
Rerolled with build system integration and suppressed ALSA messages.

m9710797:
Thanks for the update. I can confirm that the error messages I saw earlier are suppressed now. Though the problem of the occasional slow startup remains (obviously).

Did anyone else already try this patch? Do any of you experience these problems?

I don't have any better solution, but I don't like the layering violation in this latest patch. To suppress the error messages you assume that (on linux) PortAudio is configured to use the ALSA backend. From what I read in the documentation, this will often be the case, but not always.

@mthuurne: Can you review the build system changes I made? It works for me, but I'm not confident it's completely correct.

Shall we merge this patch? Even with the known problems? If we decide to merge, my personal preference would be to not include the ALSA-error-suppression-hack, then you have at least an idea what's going on during slow startup.

D15C0DE:
To solve the layering violation, which you correctly point out is problematic in some configurations, it should be possible to patch PortAudio itself using the 3rdparty system, correct?
As for the slower startup, I'm thinking this might be due to PortAudio trying to establish the optimal buffer sizes, although I have not confirmed this.

D15C0DE:
Rerolled without ALSA-error-suppression-hack ;-)

D15C0DE:
Any news?

mthuurne:
I created a [portaudio](https://github.com/openMSX/openMSX/tree/portaudio) branch and added your patch and 3rdparty build support there.

PortAudio took some patching before it would compile. I picked the WASAPI backend for Windows, since that is supposed to offer the lowest latency.

Building for Windows using Visual Studio isn't implemented yet. I did test building using mingw-w64 and that at least produces an executable after the last round of patching. I don't have a Windows system to test it on and testing under Wine probably isn't representative of latency on Windows.

I don't have build results from macOS yet.

mthuurne:
By the way, I came across another low-latency audio library named [RtAudio](http://www.music.mcgill.ca/~gary/rtaudio/) when reading about PortAudio. If we cannot get PortAudio to work on all platforms, RtAudio might be an alternative.

D15C0DE:
I added a commit to suppress ALSA messages in PortAudio itself. I also changed the base branch to the new portaudio branch. So if you will accept this pull request, it should free my master branch for other things :-)

mthuurne:
Suppressing ALSA messages in PortAudio itself will only help if PortAudio is built via 3rdparty. ALSA only exists on Linux and Linux is a platform where static linking is the exception rather than the rule. So while this is a clean approach, it only solves the issue for a minority of use cases.

mthuurne:
By the way, PortAudio [doesn't build out of the box on macOS](https://softfab.openmsx.org/Task?report=wrapper_log.txt&taskName=build_openmsx_make&jobId=190923-0818-9280). So that's 3 out of 3 major platforms where it needs patching :(

D15C0DE:
Yet it has packages for most Linux distros, MXE (as you found), MSYS2, et cetera. How about forgetting about 3rdparty for this component? Pkg-config works great.

I'm closing this pull request to free up my master branch.

mthuurne:
Note that you can create pull requests from other branches than master. That way you don't have to limit yourself to one pull request at a time.

PR #1179, opened 2019-08-14
 Added Sony HBI-J1 extension

sndpl:
Config based on info from https://uniabis.net/pico/msx/hbij1/


MBilderbeek:
Thanks! What kind of testing did you do to verify it?

sndpl:
I only tested with kanji basic, CALL KANJI and CALL KNJ on the Sony HB-F1. It did give the same results when I executed the same calls on a Sony HB-F1XV.

PR #1185, opened 2019-09-19
Reimplement debug step/cont as a condition

D15C0DE:
Using the existing facilities for breakpoints allows us to optimize the code and greatly simplify it, while reducing global state.


m9710797:
I like this very much!

But before merging this patch, I need to sit down and carefully review it, and run some tests. Hopefully I can make time for it this weekend.  (A quickly cursory review looked fine).

Issue #1091, opened 2017-11-05
Support for undocumented V99x8 register 1 bit 2


MBilderbeek:
Apparently, this undocumented bit has at least the following effects on V9938 and V9958:
- Seems to change the clock source for the blink/page switch counter decrease from VSYNC to HSYNC.
- Seems to make the registers EO and IL at R#9 irrelevant. Whatever is in there, the VDP behaves as if both bits are 1

Test software: 
- http://msx2.org/Cadari/VDP.dsk
- https://pastebin.com/FEzBbcgA

See https://www.msx.org/forum/msx-talk/emulation/new-online-msx-emulator?page=47
See https://github.com/ppeccin/WebMSX/commit/b930be072e24c90e1356ce6f00fe5fdbbe2014ae

MBilderbeek:
Implemented via 0117dfd2f3daf66def88468af2b756bc8bfd2676

Thanks to @ducasp and @m9710797 .

MBilderbeek:
Thanks to the Wayback machine I was able to find back the vdp.dsk.... attached. Supposed to run on V9958.

[vdp.dsk.gz](https://github.com/user-attachments/files/30282592/vdp.dsk.gz)

The pastebin is this:
```basic
10 SCREEN 5
20 COLOR 0,0:VDP(1)=116
30 COLOR 15:VDP(14)=28
40 SETPAGE 0,1:CLS:SETPAGE 1:CIRCLE(120,106),100
50 GOTO 50
```

On my GT in Z80  mode it produces the same as in openMSX: 

https://github.com/user-attachments/assets/c6ec46f1-8f39-4318-ada7-850f24804ce2

grauw:
> Seems to make the registers EO and IL at R#9 irrelevant. Whatever is in there, the VDP behaves as if both bits are 1

That’s not true. The following test program results in flickering horizontal bands on a real MSX.

```basic
10 COLOR 15,1,1:SCREEN 5
20 SET PAGE 1,1:COLOR ,15:CLS
30 VDP(10)=VDP(10) OR 4
40 VDP(14)=&H11
50 VDP(1)=VDP(1) OR 4
60 I$=INPUT$(1):COLOR 15,4,4
```

It displays `page AND blink AND EO`, just like when the “caldari” bit is not set.

I think the IL flag is irrelevant to the behaviour since the only thing it does is slightly change the field timing so that monitors display it as interlaced rather than progressive.

grauw:
[...]

PR #1186, opened 2019-09-19
Support for undocumented v99x8 feature (3rd bit of register 1)

ducasp:
This implements the request #1091  , supporting the 3rd bit of VDP register 1 to change the clock source of blink.


src/video/VDP.hh
```diff
@@ -372,6 +372,31 @@ public:
 		       (!blinkState << 8);
 	}
 
+	/** Expresses the state of even/odd page interchange in a mask
+	  * on the line number. If even/odd interchange is active, for some
+	  * frames lines 256..511 (page 1) are replaced by 0..255 (page 0)
+	  * and 768..1023 (page 3, if appicable) by 512..767 (page 2).
```
D15C0DE: applicable

m9710797:
@ducasp: Thanks a lot for your patch!
I didn't have time yet to review it in detail (and I'm afraid I also won't have time in the following days). But I did have a quick look during lunch break. (Because I only took a quick look my following comments may be completely wrong).

Seems like a pretty straight-forward implementation. But I'm wondering about two things:

1) Is it needed to explicitly store the blinkStateMatrix[] array? Or could it's content be calculated on demand in the getEvenOddMask() function? At first sight i *think* it should be sufficient to store a reference line number (e.g. some line at which blinkCount==0 and blinkState=false) and from that you can deduce the full content of the array. This new state (either blinkStateMatrix[] or the reference line) should become part of the savestate (via VDP::serialize()).

2) After the previous point is addressed, is it still needed to put a synchronization point on every line? Currently it's still needed to call renderer->updateBlinkState(), but I wonder if that's really required, or if it can be handled in a different way (e.g. in SDLRasterizer<Pixel>::drawDisplay() you already query the state per line). The reason why I ask is because frequent synchronization points are relatively expensive to emulate (by itself a sync point or the code that it triggers is not expensive, but it does prevent optimizations in other sub-systems).

src/video/VDP.hh
ducasp: Last commit fixes this and the original comment

ducasp:
> @ducasp: Thanks a lot for your patch!
> I didn't have time yet to review it in detail (and I'm afraid I also won't have time in the following days). But I did have a quick look during lunch break. (Because I only took a quick look my following comments may be completely wrong).
> 
> Seems like a pretty straight-forward implementation. But I'm wondering about two things:
> 
> 1. Is it needed to explicitly store the blinkStateMatrix[] array? Or could it's content be calculated on demand in the getEvenOddMask() function? At first sight i _think_ it should be sufficient to store a reference line number (e.g. some line at which blinkCount==0 and blinkState=false) and from that you can deduce the full content of the array. This new state (either blinkStateMatrix[] or the reference line) should become part of the savestate (via VDP::serialize()).
> 2. After the previous point is addressed, is it still needed to put a synchronization point on every line? Currently it's still needed to call renderer->updateBlinkState(), but I wonder if that's really required, or if it can be handled in a different way (e.g. in SDLRasterizer::drawDisplay() you already query the state per line). The reason why I ask is because frequent synchronization points are relatively expensive to emulate (by itself a sync point or the code that it triggers is not expensive, but it does prevent optimizations in other sub-systems).

@m9710797 :

The main issue is that updateBlinkState currently does nothing, and the renderer currently is the one responsible for "blink" and "EO" related changes, using the EO related call to this function.... If EO and Blink got moved to the pixelrenderer, then probably getEvenOddMask would not be needed. But this is a way larger changer that might require much more time invested.

So, considering that we have the rasterizer checking the vram page when it is presenting, not when image is being built, and that the original approach worked fine as the EO and BLINK (not considering the undocumented feature) occur at most once per frame, that was fine. Once considering the undocumented feature and that blink might occur several times in the same frame, the approach failed to get the desired effect.

I'll check how difficult it would be to have an algorithm calculating it based on the line being requested, instead of having the matrix / synchronization point every frame line, I understand this might be way easier on the CPU.

ducasp:
@m9710797 

I think the latest commit achieves what you have suggested and is faithful to the real effect.

m9710797:
@ducasp : I merged I reworked version of this pull request. See the commit message of 0117dfd2f3da where I mentioned the most important differences.

Thanks again.

PR #1187, opened 2019-09-21
Remove redundant emulation of OPL4 NEW2 status

D15C0DE:
Because the YMF262 class already emulates this and the MSXMoonSound class reads the YMF262 status, the emulation of this  one-time status bit in MSXMoonSound is redundant.

### DIFF
--- a/src/sound/MSXMoonSound.cc
+++ b/src/sound/MSXMoonSound.cc
@@ -71,7 +71,6 @@ void MSXMoonSound::reset(EmuTime::param time)
 
 	opl4latch = 0; // TODO check
 	opl3latch = 0; // TODO check
-	alreadyReadID = false;
 
 	ymf278BusyTime = time;
 	ymf278LoadTime = time;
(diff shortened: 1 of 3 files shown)


m9710797:
Agreed. I'm about to merge this pull request. But first I'd like to ask for clarification on one detail:

In YMF262::serialize() when loading an old savestate (without info on alreadySignaledNEW2) you changed the 'guessed' value from false to true. Why is that? I should have documented it in the source code, but my reasoning was: "it's safer to allow querying NEW2 twice than not all". Scenario: MSX boots and software queries NEW2 (once). Savestate before this query still works. Savestate after the query is technically in the wrong state, but since software only queries NEW2 once this doesn't really matter. What am I missing?

D15C0DE:
This behavior in serialize() was in the MSXMoonSound class so I merged the implementation. I figured your reasoning was "a savestate is probably somewhere after the first initialization, so setting it to already initialized is more likely". Although it really doesn't matter much, because an already running program probably won't query that status bit. Although if it uses fixed value compare, it might be significant. In any case it's a tiny backwards compatibility issue and I don't have a strong opinion either way.

D15C0DE:
Also, because this only concerns savestates where this function was not yet implemented, setting it to true is actually more correct, because it hides the function just like it was when the savestate was created.

m9710797:
Good point. I've merged your pull request.
(No idea why github doesn't automatically sees this merge, so I'm manually closing this pull request).

Issue #961, opened 2015-06-24
[Bug] MSX turboR is going too slow [sf#570]


openMSX-import:
**Reported by manuelbi on 2015-03-04 20:58 UTC**
The program posted here http://www.msx.org/forum/msx-talk/development/basic-hat?page=1:

```
10 DIM RR(320)
100 COLOR 15,1,1:SCREEN 6
110 FOR I=0 TO 320:RR(I)=193:NEXT I
130 XP=144:XR=4.71238905#:XF=XR/XP
140 FOR ZI=64 TO -64 STEP -1
150 ZT=ZI*2.25:ZS=ZT*ZT
160 XL=INT(SQR(20736-ZS)+.5)
170 ZX=ZI+160:ZY=90+ZI
180 FOR XI=0 TO XL
190 XT=SQR(XI*XI+ZS)*XF
200 YY=(SIN(XT)+SIN(XT*3)*.4)*56
210 X1=XI+ZX:Y1=ZY-YY
220 IF RR(X1)>Y1 THEN RR(X1)=Y1:PSET (X1,Y1),15
230 X1=ZX-XI
240 IF RR(X1)>Y1 THEN RR(X1)=Y1:PSET (X1,Y1),15
250 NEXT XI:NEXT ZI
260 GOTO 260
```

is supposed to run in about 22 seconds on a turboR GT in R800 mode with MSX-BASIC-kun 2.1. But in openMSX it takes much longer: about 58 seconds.

Interesting, because we also have a ticket that R800 emulation is too slow!

Looks like on blueMSX the time is much more like real hw: about 23 seconds.

openMSX-import:
**Commented by m9710797 on 2015-03-05 08:29 UTC**
There certainly are timing issues in the R800 emulation in openMSX. But a factor 2x-3x is very large. Therefore I'd suggest to first investigate other possible causes for this difference.

The 3 measurements discussed in the MRC forum where performed by 3 different people. Did all 3 test exactly the same configuration? E.g.
- Z80 vs R800-ROM vs R800-DRAM
- The Basic program typed in exactly the same way, including spaces (though maybe for basic-kun this matters less).
- But my prime suspect is the basic-kun code. Was it loaded in the same way? It makes a huge difference whether it's located in an external ROM vs loading it in RAM (IIRC with something like  bload"xbasic.bin",r)

[...]

Issue #989, opened 2015-12-18
[bug] Missing I/O slowdown on Panasonic MSX2+ machines in turbo mode


sdsnatcher:
I ran hit9918's VDPSPEED.BAS on both Panasonic FS-A1ST and FS-A1WX, and clearly the Panasonic MSX2+ machines have an I/O slowdown imposed by the engine (the M60014, in this case) with the turbo on as the S1990 does for the R800 mode of the TR.

The results are shown here:
http://www.msx.org/forum/msx-talk/hardware/vdp-speed-curiousity?page=7

The results of the WX were very close to the ST when both machines were on turbo mode. My suggestion is to use the S1990 delays for the Panasonic MSX2+ machines too. Maybe it isn't perfect, but it's certainly better than the broken machines we have currently. It can be perfected some day.

MBilderbeek:
Do I understand correctly that this delay
- is only implemented in MSX2+  machines with the turbo feature (5.37MHz)
- is only active when the turbo mode is active?

sdsnatcher:
When I opened the bug report, I meant it exactly as you just described. The delay only for turbo machines and only when the turbo is active.

But after that, I discovered that both Panasonic and Sanyo MSX2+ models also have another type of VDP I/O delay that takes place when there's no turbo activated. (*1) This delay causes the Unknown Reality sound scope demo to fail.

The relationship between the non turbo delay and the turbo delay is interesting: the T9769A always impose a delay for VDP access. Then, the gate array (M60014 chip) on Panasonics impose a longer delay only when the turbo is enabled. 

I don't know if you prefer this non-turbo delay as part of this bug report, or if you prefer it as a separate issue.

*1: Since on Sanyo machines the gate array has the turbo feature almost completely stripped, the only delay that exists on those machines is the one imposed by the T9769A.

Note: it seems that the T9769A used on MSX2+ models imposes a slightly longer delay than the T9769C used on Turbo-R models. This could explain why the Unknown Reality demo runs fine on TRs, but not on MSX2+ from Panasonic & Sanyo. It's either that, or the gate array of those MSX2+ machines also impose a delay when the turbo is off. 

Anyway, for emulation purposes it doesn't matter if the non-turbo delay is imposed by the gate array or the T9769A: it only matters to replicate the fact that those machines won't run the Unknown Reality 2.0 sound scope demo without glitches.     :)

MBilderbeek:
OK, so did I understand correctly that:
[...]

Issue #563, opened 2015-06-22
[Bug] OUT timing on machines with Toshiba MSX Engine is slower [sf#262]


openMSX-import:
**Reported by manuelbi on 2007-03-23 19:34 UTC**
VDP out timing in MSX2+ and turboR machines (with T9769B and T9769C MSX-engines) takes one extra cpu cycle. Some programs rely on this and some programs don't work on these machines because of this.

openMSX doesn't emulate this.

openMSX-import:
**Commented by manuelbi on 2010-07-18 18:55 UTC**
Update from Wouter: openMSX already emulates this on turboR's, but not on the other machines using such an MSX-ENGINE.

openMSX-import:
**Commented by manuelbi on 2014-09-18 20:18 UTC**
Mar 23 2007 20:27:40 <dvik>     ok, just wanted to know what to expect. I'm test
ing our new demo
Mar 23 2007 20:28:02 <dvik>     its supposed not to run on MSX2+ and TR machines
 but it seems to do so in openMSX

....

Mar 23 2007 20:30:51 <dvik>     its kindof nice to have that emulated. I think t
here are some (not many) apps that needs it
Mar 23 2007 20:31:21 <Quibus>   Can you name some?
Mar 23 2007 20:31:28 <dvik>     there is I think a turbor video player that needs it (can't remember the name)
Mar 23 2007 20:32:20 <dvik>     That player was one of the reasons white cat fixed the timing on a couple of instructions a month or two ago. but I think it requires correct out timing as well
Mar 23 2007 20:32:33 <dvik>     so I guess it will play slightly too fast in openMSX
Mar 23 2007 20:32:40 <Quibus>   would be nice to have that soft
Mar 23 2007 20:32:47 <dvik>     I can ask him
Mar 23 2007 20:32:53 <Quibus>   please do
Mar 23 2007 20:32:59 <Quibus>   it's good to have test cases
Mar 23 2007 20:33:06 <BiFiMSX>  the eva player or something?
Mar 23 2007 20:33:13 <dvik>     sounds familiar
Mar 23 2007 20:33:38 <dvik>     it supports scsi (not sure if it supports ide as well)

So, I guess we'll have to ask dvik for the test cases :S

MBilderbeek:
This delay on MSX2+ sounds like it's the fixed delay mentioned in #989 .
[...]

PR #1188, opened 2019-09-23
T9769 applies VDP I/O waitcycles only on Z80

D15C0DE:
The reason for this is that both the Z80 and the waitcycle circuit are internal to the T9769 MSX-ENGINE and do not manifest externally. The situation before this patch caused a 1 or 2 cycle wait on R800 which does not exist in real hardware.

### DIFF
--- a/src/cpu/MSXCPU.cc
+++ b/src/cpu/MSXCPU.cc
@@ -188,10 +188,10 @@ void MSXCPU::wait(EmuTime::param time)
 	          : r800->wait(time);
 }
 
-EmuTime MSXCPU::waitCycles(EmuTime::param time, unsigned cycles)
+EmuTime MSXCPU::waitCyclesZ80(EmuTime::param time, unsigned cycles)
 {
 	return z80Active ? z80 ->waitCycles(time, cycles)
-	                 : r800->waitCycles(time, cycles);
+	                 : time;
 }
 
 EmuTime MSXCPU::waitCyclesR800(EmuTime::param time, unsigned cycles)
--- a/src/cpu/MSXCPU.hh
+++ b/src/cpu/MSXCPU.hh
@@ -110,7 +110,7 @@ public:
 	void setNextSyncPoint(EmuTime::param time);
 
 	void wait(EmuTime::param time);
-	EmuTime waitCycles(EmuTime::param time, unsigned cycles);
+	EmuTime waitCyclesZ80(EmuTime::param time, unsigned cycles);
 	EmuTime waitCyclesR800(EmuTime::param time, unsigned cycles);
 
 	CPURegs& getRegisters();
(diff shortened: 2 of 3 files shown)


m9710797:
Could you tell a bit more about how you tested this? E.g. share your test code. Did you take the wait-for-even-cycle R800 behavior into account? Thanks.

D15C0DE:
There is no test code and nothing to take into account. The T9769 wait cycles simply having nothing to do, at all, with the R800.

m9710797:
What you say makes sense, but I'm not confident with applying this change without first verifying it on a real machine.
Maybe this weekend I can make some time to perform such tests myself. I suspect those tests will confirm your claim, and then we can apply this patch (+extra comments).

D15C0DE:
Unless you believe in signal telepathy, there is no way for this internal T9769 signal to propagate outside of the chip. Additionally, as you no know, the S1990 applies its own VDP wait mechanism on the R800.

Looking at the https://github.com/openMSX/openMSX/issues/563 and https://github.com/openMSX/openMSX/issues/989 issues which are referenced by the code, the current implementation stems from a time when the exact behavior was not know to you.

But by all means, waste your time by writing a test case. You'll have to make sure you don't trip the S1990 VDP wait. Otherwise the effect will be masked.

MBilderbeek:
I agree with Wouter it would be interesting to know what the effect of these changes are on existing software. E.g. which software will run wrong now and correct after the patch is applied? How can we see that?
So, what is the (positive) effect of this patch? The visible fix?
Or, if you like, how can we easily explain our users that this is a good thing to have? 

For instance, is the turboR VDP I/O running too slow now? Could it fix #961 ?

D15C0DE:
Of course it is interesting to see if it could fix https://github.com/openMSX/openMSX/issues/961 but I don't understand what seems to be a resistance to this obviously correct patch. I certainly don't understand why this would need to be marketed/explained to users.

It feels like you don't appreciate having this long-standing bug fixed.

MBilderbeek:
Sure it's interesting. I'd just like to know what the visible effect is. Is that too much to ask?

Vampier:
This is exactly the reason why some people are not participating in open source projects. They jump through all the hoops that have been put up - they conform according to the rules and the when push comes to shove... well sorry... come on.

D15C0DE:
I was expecting "Oh thanks! How could we have missed this! What a silly oversight."
Instead I get a response that seems to indicate you don't actually understand how a MSX turboR works.

Whether or not the effects of this patch are interesting or visible, or whether it's marketable to your users, is secondary to its primary function, which is to emulate the hardware properly.

m9710797:
Sigh. I do welcome this patch. I do think you're right about this behavior. The only thing I'd like to do is to double check this on real hardware. And I volunteered to do that check myself (when time permits). And after that I will merge your patch.
Isn't this all standard practice? I mean at work, even if you're pretty sure your code change is correct, you still do code review, and you still write tests, no?  (Actually I don't know if your day job involves writing code ;-)
I hope this is settled now.

D15C0DE:
If you had said so immediately, that would've been fine. Except you didn't. And MBilderbeek didn't.
At the very least you should work on your communication skills and don't send conflicting messages to contributors (like with the PortAudio patch), because as a contributor, I'm starting to wonder why I even bother making pull requests.

m9710797:
(As expected) I've confirmed this behavior. See 8d986c47dc13cb for details.

PR #1190, opened 2019-10-11
Fix V9990 P2 background rendering

D15C0DE:
In P1, each layer has its own palette offset. However, the coupling between palette offset and layer really comes from the fact that layer A patterns are in VRAM0, and layer B patterns are in VRAM1. This coupling between palette offset en VRAM chips also exists in P2, in which pattern data is interleaved between VRAM0 and VRAM1. Thus, even and odd bytes each have their own palette offset.

This patch implements that by always alternating between two palette offsets for even and odd bytes. In the case of P1 both offsets are the same for each respective layer.


m9710797:
Had a quick look during lunch break: looks good.
I'll try to make more time for this tomorrow (no promise).
Thanks!

m9710797:
Merged. Thanks again.
I only re-added a small comment I removed in an earlier patch (wasn't needed anymore, but now it's needed again).

PR #1194, opened 2019-10-21
Remove SDLGL-FB16/32 renderers

D15C0DE:
These have been disabled in release builds 14 years ago, and have been deprecated for over 10 years.

The second commit in this pull request dissolves the confusingly named SDLGLOutputSurface class, which lost most of its function after the removal of the SDLGL-FB renderers. For the common saveScreenshot code I used a static member function, just like in the SDL renderer.


mthuurne:
In what way are they deprecated? We kept them around in developer builds as a debug tool, if I recall correctly.

That said, I've been arguing that software scaling is not viable long-term, with hi-DPI screens becoming more common. So I'd like to keep a 1x and perhaps a 2x software scaler and drop the rest. In that case, the FB16/32 renderers become less useful. This PR shows the code can be simplified by dropping them, so that's probably a good move then.

One thing in the patch that seems a bit strange at a first glance is giving `clearScreen()` an empty default implementation. If that's not necessary, I'd rather keep it abstract, or at least change it in a separate commit if there is a strong reason to do so.

D15C0DE:
Deprecated as in commit messages as early as 2009 mentioned them likely to be removed in the future.

For clearScreen() I used the same empty default mechanism that was used for other methods in that class which only apply to frame buffers.

m9710797:
Occasionally I still used the FB16 renderer to check if 16bpp software rendering still works. Probably no current machine (desktop, laptop, mobile) is still using a 16bpp output mode. But we once had plans to try using 16bpp to speedup the openGL renderer: do MSX rendering in 16bpp (should be enough), upload this (half as much) data to a texture, and do color conversion (to 32bpp) with pixel shaders. Never got around to this, so not a strong enough reason to keep FB16.

Another thing the FBxx renderers can do, but SDLGL_PP cannot is running the 'SaI' and 'MLAA' scalers. At one point in time FBxx seemed like a viable approach to keep these scalers working in an openGL-only openMSX version. Unfortunately still today(!) there are systems that have trouble running openGL (or GL-ES), so dropping the software rendering is not yet an option :(  But apparently not much work has been done on this in 10 years. So I agree it's probably fine to remove the FBxx renderers now.

So I agree in principle with this PR. I didn't have time yet to look at the actual code. I'll try to make time for that later this week. (@mthuurne: if you find time before me, then of course feel free to already merge).

m9710797:
Merged (I also adjusted the msvc project files).
Review was non-trivial, not because of this PR, but because the original code was rather complex. Thanks to these patches it's a little bit simpler now :)
Thanks.

PR #1203, opened 2019-12-21
Make it possible to use non-SDL video backends

D15C0DE:
Historically openMSX has provided drivers and renderers that seem to allow different video backends. In practice however, SDL-specific functionality has seeped into the very core of the video classes.

This set of commits make it possible to use non-SDL video backends by turning the core video classes into (abstract) base classes and moving the SDL functionality into derivations of them.

The first and last commit of this set, though strictly not SDL-specific, are necessary to allow videosystems such as GUI toolkits to work.

### DIFF
--- a/src/Reactor.cc
+++ b/src/Reactor.cc
@@ -555,21 +555,24 @@ void Reactor::run(CommandLineParser& parser)
 		}
 	}
 
-	while (running) {
-		eventDistributor->deliverEvents();
-		assert(garbageBoards.empty());
-		bool blocked = (blockedCounter > 0) || !activeBoard;
-		if (!blocked) blocked = !activeBoard->execute();
-		if (blocked) {
-			// At first sight a better alternative is to use the
-			// SDL_WaitEvent() function. Though when inspecting
-			// the implementation of that function, it turns out
-			// to also use a sleep/poll loop, with even shorter
-			// sleep periods as we use here. Maybe in future
-			// SDL implementations this will be improved.
-			eventDistributor->sleep(20 * 1000);
-		}
+	while(iterate());
+}
+
+bool Reactor::iterate() {
+	eventDistributor->deliverEvents();
+	assert(garbageBoards.empty());
+	bool blocked = (blockedCounter > 0) || !activeBoard;
+	if (!blocked) blocked = !activeBoard->execute();
+	if (blocked) {
+		// At first sight a better alternative is to use the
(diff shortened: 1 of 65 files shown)


m9710797:
Thanks a lot for this work. I had a quick first look at this series and it looks very reasonable. But before merging I'd like to do a more detailed review first. A large part of that work will be to re-understand my own design ;-) I think I can make time for this on Monday or Tuesday. 

Already some initial feedback:

In part 4 'Move SDL-specific functionality out of OutputSurface' you add some dynamic_cast's to the code. I understand why a cast is needed, but I wonder why you choose 'dynamic_cast' instead of 'static_cast'. Would using our 'checked_cast' utility be a good compromise? (It acts like a static_cast, but for extra safety, and only when asserts are enabled, it checks that static_cast and dynamic_cast give the same result).

In part 5 'Make it possible to use non-SDL video backends' you introduce an abstract PixelFormat class. My first reaction is that this is overkill. But it's hard to judge the API of a class hierarchy when only one concrete implementation of the base class is visible, so give me some slack ;-) Wouldn't it be possible to make PixelFormat a concrete class? So no subclasses, instead of abstracting SDL_PixelFormat, PixelFormat would replace SDL_PixelFormat. The current virtual methods in PixelFormat then become simple getters for concrete data members. And the PixelFormat::map() method will do the calculation itself instead of delegating to SDL_MapRGB(), but that calculation is fairly easy.

Do you mind if I already cherry pick some easy parts (e.g. part 1)? Or do you prefer that I wait and merge this whole series in one go?

D15C0DE:
About dynamic_cast; any cast that works is fine by me.

About SDLPixelFormat; I didn't know to what degree the openMSX 2D engine depends on the specifics of SDL_MapRGB(), particularly with SDL ports to more exotic platforms. If you think it's ok to replace SDL_PixelFormat with PixelFormat completely, that might work. Since OpenGL has a fixed pixel format, the PixelFormat class is really mostly there to make the rest of the code work with both the GL and the SDL 2D engine. In fact, in my GL-only WIP I use a read-only pixelformat.

You can cherry pick from this series, sure!

m9710797:
I've merged this PR with small edits (see commit messages for details).

Apart from the 2 items I mentioned above (dynamic_cast and PixelFormat), I also:
- Made OutputSurface aware of its logical dimension (it already knew about its physical dimension).
- Fixed a bug related to the 'screenshot' command with renderer=sdl.

Thanks again.

Issue #949, opened 2015-06-24
[Bug] Graphics corruption on MSX1 with Karateka [sf#558]


openMSX-import:
**Reported by manuelbi on 2014-12-06 22:34 UTC**
Regression with VRAM corruption: Karateka.

This game gives graphics corruption when emulating an MSX1.

I still need to double check on my real MSX. Attached ROM (based on Amusement Factory's disk version) shows the issue.

openMSX-import:
**Commented by manuelbi on 2014-12-08 21:51 UTC**
OK, I tested the game (in the attached ROM form) with my MegaFlashROM SCC on two real MSX1's: a Yashica YC-64 (with TMS9929A) and a Sony HB-10P (with T6950). And... there is _some_ corruption noticable. Not as much by far as you can see in openMSX, but it's clearly visible. Especially in the opening texts and also a few wrong lines in the logo. During the scroll after the logo you can also see some wrong pixel or groups of pixels every now and then (but constantly). All of this is not visible on an MSX2.

So, there is corruption on real MSX1 too, but it's a bit over the top in openMSX at the moment.

openMSX-import:
**Commented by manuelbi on 2014-12-14 22:11 UTC**
Experiment results: changing the delta from 7 to 6 cycles gives too little corruption compared to real MSX. Now only real corruption is seen in the word 'presenta' below "Broderbund Software" in the first screen of the title sequence. On the real machine there was some corruption in almost all of these screens, see previous post.

So, changing it to 6 may be on the safe(r) side, but it's not giving the same results as on real MSX, so it's not the definite solution.

mcolom:
Hi all!
I suspect that this is caused because the VDP access slots don't appear exactly at the same moment they're scheduled in the emulator. After comparing with some videos of the real machines with the emulated counterparts, one can infer that the timing windows open too late. Adding a small negative time shift (but not a DELTA integer multiple) fixes Karateka and produces the same artifacts as in the real machines, in the few tests I could do. Of course, more tests in real machines are needed to confirm 100% this, but I'm quite confident that's the reason.
There's a discussion about this in this forum: https://msx.org/forum/msx-talk/openmsx/mystery-game-bestial-warrior-with-corrupted-colors-in-msx1?page=1

[...]

PR #1206, opened 2019-12-27
Fix MSX1's VDP window timing

mcolom:
It seems that the actual slot time access happens slightly sooner in a real MSX 1. Thus, correct the scheduled time for the next access window.
This fixes Karateka and gives the same color corruption that can be seen in https://youtu.be/7HXZHeZIuJY?t=1867
[ToDo] Check if the timing needs also to be corrected in MSX2.
Related bug: https://github.com/openMSX/openMSX/issues/949

### DIFF
--- a/src/video/VDP.cc
+++ b/src/video/VDP.cc
@@ -131,8 +131,8 @@ VDP::VDP(const DeviceConfig& config)
 		version = TMS91X8;
 		defaultSaturation = 100;
 	} else if (versionString == "TMS9128") version = TMS91X8;
-	else if (versionString == "TMS9929A") {version = TMS9929A; slotTimeShift = 2500.0;}
-	else if (versionString == "TMS9129") version = TMS9129;
+	else if (versionString == "TMS9929A") version = TMS9929A;
+	else if (versionString == "TMS9129") {version = TMS9129; slotTimeShift = 2420.0;}
 	else if (versionString == "V9938") version = V9938;
 	else if (versionString == "V9958") version = V9958;
 	else throw MSXException("Unknown VDP version \"", versionString, '"');
--- a/src/video/VDP.cc
+++ b/src/video/VDP.cc
@@ -114,13 +114,15 @@ VDP::VDP(const DeviceConfig& config)
 
 	int defaultSaturation = 54;
 
+    slotTimeShift = 0;
+
 	std::string versionString = config.getChildData("version");
 	if (versionString == "TMS99X8A") version = TMS99X8A;
 	else if (versionString == "TMS9918A") {
 		version = TMS99X8A;
 		defaultSaturation = 100;
 	} else if (versionString == "TMS9928A") version = TMS99X8A;
-	else if (versionString == "T6950PAL") version = T6950PAL;
+	else if (versionString == "T6950PAL") {version = T6950PAL; slotTimeShift = 4930.0;}
 	else if (versionString == "T6950NTSC") version = T6950NTSC;
 	else if (versionString == "T7937APAL") version = T7937APAL;
 	else if (versionString == "T7937ANTSC") version = T7937ANTSC;
@@ -129,7 +131,7 @@ VDP::VDP(const DeviceConfig& config)
 		version = TMS91X8;
 		defaultSaturation = 100;
 	} else if (versionString == "TMS9128") version = TMS91X8;
-	else if (versionString == "TMS9929A") version = TMS9929A;
+	else if (versionString == "TMS9929A") {version = TMS9929A; slotTimeShift = 2500.0;}
 	else if (versionString == "TMS9129") version = TMS9129;
 	else if (versionString == "V9938") version = V9938;
 	else if (versionString == "V9958") version = V9958;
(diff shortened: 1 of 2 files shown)


D15C0DE:
Does this time shift really originate from the VRAM slots? It seems to me this same shift could possibly be explained by (a combination of) any of these things:
1. start-up times of CPU and VDP
2. exact timing of vblank interrupt signal
3. (de)synchronisation of CPU and VDP clock crystals
4. The fact that PAL MSX1 VDP are generally clocked at 10.68MHz instead of 10.74MHz

I would personally prefer the root cause to be found, rather than implement this symptom-based solution, which might interfere with carefully timed demoscene software.

cross-posted to #949

m9710797:
Thanks for your PR! I'm pretty busy these days. So I'm afraid I won't have time to look into this PR till Tuesday.

I've also only been following the discussion on MRC from a distance, so I have to admit that I'm not really familiar yet with the technical details and the investigations that have already been done. But I do share D15C0DE's concerns. Is there any way we could test in detail what's really going on?

mcolom:
Yes... perhaps it's better to wait until more tests are performed with real machines before modifying the emulator.
At this moment it's clear to me that the inaccuracy is caused by a time shift, but as D15C0DE says it's not clear where it originates from.
What we were discussing at MRC is to make some tests programatically to write to VRAM as fast as possibly and to compare the real and emulated machines. That would only give a precise idea of the time shifts of the VDPs (which is not bad at all!) but I'm afraid it won't tell much about the origin.
We'll have to explore more to figure out where the time shifts come from, but the fact the adding the correct one supposedly make the artifacts match, it's a good sign or hint. I guess!

MBilderbeek:
@m9710797 Is it possible that there is a bug in the implementation or an error in the research which could explain that a different timing of these windows would lead to more realistic results?

Or are more measurements needed on different types of VDP's? (Measurements similar to the research you did.)

Or are different kinds of measurements needed?

D15C0DE:
@mcolom The discovery that this is problem is caused by synchronization skew between the CPU and VDP is very good indeed! The research about the amount of shift on different MSX models is also good, because once the origins are known these findings can be applied immediately.

D15C0DE:
One thing we have to be aware of is that 100% the same output is likely not achievable, because part of it is caused by analogue effects. I'm talking about what happens when data changes during signal edges and such. There is probably also at least one on-chip buffer involved.

In tests in screen0:width80 (on V9958) I've seen the wildest effects. Even repeating values! In that particular case the VDP is able trigger the write, but it was not able to update its buffer with the new value in time. Or you see values that were written partially, with some bits new and some bits old. In this case maybe the buffer was being updated while the VRAM interface was still sampling it.

In short, while it should be possible to emulate the _location_ of the VRAM faults due to too fast acces, it likely won't be possible to emulate the exact _values_ for the faulty writes.

mcolom:
> One thing we have to be aware of is that 100% the same output is likely not achievable, because part of it is caused by analogue effects.

I'm afraid so. Indeed, these two pictures that Manuel posted on the MRC show two different executions (I guess) of the same game in the same machine. The program changes the screen mode, fill in a bitmap (at a good -slow- speed), and the updates the color map (too fast). The result: both executions of exactly the same operations give different results (look at the colors in the frame):
https://msx.pics/images/2019/12/28/TV2019122814174400.jpg
https://msx.pics/images/2019/12/28/TV2019122814160400.jpg
The colors in the frame are updated only once, at the beginning.

I agree it won't be possible to emulate the exact behavior if it's a corner case driven by analog signals.

What I don't get is where it comes from that time shift and (I assume) makes the VDP access windows happen earlier than they should be.

mcolom:
Some thoughts about the time shift.
With an emulated Sony_HB-20P, my minimal constant that makes Karateka work is slotTimeShift=1280. This constant is the number of ticks of the main clock (defined in EduDuration.hh (MAIN_FREQ = 3579545ULL * 960).
The duration of the time shift is shift_segs_per_cycle = slotTimeShift / MAIN_FREQ = 372.486820 ns.
Let's see now how many VDP ticks is that time shift: 372.486820 ns / 46.560852 ns = 8.000000.
It's "exactly" 8. The code in openMSX assumes that it's a V99x8 running, so we need to divide by 4 to obtain the number of advanced cycles in the case of the T6950A. So, it's an advance of 2 cycles for the T6950A.

I hope there's no mistake in what I wrote. I'll go on figuring out where this time shift might come from...

mcolom:
I think I have an explanation for the advance of two cycles for the 9918 VDP.

In the document about the timing measurements that was made some time ago [1], we can read (in "general (memory) timings"): "The TMS9918 runs at 5.37MHz (1.5×3.58MHz, 4× slower than the V9938). One display line takes 342 cycles (as expected, 4× less than on V9938). One memory access takes only 2 cycles or 372ns. So compared to V9938 each memory access takes slightly longer (on V9938 one access takes 6 cycles or 279ns)".

So, they have different memory access times: 372 ns vs. 279 ns.
The number of ticks of the difference between these two times is (372 - 279) 1e-9 21477270 = 1.9973861100000003 ticks.
We can compute the access times ourselves from the clock frequency and assuming that there will be always 1368 ticks/line. This gives us 372.49 ns and 279.37 ns. With this we obtain (372.49 - 279.37) * 1e-9 * 21477270 = 1.9999633824000003 ticks.
In practice, both mean 2 ticks.

When the 9918 is used in the emulator, it is expecting to see the access windows 2 ticks before, so they need to be advanced in the emulation. Actually, the current VDP code in openMSX assumes a 9938 and then it makes some adjustments (like multiplying the number of ticks by 4 in VDPAccessSlots.cc for the MSX1 case). This is just an additional adjustment.

In the case of a PAL 9918, we'd have  (360.88 - 279.37) * 1e-9 * 22168000 = 1.80691368, which accounts for 2 ticks also.

[1] http://map.grauw.nl/articles/vdp-vram-timing/vdp-timing-2.html

MBilderbeek:
So, I guess I missed something: what is the conclusion? Do we have new information on the root issue? Do we need to change something in the emulation?

mcolom:
Well, the explanation I wrote two years ago I think it's correct, but since all the experimental data we have comes from a VDP different from the 9918 I agree with D15C0DE that it'd be better to obtain more data from a real device rather than making assumptions. Even if the modifications proposed in this PR make the results resemble much more to what we observe in the real machine, still I agree on being conservative and wait until more data is gathered.

PR #1207, opened 2019-12-28
Upgrade OpenGL drawing code to version 3.3 Core

D15C0DE:
The main purpose of this series is to upgrade the OpenGL drawing code in openMSX to OpenGL 3.3 Core profile. This version should be supported on systems sold from 2008 onwards. I did a basic conversion, optimizing only obvious cases. I'm sure there is a lot more room for improvement.

### DIFF
--- a/share/shaders/default.frag
+++ b/share/shaders/default.frag
@@ -1,16 +1,18 @@
 uniform sampler2D tex;
 uniform sampler2D videoTex;
 
-varying vec2 texCoord;
-varying vec2 videoCoord;
+in vec2 texCoord;
+in vec2 videoCoord;
+
+out vec4 fragColor;
 
 void main()
 {
 #if SUPERIMPOSE
-	vec4 col = texture2D(tex, texCoord);
-	vec4 vid = texture2D(videoTex, videoCoord);
-	gl_FragColor = mix(vid, col, col.a);
+	vec4 col = texture(tex, texCoord);
+	vec4 vid = texture(videoTex, videoCoord);
+	fragColor = mix(vid, col, col.a);
 #else
-	gl_FragColor = texture2D(tex, texCoord);
+	fragColor = texture(tex, texCoord);
 #endif
 }
--- a/share/shaders/default.vert
+++ b/share/shaders/default.vert
@@ -1,11 +1,11 @@
 uniform mat4 u_mvpMatrix;
 uniform vec3 texSize; // not used
 
-attribute vec4 a_position;
-attribute vec3 a_texCoord;
+in vec4 a_position;
+in vec3 a_texCoord;
 
-varying vec2 texCoord;
-varying vec2 videoCoord;
+out vec2 texCoord;
+out vec2 videoCoord;
 
 void main()
 {
--- a/share/shaders/fill.frag
+++ b/share/shaders/fill.frag
@@ -1,6 +1,8 @@
-varying vec4 v_color;
+in vec4 v_color;
+
+out vec4 fragColor;
 
 void main()
 {
-	gl_FragColor = v_color;
+	fragColor = v_color;
 }
--- a/share/shaders/fill.vert
+++ b/share/shaders/fill.vert
@@ -1,9 +1,9 @@
 uniform mat4 u_mvpMatrix;
 
-attribute vec4 a_position;
-attribute vec4 a_color;
+in vec4 a_position;
+in vec4 a_color;
 
-varying vec4 v_color;
+out vec4 v_color;
 
 void main()
 {
--- a/share/shaders/hq.frag
+++ b/share/shaders/hq.frag
@@ -4,35 +4,37 @@ uniform sampler2D offsetTex;
 uniform sampler2D weightTex;
 uniform sampler2D videoTex;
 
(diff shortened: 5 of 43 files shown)


m9710797:
Thanks!  I won't have time today or tomorrow, but I'll try to review and merge on monday or tuesday.

Just for my understanding a few quick questions (i didn't really look at the patches yet).
- One of the main changes is to use Vertex Buffer Objects, right? Or are there other openGL-3 features that are useful? (in the context of openMSX).
- I'm slightly confused by the way you numbered your patches. E.g. 1/9 sits between 9/9 and 7/9.
- Do you know if this change brings us closer or further away from openGL ES (2 or 3)? We once had the idea to use openGL ES for the android port.  (Although nobody stepped up to maintain this port, so I'm only asking out of curiosity, openGL ES compatibility is not a requirement for this PR).

MBilderbeek:
My 2 cents: yes, our long term plan was to move to OpenGL ES. So Question 3 is important, I think. I would strongly prefer not to push to newer OpenGL standards, but to move to a common OpenGL ES and stick to it for now.

D15C0DE:
1. As I said I haven't looked into optimizing the drawing code much further. The main change in OpenGL 3.3 is not just Vertex Buffer Objects, but keeping the Vertex Buffer Objects, as well as other state information, in Vertex Array Objects. This prevents you from having to upload data to the GPU all the time. There are some situations that currently still upload new data each frame, because of possible altered window dimensions and such, which with more work could be rewritten to use cached data.
2. That's just because github is ordering by date in the pull request display. The commits themselves are properly ordered.
3. This definitely brings us closer to OpenGL ES 3. It might already be compatible, but assessing that would require investigation.

m9710797:
Merged.
- I worked a bit further on removing SDL dependencies from the VisibleSurface and OutputSurface (sub)classes.
- I added an error check on creating the openGL context.
But apart from that I merged your changes unchanged.
This patch series was very clean and easy to follow. Thanks!

m9710797:
It seems that the first commit in this series ("SDLGL does not need renderer, surface and texture") had an unexpected consequence. It's not necessarily a good or bad change, just unexpected. I suspect it has something to do with the exact order in which stuff gets initialized.

But I don't have time right now to investigate. In order to not forget about it I'm pasting all info I have so far in this thread:

Here's an extract from a discussion on our IRC channel:
```
[13:33] <p_gimeno> I'm experiencing what seems to be a severe performance regression, I'm investigating.
[13:34] <p_gimeno> It seems to be introduced in this commit: https://github.com/openMSX/openMSX/commit/706793b5e27a012b71fcb02d208d225eef32fb47
[13:35] <p_gimeno> When I press the Throttle button, it runs at about 2x at most. Before that commit, it ran at maybe 50x or so.
[13:38] <p_gimeno> I can even type almost normally with the throttle active.
[13:48] <p_gimeno> The CPU stays at low % use, which suggests it may have to do with waiting for VSYNC
[13:49] <p_gimeno> previous to the regression, the CPU was at 100% when I disabled the throttle (which I expected)
[13:49] <p_gimeno> wouterv: can you reproduce?
[14:04] <wouterv> p_gimeno: that's unexpected.  I can't easily test right now. I'll try later (possibly only tomorrow or the day after)
[14:04] <p_gimeno> ok, thank you
[14:04] <wouterv> p_gimeno: does 'set maxframeskip 99' make any difference?
[14:05] <p_gimeno> yes it does
[14:06] <p_gimeno> it runs at 60% CPU and I can no longer type
[14:06] <wouterv> Can you try disabling vsync in your video drivers?
[14:07] <p_gimeno> uh... I think there was an environment variable for it but I can't remember which, let me look
[14:07] <wouterv> i'm using KDE and in the video settings I have 'Tearing prevention ("vsync")' set to 'Automatic' ... whatever that means
[14:08] <wouterv> But in any case.  I would not have expected that commit to have an influence on this.  So i'll investigate (later)
[14:09] <p_gimeno> found it, yes, disabling VSYNC makes it run at full speed
[14:10] <p_gimeno> (the setting in my case is __GL_SYNC_TO_VBLANK=0 in the environment)
[14:27] <p_gimeno> wouterv: yes, it's definitively that commit. The regression disappears as soon as I run the immediately previous commit, and reappears with that commit.
```

@D15C0DE : (Since you wrote that commit). Any idea?

D15C0DE:
Unfortunately I can't reproduce this; on my system before and after is the same (fast) speed.

Investigating the SDL source code I see that a SDL_HINT_RENDER_VSYNC exists, but it doesn't get enabled without an explicit call to SDL_SetHint which I can't find anywhere. This hint would cause the SDL_RENDER_PRESENTVSYNC flag to become enabled. Neither the default renderer created in SDL_InitVideo() nor the openMSX SDL_CreateRenderer() call enable this flag.

I've also investigated the SDL_CreateTexture() call, but it has nothing to do with VSYNC at all.

Concluding, I can't find any reason why there should be any difference in behavior, but SDL_CreateRenderer() looks the most suspicious. It looks like it's not required even for 2D, but maybe I'm overlooking something? If necessary this commit could be reverted as it's just a simplification.

PR #1220, opened 2020-02-02
Fix MIDI In not working under Windows

### LINKED ISSUE #999: MIDI-in on Windows not working
See comments on https://www.msx.org/forum/msx-talk/general-discussion/converting-cx5m-cassettes-cas-format?page=4 and https://www.msx.org/forum/msx-talk/general-discussion/converting-cx5m-cassettes-cas-format?page=5

The user plugged in a MIDI keyboard to the MIDI-in, but it wasn't working.

ValleyBell:
Formerly: Fix deadlock when plugging Windows MIDI-In device

When plugging a MIDI Input device into the MSX via
`plug msx-midi-in midi-in-0`
then the emulator will freeze under Windows.

What happens is:
- The main thread creates a worker thread in `MidiInWindows::plugHelper()`.
- The worker thread `MidiInWindows::run()` runs and calls `threadIdCond.notify_all();` before the main thread continues.
- The main thread eventually calls `threadIdCond.wait(threadIdLock);`, but the worker thread already issued the signal before the main thread started waiting for it, creating a deadlock.

Creating the thread while `threadIdMutex` is locked resolves the issue. (first commit)
The bug was introduced with ad85714b24a9c27fb23a5f60cc990e2ba847bec5.

Also, `midiInStart` was never called, thus it never started receiving events. (fixed by second commit)

fixes #999

### DIFF
--- a/src/serial/MidiInWindows.cc
+++ b/src/serial/MidiInWindows.cc
@@ -63,10 +63,10 @@ void MidiInWindows::plugHelper(Connector& connector_, EmuTime::param /*time*/)
 
 	setConnector(&connector_); // base class will do this in a moment,
 	                           // but thread already needs it
-	thread = std::thread([this]() { run(); });
 
 	{
 		std::unique_lock<std::mutex> threadIdLock(threadIdMutex);
+		thread = std::thread([this]() { run(); });
 		threadIdCond.wait(threadIdLock);
 	}
 	{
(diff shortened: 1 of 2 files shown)


m9710797:
I couldn't compile/test this myself. But the fix seems plausible.
Thanks a lot!

PR #1223, opened 2020-02-07
fix Windows MIDI output

ValleyBell:
Windows MIDI used a custom MIDI byte stream parser, which failed to
handle the "running status" correctly.
This bug is fixed by using the parser from MidiOutDevice. It also
greatly simplifies the Windows MIDI logic.


m9710797:
I couldn't compiler nor test this myself. But I trust you did.
Also it's a very nice cleanup.
Thanks!

PR #1230, opened 2020-02-28
Added icon/button to reverse bar for viewonlymode

FiXato:
This clickable icon to the right of the reverse bar will display the current 'reverse viewonlymode' status, and allow you to toggle between both states.

This change is part of my suggestion to a discussion on https://www.msx.org/forum/msx-talk/openmsx/how-to-avoidstop-video-recording#comment-375527. 

While this change doesn't change the default setting of `reverse viewonlymode` from `false` to `true`, it at least provides a visual indication of its current state.

The icons are quick images created in Inkscape as SVG, and exported to PNG, and could probably be decreased in size, though I also don't mind them being replaced by something else; I just didn't feel like looking for images that suit the project's license.
Source SVGs can also be added if desired.

I wasn't sure how to implement these icons in a way that they actually support different skins either, so suggestions for that are welcome too.


FiXato:
### Relevant screenshots
**`reverse viewonlymode false`** 
image

**`reverse viewonlymode true`** 
image

MBilderbeek:
So, do you think the extra icon is necessary? The color of the reverse bar also already indicates the viewonly status, as you can see in these screenshots.

MBilderbeek:
Ah wait, it's not an indicator, it's a way to switch view only mode.

Would be quite a change of defaults. I'm very used to be able to reverse to fix mistakes.
So changing to default 'viewonly mode true' is quite a backwards incompatible change. But it may still be worth it. What do you think @m9710797?

m9710797:
I like this idea.

I can understand the pain of people loosing work by accidentally clicking in the reverse-bar. It never happened to me, but I also rarely use MSX software that uses the mouse.
We already have a solution/workaround for this: either disable the reverse bar (or even the whole reverse feature) or put replay in view-only mode. But apparently not many people are aware of these solutions.
This patch helps a lot in the 'discoverability' of view-only mode. I like it.

Now about the default value for view-only mode. Personally I'd use viewonly=false, but I don't know if that's also the best default for other users. Maybe we could make view-only mode a persistent setting (I mean so that it's kept when you exit/restart openMSX, currently openMSX always starts with viewonly=false). Then users only have to pick their preferred value once and we have to worry less about the best default value.

FiXato:
@MBilderbeek wrote:
> So, do you think the extra icon is necessary? The color of the reverse bar also already indicates the viewonly status, as you can see in these screenshots.

To be honest, I hadn't actually noticed the colour change in the bar. Well, I notice it changes colour between 'replaying' and 'recording' (which is what is primarily displayed in the screenshot)

But while **replaying with `viewonlymode=true`:**
image

versus **replaying with `viewonlymode=true`:**
image

you'd have to know that seagreen is when it's 'locked' and blue is 'unlocked', and I can't say I ever noticed nor realised that difference. :)

While it is at the end of the reverse and thus recording rather than replaying, there actually is no such colour difference (though I guess that theoretically it doesn't apply while it isn't replaying anyway).
**`viewonlymode=true` while recording:**
image

**`viewonlymode=false` while recording:**
image


Now as for the defaults, I'm also quite used to it being 'unlocked' by default, but I definitely can see the frustration from people who aren't interested in the reverse feature, and 'just want to play a game'.

On the one hand the current defaults allow for discoverability of an (imho quite handy) feature they perhaps otherwise wouldn't be aware of.
On the other hand it's quite a unique and thus not easily recognised feature by non-power users, especially those that are just looking for a semi-authentic quick-fix of MSX nostalgia, which apparently can cause frustration.

I think that adding an icon/button such as this, combined with setting the default to `viewonlymode=true` (locked), and @m9710797's suggestion of making that setting persistent, would be a worthwhile compromise.

We'd still have the discoverability of the reverse bar being visible and clickable, but those unfamiliar with it will at least not as easily lose progress accidentally. Power users will know how to re-enable it anyway, and having a (recognisable?) visible persistent toggle should make it easy enough for those that aren't.

Two more notes:

1. might it be an idea to automatically set viewonlymode=true when a virtual mouse is connected to the emulated machine, as it seems you can't really interact with a mouse on the reverse bar anyway without truncating? If a power user needs to truncate after interacting with the reverse bar, they can just bind `reverse truncatereplay` to a key?

2. Do settings start off as set? Or do they also have an unset state?
If they have an unset state, we could actually detect that they are still default, and pop up a short OSD wizard perhaps that runs the users through the most common/impactful settings (auto-enable reverse and/or its bar, setting viewonlymode, accuracy, resampler, scalefactor, scale algorithm, etc) on first run?
Or perhaps introduce a wizard.has_run boolean (or perhaps even version, so new wizard options can still be introduced to the user at a later point?) setting which is set once the wizard has run?
To be as little in the way of users familiar with openMSX, it should of course be possible to accept all current/default settings right from the start.
If the idea of a wizard is interesting, I can open a separate issue for it in which we could discuss this in more detail.

FiXato:
Also, would it perhaps be a good idea to ask for further input here from the users at https://www.msx.org/forum/msx-talk/openmsx/how-to-avoidstop-video-recording who actually have issues with the current defaults?

MBilderbeek:
My  main doubt: the whole value of the reverse feature in gaming is that you can redo parts of games where you made a mistake. For that the viewonly mode needs to be off. So, by turning viewonly ON by default, actually renders the main use case inoperable... Wouldn't it then be better to turn off the whole feature by default?

FiXato:
Sure it is a big part of it, but I wouldn't call it the whole value; after all, it's also very useful for recording the entire replay or just parts of it.
While playing games such as RPGs, it can also be used to quick look back at conversation to for instance see if you missed a hint, without having to walk all the way back.
Or for instance when watching a demo, to rewind a bit to watch a specific section again.

PR #1231, opened 2020-02-28
removed HTML from save_msx_screen help

FiXato:
The help_text for save_msx_screen contained an HTML anchor tag, which AFAIK has no use in the tcl code?

Stripped the HTML and left only the URL.


MBilderbeek:
Thanks!

Issue #1229, opened 2020-02-27
Possibility to disable dosbox window with status info without use catapult?


ghost:
When starting the Windows build of OpenMSX it always shows a dosbox window with status info and the emulator itself. I would like to disable the dosbox window. Is this possible? I don't use Catapult, but the build-in OSD of OpenMSX.

mthuurne:
It seems to be possible to suppress the creation of the console window by [selecting the `WINDOWS` subsystem](https://docs.microsoft.com/en-us/cpp/build/reference/subsystem-specify-subsystem).

What I don't know though is how openMSX would then behave when started from the command line. We wouldn't want to lose our logging, since that would make troubleshooting very difficult.

MBilderbeek:
For Android we also use something special for the logging, as there's no terminal there. See main.cc:
```
// Set LOG_TO_FILE to 1 for any platform on which stdout and stderr must
// be redirected to a file
// Also, specify the appropriate file names, depending on the platform conventions
#if PLATFORM_ANDROID
#define LOG_TO_FILE 1
#define STDOUT_LOG_FILE_NAME "openmsx_system/openmsx.stdout"
#define STDERR_LOG_FILE_NAME "openmsx_system/openmsx.stderr"
#else
#define LOG_TO_FILE 0
#endif
```

MBilderbeek:
Oops, accidentally closed :)

cmitu:
It's possible to start the console window even when selecting the WINDOWS subsystem during compilation, toggled by a command line option (i.e. `--debug`, `--log`). An example is [here](https://github.com/RetroPie/EmulationStation/blob/55df3d833235976585d549d0a0548244ab838ebf/es-app/src/main.cpp#L258-L288).

ghost:
I close this issue. Maybe it can be done in a future release for Windows.
I now use Android version instead which doesn't have this console window.

MBilderbeek:
Please don't just close issues. 
Anyway, I couldn't find back this issue, but it has actually been solved by @peterhanraets via a recent set of commits. 
See 7f73527b0d964d1af08c4a0d5ea2ce14c8e7b9df, 9986d9a5ef76270c76c27c236397327b2894abac, 2ace7fd5401f5191df1f23a237552a03b97e447c, 20d0bae9c20f883c07e0d0077684fa934a0b5154, 6ad65ef242e7fe3695bc73208687b81ce7c3e76a for example.

PR #1263, opened 2020-05-20
Disable the console

Anths2016:
A option to disable the console like Nulldc?


mthuurne:
Eh, this pull requests contains 4000 commits... Is there a relevant commit between them or did you accidentally open a pull request instead of an issue?

Which console do you mean, an OS terminal or the openMSX OSD console? In the former case, I'm assuming this is about Windows, in which case the question is already covered in #1229. If not, please open a new issue in the tracker.

I'm closing this pull request now.

PR #1268, opened 2020-06-01
Fix asserts on string_view() on MSVC 2019

wbahnassi-if:
Currently string_view operations (like remove_prefix) manipulate string_view::_Mydata and _Mysize, yet the subtract operator (-) asserts that both operands share the same _Mydata and _Mysize, and it only subtracts on _Myoff, which is never manipulated actually. If we just ignore the assert we're sure to get no tokens. This fix avoids using the string_view subtract operator altogether and counts the tokenLength manually instead.

### DIFF
--- a/src/input/UnicodeKeymap.cc
+++ b/src/input/UnicodeKeymap.cc
@@ -76,11 +76,13 @@ static string_view nextToken(string_view& str)
 {
 	skipSep(str);
 	auto tokenBegin = str.begin();
+	size_t tokenLength = 0;
 	while (!str.empty() && str.front() != '\n' && !isSep(str.front())) {
 		// Pop non-separator character.
 		str.remove_prefix(1);
+		tokenLength++;
 	}
-	return string_view(&*tokenBegin, str.begin() - tokenBegin);
+	return tokenLength ? string_view(&*tokenBegin, tokenLength) : string_view();
 }
 
 


wbahnassi-if:
FYI, here is the VS2019 implementation of stringview::operator-
`_NODISCARD constexpr difference_type operator-(const _String_view_iterator& _Right) const noexcept
{
#if _ITERATOR_DEBUG_LEVEL >= 1
        _STL_VERIFY(_Mydata == _Right._Mydata && _Mysize == _Right._Mysize,
            "cannot subtract incompatible string_view iterators");
        return static_cast<difference_type>(_Myoff - _Right._Myoff);
#else // ^^^ _ITERATOR_DEBUG_LEVEL >= 1 ^^^ // vvv _ITERATOR_DEBUG_LEVEL == 0 vvv
        return _Myptr - _Right._Myptr;
#endif // _ITERATOR_DEBUG_LEVEL
}`

And here is the implementation of stringview::remove_prefix:
` constexpr void remove_prefix(const size_type _Count) noexcept /* strengthened */
{
#if _CONTAINER_DEBUG_LEVEL > 0
        _STL_VERIFY(_Mysize >= _Count, "cannot remove prefix longer than total size");
#endif // _CONTAINER_DEBUG_LEVEL > 0
        _Mydata += _Count;
        _Mysize -= _Count;
}`

Clearly the two functions don't agree. One expects changes to _Myoff only, and the second leaves _Myoff but changes _Mydata and _Mysize. The assert is fired by STL_VERIFY.

m9710797:
I've pushed an alternative implementation of this PR. But I couldn't test this myself on msvc.
@wbahnassi-if: Could you check if this patch also fixes the original problem?
Thanks!

m9710797:
wbahnassi-if confirmed the issue is fixed. I'm closing this PR.

PR #1275, opened 2020-07-04
Fix very broken Meson build

chewi:
Some required sources were just missing. Other conditional sources hadn't been made conditional.

I'm really glad you've started using Meson and I understand it's a work in progress but you might want to test it once in a while. :stuck_out_tongue: I don't think it's worked since at least April.


mthuurne:
Thanks for the patch!

The Meson build is still experimental and isn't part of our CI infrastructure yet, so it easily gets out of date. After the 16.0 release I want to resume working on it and I'll integrate your patch then.

PR #1285, opened 2020-08-29
build: Fix unmountable DMG images on macOS < 10.13

grauw:
The default either changed to APFS or depends on the local file system.
Since APFS is not supported on macOS versions before 10.13, we tell
hdiutil explicitly to use the HFS+J file system.

Pending confirmation by one of the affected users here:
https://www.msx.org/news/en/openmsx-160-released#comment-387833

You can use the freshly built DMG here for distribution on the openmsx.org:
http://www.grauw.nl/etc/msx/openmsx-16.0-mac-x86_64-bin.dmg


grauw:
gdx> “Thank you Grauw! You have found the solution, your image works.”

@MBilderbeek

MBilderbeek:
Thanks, I replaced the binary on the site with this one and merged the PR.

PR #1289, opened 2020-10-02
Automatically msx_update widgets msx_init created

FiXato:
Trace changes to ::horizontal_update, as well as watch writes to REG18SAV memory address 0xFFF1.
Once detected, call ::osd_widgets::msx_update on each OSD widget MSX layer initialised by ::osd_widget::msx_init.

This should make sure that OSD widgets that should be relative to the MSX screen, stay properly positioned even when `set horizontal_stretch` is called from openMSX, or `set adjust(x,y)` is called from the MSX.

Theoretically changes to the amount of MSX lines (vdpreg 9) could also be traced I guess, but I couldn't think of a good way to test this, or implement. It's probably also best to be left to future script writers who actually run into the issue.


m9710797:
I merged this patch and applied some small tweaks.
Thanks!

PR #1304, opened 2020-12-04
Updated workflows for Linux and Windows to use Ubuntu 20.04 to suppor…

sdsnatcher73:
…t std c++ 17


m9710797:
Thanks!
One clarification: the upgrade is needed for gcc-8 (which has more complete c++17 support).

PR #1305, opened 2020-12-11
Set macOS target to 10.13, and reimplement FileOperations in Objective-C without deprecated APIs

grauw:
Upgrade macOS target version to 10.13 High Sierra. This version supports macs from 2009/2010.

Add Objective-C compilation support. I’m not sure the makefile rules in main.mk work cross-platform; the .mm files should only be compiled on macOS. Maybe someone with stronger mk-fu can take a look at this.

Lastly, reimplement FileOperations::findShareDir() in Objective-C. The old implementation uses functions which were deprecated in macOS 10.8 and 10.9. This serves both to eliminate those deprecation warnings, as well as be a first user for the Objective-C support. It’s also significantly more compact.

In the future the Objective-C support is useful to add e.g. Metal and touch bar support.

### DIFF
--- a/build/main.mk
+++ b/build/main.mk
@@ -256,6 +256,9 @@ endif # goal requires dependencies
 SOURCE_DIRS:=$(sort $(shell find src -type d))
 
 SOURCES_FULL:=$(foreach dir,$(SOURCE_DIRS),$(sort $(wildcard $(dir)/*.cc)))
+ifeq ($(OPENMSX_TARGET_OS),darwin)
+SOURCES_FULL+=$(foreach dir,$(SOURCE_DIRS),$(sort $(wildcard $(dir)/*.mm)))
+endif
 SOURCES_FULL:=$(filter-out %Test.cc,$(SOURCES_FULL))
 
 # TODO: This doesn't work since MAX_SCALE_FACTOR is not a Make variable,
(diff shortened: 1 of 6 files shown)


grauw:
The macOS build failed because `TARGET_FLAGS+=-fobjc-arc` (enable Objective-C automatic reference counting) is not only applied to the openMSX source code but also seems to propagate to the SDL2 build, which uses manual reference counting.

The Windows and Linux builds indeed seem to fail on the inclusion of .mm files in their build process. Only Clang supports Objective-C, and the frameworks are only available on macOS.

Some mk-fu is needed indeed 🙂.

grauw:
Build issues resolved.

grauw:
For additional context and future reference;

When OSX_VER is set to 10.13 in the first change, some deprecation warnings surface:

> src/file/FileOperations.cc:87:6: warning: 'GetCurrentProcess' is deprecated: first deprecated in macOS 10.9
> src/file/FileOperations.cc:91:6: warning: 'GetProcessBundleLocation' is deprecated: first deprecated in macOS 10.9
> src/file/FileOperations.cc:97:6: warning: 'FSGetCatalogInfo' is deprecated: first deprecated in macOS 10.8
> src/file/FileOperations.cc:106:6: warning: 'FSGetVolumeInfo' is deprecated: first deprecated in macOS 10.8
> src/file/FileOperations.cc:119:7: warning: 'FSOpenIterator' is deprecated: first deprecated in macOS 10.8
> src/file/FileOperations.cc:130:16: warning: 'FSGetCatalogInfoBulk' is deprecated: first deprecated in macOS 10.8
> src/file/FileOperations.cc:162:24: warning: 'FSCloseIterator' is deprecated: first deprecated in macOS 10.8
> src/file/FileOperations.cc:166:11: warning: 'FSRefMakePath' is deprecated: first deprecated in macOS 10.8
> src/file/FileOperations.cc:176:20: warning: 'FSCloseIterator' is deprecated: first deprecated in macOS 10.8
> src/file/FileOperations.cc:179:7: warning: 'FSCompareFSRefs' is deprecated: first deprecated in macOS 10.8
> src/file/FileOperations.cc:183:7: warning: 'FSGetCatalogInfo' is deprecated: first deprecated in macOS 10.8

These are fixed by the new Objective-C implementation.

I chose Objective-C + Foundation over C + Core Foundation because the former exposes more functionality and is safer to use wrt reference counting. And looking at the long term, using Objective-C to interface with macOS APIs is the best and often the only way, most things are not exposed through C APIs.

PR #1317, opened 2021-01-09
Updated Sony_HB-G900D.xml 

sdsnatcher73:
With disk rom shasum where register area wa…s cleaned, same ROM was used in F700D. When I originally dumped this I did not know about the registers...

### DIFF
--- a/share/machines/Sony_HB-G900D.xml
+++ b/share/machines/Sony_HB-G900D.xml
@@ -77,7 +77,7 @@
           <broken_fdc_read>false</broken_fdc_read>
           <drives>1</drives>
           <rom>
-            <sha1>0e081572f84555dc13bdb0c7044a19d6c164d985</sha1>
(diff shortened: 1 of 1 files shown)


m9710797:
Thanks for your PR. Though I wonder if instead of replacing the 'sha1' value, we should add the new value as an alternative? Possibly the primary alternative. (For backwards compatibility, to not break (released?) machine configs that were working before.)

sdsnatcher73:
I don't think it will break the config for anyone. The rom that is used now is used in the following machine configs already:

Sony_HB-F1XD.xml
Sony_HB-F1XDJ.xml
Sony_HB-F700D.xml
Sony_HB-G900AP.xml
Sony_HB-G900P.xml

m9710797:
Ok, that's a convincing argument. Merged.
Thanks.

PR #1319, opened 2021-01-17
Updated workflows to use environment variables and added aarch64 builds for macOS and Linux

sdsnatcher73:


MBilderbeek:
Thanks for the contribution! The checks were all successful, so I guess it's OK to merge this :)

Issue #1325, opened 2021-02-06
Make openmsx.exe a GUI application (but still allow --help to output to a calling console)


sdsnatcher73:
openMSX on Windows is built as a console application instead of a windowed application. Not sure about why this was done, I suspect to be able to print the --help output. On macOS and on Linux starting the application from whatever menu is present just brings up the openMSX window. On Windows it brings up a Command Prompt window as well (which I personally find annoying, if I want console output I can start it from the Command Prompt / Terminal as on other OS's). Either way it is inconsistent with other Desktop OS platforms supported by openMSX.

With a minor change to platform-mingw32.mk (and probably openmsx.vcproj) the resulting executable is no longer a console applictation. With a small addition to main.cc we can attach to the calling console (if present) and print to it, this way openmsx.exe --help will work.

I will create a pull request for the mingw builds, for the visual studio build it will probably be sufficient to change <SubSystem>Console</SubSystem> to <SubSystem>Windows</SubSystem>.

MBilderbeek:
As we merged the pull request, I guess we can close this issue :) Thanks!

PR #1326, opened 2021-02-06
Peter Hanraets Make openmsx.exe a GUI application (but still allow --help to output

sdsnatcher73:


sdsnatcher73:
Implements #1325 for mingw builds, if accepted a change shoiuld be made to the visual studio build.

MBilderbeek:
Are you also going to submit a patch for the VS build? (Which is the main/mainstream build for Windows...)

sdsnatcher73:
I am trying to build the Visual Studio solution but am running into issue. Started with VS Community 2019 that had issues with the zlib build. So I thought maybe I need 2017 but that seems to old as it returns error MSB8020: The build tools for v142 (Platform Toolset = 'v142') cannot be found. Still figuring this out...

sdsnatcher73:
Okay it was just a huge list of warnings...

sdsnatcher73:
msvc build also updated

sdsnatcher73:
I changed the Subsystem from Console to Windows from within the IDE (I tried first manually with VSCode but it said the root structure was not valid or something like that). The change in the IDE also did the changes to the filters file.

sdsnatcher73:
I don't understand why there are so many duplicates in that file, I guess optimisation is not a priority on MS's shortlist...

MBilderbeek (approved):
OK, let's just see. We can always (partially) revert things if they don't work out as we expected.

sdsnatcher73:
Yeah makes sense. if anything pops up you'll know where to find me ;)

PR #1328, opened 2021-02-17
Properly fix drawing of OSD progress info

D15C0DE:
The original solution in b6fd832980 was to call repaintDelayed with time zero, but that had unintended side-effects. The partial 'fix' of ca95ac84 broke the actual intention of b6fd832980. This patch properly fixes both commits.

### DIFF
--- a/src/video/Display.cc
+++ b/src/video/Display.cc
@@ -176,7 +176,7 @@ Display::Layers::iterator Display::baseLayer()
 
 void Display::executeRT()
 {
-	videoSystem->repaint();
+	repaint();
 }
 
 int Display::signalEvent(const std::shared_ptr<const Event>& event) noexcept
@@ -184,7 +184,7 @@ int Display::signalEvent(const std::shared_ptr<const Event>& event) noexcept
 	if (event->getType() == OPENMSX_FINISH_FRAME_EVENT) {
 		const auto& ffe = checked_cast<const FinishFrameEvent&>(*event);
 		if (ffe.needRender()) {
-			videoSystem->repaint();
+			repaint();
 			reactor.getEventDistributor().distributeEvent(
 				std::make_shared<SimpleEvent>(
 					OPENMSX_FRAME_DRAWN_EVENT));
@@ -212,7 +212,7 @@ int Display::signalEvent(const std::shared_ptr<const Event>& event) noexcept
 		//  port discovers that the graphics context is gone.
 		// -When gaining the focus, this repaint does nothing as
 		//  the renderFrozen flag is still false
-		videoSystem->repaint();
+		repaint();
 		const auto& focusEvent = checked_cast<const FocusEvent&>(*event);
 		ad_printf("Setting renderFrozen to %d", !focusEvent.getGain());
 		renderFrozen = !focusEvent.getGain();
@@ -325,7 +325,7 @@ void Display::doRendererSwitch2()
 	}
 }
 
-void Display::repaint()
+void Display::repaintImpl()
 {
 	if (switchInProgress) {
 		// The checkRendererSwitch() method will queue a
@@ -343,7 +343,7 @@ void Display::repaint()
 	if (!renderFrozen) {
 		assert(videoSystem);
 		if (OutputSurface* surface = videoSystem->getOutputSurface()) {
-			repaint(*surface);
+			repaintImpl(*surface);
 			videoSystem->flush();
 		}
 	}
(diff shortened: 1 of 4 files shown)


m9710797:
Hi, thanks for you patch. But I have some comments.

I understand that you need this to make integration with a GUI toolkit easier. But nothing in the openMSX code documents that dependency. And without that extra knowledge your patch doesn't make sense.

More specifically:
* Your patch changes (some/all?) calls to Display::repaint() into Display::repaintNow().
* The implementation of DisplayRepaintNow() (only) calls VideoSystem::repaint().
* And VideoSystem::repaint() (only) calls Display::repaint(), the original method.
So, your patch adds a detour via VideoSystem, but there's nothing in the code explaining why that detour is needed. Because functionally it really does nothing.

I'm also left wondering if all or only some calls to Display::repaint() should be transformed to Display::repaintNow(). In other words: is it a bug if some code still calls Display::repaint() directly. If so, then it's a good idea to rename this method, and possibly make it private (and make VideoSystem a friend). So make the interface impossible to use wrong.

I also wonder why you've chosen this detour approach. I assume you want to do extra work in VideoSystem::repaint() in your integration. But wouldn't it be simpler (as in a smaller diff compared to openMSX:master) to _only_ add a call to a new function VideoSystem::initialize() as the first line in the implementation of Display::repaint()?   (Just a guess because I can't see your code).

So in summary: I can't accept this patch because in isolation it doesn't make sense (without extra, now hidden, knowledge). At the very least you need to add comments explaining why this weird detour is needed. And explain how the interface should be used correctly (ideally make it impossible to misuse).

D15C0DE:
Hi!

Display.hh already has this comment that documents that repaint() should only be called by the VideoSystem:
	/** Redraw the display.
	  * repaint() should only be called from the VideoSystem.
	  */
Maybe it's too terse? In any case it's not hidden as you allege.

There are no other calls to Display::repaint() and indeed it is a bug to call Display::repaint() directly. Your suggestion of making it private and making VideoSystem a friend crossed my mind, but I thought that would be too invasive. I wanted to stay close to my original patch which was already accepted, where all Display::repaint() were changed to Display::repaintDelayed(0) instead. (Which turned out to be a bad idea, sorry for that!)

The "detour approach" is necessary because, when using the GUI renderer, VideoSystem::repaint() only queues a rendering. The actual call to Display::repaint() then happens in a render signal callback. It's a really really small detour that has no difference in performance compared to your non-working suggestion (one extra virtual method call).

As you should already know, the point of this patch and many of my previous (and upcoming) patches is to move specifics of the SDL implementation out of various classes and consolidate them in the VideoSystem. I realize this patch seems unnecessary for the current openMSX code that depends on the specific workings of SDL, but the only alternative would be for me to push half-broken work-in-progress stuff that will be in flux as I work on it. I don't think that's desirable or reasonable.

[rant] Instead I'm left to wonder why there is so much resistance against abstracting away any SDL specifics! I'm doing you all a great favor by doing this work, and the thanks I get is that my work, after being accepted, eventually gets partially yet significantly reverted without any communication or discussion! And now when I provide a proper fix to **your botched hot-fix**, it's not accepted because you want to stay as close as possible to the current openMSX:master that's so highly dependent on the (inner) workings of SDL!
Do you even want contributions? Given all the extra work I have to do to repair the damage you're mindlessly and disrespectfully doing to my earlier work, I'm starting to think that it would be more economic for me to properly fork openMSX as a separate project and see how many users you will be left with while you constantly blame SDL for actual user problems and discard user calls for a proper GUI while again blaming SDL. Stockholm syndrome anyone? [/rant]

D15C0DE:
To be perfectly clear: what you complain about is already in openMSX:master. This patch only corrects the inconsistency that you yourself created by only partially reverting the original patch. Blocking this patch by complaining about an earlier patch is utterly ridiculous.

m9710797:
The only reason I ask for extra changes is to avoid unintentionally breaking your work again in the future. Your integration adds extra constraints on the code, but currently those constraints are invisible in our code.

These changes could be minimal. I propose to:
- Rename Display::repaint() and Display::repaint(OutputSurface&) to something like 'repaintImpl()' (even without making it private/friend). That should already make it clearer that these renamed methods are implementation details and shouldn't be called directly.
- Add some comments in SDLVideoSystem::repaint() that explain why this method exists (why it can't be replaced with a direct call to Display::repaint()). Your comment that in your integration this method only queues rendering is new information to me, and already very useful.

Will you update your PR or shall I make these proposed changes? Though it would still be useful if you could write the comment for SDLVideoSystem::repaint(), maybe add a link to the GUI toolkit documentation or even to your work-in-progress source code.

D15C0DE:
Ok, the repaintImpl() rename cleans up nicely, thanks for the suggestion. I've sprinkled some comments here and there which I hope are sufficient.

m9710797:
Thanks.

PR #1330, opened 2021-02-18
Get mouse cursor info via VideoSystem

D15C0DE:
This abstracts away the last SDL dependency in the GL execution path for OSDWidget.

### DIFF
--- a/src/console/OSDWidget.cc
+++ b/src/console/OSDWidget.cc
@@ -1,6 +1,7 @@
 #include "OSDWidget.hh"
 #include "SDLOutputSurface.hh"
 #include "Display.hh"
+#include "VideoSystem.hh"
 #include "CommandException.hh"
 #include "TclObject.hh"
 #include "GLUtil.hh"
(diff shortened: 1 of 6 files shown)


m9710797:
Thanks. Look good at first glance. I'll probably merge tomorrow when I have more time.

I really only looked at this for a few minutes, but these questions came to mind:

VideoSystem::getMouseCoord() returns floating point coordinates, but typically mouse coordinates are integers. I know that in OSDWidget we need them as floating point. But maybe for an abstraction layer it makes sense to stick with the traditional types? So return a ivec2 instead of vec2?   (And ivec2 and easily be converted to vec2 in OSDWidget).

Would it make sense to combine VideoSystem::getCursorEnabled() and VideoSystem::getMouseCoord() into a single call? Maybe getMouseCoord() that returns a std::optional<vec2>?  Probably not: OSDWidget specifically only needs the coordinates when the cursor is visible, but in general these two should remain independent.

In src/input/Mouse.cc there's another call to SDL_GetMouseState(). Not for the coordinates but for the button-press status. Should that also be abstracted? In a combined coordinate+button query (like in SDL) or as separate calls?

I'll do a proper review tomorrow, and think about my own questions some more.
But in general I like this patch.

D15C0DE:
Good points.

Using ivec2 makes sense. I've made this modification.

Actually I don't know if the SDL_GetMouseState() call in Mouse.cc needs abstraction. I suppose it does! It might be an idea to have it listen to the mouse button up/down events? I think that's something for a separate patch though.

m9710797:
Thanks.
Merged with minor edits.

PR #1332, opened 2021-02-20
Handle clipboard via VideoSystem

D15C0DE:


D15C0DE:
After this there are only three things left that need abstraction:
- SDL_GetMouseState() in Mouse.cc
- SDL_GetModState() in Keyboard.cc
- input grabbing in InputEventGenerator.cc

Any thoughts on how to approach those?

m9710797:
Looks good, except for the 'assert(display)' in {Get,Set}ClipboardCommand. Probably a leftover from some earlier version? I haven't actually tried, but think this will cause compilation errors.
(Detail: I think it makes sense to store 'Display&' instead of 'Reactor&' in these commands).

I'll merge your patch later (tomorrow or the day after).

I don't have an immediate answer to your other questions. But I'll check the code and think about it.

Thanks.

D15C0DE:
assert(display) is used in Reactor::getDisplay() actually, but because it's there these calls are extraneous. I didn't know if it was safe to getDisplay(), but it appears to work fine so I modified it.

m9710797:
Found some time anyway. Merged but with some bug fixes.
My suggestion to store Display instead of Reactor was wrong, sorry about that.

Could it be that you're testing without assertions enabled?
That would explain why you didn't see the compilation error in the first version of your patch. And the crash at startup in your second version. Though it would still have crashed when you would actually execute the 'set_clipboard_text' command.

PR #1338, opened 2021-03-27
Updated code to attach stderr as well as stdout to an existing console

### LINKED ISSUE #1337: No error message is shown when selecting a non-existing machine
* openMSX version: 16.0-412-gf6ee2a4fc
* OS: Windows 7 X64
* Description of the issue: if openMSX is invoked via the command line and a machine whose configuration does not exist is selected, nothing happens. The emulator does not boot, and no error message is shown.

image

sdsnatcher73:
This fixes #1337

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -58,10 +58,8 @@ int setenv(const char *name, const char *value, int overwrite)
 void EnableConsoleOutput()
 {
     if (AttachConsole(ATTACH_PARENT_PROCESS)) {
-        FILE* pCout;
-        freopen_s(&pCout, "CONOUT$", "w", stdout);
-        std::cout.clear();
-        std::wcout.clear();
+        freopen("CONOUT$", "w", stdout);
+        freopen("CONOUT$", "w", stderr);
     }
 }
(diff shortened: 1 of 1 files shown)


m9710797:
This patch seems to do more than only attaching stderr.
It also changes the way how stdout is attached. And it removes stuff related to std::cout and std::wcout.
(If correct) this is a nice cleanup .. but just checking that this was indeed the intention.

sdsnatcher73:
When looking for how to also connect stderr to console I found this page:

https://stackoverflow.com/questions/54536/win32-gui-app-that-writes-usage-text-to-stdout-when-invoked-as-app-exe-help/113032

The solution there is indeed shorter than what I 'borrowed' earlier (from a similar page ;)

I tested and -help still works

m9710797:
Thanks.

MBilderbeek:
Does catapult still work as well? It calls openMSX with parameters to find
out which mapper types and machines exist.

Op za 27 mrt. 2021 10:49 schreef Wouter Vermaelen ***@***.***

PR #1353, opened 2021-04-17
Renamed National FS-4600 and FS-4700 to include the F behind their names 

sdsnatcher73:
The FS-4600F and FS-4700F both had floppy drives where the FS-4500 did not. The added F likely indicates this.
The FS-5000's official name is also FS-5000F2.


MBilderbeek:
@peterhanraets Are you sure about this? The pictures here: http://usbsecretbase.michikusa.jp/fs4600f_fs4700/index.html show that the FS-4700 has no F behind it. (But the FS-4600F does.) Although, perhaps on the box it does have FS-4700F?

For FS-5500 it's also without the F1/F2 on the front http://usbsecretbase.michikusa.jp/fs5500/index.html - but on the type plate there is the detail of F1 on this site.

So, the question is: do we use the name as on the machine? As on the box? As on the type-plate? They seem to conflict in these cases.

sdsnatcher73:
Here is the type plate of the FS-4700F: 


Here of the FS-5000F2:


The branding in these machines on the front indicates the overall model. National may have considered versions without floppy drive (like with the FS-4500, where there is actually place for the disk drive in the case, see this picture:


Why they produced an FS-5500F1 AND F2 but only an FS-5000F2 we won’t know (or maybe at some point we’ll find an F1). But at the moment openMSX is just inconsistent on these machines and because we have F1 and F2 for the 5500 (and you can’t really leave it out there) I rhink we should do it for all.

Another example is of course the NMS8255 which also showed NMS8250 on the front but NMS8255 on the label.

MBilderbeek (commented):
Strange, this commit on the FS-5000 doesn't show the original file... Did you perhaps only add the new file and not rename the original?

PR #1364, opened 2021-07-07
The scope of the variable  `X` can be reduced.

ZanyXDev:
L242,The scope of the variable 'slot' can be reduced.
L330,The scope of the variable 'channel' can be reduced.
L468,The scope of the variable 'rm_bit' can be reduced.
L838,The scope of the variable 'sign' can be reduced.
L842,The scope of the variable 'mute_r' can be reduced.
L1022,The scope of the variable 'vib_step' can be reduced.

### DIFF
--- a/src/sound/opll.cc
+++ b/src/sound/opll.cc
@@ -239,9 +239,8 @@ static void OPLL_DoIO(opll_t *chip) {
 }
 
 static void OPLL_DoModeWrite(opll_t *chip) {
-    uint8_t slot;
     if ((chip->write_mode_address & 0x10) && chip->write_d_en) {
-        slot = chip->write_mode_address & 0x01;
+        uint8_t slot = chip->write_mode_address & 0x01;
         switch (chip->write_mode_address & 0x0f) {
         case 0x00:
         case 0x01:
@@ -327,7 +326,6 @@ void OPLL_Reset(opll_t *chip, uint32_t chip_type) {
 }
 
 static void OPLL_DoRegWrite(opll_t *chip) {
-    uint32_t channel;
 
     /* Address */
     if (chip->write_a_en) {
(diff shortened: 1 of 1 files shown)


m9710797:
Hi ZanyXDev,

Thanks for your PR. I fully agree with the kind of changes you made ...
... however, the file 'src/sound/opll.cc' is a bit special. That file is as much as possible the same as the upstream version:
   https://github.com/nukeykt/Nuked-OPLL/blob/master/opll.c

In openMSX we made a re-implementation of 'opll.c', you can find that implementation in 'src/sound/YM2413NukeYKT.cc'. That new code should generate 100% identical output, but run quite a bit faster. I also checked in the original code to be able to verify that it is indeed still generates identical output.

So although I fully agree your PR improves the code, I still prefer to keep this code as is. Maybe you could propose your patch for 'https://github.com/nukeykt/Nuked-OPLL'?

For now I'm closing this PR, but thanks anyway.

How did you find these issues? Via some static code analysis tool? Found any issues in other files?

Wouter

ZanyXDev:
I'm use
` /usr/bin/cppcheck  -q --enable=warning,style,performance,portability,information --inconclusive --template={severity}:{id},{file},{line},{message} --std=c++11 --language=c++  ./ 2> ./err.txt`
[err.txt](https://github.com/openMSX/openMSX/files/6782953/err.txt)

MBilderbeek:
Here's the outout with --std=c++17 and filtering out the 'names different' warnings....
[err_filtered.txt](https://github.com/openMSX/openMSX/files/6787107/err_filtered.txt)
@m9710797 Is it useful?

ZanyXDev:
https://github.com/nukeykt/Nuked-OPLL/commit/5d287db36d58eab6edc157472a262eb1ccb9b387

m9710797:
@ZanyXDev: Thanks! I've synced with upstream opll.c version.

m9710797:
Hmm, the new opll.c version doesn't compile anymore. That is: it does compile in C mode (it is a C file after all), but in openMSX we compile this file in C++ mode, and that failed. But easy to fix: 
https://github.com/openMSX/openMSX/commit/fd9f21f115ff16533e0512f7d016f39c1a02164c

PR #1365, opened 2021-07-13
Removed some warnings for Windows build

sdsnatcher73:


m9710797:
Thanks!
I merged your patch. At least I merged your idea, but implemented it in a slightly different way. See de49887191fafe7ca176b0783c496f552c366ae0.

PR #1367, opened 2021-07-24
Really enable build on riscv64

ginggs:
https://github.com/openMSX/openMSX/commit/b06f755fa5654b3b17643feeaf74497a64755cf0 adds support for riscv64 in build/cpu.py, but misses a change needed in build/detectsys.py.


ginggs:
With this change in place, I was able to get a successful build on an emulated riscv64 qemu, but have not yet tested on actual hardware.

m9710797:
Thanks.

PR #1368, opened 2021-07-26
Renamed VG 8235 to VG 8235/00 and removed shasums that did not belong

sdsnatcher73:
Now that we have a confirmed .xml for the VG 8235/20 the original VG 8235 xml needed some cleaning. I renamed the xml, machine code and identifiers in the xml. I removed all shasums of ROMs that were dubious and only kept the ones that were confirmed (basically from Luc's machine).


MBilderbeek:
I manually included the changes from this Pull Request into a separate commit. Thank you all for the contribution!

PR #1379, opened 2021-10-19
Fixes 'Initializer not allowed' error in shader.

gilbertfrancois:
The video setting scaler: ScaleNx gives a shader error on macOS:

```
Error(s) compiling shader "scale2x.vert":
ERROR: 0:14: Initializer not allowed
ERROR: 0:15: Use of undeclared identifier 'pi'
ERROR: 0:21: Use of undeclared identifier 'pi2'
```

Most likely reason for the error is that the compiler cannot evaluate the (constant) value at compile time. Removing the `const` type qualifier fixes the problem and the shader works as intended. 

Before the fix, **scale2x** shader is ignored:
before fix

After the fix, **scale2x** shader works as intended:
after fix

### DIFF
--- a/share/shaders/scale2x.vert
+++ b/share/shaders/scale2x.vert
@@ -9,8 +9,8 @@ varying vec2 coord2pi;
 varying vec2 texCoord;
 varying vec2 videoCoord;
 
-const float pi = 4.0 * atan(1.0);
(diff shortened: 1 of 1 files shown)


m9710797:
Merged.  Thanks!

PR #1382, opened 2021-10-29
Move build to src_build

### LINKED ISSUE #1381: Move build folder
Hi,

Are you ok to move `build` folder to `src_build` folder ?
That can help for compilation on  batocera with buildroot = br is remove automatically `build` folder.
I think is better to just move an not to make a big patch for fixe that is not good solution for later.

Best Regards

liberodark:
Build : OK
Fix Batocera support (All OS based on Buildroot).
Fix : https://github.com/openMSX/openMSX/issues/1381


liberodark:
We have make a workaround for fix that

PR #1388, opened 2021-11-28
Merge Github Actions Workflow into single file

duhow:
By using **matrix**, you can reuse the same code and have all workflows into a single Action, which also includes uploading all artifacts into a single place, instead of having them on separate Actions.


duhow:
You can check https://github.com/duhow/openMSX/actions/runs/1512594553 as example.

MBilderbeek:
Hm, I somehow completely missed this pull request. Can you please updated to make it work against the latest yml files?

MBilderbeek:
Thanks for this great build. I merged your code manually, as I am not good enough with git yet to do it automatically and resolve the conflicts with the files that were deleted and such. Much appreciated though!
More improvements to our workflows are most welcome.

PR #1405, opened 2022-02-15
Update platform-freebsd.mk

ghost:
Fixed an issue in FreeBSD where the binary generates a segmentation fault.

### DIFF
--- a/build/platform-freebsd.mk
+++ b/build/platform-freebsd.mk
@@ -8,3 +8,5 @@ EXEEXT:=
 LIBRARYEXT:=.so
 
 COMPILE_FLAGS+=-D_REENTRANT -D_THREAD_SAFE
+
+LINK_FLAGS:= -static-libgcc -static-libstdc++ $(LINK_FLAGS)


m9710797:
Thanks. I've merged your PR.

Out of curiosity, could you give some background information: what was the problem, and how does this patch fix it?

mthuurne:
I'm curious as well: statically linking to the C++ standard library shouldn't be necessary in theory and if it's required to work around a particular issue, we should probably not keep doing it forever after that issue is fixed.

By the way, these flags seem to imply that we're compiling with GCC, but I thought Clang was the default compiler on FreeBSD these days. Am I mistaken or is there something in our build that selects GCC explicitly?

ghost:
@mthuurne I've choose to keep the build with gcc because it was originally used by openmsx developers (older releases), it's possible to use any compiler as long you indicates what you use in the port Makefile. I have plans to use Clang, I'll do some tests first and maybe I'll use it for 0.18 and beyond.

@m9710797 It managed to compiles without any issue but it's segfaults when executing it.
Here's the gdb backtrace:

```
Program received signal SIGSEGV, Segmentation fault.
Invalid permissions for mapped object.
0x0000000802424a78 in vtable for __cxxabiv1::__si_class_type_info () from /lib/libcxxrt.so.1
(gdb) backtrace
#0  0x0000000802424a78 in vtable for __cxxabiv1::__si_class_type_info () from /lib/libcxxrt.so.1
#1  0x00000008014629ea in __dynamic_cast () from /usr/local/lib/gcc10/libstdc++.so.6
#2  0x00000008014e8d36 in bool std::has_facet<std::ctype<char> >(std::locale const&) () from /usr/local/lib/gcc10/libstdc++.so.6
#3  0x00000008014daac5 in std::basic_ios<char, std::char_traits<char> >::_M_cache_locale(std::locale const&) () from /usr/local/lib/gcc10/libstdc++.so.6
#4  0x00000008014daf71 in std::basic_ios<char, std::char_traits<char> >::init(std::basic_streambuf<char, std::char_traits<char> >*) () from /usr/local/lib/gcc10/libstdc++.so.6
#5  0x000000080147707b in std::ios_base::Init::Init() () from /usr/local/lib/gcc10/libstdc++.so.6
#6  0x000000000043ab9e in ?? ()
#7  0x0000000800c45e8d in ?? () from /libexec/ld-elf.so.1
#8  0x00007fffffffe528 in ?? ()
#9  0x0000000000000011 in ?? ()
#10 0x0000000000000000 in ?? ()
```

Issue #1400, opened 2022-01-22
Moonsound/YMF278 issue: high notes sound bad (aliasing?)


MBilderbeek:
As Wolf reported.

To test: Run MBWave -> Select Waves -> Octave on C5 -> select Harmonica voice -> press U

The sound is bad, it's like aliasing. He tested also some other versions: 0.15 is the same, but 0.14 sounds clean(er).  With 0.14 the issue does show up when the first keyboard note is set to C6 (so at higher frequencies). Major changes in 0.15 were the great improvements authored by @ValleyBell . (Maybe he can help to analyze what's going on?)

The problem does not go away if I use the channel recorder [1], so it is in the actual sound output by the chip emulation at the native frequency (so, it's not a resampling issue).

[1] [channelrecorded_u.zip](https://github.com/openMSX/openMSX/files/7918926/channelrecorded_u.zip)
This is a small recording of the test case with different notes using Q W E R T Y U I O P [harmonica.avi.zip](https://github.com/openMSX/openMSX/files/7918925/harmonica.avi.zip) The issue is mostly present from U onwards.

ValleyBell:
You don't happen to have a hardware recording of that, do you?

I'd like to rule out the possibility of this being a flaw with the actual chip.

MBilderbeek:
@ValleyBell A friend of mine (Meits) checked on real hardware and there it did sound clean. I've requested him to record it, so I can attach it to this issue. 

FYI, how it sounds in openMSX 0.14.0 is attached here:
[openmsx0003.wav.zip](https://github.com/openMSX/openMSX/files/7919124/openmsx0003.wav.zip)
And this is the source (requires MSX-DOS2):
[mbwave.zip](https://github.com/openMSX/openMSX/files/7919123/mbwave.zip)

MBilderbeek:
I made a VGM file for the test case provided here, so that it's easy to check for @ValleyBell whether his implementation has the same issue/difference.
[music0007.vgm.zip](https://github.com/openMSX/openMSX/files/7919341/music0007.vgm.zip)

MBilderbeek:
@grauw Could you make a recording of this on real hardware, just to be sure?

ValleyBell:
I can confirm that the aliasing got worse in libvgm/VGMPlay as well.
In the 2017 version (with the old core) there is almost no aliasing, but it is very audible with the rewritten OPL4 emulation.

m9710797:
[...]

PR #1413, opened 2022-03-25
YMF278B: Fix aliasing in high pitched notes.

grauw:
See issue #1400.


m9710797:
Thanks a lot grauw to work on this!

If I understand correctly the difference between 'pos1' and 'pos2' should always be 1, except for the case right after looping, correct? If so, would it be an option to only store one of these two variables? Maybe something like this:
```
   pos2 = wrap(pos + 1);
   fetchAndInterpolateUsing(pos, pos2);
   pos = wrap(pos + incr);
```

m9710797:
Thanks!

PR #1420, opened 2022-05-26
YMF278: Restart sample position when wave changes during release.

grauw:
I verified on real OPL4 that changing wave during key on indeed retriggers the EG, however doing so during key off does not. The EG state is unaffected other than the new wave table parameters being applied to it. The sample position does get reset though, so it plays from the beginning.

Test VGM and recording: [opl4test.zip](https://github.com/openMSX/openMSX/files/8780808/opl4test.zip)


m9710797:
Thanks!

PR #1445, opened 2022-09-07
add explicit ctime include

ajakk:
Without it, we hit build failures related to missing time_t in some cases:

https://bugs.gentoo.org/868726

Signed-off-by: John Helmert III <ajak@gentoo.org>


m9710797:
Thanks!

PR #1457, opened 2022-11-02
Updates build process for macOS

gilbertfrancois:
Updates the build process for macOS.

- The folder `share` in the app bundle has been relocated to `Content/Resources` inside the app bundle. This allows codesign to work without errors and enhances compatibility for Macs with Apple Silicon.
- The `build/package-darwin/app.mk` has been adapted to automatically sign the app bundle in case the user has an Apple Developers ID. 
- Small addition to the documentation, on code signing for macOS.

### DIFF
--- a/build/package-darwin/app.mk
+++ b/build/package-darwin/app.mk
@@ -10,7 +10,7 @@ APP_ICON:=$(APP_RES)/openmsx-logo.icns
 # Override install locations.
 DESTDIR:=$(BINDIST_DIR)
 INSTALL_BINARY_DIR:=$(APP_EXE_DIR)
-INSTALL_SHARE_DIR:=$(APP_DIR)/share
+INSTALL_SHARE_DIR:=$(APP_RES)/share
 INSTALL_DOC_DIR:=Documentation
 
 PACKAGE_FULL:=$(shell PYTHONPATH=build $(PYTHON) -c \
@@ -23,7 +23,7 @@ BINDIST_LICENSE:=$(INSTALL_DOC_DIR)/GPL.txt
 # TODO: What is needed for an app folder?
 app: install $(DESTDIR)/$(APP_PLIST) $(DESTDIR)/$(APP_ICON)
 
-bindist: app $(DESTDIR)/$(BINDIST_README) $(DESTDIR)/$(BINDIST_LICENSE)
+bindist: app codesign $(DESTDIR)/$(BINDIST_README) $(DESTDIR)/$(BINDIST_LICENSE)
 	@echo "Creating disk image:"
 	@hdiutil create -srcfolder $(BINDIST_DIR) \
 		-fs HFS+J \
@@ -42,6 +42,14 @@ $(DESTDIR)/$(APP_ICON): $(DESTDIR)/$(APP_RES)/%: $(APP_SUPPORT_DIR)/% bindistcle
 	@echo "  Copying resources..."
 	@mkdir -p $(@D)
 	@cp $< $@
+
+codesign:
+	@if [ -z "$(CODE_SIGN_IDENTITY)" ]; then \
+		echo "  Skipping code sign, CODE_SIGN_IDENTITY not set."; \
+	else \
+		echo "  Signing the application bundle..."; \
+		codesign --deep --force --verify --verbose --options runtime --sign "$(CODE_SIGN_IDENTITY)" $(DESTDIR)/$(APP_DIR); \
+	fi
 
 $(DESTDIR)/$(BINDIST_README): $(APP_SUPPORT_DIR)/README.html
 	@echo "  Copying README..."
--- a/doc/manual/compile.html
+++ b/doc/manual/compile.html
@@ -730,6 +730,27 @@ The final output files will be produced in <code>derived/&lt;cpu&gt;-&lt;os&gt;-
 <p>
 The final output is a DMG file (Mac disk image) containing the openMSX application folder and documentation. This file is internet-enabled, which means it will be automatically mounted after it is downloaded.
 </p>
+<p>
+If you have a Apple Developers ID, you can automatically sign the app bundle by providing the following environment variable in your shell, before building the app bundle:
+</p>
+<div class="commandline">
+    <pre>
+export CODE_SIGN_IDENTITY=&lt;your Apple Application Developer ID here&gt;
+make staticbindist
+    </pre>
+</div>
+<p>
+To find the exact string, needed for the CODE_SIGN_IDENTITY, you can use the command:
+</p>
+<div class="commandline">
+security find-identity
+</div>
+<p>
+Check if the code signing was successful with:
+</p>
+<div class="commandline">
+codesign -dv --verbose=4 derived/&lt;cpu&gt;-&lt;os&gt;-&lt;flavour&gt;-3rd/bindist/openMSX.app
+</div>
 
 <h5>Microsoft Windows (MinGW)</h5>
 
--- a/src/file/FileOperationsMac.mm
+++ b/src/file/FileOperationsMac.mm
@@ -14,9 +14,9 @@ std::string findShareDir()
 			throw FatalError("Failed to get main bundle");
 		}
 
-		NSURL* url = [mainBundle executableURL].URLByResolvingSymlinksInPath.URLByDeletingLastPathComponent;
+		NSURL* url = [mainBundle bundleURL].URLByResolvingSymlinksInPath;
 		while (url != nil) {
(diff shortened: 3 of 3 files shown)


MBilderbeek:
@mthuurne can you please comment on this, or approve if you agree?

MBilderbeek:
Or perhaps I should ask @grauw ?

grauw:
I’m by no means an expert, but think it looks fine, it makes sense. I will complain if I encounter any issues with compiling locally in the future :).

MBilderbeek:
Merged, as that comment satisfies me more than enough :)

MBilderbeek:
Thanks @gilbertfrancois !

grauw:
@gilbertfrancois I did find find a problem; if I run from the command line (`/usr/local/bin/openmsx` symbolic links to `/Users/.../Development/openmsx/derived/x86_64-darwin-devel-3rd/bindist/openMSX.app/Contents/MacOS/openmsx`), now it does not find the `share` directory anymore.

grauw:
The problem is that when running it on the command line, `[mainBundle bundleURL]` returns `/usr/local/bin/`.

If I change it back to `[mainBundle executableURL]` it will work for me in this case:

```diff
diff --git a/src/file/FileOperationsMac.mm b/src/file/FileOperationsMac.mm
index c4f72b37f..d42b6d3f3 100644
--- a/src/file/FileOperationsMac.mm
+++ b/src/file/FileOperationsMac.mm
@@ -14,7 +14,7 @@
                        throw FatalError("Failed to get main bundle");
                }
 
-               NSURL* url = [mainBundle bundleURL].URLByResolvingSymlinksInPath;
+               NSURL* url = [mainBundle executableURL].URLByResolvingSymlinksInPath;
                while (url != nil) {
                        NSURL* shareURL = [url URLByAppendingPathComponent:@"Contents/Resources/share" isDirectory:YES];
                        NSNumber* isDirectory;
```

It works because it iteratively looks in the parent directories until it finds the `shareURL`, so after a few iterations it will get to the bundle root and find a match.

Since the executable is in `Contents/MacOS/`, having the `Contents` appended to the `shareURL` isn’t strictly necessary either.

What do you think @gilbertfrancois?

gilbertfrancois:
I see. I did not expect that macOS users would install openMSX in the Linux way, in `/usr/local/`. 

I think that change can work as long as the `share` folder will be in the `Contents/Resources` folder within the bundle. If you pack the `share` folder outside that `Content/Resources` structure, you cannot sign the app bundle and cannot distribute a pre-build binary.

grauw:
Ok, after thinking more about this, I think this is the right approach:

```diff
diff --git a/src/file/FileOperationsMac.mm b/src/file/FileOperationsMac.mm
index c4f72b37f..7cb6b86b2 100644
--- a/src/file/FileOperationsMac.mm
+++ b/src/file/FileOperationsMac.mm
@@ -14,9 +14,16 @@
                        throw FatalError("Failed to get main bundle");
                }
 
-               NSURL* url = [mainBundle bundleURL].URLByResolvingSymlinksInPath;
+               if (NSURL* shareURL = [mainBundle URLForResource:@"share" withExtension: nil]) {
+                       if ([shareURL hasDirectoryPath]) {
+                               return std::string(shareURL.fileSystemRepresentation);
+                       }
+               }
+
+               NSURL* url = [mainBundle executableURL].URLByResolvingSymlinksInPath;
+
                while (url != nil) {
-                       NSURL* shareURL = [url URLByAppendingPathComponent:@"Contents/Resources/share" isDirectory:YES];
+                       NSURL* shareURL = [url URLByAppendingPathComponent:@"share" isDirectory:YES];
                        NSNumber* isDirectory;
                        BOOL success = [shareURL getResourceValue:&isDirectory forKey:NSURLIsDirectoryKey error:nil];
                        if (success && [isDirectory boolValue]) {
```

With this, if there is a proper application bundle, it will find the `share` directory in `Resources` immediately. Note that if you run the executable path from the CLI without a symlink, `[NSBundle mainBundle]` will also recognise the bundle properly so that will also work.

And otherwise it will use the old search method. In this case, it will eventually find the share directory in the openMSX source code, which is fine at least for my purposes.

The only case left uncovered is when you symlink into an application bundle installed in /Applications… Which is a legit use case I think, to run openMSX from shell scripts (e.g. to run a ROM after assembly). Not sure what to do there yet, but at least the above change is an improvement.

Locally I’ll also update my symlink to point to:

`/Users/.../Development/openmsx/derived/x86_64-darwin-devel-3rd/bin/openmsx`

Rather than:

`/Users/.../Development/openmsx/derived/x86_64-darwin-devel-3rd/bindist/openMSX.app/Contents/MacOS/openmsx`

So that there is no expectation of an app bundle being present.

gilbertfrancois:
I think it looks good by visual inspection, but I'd like to test it tomorrow, if that is ok with you. Then I'm able to run it line by line in the debugger.

grauw:
Final patch that should cover all the bases, including a symlink to `/Applications/openMSX.app/Contents/MacOS/openmsx`:

```diff
diff --git a/src/file/FileOperationsMac.mm b/src/file/FileOperationsMac.mm
index c4f72b37f..c311d1a00 100644
--- a/src/file/FileOperationsMac.mm
+++ b/src/file/FileOperationsMac.mm
@@ -14,12 +14,19 @@
                        throw FatalError("Failed to get main bundle");
                }
 
-               NSURL* url = [mainBundle bundleURL].URLByResolvingSymlinksInPath;
+               NSURL* shareURL = [mainBundle URLForResource:@"share" withExtension:nil];
+               if ([shareURL hasDirectoryPath]) {
+                       return std::string(shareURL.fileSystemRepresentation);
+               }
+
+               // Fallback when there is no application bundle or it is hidden by a symlink
+               NSURL* url = [mainBundle executableURL].URLByResolvingSymlinksInPath;
                while (url != nil) {
-                       NSURL* shareURL = [url URLByAppendingPathComponent:@"Contents/Resources/share" isDirectory:YES];
-                       NSNumber* isDirectory;
-                       BOOL success = [shareURL getResourceValue:&isDirectory forKey:NSURLIsDirectoryKey error:nil];
-                       if (success && [isDirectory boolValue]) {
+                       shareURL = [url URLByAppendingPathComponent:@"Contents/Resources/share"];
+                       if ([shareURL hasDirectoryPath]) {
+                               return std::string(shareURL.fileSystemRepresentation);
+                       }
+                       shareURL = [url URLByAppendingPathComponent:@"share"];
+                       if ([shareURL hasDirectoryPath]) {
                                return std::string(shareURL.fileSystemRepresentation);
                        }
                        if (![url getResourceValue:&url forKey:NSURLParentDirectoryURLKey error:nil]) {
```

Note that I also simplified the code a bit by removing the `[shareURL getResourceValue]` call and instead relying on the fact that [`[url URLByAppendingPathComponent]`](https://developer.apple.com/documentation/foundation/nsurl/1410614-urlbyappendingpathcomponent?language=objc) (without `isDirectory`) will query the filesystem to determine whether the URL points to a directory, and adds a `/` in that case, which can be checked with `[shareURL hasDirectoryPath]`.

Let me know if this looks ok to you @gilbertfrancois, if so then I will submit. I’m happy to wait for you to test it.

grauw:
These below are the above patch split out into three commits. I’ll push to openmsx after your approval, @gilbertfrancois.

grauw:
I went ahead and pushed these changes to master since it troubles my workflow if this doesn’t work. If you still have any feedback, feel free to let me know and I’ll address it later.

gilbertfrancois:
@grauw Sorry for the delay. I noticed that gdb is not working anymore on arm64 and I had to switch to lldb. I think your patch works well. I've tested the application bundle in different locations and always gives the correct path. Also the code sign procedure still works well. I did not test it with symbolic links, but I guess that is what works because it is what you use. Thank you for taking care and working together.

```
(lldb) n
Process 69209 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = step over
    frame #0: 0x0000000101aef124 openmsx`openmsx::FileOperations::findShareDir() at FileOperationsMac.mm:18:8
   15  			}
   16
   17  			NSURL* shareURL = [mainBundle URLForResource:@"share" withExtension:nil];
-> 18  			if ([shareURL hasDirectoryPath]) {
   19  				return std::string(shareURL.fileSystemRepresentation);
   20  			}
   21
Target 0: (openmsx) stopped.
(lldb) print shareURL
(NSURL *) $1 = 0x000060000260c000 "file:///Applications/openMSX.app/Contents/Resources/share/"
```

grauw:
No problem, thanks!

PR #1464, opened 2022-11-23
Add MIDI input support to ALSA

Cacodemon345:
This PR adds MIDI input support to ALSA.

I couldn't ever get MIDI In Reader to work properly with my Fedora installation since the OSS emulation layer of ALSA do not work well with applications expecting `/dev/midi*` to be there and readable. Therefore, this is meant largely as a replacement for it on Linux systems.

### DIFF
--- a/src/serial/MidiSessionALSA.cc
+++ b/src/serial/MidiSessionALSA.cc
@@ -1,10 +1,18 @@
 #include "MidiSessionALSA.hh"
 #include "CliComm.hh"
 #include "MidiOutDevice.hh"
+#include "MidiInDevice.hh"
+#include "MidiInConnector.hh"
+#include "EventListener.hh"
+#include "EventDistributor.hh"
 #include "PlugException.hh"
+#include "Poller.hh"
 #include "PluggingController.hh"
+#include "Scheduler.hh"
 #include "narrow.hh"
 #include "serialize.hh"
+#include "circular_buffer.hh"
+#include "checked_cast.hh"
 #include <iostream>
 #include <memory>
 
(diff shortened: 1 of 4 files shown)


MBilderbeek:
Thanks a lot for this PR! I hope @mthuurne (who wrote the ALSA MIDI-out code) can review it.

MBilderbeek:
@Cacodemon345 your patch seems to introduce compilation errors... can you please update the patch to fix them?

src/serial/MidiSessionALSA.cc
```diff
@@ -1,10 +1,18 @@
...
 #include "PlugException.hh"
+#include "Poller.hh"
 #include "PluggingController.hh"
+#include "Scheduler.hh"
 #include "narrow.hh"
 #include "serialize.hh"
+#include "circular_buffer.hh"
+#include "checked_cast.hh"
 #include <iostream>
 #include <memory>
```
m9710797, line 17: There's a missing `#include <thread>` here. That'll probably fix the compilation errors in the buildloop.

src/serial/MidiSessionALSA.cc
```diff
@@ -166,13 +174,253 @@ void MidiOutALSA::serialize(Archive& /*ar*/, unsigned /*version*/)
...
+{
+	assert(isPluggedIn());
+
+	auto npfd = snd_seq_poll_descriptors_count(&seq, POLLIN);
+	std::vector<struct pollfd> pfd(npfd);
+	snd_seq_poll_descriptors(&seq, pfd.data(), npfd, POLLIN);
+
+	while (true) {
+		if (poll(pfd.data(), npfd, 1000) > 0) {
+			do {
```
m9710797: I was confused why the `do { ... } while (0);` loop is needed.
I *think* it's to direct where the 'break' statements in this loop should go to, correct?
Maybe an alternative could be to change those 'breaks' into 'continue' together with moving `if (stop) break;` before `if (poll(...)) {`.

src/serial/MidiSessionALSA.cc
```diff
@@ -166,13 +174,253 @@ void MidiOutALSA::serialize(Archive& /*ar*/, unsigned /*version*/)
...
+				snd_seq_event_t *ev = NULL;
+				long err, size;
```
m9710797: Detail (coding style).
I personally try to avoid forward declaring variables and leaving them uninitialized. In this case it's easy to avoid:

Write the if below like:
`if (auto err = snd_seq_event_input(...); err < 0) { ...`
And further down:
`if (auto size = snd_midi_event_decode(...); size < 0) { ...`

Bonus:
* This limits the scope of the `err` and `size` variables.
* It allows to use `auto` to deduce the type. Direct use of the type `long` is problematic in portable code (linux/windows, 32bit/64bit) (i know that this code is not supposed to compile on windows, but it's a good habit to avoid `long` anyway). And in this case we don't really care about the exact type of these variable, we only care that it's the same type as is returned from these functions, that's a perfect fit for `auto`.

src/serial/MidiSessionALSA.cc
```diff
@@ -166,13 +174,253 @@ void MidiOutALSA::serialize(Archive& /*ar*/, unsigned /*version*/)
...
+				}
+
+				if (ev->type == SND_SEQ_EVENT_SYSEX) {
+					std::lock_guard<std::mutex> lock(mutex);
+
+					for (auto i : xrange(ev->data.ext.len)) {
+						queue.push_back(static_cast<uint8_t*>(ev->data.ext.ptr)[i]);
+					}
+				} else {
+					uint8_t bytes[12] = {};
```
m9710797: Very much a detail / coding style:
Recently (last couple of weeks) I replaced all C-style arrays with c++ std::array. Rationale: c++ inherited arrays from the C language, including to associated problems (implicit array-to-pointer decay,  can't pass/return an array by value to/from a function (instead you pass a pointer to the array), ...). std::array solves some of these problems.

This line then becomes: `std::array<uint8_t, 12> bytes = {};`
And because there is no implicit array-to-pointer decay anymore, the call below then becomes:
`snd_midi_event_decode(event_parser, bytes.data(), bytes.size(), ev);`
The usage further down `queue.push_back(bytes[i]);` remains unchanged.

src/serial/MidiSessionALSA.cc
```diff
@@ -166,13 +174,253 @@ void MidiOutALSA::serialize(Archive& /*ar*/, unsigned /*version*/)
...
+{
+	auto* conn = static_cast<MidiInConnector*>(getConnector());
+	if (!conn->acceptsData()) {
+		std::lock_guard<std::mutex> lock(mutex);
+		queue.clear();
+		return;
+	}
+	if (!conn->ready()) return;
+
+	byte data;
```
m9710797: Detail. I see the code uses a mix of `byte` and `uint8_t`, both are exactly the same.
A lot of code in openMSX uses a mix of these two types (we're slowly migrating to `uint8_t`). Though within one file we try to be consistent.

m9710797 (commented):
Thanks a lot for this patch!

I added some review comments. All are really about details. (Would be nice if you can fix them, but not strictly necessary).

I could only review the patch from a C++ point of view. I can't review (nor test) the actual functionality. It would be nice if @mthuurne  could review the functionality.
OTOH I assume you already tested this code well, So if @mthuurne isn't available short term, we can pull without extra review.

Thanks again!

src/serial/MidiSessionALSA.cc
Cacodemon345: It's to break out of the if block without exiting the function entirely, since C/C++ doesn't allow you to use `break` statements outside for and while loops.

src/serial/MidiSessionALSA.cc
Cacodemon345: Done.

src/serial/MidiSessionALSA.cc
Cacodemon345: Done.

src/serial/MidiSessionALSA.cc
Cacodemon345: Done except for `size`, since that one needs to be accessed in the later iteration loop.

Cacodemon345:
@MBilderbeek Compile error fixed.

MBilderbeek:
Yeah, I noticed, new builds are already running...

src/serial/MidiSessionALSA.cc
m9710797: Yes, hence my suggestion to write it like:
```
while (true) {
    if (stop) break;
    if (poll(...)) {
        if (...) {
            continue;
        }
    }
}
```
But I'll leave it up to you.

Thanks for addressing all my other review comments so quickly.

Hopefully @mthuurne can make time for a functional review in the coming days. If not, then I'll pull it anyway. I trust you tested it well.

MBilderbeek:
All builds OK now! :) I'll leave this to @mthuurne or @m9710797 to pull whenever they see fit.

src/serial/MidiSessionALSA.cc
Cacodemon345: Done.

Cacodemon345: Done.

src/serial/MidiSessionALSA.cc
```diff
@@ -166,13 +175,250 @@ void MidiOutALSA::serialize(Archive& /*ar*/, unsigned /*version*/)
...
+{
+	assert(isPluggedIn());
+
+	auto npfd = snd_seq_poll_descriptors_count(&seq, POLLIN);
+	std::vector<struct pollfd> pfd(npfd);
+	snd_seq_poll_descriptors(&seq, pfd.data(), npfd, POLLIN);
+
+	while (true) {
+		if (stop)
+			break;
```
mthuurne:
```suggestion
	while (!stop) {
```

src/serial/MidiSessionALSA.cc
```diff
@@ -166,13 +175,250 @@ void MidiOutALSA::serialize(Archive& /*ar*/, unsigned /*version*/)
...
+				}
+			} else {
+				std::array<uint8_t, 12> bytes = {};
+
+				auto size = snd_midi_event_decode(event_parser, bytes.data(), bytes.size(), ev);
+				if (size < 0) {
+					std::cerr << "Error decoding MIDI event: "
+						<< snd_strerror(size) << '\n';
+					snd_seq_free_event(ev);
+					break;
```
mthuurne: Why `break` rather than `continue` here?

src/serial/MidiSessionALSA.cc
```diff
@@ -166,13 +175,250 @@ void MidiOutALSA::serialize(Archive& /*ar*/, unsigned /*version*/)
...
+				}
+
+				std::lock_guard<std::mutex> lock(mutex);
+				for (auto i : xrange(size)) {
+					queue.push_back(bytes[i]);
+				}
+			}
+			snd_seq_free_event(ev);
+			eventDistributor.distributeEvent(
+				Event::create<MidiInALSAEvent>());
```
mthuurne, line 356: This sends a `MidiInALSAEvent` for each MIDI message put in the buffer, but the signal handler method only retrieves a single byte from that buffer, while more than one byte could have been inserted. Unless I'm missing something, this would lead to multi-byte MIDI messages to be processed too late.

I checked the Windows MIDI-in code and the same pattern is used there. However, I'm not convinced that the Windows MIDI-in code is bug free; it's really old code and I don't think it has seen a lot of testing.

mthuurne (commented):
I did a quick review. For the most part it looks good, but I'm not certain multi-byte messages are handled correctly. If they aren't, that means the Windows version of the code has the same problem, but I wouldn't rule that out.

src/serial/MidiSessionALSA.cc
Cacodemon345: Good catch.

m9710797:
@mthuurne If I understand correctly this PR is now on par with the corresponding windows MIDI in code. (And both may have similar bugs). Could we then maybe already merge this PR?

@Cacodemon345 Would you be interested in investigating/fixing these bugs (in both classes) after this PR has been merged?

Cacodemon345:
Yes.

mthuurne:
This is certainly much better than relying on OSS emulation, so I'm fine with merging it. And maybe it's even correct and I'm just mistaken about how the code works, but I think it's worth investigating whether the buffer gets emptied quickly after receiving multi-byte events.

Issue #170, opened 2015-06-22
[Feature] Initialize existing harddisk image / create it initialized [sf#146]


openMSX-import:
**Reported by manuelbi on 2007-03-27 19:31 UTC**
&lt;Quibus&gt; turbor: is there a way to format an existing hard disk iamge with diskmanipulator?
&lt;turbor&gt; you should be able to format the partitions yes, repartitioning.. no.
&lt;Quibus&gt; I mean, when you start openMSX you get an empty 100MB disk image
&lt;Quibus&gt; can you partition and initialize this existing image?
&lt;turbor&gt; ah, I don't think so, since that would require creating partitions
&lt;turbor&gt; I wonder if you can format hda itself, so no partitions
&lt;turbor&gt; but then you end up with at max 32M usable of those 100M
&lt;turbor&gt; but I guess you can 'set power off' and then 'create' a new file over the old 100 MB one...
&lt;Quibus&gt; it's a bit weird that you need the MSX FDISK.COm to initialize an existing harddisk image, while you can create new ones which are already partitioned/iniitialzed directly with dm
&lt;turbor&gt; So we need to alter openMSX, so that if it detects that no HD image exists it uses dm to create one instead of simply creating a 100MB file...

Or, make it possible to format/partition/initialize an existing harddisk image.

grauw:
Resolved by https://github.com/openMSX/openMSX/commit/eca8328384ce43277c0e3675f8be40ec502beb37.

PR #1475, opened 2023-01-07
Add FAT16 support to disk manipulator

grauw:
This patch series adds FAT16 support to diskmanipulator, as well as contains some other things I ran into along the way. In order to create FAT16 file systems, the new `-extended` flag must be passed when creating or formatting the volume. Or, create the volume with `_FDISK` on the MSX itself, of course.

For the implementation I referenced [Wikipedia](https://en.wikipedia.org/wiki/Design_of_the_FAT_file_system) and the [Nextor source code](https://github.com/Konamiman/Nextor/) (e.g. [fdisk2.c](https://github.com/Konamiman/Nextor/blob/v2.1/source/kernel/bank5/fdisk2.c)).

I hope the patches are fairly self-evident, but let me know any kind of questions or review remarks.

One thing I am thinking about myself to add as an additional change is to rename `-extended` to `-nextor`. It is a bit of a matter of perspective, Nextor does use a standard extended boot sector, so why name it after Nextor if it’s standard. However I think `-nextor` will be more intuitive to the user, but in that case I would probably also want to apply Nextor’s format for FAT12 volumes when specified. But since you can do the same from within `_FDISK`, does that really add value or just more code to maintain.

A second question is whether `-extended` should become the default or be left explicit. From a usability point of view probably the former, but maybe we don’t want to create extended boot sectors for FAT12 volumes (I think it’s harmless but not 100% sure since I didn’t test all devices). Or, the default could depend on the size that’s specified (>32M = extended).


m9710797:
Great! This has been on my TODO list for a long time, but always with very low priority.
Thanks a lot for picking this up!

I'll start reviewing this, but I'll probably won't finish it today. I'll leave review comments on the individual commits.

After the review I might have more meaningful answers to the questions in your message. But for now I think that supporting **both** '-extended' and '-nextor' might be a good compromise? Making '-extended' default for sizes >32M also seems like a good idea.

At the moment I don't know what this "extended" stuff is. Hopefully that'll become clear to me when I read your patches. Though my initial thought was: why not name this flag "-fat16". That's more expressive and leaves the way open to future (though unlikely) "-fat32" support.

It's a bit unfortunate that currently many of the (FAT12) handling openMSX code is duplicated between MSXtar and DirAsDSK. I assume this PR only deals with the MSXtar (and DiskManipulator) code?

grauw:
> At the moment I don't know what this "extended" stuff is. Hopefully that'll become clear to me when I read your patches. Though my initial thought was: why not name this flag "-fat16". That's more expressive and leaves the way open to future (though unlikely) "-fat32" support.

I had that at first, but I changed it with the consideration that like the `-dos1` flag it is more about the format of the boot record, where the Extended BIOS Parameter Block enables volumes larger than 32 MB. You can create FAT16 volumes on DOS2.20 boot records, as long as they are smaller than 32 MB. And conversely, using `-extended` with a size smaller than 32 MB still creates a FAT12 file system.

It comes from a choice to keep the concern of whether a FAT12 or FAT16 system is created out of the users’s hands and just base it on the partition size. But perhaps there could be an additional `-fat16` option which forces a FAT16 partition even on sizes smaller than 32 MB (down to 2 MB), separate from the format of the boot record. Since it currently is not possible to create those.

However generally speaking I think FAT12 is preferable for <= 32 MB volumes for the sake of compatibility, not much reason to use FAT16 other than somewhat less space waste due to smaller cluster sizes, in exchange for slower disk space calculations when you `DIR`. So is that really serving a use case, and not just making the command line interface more complicated to understand for those not so familiar with file system details? And when we go down that route, would an option to manually decrease or increase the cluster size not be better instead?

But yeah, your sentiment “I don't know what this "extended" stuff is” probably applies to more people. So either a good default so that you never have to specify it, or `-nextor` or `-fat16` (with somewhat different semantics), should probably be considered.

> It's a bit unfortunate that currently many of the (FAT12) handling openMSX code is duplicated between MSXtar and DirAsDSK. I assume this PR only deals with the MSXtar (and DiskManipulator) code?

Yes, maybe that’s something to look at later. For DirAsDsk I would first try to make it work with FAT12 volumes on HD interfaces, then FAT16 volumes, and then (or along the way) make it share more code with MSXtar.

m9710797 (commented):
This commit is fine. Though I'd personally would like to ask for some more background information (but no need to make any changes to the commit).

After reading the comment block of your PR i was expecting to see FAT16 related changes. So my assumption while reading this commit was that these reserved sectors were somehow needed to create a FAT16 filesystem. After reading more of your patches, I no longer think that's the case. Is this correct?  In other words: the reserved sectors stuff is fully orthogonal to the FAT12/FAT16 choice.

Do you have more info on this reserved sector stuff. In particular in the case when there is more than one, where are those extra sectors located. Because this patch (only) changes the location of the root directory, I assume they must be located between the FAT(s) and the (root)directory sectors?

m9710797 (commented):
MSXtar: Add check for sector size.

This commit is fine

m9710797 (commented):
MSXtar: Improve sectors per cluster check.

This commit is fine.

grauw:
> After reading the comment block of your PR i was expecting to see FAT16 related changes. So my assumption while reading this commit was that these reserved sectors were somehow needed to create a FAT16 filesystem. After reading more of your patches, I no longer think that's the case. Is this correct? In other words: the reserved sectors stuff is fully orthogonal to the FAT12/FAT16 choice.

Yes, sorry, I was just cleaning up things along the way. Only the last two patches in the series are really about FAT16.

> Do you have more info on this reserved sector stuff. In particular in the case when there is more than one, where are those extra sectors located. Because this patch (only) changes the location of the root directory, I assume they must be located between the FAT(s) and the (root)directory sectors?

They are counted from the beginning of the partition, to support more extended boot records. See [Wikipedia](https://en.wikipedia.org/wiki/Design_of_the_FAT_file_system#RSVD_SECTORS). E.g. FAT32 apparently sets it to 32 to hold the extended boot sector, FS info sector and backup boot sectors. In practice for us it’s probably always 1, but I guess I could put an awesome copy protection there, or hidden data, or something, and diskmanipulator should then still work.

src/fdc/MSXtar.cc
```diff
@@ -101,9 +102,9 @@ void MSXtar::parseBootSector(const MSXBootSector& boot)
 		throw MSXException(
 			"Illegal number sectors per FAT: ", sectorsPerFat);
 	}
-	if (nbRootDirSectors == 0) { // TODO: check limits more accurately
+	if (nbRootDirSectors == 0 || nbRootDirSectors * DIR_ENTRIES_PER_SECTOR != boot.dirEntries) {
```
m9710797: Is the first part of the check `nbRootDirSectors == 0` still needed?

Hmm, when I think about it more. You check that `((boot.dirEntries / 16) * 16) == boot.dirEntries`. So you effectively check that it's a non-zero multiple of 16. Might be clearer to express it like that?

m9710797:
> > Do you have more info on this reserved sector stuff. In particular in the case when there is more than one, where are those extra sectors located. Because this patch (only) changes the location of the root directory, I assume they must be located between the FAT(s) and the (root)directory sectors?
> 
> They are counted from the beginning of the partition, to support more extended boot records. See [Wikipedia](https://en.wikipedia.org/wiki/Design_of_the_FAT_file_system#RSVD_SECTORS). E.g. FAT32 apparently sets it to 32 to hold the extended boot sector, FS info sector and backup boot sectors. In practice for us it’s probably always 1, but I guess I could put an awesome copy protection there, or hidden data, or something, and diskmanipulator should then still work.

My question was more about how `reserved sectors > 1` shifts around the other sectors. In your patch you shift the root directory sectors down, but you leave the location of the FAT sectors unchanged.

But from your previous paragraph I get the impression that these extra reserved sectors are located between the boot sector and the FAT sectors.

m9710797 (commented):
MSXtar: Use FREE_FAT constant where appropriate. 

This commit is fine.

m9710797 (commented):
MSXtar: Do not crash on FREE_FAT in FAT chain. 

This commit is fine.

m9710797 (commented):
MSXtar: Put exceptions on a single line. 

OK

grauw:
> My question was more about how `reserved sectors > 1` shifts around the other sectors. In your patch you shift the root directory sectors down, but you leave the location of the FAT sectors unchanged.
> 
> But from your previous paragraph I get the impression that these extra reserved sectors are located between the boot sector and the FAT sectors.

Oh, then that’s a bug :). I’ll fix that. They come after (and include) the boot sector.

Here’s a clear table of the layout: [Design_of_the_FAT_file_system#Layout](https://en.wikipedia.org/wiki/Design_of_the_FAT_file_system#Layout)

src/fdc/DiskImageUtils.cc
```diff
@@ -104,6 +104,10 @@ static SetBootSectorResult setBootSector(
 		nbDirEntry = 256;
 		descriptor = 0xF0;
 		nbHiddenSectors = 16;	// override default from above
+		// for a 32MB disk or greater the sectors would be >= 65536
+		// since MSX use 16 bits for this, in case of sectors = 65536
+		// the truncated word will be 0 -> formatted as 320 Kb disk!
+		if (nbSectors > 65535) nbSectors = 65535; // this is the max size for fat12 :-)
```
m9710797, line 285: nitpick: `nbSectors = std::min(nbSectors, 65536);`

Though I wonder if moving this line from Diskmanipulator to here doesn't have any other consequences.
For example: before the size of the generated disk image would be truncated to max 65535 sectors. But now we create a possibly larger image with truncated sector count information in the boot sector.

Maybe this is fine, but I wanted to point you to this (maybe unintended) difference.

Maybe instead of silently modifying the sector count, we should give an error then something doesn't fit?

grauw: That’s intentional, now it is possible to create any size of partition with the diskmanipulator so you can format it with `_FDISK` (absent diskmanipulator support). I thought it was weird that the file was smaller than I indicated.

Modifying the sector count before storing it I think is normal, ideally the sector count matches data start + cluster count * cluster size precisely, which is what Nextor’s `_FDISK` actually does.

If not an exact match, it relies on the code that uses the file system’s integer division flooring the cluster count, as well as checking against the max cluster number. That will probably happen, since due to all the FAT variations stuff dealing with FAT is written to be quite tolerant, but it doesn’t hurt to be precise.

grauw: > nitpick: `nbSectors = std::min(nbSectors, 65536);`

Then it becomes `nbSectors = std::min(nbSectors, size_t(65535));` because `std::min` is stubborn about type widening. I had that at first but decided to keep the original code as-is, but I can change it.

src/fdc/MSXtar.cc
```diff
@@ -75,26 +83,28 @@ unsigned MSXtar::sectorToCluster(unsigned sector) const
...
-		throw MSXException("Illegal number of sectors: ", boot.nrSectors);
+	if (nrSectors == 0) { // TODO: check limits more accurately
+		throw MSXException("Illegal number of sectors: ", nrSectors);
 	}
 	if (boot.resvSectors == 0) {
 		throw MSXException("Illegal number of reserved sectors: ", boot.resvSectors);
 	}
-	if (boot.nrSides == 0) { // TODO: check limits more accurately
-		throw MSXException("Illegal number of sides: ", boot.nrSides);
-	}
```
m9710797: Was this intentional? Why is checking `nrSides` wrong or no longer needed?

src/fdc/MSXtar.cc
```diff
@@ -31,11 +31,19 @@ namespace openmsx {
...
 static constexpr unsigned FIRST_CLUSTER = 0x002;
-static constexpr unsigned BAD_FAT = 0xFF7;
-static constexpr unsigned EOF_FAT = 0xFFF; // actually 0xFF8-0xFFF, signals EOF in FAT12
+static constexpr unsigned MAX_FAT12 = 0xFF6;
+static constexpr unsigned BAD_FAT12 = 0xFF7;
+static constexpr unsigned EOF_FAT12 = 0xFFF; // actually 0xFF8-0xFFF, signals EOF in FAT12
+static constexpr unsigned MAX_FAT16 = 0xFFF6;
+static constexpr unsigned BAD_FAT16 = 0xFFF7;
+static constexpr unsigned EOF_FAT16 = 0xFFFF; // actually 0xFFF8-0xFFFF, signals EOF in FAT16
+static constexpr unsigned EOF_FAT = 0x0FFFFFFF; // coinciding with FAT32's EOF
```
m9710797: Just checking: you use `EOF_FAT` as a stand-in to either mean `EOF_FAT12` or `EOF_FAT16`, right?
So technically any numerical value > 0xffff would do, and you picked the value 0x0fffffff (which happens to be the EOF value for FAT32, but that doesn't really matter here).
This is fine, I just want to make sure I understand.

src/fdc/MSXtar.cc
```diff
@@ -107,10 +117,18 @@ void MSXtar::parseBootSector(const MSXBootSector& boot)
 	chrootSector = rootDirStart;
 	rootDirLast = rootDirStart + nbRootDirSectors - 1;
 
+	// Whether to use FAT16 is strangely implicit, it must be derived from the
+	// cluster count being higher than 0xFF4. Let's round that up to 0x1000,
+	// since FAT12 uses 1.5 bytes per cluster its maximum size is 12 sectors,
+	// whereas FAT16 uses 2 bytes per cluster so its minimum size is 16 sectors.
+	// Therefore we can pick the correct FAT type with sectorsPerFat >= 14.
+	unsigned fat16Threshold = (0x1000 * 3 / 2 + 0x1000 * 2) / 2 / SECTOR_SIZE;
```
m9710797: Very much a detail: this variable could be `constexpr`.

src/fdc/DiskImageUtils.cc
grauw: As for giving an error, I don’t think it is an error. Especially not in this case where you just can’t quite format the entire space so there’s a small deviation.

However even in the case that you have an 64 MB drive but just want one volume on it without partition table, I think that’s legit. Maybe an informational message would be nice, but I couldn’t find a simple way to do that (as simple as throwing). Or a way to query the used vs available space (a new `info` command of sorts), but that’s a separate feature.

src/fdc/MSXtar.cc
```diff
@@ -107,10 +117,18 @@ void MSXtar::parseBootSector(const MSXBootSector& boot)
...
+	fat16 = sectorsPerFat >= fat16Threshold;
+
 	// Bypass max cluster check in sectorToCluster
 	maxCluster = std::numeric_limits<decltype(maxCluster)>::max();
-	maxCluster = sectorToCluster(boot.nrSectors);
-	maxCluster = std::min(maxCluster, BAD_FAT);
+	maxCluster = sectorToCluster(nrSectors);
+	maxCluster = std::min(maxCluster, fat16 ? MAX_FAT16 : MAX_FAT12);
```
m9710797: This changes from `BAD_FAT` to `MAX_FAT`. That was probably a bug in the old code which you fixed?

Maybe we should add some comments to clarify:
* Is `MAX_FAT` inclusive or exclusive. IOW is it the last still valid or the first invalid number.
* Similar for the `maxCluster` variable, is it inclusive or exclusive?
Hmm, but from the formula `sectorToCluster(nrSectors)` I get the impression that `maxCluster` is exclusive because also `nrSectors` is exclusive.

src/fdc/MSXtar.cc
```diff
@@ -194,38 +212,48 @@ MSXtar::~MSXtar()
...
-	assert(val < 4096); // FAT12
-	std::span data{fatBuffer[0].raw.data(), sectorsPerFat * size_t(SECTOR_SIZE)};
-	auto p = subspan<2>(data, (clNr * 3) / 2);
-	if (clNr & 1) {
-		p[0] = narrow_cast<uint8_t>((p[0] & 0x0F) + (val << 4));
-		p[1] = narrow_cast<uint8_t>(val >> 4);
+	if (val == EOF_FAT) {
+		val = fat16 ? EOF_FAT16 : EOF_FAT12;
 	} else {
+		assert(fat16 ? val < MAX_FAT16 : val < MAX_FAT12);
```
m9710797: Detail. This confused me for a second. Maybe rewrite as:
`assert(val < (fat16 ? MAX_FAT16 : MAX_FAT12));`
This check implies `MAX_FAT` is exclusive (see above), is that correct?

src/fdc/MSXtar.cc
```diff
@@ -265,10 +293,12 @@ unsigned MSXtar::getNextSector(unsigned sector)
 }
 
 // get start cluster from a directory entry,
-// also takes care of BAD_FAT and EOF_FAT-range.
-static unsigned getStartCluster(const MSXDirEntry& entry)
+// BAD FAT and EOF FAT-range are normalized to a single EOF_FAT value (0xFFFFFFF)
```
m9710797: Detail. I would suggest to remove the numerical value `0xfffffff` from the comment. Or else write it as `0x0fffffff` (currently it's too easy to read it as all-1-bits).

src/fdc/MSXtar.cc
```diff
@@ -265,10 +293,12 @@ unsigned MSXtar::getNextSector(unsigned sector)
...
+unsigned MSXtar::getStartCluster(const MSXDirEntry& entry) const
 {
-	return normalizeFAT(entry.startCluster);
+	unsigned cluster = entry.startCluster;
+	unsigned badFat = fat16 ? BAD_FAT16 : BAD_FAT12;
+	return cluster < badFat ? cluster : EOF_FAT;
```
m9710797: Should this be `BAD_FAT` or `MAX_FAT`?

src/fdc/MSXtar.hh
```diff
@@ -66,9 +66,12 @@ private:
 	void chroot(std::string_view newRootDir, bool createDir);
 
 private:
+	unsigned getStartCluster(const MSXDirEntry& entry) const;
+
 	SectorAccessibleDisk& disk;
 	MemBuffer<SectorBuffer> fatBuffer;
 
+	bool fat16;
```
m9710797: Detail: if there's no reason for a particular order for member variables (e.g. group related variables together) then prefer to order them from large to small.

In this case that means moving the `fat16` variable from line 74 to line 80. The reason for this is alignment and (internal) padding. The current patch will enlarge the size of the `MSXtar` object with 4 bytes (1 for the `bool`, 3 for padding to keep the next `unsigned` variable aligned). But when moving the variable, the size remains the same (the `bool` now consumes a byte that was already used for padding).

src/fdc/MSXtar.cc
grauw: Because disk images formatted by Nextor _FDISK (at least the FAT16 ones) have this set to 0, as I assume is normal for LBA type file systems.

Since it is not relevant to the correctness of the MSXtar implementation, there should be no need to check this one, otherwise there’d be a whole lot more complicated [CHS-related things](https://en.wikipedia.org/wiki/Design_of_the_FAT_file_system#FATID) we could check.

src/fdc/MSXtar.cc
grauw: Yep.

src/fdc/MSXtar.cc
grauw: Incidentally I did see this one (normally I probably wouldn’t think about it too readily), however I figured the compiler would do it for me anyway so the line would be shorter without.

src/fdc/DiskImageUtils.cc
```diff
@@ -6,11 +6,18 @@
...
 #include "xrange.hh"
+#include <algorithm>
+#include <bit>
 #include <cassert>
 #include <ctime>
 
 namespace openmsx::DiskImageUtils {
 
+static constexpr unsigned MAX_FAT12_CLUSTER_COUNT = 0xFF4;
+static constexpr unsigned MAX_FAT16_CLUSTER_COUNT = 0xFFF4;
```
m9710797: Just making sure I understand: this value comes from `MAX_FAT - 2` (which implies that MAX_FAT is exclusive (the first value that's no longer a valid cluster number)).

Detail: MAX_FAT12_CLUSTER_COUNT is not used. But fine if you leave it for symmetry. Or alternatively you could move `MAX_FAT16_CLUSTER_COUNT` closer to the only place where it's used (inside the then-branch in the `setBootSector()` function, around line 120).

src/fdc/DiskImageUtils.cc
```diff
@@ -55,14 +62,14 @@ void checkValidPartition(SectorAccessibleDisk& disk, unsigned partition)
...
-void checkFAT12Partition(SectorAccessibleDisk& disk, unsigned partition)
+void checkSupportedPartition(SectorAccessibleDisk& disk, unsigned partition)
 {
 	SectorBuffer buf;
 	Partition& p = checkImpl(disk, partition, buf);
 
 	// check partition type
-	if (p.sys_ind != 0x01) {
-		throw CommandException("Only FAT12 partitions are supported.");
+	if (p.sys_ind != 0x01 && p.sys_ind != 0x04 && p.sys_ind != 0x06 && p.sys_ind != 0x0E) {
```
m9710797: Detail: `if (p.sys_ind != one_of(0x01, 0x04, 0x06, 0x0E)) {`

src/fdc/DiskImageUtils.cc
```diff
@@ -71,30 +78,58 @@ void checkFAT12Partition(SectorAccessibleDisk& disk, unsigned partition)
...
+	bool fat16;
 };
 static SetBootSectorResult setBootSector(
-	MSXBootSector& boot, size_t nbSectors, bool dos1)
+	MSXBootSector& boot, MSXBootSectorType bootType, size_t nbSectors)
 {
 	// start from the default boot block ..
-	const auto& defaultBootBlock = dos1 ? BootBlocks::dos1BootBlock : BootBlocks::dos2BootBlock;
+	const auto& defaultBootBlock = bootType == MSXBootSectorType::DOS1 ? BootBlocks::dos1BootBlock :
+		bootType == MSXBootSectorType::DOS2 ? BootBlocks::dos2BootBlock : BootBlocks::extendedBootBlock;
```
m9710797: Very much a detail. I'd prefer to format this code as:
```
const auto& defaultBootBlock = bootType == MSXBootSectorType::DOS1 ? BootBlocks::dos1BootBlock
                             : bootType == MSXBootSectorType::DOS2 ? BootBlocks::dos2BootBlock
                                                                   : BootBlocks::extendedBootBlock;
```

src/fdc/DiskImageUtils.cc
```diff
@@ -71,30 +78,58 @@ void checkFAT12Partition(SectorAccessibleDisk& disk, unsigned partition)
...
+		nbHiddenSectors = 0;
+		fat16 = true;
+
+		// <= 128 MB: 4, <= 256 MB: 8, ..., <= 4 GB: 128
+		nbSectorsPerCluster = narrow<uint8_t>(std::clamp(std::bit_ceil(nbSectors) >> 16, size_t(4), size_t(128)));
+
+		// calculate fat size based on cluster count estimate
+		unsigned fatStart = nbReservedSectors + nbDirEntry / DIR_ENTRIES_PER_SECTOR;
+		unsigned estSectorCount = narrow<unsigned>(nbSectors - fatStart);
+		unsigned estClusterCount = std::min(estSectorCount / nbSectorsPerCluster, MAX_FAT16_CLUSTER_COUNT);
```
m9710797: I understand the calculation, but I don't know what `est` stands for (my first idea was 'estimation' but that doesn't make sense).

src/fdc/MSXtar.cc
grauw: Yes indeed.

I added `BAD_FAT` in an earlier submit (before then it was unclamped), but since then I got more detail about what the maximum legal (non-special) cluster number is. `MAX_FAT12` is the maximum cluster number + 1 (so, exclusive) for FAT12 and ditto for `MAX_FAT16`. And, because of the encoding of cluster numbers, the maximum number of clusters is `MAX_FAT*` - 2.

The `maxCluster` variable is also exclusive, that is, it is the maximum cluster number + 1 or the maximum cluster count + 2.

It was also not clear to me at first, so I can see where you’re coming from, I’ll add some comments. Or maybe I can do something with the name or the way they’re used, e.g. renaming to `clusterCount` and then doing +2 when it’s used to compare with a cluster number. Counts are unambiguously exclusive, after all.

src/fdc/DiskImageUtils.cc
```diff
@@ -170,32 +205,49 @@ static SetBootSectorResult setBootSector(
...
 	boot.dirEntries = nbDirEntry;
 	boot.descriptor = descriptor;
 	boot.resvSectors = nbReservedSectors;
-	boot.hiddenSectors = nbHiddenSectors;
 
-	if (!dos1) {
-		// set random volume id
-		boot.vol_id = random_32bit() & 0x7F7F7F7F; // why are bits masked?
+	if (bootType == MSXBootSectorType::DOS1) {
+		auto params = boot.params.dos1;
```
m9710797: This seems like a bug. It should be:
`auto& params = boot.params.dos1;`
You want to refer to part of the boot sector and make changes there. Instead of copying part in a local variable and changing that copy (and then immediately discarding that changed copy).

Same on line 225 and 229.

src/fdc/DiskImageUtils.hh
```diff
@@ -24,13 +30,31 @@ struct MSXBootSector {
...
+			std::array<uint8_t, 9>      pad1;          // +30
+			Endian::UA_L32              vol_id;        // +39
+			std::array<uint8_t, 512-43> pad2;          // +43
+		} dos1;
+		struct {
+			Endian::L16                 hiddenSectors; // +28 not used
+			std::array<uint8_t, 9>      pad1;          // +30
+			Endian::UA_L32              vol_id;        // +39
+			std::array<uint8_t, 512-43> pad2;          // +43
+		} dos2;
```
m9710797, line 43: Currently the `dos1` and `dos2` fields of the union are identical. But I assume the intention was to leave out the `vol_id` field from `dos1`?

src/fdc/DiskImageUtils.hh
```diff
@@ -111,27 +135,28 @@ namespace DiskImageUtils {
...
 
 	/** Format the given disk (= a single partition).
 	 * The formatting depends on the size of the image.
-	 * @param disk the disk/partition image to be formatted
-	 * @param dos1 set to true if you want to force dos1 formatting (boot sector)
+	 * @param disk The disk/partition image to be formatted.
+	 * @param bootType The boot sector type to use.
 	 */
-	void format(SectorAccessibleDisk& disk, bool dos1 = false);
+	void format(SectorAccessibleDisk& disk, MSXBootSectorType bootType = MSXBootSectorType::DOS2);
```
m9710797: I know the old code already used a default value. But I think it's better to make `bootType` a mandatory parameter (leave out the default value). I quickly checked and I think all callers do already pass a value.

And the same for `partition()` on line 159.

src/fdc/DiskManipulator.cc
```diff
@@ -186,18 +186,20 @@ void DiskManipulator::execute(std::span<const TclObject> tokens, TclObject& resu
...
+			if (drive == one_of("-dos1", "-dos2", "-extended")) {
+				bootType = drive == "-dos1" ? MSXBootSectorType::DOS1 :
+					drive == "-dos2" ? MSXBootSectorType::DOS2 : MSXBootSectorType::EXTENDED;
 				drive = tokens[3].getString();
-			} else if (tokens[3] == "-dos1") {
-				dos1 = true;
+			} else if (tokens[3] == one_of("-dos1", "-dos2", "-extended")) {
+				bootType = tokens[3] == "-dos1" ? MSXBootSectorType::DOS1 :
+					tokens[3] == "-dos2" ? MSXBootSectorType::DOS2 : MSXBootSectorType::EXTENDED;
 			}
```
m9710797, line 205: This was already a problem in the code before your changes. There is no good error checking, for example an invalid flag in the last position is interpreted as `DOS2` instead of reported as an error.

Here's one way to improve this (not tested):
```
if (tokens.size() == 4) {
    auto parseType = [](std::string_view s) -> std::optional<MSXBootSectorType> {
        if (s == "-dos1")     return MSXBootSectorType::DOS1;
        if (s == "-dos2")     return MSXBootSectorType::DOS2;
        if (s == "-extended") return MSXBootSectorType::EXTENDED;
        return {};
    };
    if (auto t = parseType(drive)) {
        bootType = *t;
        drive = tokens[3].getString();
    } else {
        auto t = parseType(tokens[3].getString())) {
        if (!t) {
            throw ... some error message ...
        }
        bootType = *t;
    }
}
```
This also has the advantage that we don't repeat the flag names 4(!) times. (And reduce the number of string comparisons).

If this gets any more complicated we might want to use the `parseTclArgs()` utility function (though at the moment it doesn't have good support yet for enum-like flags).

Could you also update the help text around line 255. And the manual in `doc/manual/diskmanipulator.html`.

src/fdc/DiskManipulator.cc
```diff
@@ -346,11 +348,19 @@ void DiskManipulator::create(std::span<const TclObject> tokens)
...
+			continue;
+		}
+		if (token == "-dos2") {
+			bootType = MSXBootSectorType::DOS2;
+			continue;
+		}
+		if (token == "-extended") {
+			bootType = MSXBootSectorType::EXTENDED;
 			continue;
 		}
```
m9710797, line 368: Hmm, here we again convert a string into a `MSXBootSectoType` enum value.

Maybe we should factor out the helper I used above (move the `parseType` lambda function out and make it a proper free function). Then use it here like:
```
if (auto t = parseBootSectorType(token)) {
    bootType = *t;
    continue;
}
```

src/fdc/MSXtar.cc
m9710797: OK. This is useful information. Maybe add it as a comment?

src/fdc/MSXtar.cc
m9710797: You're right.
It's not useful for the compiler but for human readers of the code.

m9710797:
@grauw I finished the review (1st pass). Thanks again for your work!
I'll still have to think about the questions in your message at the very top (the flag name `-extended` vs `-nextor` vs `-fat16`).

src/fdc/MSXtar.cc
grauw: Still needs a check for zero, no?

src/fdc/DiskImageUtils.cc
grauw: Indeed it’s estimation (as per the comment above). It does not know yet how big the FAT will be, so it calculates the available sector & cluster count without considering space needed for the FAT. This means there is a chance the FAT becomes a sector larger than absolutely needed. After that it calculates the real cluster count taking into account that FAT size, which will always be lower than the prior count so it will still fit in the FAT.

This is the algorithm that Nextor `_FDISK` implements, albeit rewritten a bit for clarity since we don’t need to optimise for Z80.

src/fdc/MSXtar.cc
m9710797: You're right.
I think I would be less confused if it were written like: `dirEntries == 0 || (dirEntries % DIR_ENTRIES_PER_SECTOR) != 0`.

src/fdc/DiskImageUtils.cc
grauw: Oops. My testing didn’t catch it because I based the boot block template on a 64 MB image. Let me zero out the sector count in the boot block template to make sure it fails if incorrect.

src/fdc/MSXtar.cc
grauw: Yes, cluster number 0xFF6 / 0xFFF6 is [reserved](https://en.wikipedia.org/wiki/Design_of_the_FAT_file_system#cite_ref-Ecma-107_5-8).

src/fdc/MSXtar.cc
grauw: As per https://en.wikipedia.org/wiki/Design_of_the_FAT_file_system#cite_ref-Ecma-107_5-8, we should be tolerant of reserved cluster number 0xFF6 / 0xFFF6 when reading, and treat it as a normal cluster number. So we do not include it in the normalisation (also not in `readFAT`).

But we should never write it, which is why `writeFAT` asserts the narrower range.

src/fdc/DiskImageUtils.hh
grauw: Agreed. Fyi it was called with default argument from `RamDSKDiskImage.cc`.

src/fdc/DiskImageUtils.cc
grauw: > Just making sure I understand: this value comes from `MAX_FAT - 2` (which implies that MAX_FAT is exclusive (the first value that's no longer a valid cluster number)).

Indeed, where 2 is `FIRST_CLUSTER`.

> Detail: MAX_FAT12_CLUSTER_COUNT is not used. But fine if you leave it for symmetry. Or alternatively you could move `MAX_FAT16_CLUSTER_COUNT` closer to the only place where it's used (inside the then-branch in the `setBootSector()` function, around line 120).

I think instead I’ll add support for the `-nextor` option as `MSXBootSectorType::NEXTOR` (instead of extended), which will use the FAT12 version for volumes smaller than 32 MB.

src/fdc/MSXtar.cc
grauw: I did some further investigation, and it is actually not illegal for the number of root dir entries not to be a multiple of 16. However in practice it always is, and the code is currently not equipped to handle it so better to fail early.

If the number of root dir entries is not a multiple of 16, it should be rounded up to determine the sector count, and the places where it scans the root directory should stop when the count is reached.

I put an item on my TODO list to support arbitrary root dir entry counts.

grauw:
I pushed the changes that were ok to master, and rebased the rest. No need for a new review yet, I’ll do some more changes to it later.

src/fdc/MSXtar.cc
m9710797: OK.
It's also fine for me to not support such obscure corner cases. (Don't complicate the code for something that will never trigger in practice).

src/fdc/MSXtar.cc
m9710797: Ah OK.
This asymmetrical treatment of 0xFF6 was unexpected to me. Might be worth adding a comment.

src/fdc/MSXtar.cc
grauw: Also I just read on Wikipedia:

> This value must be adjusted so that directory entries always consume full logical sectors, given that each [directory entry](https://en.wikipedia.org/wiki/Design_of_the_FAT_file_system#DIR) takes up 32 bytes. MS-DOS/PC DOS require this value to be a multiple of 16. The maximum value supported on floppy disks is 240, the maximum value supported by MS-DOS/PC DOS on hard disks is 512.

So even though MSX-DOS does support odd-numbered directory entry counts (which I based my previous remark on), other systems do have this restriction, so all the more reason not to bother.

grauw:
Just rebased on top of master. The FAT16 support in MSXtar has been pushed to master. I will continue to work on the formatting and partitioning support. For that, these are my current plans:

- [x] Change the Extended boot sector type to Nextor, and match Nextor’s formatting for FAT12 volumes in this mode as well.
- [x] Change the DOS1 mode to format partitions so that they work under DOS1. Meaning that the max number of FAT sectors is 3, limiting the cluster count to 1022, and the volume size to 16 MB.
- [x] Take a look at the differences in the partition format between Sunrise and Nextor. It was mentioned that diskmanipulator couldn’t always handle Nextor drives, perhaps it’s related to that. (I’ve used it before with MFR without issues, but those were 32MB FAT12 volumes without partitions).
- [x] Determine good defaults for file system and partitioning format. Should it be DOS2? Nextor? Dynamic based on partition size? What of (non-Nextor) Sunrise IDE images, when the default mounted disk image size is 100 MB, a FAT16 partition will not work.
- [x] Update documentation & help texts.

grauw:
Here’s an updated series of changes which adds FAT16 support as well as Nextor partitioning support (standard MBR / EBR format).

I still need to do a final thorough test, since I did some git interactive rebasing to split things out into smaller digestible commits since my last test, but it’s ready for review.

src/fdc/DiskImageUtils.cc
```diff
@@ -77,21 +77,25 @@ struct SetBootSectorResult {
...
+	MSXBootSector& boot, MSXBootSectorType bootType, size_t nbSectors)
 {
 	// start from the default boot block ..
-	const auto& defaultBootBlock = dos1 ? BootBlocks::dos1BootBlock : BootBlocks::dos2BootBlock;
-	boot = defaultBootBlock.bootSector;
+	if (bootType == MSXBootSectorType::DOS1) {
+		boot = BootBlocks::dos1BootBlock.bootSector;
+	} else if (bootType == MSXBootSectorType::DOS2) {
+		boot = BootBlocks::dos2BootBlock.bootSector;
+	}
```
m9710797, line 174: Detail: this code makes me a bit nervous. What if `bootType` is neither of these two values?
I know that in this commit that isn't possible. But it's not robust against future changes (I haven't yet looked in later patches in this series how you adapt this code when the enum does get extended).

I'd suggest one of the following alternatives:
```
if (bootType == ...) {
} else {
    assert(bootType = ...)
}
```
```
if (bootType == ...) {
} else if (bootType = ...)
} else {
    UNREACHABLE;
}
```
```
switch (bootType) {
case ...
case ...
// NO default branch. And then rely on compiler warnings for unhandled enum values
```
My personal preference is the last alternative (if that's warning free in the 3 major compilers, I haven't checked this).

src/fdc/DiskImageUtils.hh
```diff
@@ -11,11 +11,16 @@ namespace openmsx {
 
 class SectorAccessibleDisk;
 
+enum class MSXBootSectorType : uint8_t {
```
m9710797: Very much a detail. No need to change, I was just wondering...

Why did you explicitly specify the underlying type here? It's only used as a parameter (not stored in a larger struct), so the default `int` should be fine, no?

On modern CPU architectures passing/manipulating `int` is generally at least as fast and sometimes faster than smaller types. So if there's no specific reason for a smaller type (e.g. less storage) than just use `int`.  (E.g. x86 16-bit instructions have a longer encoding than 8/32 bit instructions. ARM supports more addressing modes for 32-bit load/store instructions than for 8/16-bit).

m9710797 (commented):
DiskManipulator: Refactor DOS1 / DOS2 boot sector types.

Looks fine.
I made 2 remarks about some details you may (or may not) want to address.

m9710797 (commented):
DiskManipulator: Limit DOS1 volumes to 3 sectors per FAT. 

This commit is fine.

m9710797 (commented):
DiskManipulator: Extract Sunrise IDE-specific bits. 

OK.

m9710797 (commented):
DiskManipulator: Get max partitions from array. 

OK.

src/fdc/DiskManipulator.cc
```diff
@@ -406,7 +409,7 @@ void DiskManipulator::create(std::span<const TclObject> tokens)
 	if (sizes.empty()) {
 		throw CommandException("No size(s) given.");
 	}
-	if (sizes.size() > 1) {
+	if (sizes.size() > 1 && bootType != MSXBootSectorType::NEXTOR) {
 		// extra sector for partition table
 		++totalSectors;
 	}
```
m9710797, line 417: This is unexpected to me.
I understand that for a Sunrise partition table there's only one extra sector for the whole disk. While for Nextor there's one extra sector per partition (right?).
But then I would have expected to see: `totalSectors += sizes.size()`.

m9710797 (commented):
DiskManipulator: Support Nextor partitions.

From a C++ point of view it looks fine. But I'm not claiming to understand (and thus have checked) the details of the Nextor partition format.

m9710797 (commented):
DiskManipulator: Support Nextor FAT12 formatting. 

OK.

m9710797 (commented):
DiskManipulator: Add FAT16 support. 

OK.

src/fdc/DiskImageUtils.cc
grauw: Once you get to [the final submit](https://github.com/openMSX/openMSX/blob/11f7098bb67473cefde4df643534ba49a43c1634/src/fdc/DiskImageUtils.cc#L163), it is no longer a simple switch / case-able list, and similarly there is an if block later down in the function which is even less switch / case like. So option 2, in both places, seems best to me.

src/fdc/DiskImageUtils.hh
grauw: Oh, I thought it was mandatory for some reason, I thought I got a warning about it the other day. But I must have been mistaken and there was another issue.

src/fdc/DiskManipulator.cc
grauw: So, what I want to do is to change the size definition for partitions to include the boot sector. It is a bit of an implementation or UX choice that can go either way, however when I ask for a disk that is 32 MB + 32 MB, I expect to get a disk that is 64 MB, not 64 MB + 512 bytes.

And it is a choice, right. We don’t add an additional 512 bytes for the boot sector either, right. And the volume space reported by DIR is also lower than you requested, so why don’t we increase the volume size so that that matches up.

There will usually be some padding at the end of the partition anyway due to the cluster size (reflected in the sector count in the boot sector being lower), so there’s already a precedent that the size you get back is not exactly what you requested. And in terms of cluster count available on the volume it often won’t even matter if there’s one less sector. So in other words, it doesn’t feel strange to me if (some) partitions are 1 sector smaller than specified.

An additional reason why I don’t like this `++totalSectors` in DiskManipulator.cc much is because now here you’ve got to add knowledge about an implementation detail of the formatting that is supposed to be handled completely by another subsystem, and if you don’t do it right you hit an exception, possibly even only much later when you completely fill the disk image and it tries to access a sector that doesn’t exist. And I don’t really like the possible alternative to add a function to query the computed size to DiskImageUtils either.

Perhaps it makes more sense in the context of a later submit which I will send in as a separate series, in which I add a "partition" command (ref issue #170). For that I need to introduce the ability to clamp the partition size if it is larger than what is available (as well as produce less partitions than requested), because if it wouldn’t do that it would be difficult (and annoying) as a user to specify partition sizes that exactly add up to the disk image size. With that, I can then also remove this check and `++totalSectors` from here easily and the last partition (with Sunrise MBR) will just be reduced by 1 sector automatically.

grauw: Note partitionNextor does:

    p.start = 1;
    p.size = size - 1;

m9710797: Makes sense.

Though I should have been more clear: my remark was not so much about the missing `totalSectors += ...` stuff. But about the difference between how Nextor and Sunrise partition tables are accounted. So instead of adding `totalSectors += ...` for Nextor we could also make it uniform by removing `++totalSectors` for Sunrise.

I understand that the latter is your plan?

grauw: Yeah, that’s my intention.

I didn’t want to complicate this patch series with that so I kept it to adding that check, but I figured you’d probably pick up on it anyway ;).

grauw:
Ok, pushed.

About the defaults, for now I’m keeping things as they are (Sunrise IDE, DOS2, FAT12), and the Nextor partitioning and formatting is opt-in.

m9710797:
Thanks!

MBilderbeek:
Note that current master gives me these new warnings:
```
src/fdc/DiskManipulator.cc:134:34: warning: no previous declaration for ‘std::optional<openmsx::MSXBootSectorType> openmsx::parseBootSectorType(std::string_view)’ [-Wmissing-declarations]
  134 | std::optional<MSXBootSectorType> parseBootSectorType(std::string_view s) {
      |                                  ^~~~~~~~~~~~~~~~~~~
Compiling fdc/MSXtar.cc...
Compiling fdc/NowindCommand.cc...
Compiling fdc/NowindHost.cc...
src/fdc/MSXtar.cc: In lambda function:
src/fdc/MSXtar.cc:635:61: warning: declaration of ‘cluster’ shadows a previous local [-Wshadow]
  635 |                                                 [&](Cluster cluster) { writeFAT(cluster, newCl); }
      |                                                     ~~~~~~~~^~~~~~~
src/fdc/MSXtar.cc:626:25: note: shadowed declaration is here
  626 |                 Cluster cluster;
      |                         ^~~~~~~
src/fdc/MSXtar.cc: In lambda function:
src/fdc/MSXtar.cc:640:44: warning: declaration of ‘cluster’ shadows a previous local [-Wshadow]
  640 |                                 [](Cluster cluster) { return cluster; }
      |                                    ~~~~~~~~^~~~~~~
src/fdc/MSXtar.cc:626:25: note: shadowed declaration is here
  626 |                 Cluster cluster;
      |                         ^~~~~~~
```
@grauw Can you please fix them?

grauw:
Will do. Weird that the CI didn’t pick up on it.

Edit: Right, it did, it just didn’t fail the build on it. I thought I got a CI build fail on a shadowing warning for Windows earlier, but perhaps the configuration is different.

src/fdc/DiskImageUtils.hh
grauw: Ah I probably know, I think Unreal Engine’s reflection system requires you to be explicit about the type so I must’ve picked up the habit at work.

PR #1477, opened 2023-01-11
MSXtar refactors

grauw:
Here are a few more MSXtar commits I’ve got ready.

Opening as a draft pull request for review separate from #1475, since they’re not directly related to the FAT16 support.

I’m especially curious what you think about the commit “Introduce Cluster class”. It was a bit of an experiment, to try to make conversions between cluster numbers and indices explicit, and to give the compiler a bit more to work with for checking program validity.

p.s. I haven’t had the time to thoroughly test yet, and I haven’t rebased the FAT16 stuff on top of this either.

### DIFF
--- a/src/fdc/MSXtar.cc
+++ b/src/fdc/MSXtar.cc
@@ -35,12 +35,25 @@ using FAT::Cluster;
 using FAT::DirCluster;
 using FAT::FatCluster;
 
-static constexpr unsigned FAT12_MAX_CLUSTER_COUNT = 0xFF4;
+namespace FAT {
+	static constexpr unsigned FREE = 0x000;
+	static constexpr unsigned FIRST_CLUSTER = 0x002;
+}
 
-static constexpr unsigned FREE_FAT = 0x000;
-static constexpr unsigned FIRST_CLUSTER = 0x002;
-static constexpr unsigned BAD_FAT = 0xFF7;
-static constexpr unsigned EOF_FAT = 0xFFF; // actually 0xFF8-0xFFF, signals EOF in FAT12
+namespace FAT12 {
+	static constexpr unsigned BAD = 0xFF7;
+	static constexpr unsigned END_OF_CHAIN = 0xFFF; // actually 0xFF8-0xFFF, signals EOF in FAT12
+
+	static constexpr unsigned MAX_CLUSTER_COUNT = 0xFF4;
+
+	// Functor to convert a FatCluster or DirCluster to a FAT12 cluster number
+	struct ToClusterNumber {
+		unsigned operator()(Free) const { return FAT::FREE; }
+		unsigned operator()(EndOfChain) const { return END_OF_CHAIN; }
+		unsigned operator()(Cluster cluster) const { return FAT::FIRST_CLUSTER + cluster.index; }
+	} toClusterNumber;
+}
+
 static constexpr unsigned SECTOR_SIZE = sizeof(SectorBuffer);
 static constexpr unsigned DIR_ENTRIES_PER_SECTOR = SECTOR_SIZE / sizeof(MSXDirEntry);
 
@@ -55,13 +68,6 @@ static constexpr uint8_t T_MSX_ARC  = 0x20; // Archive bit
 // to store a long Unicode file name.
 // For details, read http://home.teleport.com/~brainy/lfn.htm
 static constexpr uint8_t T_MSX_LFN  = 0x0F; // LFN entry (long files names)
-
-// Functor to convert a FatCluster or DirCluster to a FAT12 cluster number
-struct ToClusterNumberFAT12 {
-	unsigned operator()(Free) const { return FREE_FAT; }
-	unsigned operator()(EndOfChain) const { return EOF_FAT; }
-	unsigned operator()(Cluster cluster) const { return FIRST_CLUSTER + cluster.index; }
-} toClusterNumberFAT12;
 
 /** Transforms a cluster number towards the first sector of this cluster
   * The calculation uses info read fom the boot sector
@@ -121,12 +127,12 @@ void MSXtar::parseBootSector(const MSXBootSector& boot)
 	}
 
 	clusterCount = std::min((boot.nrSectors - dataStart) / sectorsPerCluster,
-	                        FAT12_MAX_CLUSTER_COUNT);
+	                        FAT12::MAX_CLUSTER_COUNT);
 
 	// Some (invalid) disk images have a too small FAT to be able to address
 	// all clusters of the image. OpenMSX SVN revisions pre-11326 even
 	// created such invalid images for some disk sizes!!
-	unsigned fatCapacity = (2 * SECTOR_SIZE * sectorsPerFat) / 3 - FIRST_CLUSTER;
+	unsigned fatCapacity = (2 * SECTOR_SIZE * sectorsPerFat) / 3 - FAT::FIRST_CLUSTER;
 	clusterCount = std::min(clusterCount, fatCapacity);
 }
 
@@ -217,19 +223,19 @@ FatCluster MSXtar::readFAT(Cluster cluster) const
 
 	std::span<const uint8_t> data{fatBuffer[0].raw.data(), sectorsPerFat * size_t(SECTOR_SIZE)};
 
-	unsigned index = FIRST_CLUSTER + cluster.index;
+	unsigned index = FAT::FIRST_CLUSTER + cluster.index;
 	auto p = subspan<2>(data, (index * 3) / 2);
 	unsigned value = (index & 1)
 	                ? (p[0] >> 4) + (p[1] << 4)
 	                : p[0] + ((p[1] & 0x0F) << 8);
 
 	// Be tolerant when reading:
-	// * FREE_FAT is returned as FREE_FAT.
-	// * Anything else but a valid cluster number is returned as EOF_FAT.
-	if (value == FREE_FAT) {
+	// * FREE is returned as FREE.
+	// * Anything else but a valid cluster number is returned as END_OF_CHAIN.
+	if (value == FAT::FREE) {
 		return Free{};
-	} else if (value >= FIRST_CLUSTER && value < FIRST_CLUSTER + clusterCount) {
-		return Cluster{value - FIRST_CLUSTER};
+	} else if (value >= FAT::FIRST_CLUSTER && value < FAT::FIRST_CLUSTER + clusterCount) {
+		return Cluster{value - FAT::FIRST_CLUSTER};
 	} else {
 		return EndOfChain{};
 	}
@@ -242,13 +248,13 @@ void MSXtar::writeFAT(Cluster cluster, FatCluster value)
 	assert(cluster.index < clusterCount);
 
 	// Be strict when writing:
-	// * Anything but FREE_FAT, EOF_FAT or a valid cluster number is rejected.
+	// * Anything but FREE, END_OF_CHAIN or a valid cluster number is rejected.
 	assert(!std::holds_alternative<Cluster>(value) || std::get<Cluster>(value).index < clusterCount);
-	unsigned fatValue = std::visit(toClusterNumberFAT12, value);
+	unsigned fatValue = std::visit(FAT12::toClusterNumber, value);
 
 	std::span data{fatBuffer[0].raw.data(), sectorsPerFat * size_t(SECTOR_SIZE)};
 
-	unsigned index = FIRST_CLUSTER + cluster.index;
+	unsigned index = FAT::FIRST_CLUSTER + cluster.index;
 	auto p = subspan<2>(data, (index * 3) / 2);
 	if (index & 1) {
 		p[0] = narrow_cast<uint8_t>((p[0] & 0x0F) + (fatValue << 4));
@@ -300,10 +306,10 @@ unsigned MSXtar::getNextSector(unsigned sector)
 DirCluster MSXtar::getStartCluster(const MSXDirEntry& entry) const
 {
 	// Be tolerant when reading:
-	// * Anything but a valid cluster number is returned as FREE_FAT.
+	// * Anything but a valid cluster number is returned as FREE.
 	unsigned cluster = entry.startCluster;
-	if (cluster >= FIRST_CLUSTER && cluster < FIRST_CLUSTER + clusterCount) {
-		return Cluster{cluster - FIRST_CLUSTER};
+	if (cluster >= FAT::FIRST_CLUSTER && cluster < FAT::FIRST_CLUSTER + clusterCount) {
+		return Cluster{cluster - FAT::FIRST_CLUSTER};
 	} else {
 		return Free{};
 	}
@@ -313,9 +319,9 @@ DirCluster MSXtar::getStartCluster(const MSXDirEntry& entry) const
 void MSXtar::setStartCluster(MSXDirEntry& entry, DirCluster cluster) const
 {
 	// Be strict when writing:
-	// * Anything but FREE_FAT or a valid cluster number is rejected.
+	// * Anything but FREE or a valid cluster number is rejected.
 	assert(!std::holds_alternative<Cluster>(cluster) || std::get<Cluster>(cluster).index < clusterCount);
-	entry.startCluster = narrow<uint16_t>(std::visit(toClusterNumberFAT12, cluster));
+	entry.startCluster = narrow<uint16_t>(std::visit(FAT12::toClusterNumber, cluster));
 }
 
 // If there are no more free entries in a subdirectory, the subdir is
--- a/src/fdc/MSXtar.cc
+++ b/src/fdc/MSXtar.cc
@@ -29,6 +29,12 @@ using std::string_view;
 
 namespace openmsx {
 
+using FAT::Free;
+using FAT::EndOfChain;
+using FAT::Cluster;
+using FAT::DirCluster;
+using FAT::FatCluster;
+
 static constexpr unsigned FAT12_MAX_CLUSTER_COUNT = 0xFF4;
 
 static constexpr unsigned FREE_FAT = 0x000;
@@ -50,26 +56,30 @@ static constexpr uint8_t T_MSX_ARC  = 0x20; // Archive bit
 // For details, read http://home.teleport.com/~brainy/lfn.htm
 static constexpr uint8_t T_MSX_LFN  = 0x0F; // LFN entry (long files names)
 
+// Functor to convert a FatCluster or DirCluster to a FAT12 cluster number
+struct ToClusterNumberFAT12 {
+	unsigned operator()(Free) const { return FREE_FAT; }
+	unsigned operator()(EndOfChain) const { return EOF_FAT; }
+	unsigned operator()(Cluster cluster) const { return FIRST_CLUSTER + cluster.index; }
+} toClusterNumberFAT12;
+
 /** Transforms a cluster number towards the first sector of this cluster
   * The calculation uses info read fom the boot sector
   */
-unsigned MSXtar::clusterToSector(unsigned cluster) const
+unsigned MSXtar::clusterToSector(Cluster cluster) const
 {
-	// check lower bound since unsigned can't represent negative numbers
-	// don't check upper bound since it might be used in calculations
-	assert(cluster >= FIRST_CLUSTER);
-	return dataStart + sectorsPerCluster * (cluster - FIRST_CLUSTER);
+	return dataStart + sectorsPerCluster * cluster.index;
 }
 
 /** Transforms a sector number towards it containing cluster
   * The calculation uses info read fom the boot sector
   */
-unsigned MSXtar::sectorToCluster(unsigned sector) const
+Cluster MSXtar::sectorToCluster(unsigned sector) const
 {
 	// check lower bound since unsigned can't represent negative numbers
 	// don't check upper bound since it is used in calculations like getNextSector
 	assert(sector >= dataStart);
-	return FIRST_CLUSTER + (sector - dataStart) / sectorsPerCluster;
+	return Cluster{(sector - dataStart) / sectorsPerCluster};
 }
 
 
@@ -200,13 +210,14 @@ MSXtar::~MSXtar()
 }
 
 // Get the next cluster number from the FAT chain
-unsigned MSXtar::readFAT(unsigned index) const
+FatCluster MSXtar::readFAT(Cluster cluster) const
 {
 	assert(!fatBuffer.empty()); // FAT must already be cached
-	assert(index >= FIRST_CLUSTER && index < FIRST_CLUSTER + clusterCount);
+	assert(cluster.index < clusterCount);
 
 	std::span<const uint8_t> data{fatBuffer[0].raw.data(), sectorsPerFat * size_t(SECTOR_SIZE)};
 
+	unsigned index = FIRST_CLUSTER + cluster.index;
 	auto p = subspan<2>(data, (index * 3) / 2);
 	unsigned value = (index & 1)
 	                ? (p[0] >> 4) + (p[1] << 4)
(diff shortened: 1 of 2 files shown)


m9710797 (commented):
MSXtar: Account for reserved sectors. 

This commit is fine. Thanks.

m9710797 (commented):
MSXtar: Replace rootDirLast by dataStart field. 

This commit is fine. Nice cleanup. Thanks.

src/fdc/MSXtar.cc
```diff
@@ -29,6 +29,8 @@ using std::string_view;
 
 namespace openmsx {
 
+static constexpr unsigned FAT12_MAX_CLUSTER_COUNT = 0xFF4;
+
```
m9710797: I agree with with this commit.
But it's unfortunate that this new value 0xFF4 cannot immediately be found in (most) documentation about the FAT filesystem. Maybe expand this to:
```
static constexpr unsigned BAD_FAT = 0xFF6;
static constexpr unsigned FAT12_MAX_CLUSTER_COUNT = BAD_FAT - FIRST_CLUSTER;
```

src/fdc/MSXtar.cc
```diff
@@ -105,16 +107,14 @@ void MSXtar::parseBootSector(const MSXBootSector& boot)
 	chrootSector = rootDirStart;
 	dataStart = rootDirStart + nbRootDirSectors;
 
-	// Bypass max cluster check in sectorToCluster
-	maxCluster = std::numeric_limits<decltype(maxCluster)>::max();
-	maxCluster = sectorToCluster(boot.nrSectors);
-	maxCluster = std::min(maxCluster, BAD_FAT);
+	clusterCount = (boot.nrSectors - dataStart) / sectorsPerCluster;
+	clusterCount = std::min(clusterCount, FAT12_MAX_CLUSTER_COUNT);
```
m9710797: Detail, style issue. This is a more a comment on the existing code (prior to your change).
I don't like reassigning the same variable. In this case it's equally clear to write this as:
```
clusterCount = std::min((boot.nrSectors - dataStart) / sectorsPerCluster,
                        FAT12_MAX_CLUSTER_COUNT);
```

m9710797 (commented):
MSXtar: Store clusterCount instead of maxCluster. 

This is fine. I have some small remarks. Feel free to apply or to ignore.

src/fdc/MSXtar.cc
```diff
@@ -264,13 +264,20 @@ unsigned MSXtar::getNextSector(unsigned sector)
...
 
-// get start cluster from a directory entry,
-// also takes care of BAD_FAT and EOF_FAT-range.
-static unsigned getStartCluster(const MSXDirEntry& entry)
+// Get start cluster from a directory entry.
+// BAD FAT and EOF FAT-range is normalized to a single EOF_FAT value.
+unsigned MSXtar::getStartCluster(const MSXDirEntry& entry) const
 {
 	return normalizeFAT(entry.startCluster);
 }
```
m9710797, line 316: Technically this could remain a non-member function, right?
But if you want to keep it symmetrical with the new setStartCluster() member function, then that's fine for me.

src/fdc/MSXtar.hh
```diff
@@ -66,6 +66,9 @@ private:
 	void chroot(std::string_view newRootDir, bool createDir);
 
 private:
+	unsigned getStartCluster(const MSXDirEntry& entry) const;
```
m9710797: Consider annotating this function with `[[nodiscard]]`.
I didn't check, maybe other functions could be annotated as well?

m9710797 (commented):
MSXtar: Add setStartCluster helper method. 

This commit is fine. Small suggestion, feel free to apply or to ignore.

src/fdc/MSXtar.cc
```diff
@@ -193,38 +193,45 @@ MSXtar::~MSXtar()
...
-{
-	return (cluster < BAD_FAT) ? cluster : EOF_FAT;
-}
-
 // Get the next cluster number from the FAT chain
-unsigned MSXtar::readFAT(unsigned clNr) const
+unsigned MSXtar::readFAT(unsigned index) const
 {
 	assert(!fatBuffer.empty()); // FAT must already be cached
+	assert(index >= FIRST_CLUSTER && index < FIRST_CLUSTER + clusterCount);
```
m9710797: This subexpression: `(FIRST_CLUSTER <= index) && (index < (FIRST_CLUSTER + clusterCount))`
is repeated 5x in this commit (maybe more if I look at the full code?)
Maybe factor it out in a `isValidClusterIndex()` helper function? (Or `clusterIndexInRange()`?)

m9710797 (commented):
MSXtar: Tolerant when reading, strict when writing. 

Very useful commit message. Might be worth repeating this somewhere in the code itself?? Though the comments in `readFAT()` and `writeFAT()` are also fine (good enough already?).

This commit is fine. I've made one suggestion to factor out `isValidClusterIndex()`, but I haven't yet checked how that would interact with the last commit in this series which introduces the `Cluster` class. So feel free to apply or to ignore.

m9710797 (commented):
MSXtar: Don’t check cluster count in clusterToSector. 

Agreed.
Might be worth documenting this in the code as well (could be a single line of comment).
Otherwise in a few years we might wonder why the code only checks the lower-bound.

src/fdc/MSXtar.cc
```diff
@@ -104,6 +104,10 @@ void MSXtar::parseBootSector(const MSXBootSector& boot)
 	chrootSector = rootDirStart;
 	dataStart = rootDirStart + nbRootDirSectors;
 
+	if (dataStart + sectorsPerCluster > boot.nrSectors) {
+		throw MSXException("Cluster count too small: ", (int(boot.nrSectors) - dataStart) / sectorsPerCluster);
+	}
+
 	clusterCount = (boot.nrSectors - dataStart) / sectorsPerCluster;
 	clusterCount = std::min(clusterCount, FAT12_MAX_CLUSTER_COUNT);
```
m9710797: Suggestion for a simplification: move the check down two lines and write it as `if (clusterCount <= 0)`.
Ah, but `clusterCount` is an unsigned variable so that won't work. Is that the reason why you wrote it like this?

The error message itself may not be the best. It will either print 0 or a very large value. Maybe formulate it as "Too few sectors: n"?

src/fdc/DiskImageUtils.cc
```diff
@@ -189,34 +192,42 @@ static SetBootSectorResult setBootSector(
 	unsigned nbRootDirSectors = nbDirEntry / 16;
 	unsigned rootDirStart = nbReservedSectors + nbFats * nbSectorsPerFat;
 	unsigned firstDataSector = rootDirStart + nbRootDirSectors;
-	return {nbReservedSectors, firstDataSector, descriptor};
+	return {nbFats, nbSectorsPerFat, nbReservedSectors, rootDirStart, firstDataSector, descriptor};
 }
```
m9710797: I don't really like this. It's fine for 2 or 3 return values. But for 6 values it becomes too easy to accidentally enumerate them in the wrong order. And the compiler can't check that indeed `nbFats` corresponds to `fatCount` etc.    ... Euhm, and I think the order is indeed already wrong :-)     You swapped the first two.

I'd suggest to define a variable `SetBootSectorResult result` inside `setBootSector()`, fill in the fields, and return that variable.

src/fdc/DiskImageUtils.cc
```diff
@@ -189,34 +192,42 @@ static SetBootSectorResult setBootSector(
...
 
 void format(SectorAccessibleDisk& disk, bool dos1)
 {
 	// first create a boot sector for given partition size
 	size_t nbSectors = disk.getNbSectors();
 	SectorBuffer buf;
-	auto [firstFATSector, firstDataSector, descriptor] =
+	auto [fatCount, sectorsPerFat, fatStart, rootDirStart, dataStart, descriptor] =
```
m9710797: Similar here. It's fine for a few variables, but maybe not for 6.
I'd suggest you do `auto bootInfo = setBootSector(...);` and then use stuff like `bootInfo.fatCount`.

Hehe, you also swapped the first two here, so the error cancels out ;-)

src/fdc/DiskImageUtils.cc
```diff
@@ -189,34 +192,42 @@ static SetBootSectorResult setBootSector(
...
 	disk.writeSector(0, buf);
 
 	// write empty FAT and directory sectors
 	ranges::fill(buf.raw, 0);
-	for (auto i : xrange(firstFATSector + 1, firstDataSector)) {
+	for (auto i : xrange(1u, sectorsPerFat)) {
+		for (auto fat : xrange(fatCount)) {
+			disk.writeSector(i + fatStart + fat * sectorsPerFat, buf);
+		}
+	}
```
m9710797: Suggestion, I haven't fully worked it out. Might be better or might not be.

You can swap the order of the inner and outer loop. Then the disk is accessed in a more linear pattern. Speed is not really important here, but it also allows to integrate the loop below that writes the first FAT sector.

Or maybe not because that requires changing the content of `buf` more often. And that might be more ugly.

src/fdc/MSXtar.cc
```diff
@@ -186,10 +188,12 @@ MSXtar::~MSXtar()
...
-			disk.writeSector(i + fatStart, fatBuffer[i]);
-		} catch (MSXException&) {
-			// nothing
+		for (auto fat : xrange(fatCount)) {
+			try {
+				disk.writeSector(i + fatStart + fat * sectorsPerFat, fatBuffer[i]);
+			} catch (MSXException&) {
+				// nothing
+			}
 		}
```
m9710797: Just checking. You unconditionally overwrite the 2nd FAT with the 1st FAT. Internally we only keep and update a single FAT copy. Right?

(Detail 1) Also here you could swap the order of the inner and outer loops.
(Detail 2) We could use the SectorAccessibleDisk::writeSectors() function to write all sectors (of a single FAT) at once. (Was already possible before your change).

src/fdc/MSXtar.cc
```diff
@@ -50,22 +50,56 @@ static constexpr uint8_t T_MSX_ARC  = 0x20; // Archive bit
 // For details, read http://home.teleport.com/~brainy/lfn.htm
 static constexpr uint8_t T_MSX_LFN  = 0x0F; // LFN entry (long files names)
```
m9710797: I 100% agree with the idea of the patch. But I do not like the API of the new `Cluster` class, nor the implementation.

I would suggest the following:
```
class Cluster
{
private:
	explicit Cluster(unsigned value) : val(value) {}

public:
    [[nodiscard]] static Cluster FREE()         { return Cluster(FREE_FAT); }
	[[nodiscard]] static Cluster END_OF_CHAIN() { return Cluster(EOF_FAT); }

	[[nodiscard]] static Cluster fromIndex(unsigned index) {
        assert(index + FIRST_CLUSTER < BAD_FAT);
        return Cluster(index + FIRST_CLUSTER);
    }
	[[nodiscard]] static Cluster fromValue(unsigned value, unsigned clusterCount) {
        assert(clusterCount < BAD_FAT);
        if (value == 0) {
		    return FREE();
	    } else if (value < FIRST_CLUSTER || value >= clusterCount) {
		    return END_OF_CHAIN();
	    } else {
            return Cluster(value);
        }
    }

	[[nodiscard]] bool isValidIndex() const {
        return (FIRST_CLUSTER <= val) && (val < BAD_FAT);
    }
	[[nodiscard]] unsigned index() const {
        assert(isValidIndex());
        return val - FIRST_CLUSTER;
    }

	[[nodiscard]] unsigned value() const {
        return val;
    }

	[[nodiscard]] bool operator==(const Cluster& other) const = default;

private:
	unsigned val;
};
```

Most important items:
* There are no public constructors. Instead a `Cluster` object must be obtained either via `Cluster::fromIndex()` or `Cluster::fromValue()`. (Or via `Cluster::FREE()` or `Cluster::END_OF_CHAIN()`).
* Internally the class stores a single `unsigned` value. There's an invariant that this is either a valid index or FREE_FAT or EOF_FAT. (So this makes the separate `type` enum obsolete).
* There are only two queries on this class: `index()` and `value()`. The former has a precondition that `isValidIndex()` must be true (OK, this is the 3rd query method). The latter simply returns the internal value.
* operator== can now simply compare the internal value.

I'll give a few examples below for how this can replace your implementation.

src/fdc/MSXtar.cc
```diff
@@ -50,22 +50,56 @@ static constexpr uint8_t T_MSX_ARC  = 0x20; // Archive bit
...
 /** Transforms a cluster number towards the first sector of this cluster
   * The calculation uses info read fom the boot sector
   */
-unsigned MSXtar::clusterToSector(unsigned cluster) const
+unsigned MSXtar::clusterToSector(Cluster cluster) const
 {
-	assert(cluster >= FIRST_CLUSTER);
-	return dataStart + sectorsPerCluster * (cluster - FIRST_CLUSTER);
+	assert(cluster.get());
+	return dataStart + sectorsPerCluster * *cluster.get();
```
m9710797: This becomes: `return dataStart + sectorsPerCluster() * cluster.index();`
The `assert` is implicit in the usage of `index()`.

src/fdc/MSXtar.cc
```diff
@@ -50,22 +50,56 @@ static constexpr uint8_t T_MSX_ARC  = 0x20; // Archive bit
...
 /** Transforms a sector number towards it containing cluster
   * The calculation uses info read fom the boot sector
   */
-unsigned MSXtar::sectorToCluster(unsigned sector) const
+Cluster MSXtar::sectorToCluster(unsigned sector) const
 {
 	assert(sector >= dataStart);
-	return FIRST_CLUSTER + (sector - dataStart) / sectorsPerCluster;
+	return Cluster((sector - dataStart) / sectorsPerCluster);
 }
```
m9710797, line 89: Becomes: `return Cluster::fromIndex((sector - dataStart) / sectorsPerCluster);`

src/fdc/MSXtar.cc
```diff
@@ -199,55 +233,59 @@ MSXtar::~MSXtar()
...
 
 // Get the next cluster number from the FAT chain
-unsigned MSXtar::readFAT(unsigned index) const
+Cluster MSXtar::readFAT(Cluster index) const
 {
 	assert(!fatBuffer.empty()); // FAT must already be cached
-	assert(index >= FIRST_CLUSTER && index < FIRST_CLUSTER + clusterCount);
+	assert(index.get() && *index.get() < clusterCount);
```
m9710797: Becomes: `assert(index.index() < clusterCount);`

src/fdc/MSXtar.cc
```diff
@@ -199,55 +233,59 @@ MSXtar::~MSXtar()
...
 
 	std::span<const uint8_t> data{fatBuffer[0].raw.data(), sectorsPerFat * size_t(SECTOR_SIZE)};
 
-	auto p = subspan<2>(data, (index * 3) / 2);
-	unsigned value = (index & 1)
+	unsigned fatIndex = index.toClusterNumber();
```
m9710797: This becomes: `... = index.value();`

m9710797 (commented):
I think it's a very good idea. Stronger types are always good. But I think the implementation could be improved. And the API could be simplified.

I made a suggestion in the review comments. But I must admit that it's something I came up with in 5 minutes. So I might have overlooked some detail.

Maybe this needs further discussion? Possibly on IRC?

src/fdc/MSXtar.cc
grauw: It’s indeed a choice to express which value is derived from which. In the end I felt a plain cluster count was the most clear to the reader. It is mentioned amongst others in [size limits on Wikipedia](https://en.wikipedia.org/wiki/Design_of_the_FAT_file_system#Size_limits).

p.s. `BAD_FAT` is `0xFF7`, while `0xFF6` is an unnamed reserved cluster number. I could name that `RES_FAT`, but then again `0x001` is also reserved so `RES1_FAT` and `RES2_FAT`? Instead of trying to name all special values, doing it like this felt simpler.

src/fdc/MSXtar.cc
grauw: I’ll do it like that, but just a side note;

I’m not a huge fan of indenting by spaces, because you often need to reindent when you e.g. rename stuff and such, and it’s also easy to forget and then it gets out of whack, or accidentally use tabs instead of spaces, etc. Same applies to lining up comments at the end of a line. So in my personal style I would write it like:

    clusterCount = std::min(
        (boot.nrSectors - dataStart) / sectorsPerCluster,
        FAT12_MAX_CLUSTER_COUNT
    );

src/fdc/MSXtar.cc
grauw: It will need to be a member function for the "Tolerant when reading" patch anyway, so might as well do it now already and keep the two methods consistent.

src/fdc/MSXtar.cc
grauw: Yes that’s the reason.

When you say a very large value, I probably got my signed / unsigned implicit conversion rules wrong? Since the intent was that the initial `int()` would make the rest of the expression operate on signed numbers.

Anyway I’ll reformulate it, yes.

src/fdc/DiskImageUtils.cc
grauw: Ah sorry I meant to do that, but rushed the commit and forgot that final step. Funny how it immediately goes wrong, too :).

src/fdc/DiskImageUtils.cc
grauw: Yes, although I thought it would look nicer to have a single loop with an if (i == 0) in it, I didn’t do it for the latter reason.

src/fdc/MSXtar.cc
grauw: Yes, it keeps the two in sync when it writes it out. But if the FAT has never been written to (only read operations like `diskmanipulator export`) then it won’t write them.

src/fdc/MSXtar.cc
grauw: Indeed, it’s factored out later in the Cluster class commit.

src/fdc/MSXtar.cc
m9710797: Ah, I overlooked that `int()` conversion. For some reason I thought the calculation was identical to the one on line 111.

But `(<int> <binary-operation> <unsigned>)` results in an `unsigned` (and I'm assuming `dataStart` and/or `sectorsPerCluster` is `unsigned`?).

grauw: Yeah it’s probably wrong, I should’ve looked up the implicit conversion rules. I rushed it at the last minute before going to bed. Anyway, changed now to your suggestion to just call out the sector count as being too few.

src/fdc/MSXtar.cc
grauw: So, the idea behind the Cluster implementation was to implement a (tagged) typed union, which represents all values completely independently in the most natural way. Since C++ doesn’t really have such a built in concept (well, could’ve used polymorphism, but I wanted to keep it simple), it needs to be explicitly tagged with an enum, where the first two are atomic types and the last has a value.

(In Haskell it would be `data ClusterNumber a : Free | EndOfChain | Cluster a`.)

What I do like about this implementation is that it has the special values and the integer completely separated. The fields have their own meaning and only that. No magic values. This way, there is no way the value can somehow overflow and mean something completely different, which gives it a level of safeness that asserts can’t provide. Which addresses an aspect of the code that I didn’t like before this change.

After I got it working, I thought about two possible ways to optimise:

1. Use bit fields to store the value in the space of a single integer

I played around a bit in compiler explorer and it did not really result in especially nice assembly code, it was not more efficient than the original. So that’s just a size optimisation, and hardly even that because the struct is passed around in registers anyway.

2. Combine the value in a single integer as you suggest to do.

I felt it was an optimisation that would optimise the comparison but not much else, and it would make the code more complicated to read. More importantly it would remove that property of safeness from overflowing.

Note that for `FREE` and `END_OF_CHAIN` I would use magic values far away from the cluster index (like `~0` and `~1`), so that makes the code more complicated than yours. Because FAT12 and FAT16 use different magic values for END_OF_CHAIN, while this object is oblivious of the FAT type, it’s got its own magic values. These far-away values would perhaps also reduce the chance of an overflow changing the type (though it’s still easy for an underflow).

I used std::optional() because it puts a stronger pressure on the programmer to check the presence of a return value before using it, even though it didn’t gel super well with me syntactically. But semantically, the meaning is, the value exists, or it does not, because it’s a different type.

So finally about your suggestion, I think it removes aspects of the design intent of this type, without which I’m not sure I still see a benefit in having this abstraction. I feel it might be clearer and simpler the way it was before.

src/fdc/MSXtar.cc
grauw: Re. detail 2, how do I convert a MemBuffer to a span?

src/fdc/MSXtar.cc
grauw: In my mind this type is similar to `std::optional` in terms of its semantics as a tagged type union.

Optional represents either the value or bottom (aka "nothing", "nil", "undefined", "error", "uncomputable").

This type represents either Cluster(index) or Free or EndOfChain.

...and I just found out about [`std::variant`](https://en.cppreference.com/w/cpp/utility/variant), which could maybe be a better way to represent this. But it can’t represent empty types I think… Or maybe it can. Let me try that.

grauw:
I tested and pushed the changes that were good to master.

I also pushed a few experimentations with changes to Cluster (now ClusterNumber, in its own file, some interface changes), as well as tried rebasing the FAT16 support on top of that to see how it would look.

I will now go do an experiment to see what it will look like if I make it based on std::variant.

Let’s catch each other on IRC tomorrow maybe, or later this weekend.

src/fdc/MSXtar.cc
m9710797: `MemBuffer` is very lightweight, it doesn't know the size of the buffer it owns. You can think of `std::span` as the combination of a (non-owning) pointer and a size. So you need to construct a `span` by combining the pointer obtained from `fatBuffer` with `sectorsPerFat`. Maybe something like this (not tested): `std::span{fatBuffer.data(), sectorsPerFat}`

src/fdc/MSXtar.cc
m9710797: Yes, the idiomatic way to implement a tagged-union in c++ is via `std::variant`.
Here's an example of how to do that:   https://godbolt.org/z/MsGvKsMYe
Under the hood a std::variant is much like the code you implemented manually (internally it has a small integer to indicate the active alternative plus the (union of the) members of those different alternatives).

I can see the appeal of why you want to model it like this.
Personally I wouldn't do it like this, because the thing you want to model really is an integer (with special meaning for certain (ranges of) values). But you decide, either is fine for me.
I must also admit that I haven't yet thought about how this class will change with future FAT16 support.

(Maybe one small thing to try to convince you: if in my implementation you store the member as a `Endian::L16` then you could use this class directly in `MSXDirEntry`.)

src/fdc/MSXtar.cc
```diff
@@ -202,16 +218,26 @@ ClusterNumber MSXtar::readFAT(ClusterNumber index) const
...
 	// * FREE_FAT is returned as FREE_FAT.
 	// * Anything else but a valid cluster number is returned as EOF_FAT.
-	return ClusterNumber::fromFAT12(value, clusterCount);
+	if (fat16) {
+		unsigned fatIndex = index.toFAT16();
+
+		auto p = subspan<2>(data, fatIndex * 2);
+		unsigned value = p[0] | p[1] << 8;
+
+		return ClusterNumber::fromFAT12(value, clusterCount);
```
m9710797: Typo: fromFAT**12**

src/fdc/MSXtar.cc
m9710797: UPDATE: I read your FAT16 follow-up patches. I can see now how the tagged-union approach makes more code independent from FAT12 vs FAT16 (the decision for how to represent "end-of-chain" can be postponed). So that's a bonus.

I'm personally undecided which approach I like better, So again, I'll let you decide.

grauw:
Yes, forget about my crappy attempt at a tagged union type. `std::variant` is where it’s at.

It gives more type safety, I was able to find three cases where the return value was not checked against special values (in `fileExtract`, `recurseDirExtract` and `alterFileInDSK`).

And it is more optimal as well, since I can use the simple `Cluster` type directly where I only expect a cluster index, specifically in `clusterToSector`, `sectorToCluster`, `findFirstFreeCluster`, `readFAT` and `writeFAT`.

…which also removes the need for some assertions, these are now statically guaranteed to be correct.

m9710797 (commented):
MSXtar: Introduce cluster types and variants. 

[We discussed some items on IRC, those are already resolved].
My only remaining concern is the too generic name "Free".

And I'm fully convinced now about the usefulness of std::variant.
Thanks.

m9710797 (commented):
MSXtar: Namespace FAT-specific stuff. 

This commit is fine.

src/fdc/MSXtar.cc
```diff
@@ -218,10 +247,17 @@ FatCluster MSXtar::readFAT(Cluster cluster) const
...
+
+	if (fat16) {
+		auto p = subspan<2>(data, index * 2);
+		value = p[0] | p[1] << 8;
+	} else {
+		auto p = subspan<2>(data, (index * 3) / 2);
+		value = (index & 1)
+		      ? (p[0] >> 4) + (p[1] << 4)
+		      : p[0] + ((p[1] & 0x0F) << 8);
+	}
```
m9710797: Very much a detail. I personally try to avoid having (even temporary) uninitialized variables. Here it can be avoided with a construct like:
```
unsigned value = [&] {
    if (fat16) {
        ...
        return p[0] | p[1] << 8;
    } else {
        ...
        return ...;
    }
}();
```
In other words use an "immediately invoked function expression (IIFE)". But I understand this can make the code less readable if you're not familiar with the idiom.

Bonus: that would allow to make `value` `const`. Personally I don't have the habit of making local variables `const` whenever possible (this rule has both proponents and opponents).

PR #1479, opened 2023-01-18
V9990: warns about writes (reads) to read-only (write-only) registers

pvmm:
Useful for debugging new v9990 software.

### DIFF
--- a/share/scripts/callbackprocs.tcl
+++ b/share/scripts/callbackprocs.tcl
@@ -107,6 +107,14 @@ proc sensorkidacquirecallback {port} {
 set sensor_kid_port_status_callback sensorkidportstatuscallback
 set sensor_kid_acquire_callback     sensorkidacquirecallback
 
+proc default_V9990_invalid_register_read_callback {reg} {
+	message [format "Reading from V9990 write-only register 0x%x" $reg] warning
+}
+proc default_V9990_invalid_register_write_callback {reg value} {
+	message [format "Writing 0x%x to V9990 read-only register %i" $value $reg] warning
+}
+#set v9990_invalid_register_read_callback  default_V9990_invalid_register_read_callback
+#set v9990_invalid_register_write_callback default_V9990_invalid_register_write_callback
 
 # show message (also) as OSD message
 interp alias {} default_message_callback {} osd::display_message
--- a/src/video/v9990/V9990.cc
+++ b/src/video/v9990/V9990.cc
@@ -63,6 +63,16 @@ V9990::V9990(const DeviceConfig& config)
 	, v9990PalDebug(*this)
 	, irq(getMotherBoard(), getName() + ".IRQ")
 	, display(getReactor().getDisplay())
+	, invalidRegisterReadCallback(getCommandController(),
+		"v9990_invalid_register_read_callback",
+		"Tcl proc to call when a write-only register was read from. "
+		"Input: register number (0-63)",
+                {}, Setting::SAVE)
+	, invalidRegisterWriteCallback(getCommandController(),
+		"v9990_invalid_register_write_callback",
+		"Tcl proc to call when a read-only register was written to. "
+		"Input: register number (0-63), 8-bit data",
+		{}, Setting::SAVE)
 	, vram(*this, getCurrentTime())
 	, cmdEngine(*this, getCurrentTime(), display.getRenderSettings())
 	, frameStartTime(getCurrentTime())
@@ -543,6 +553,7 @@ byte V9990::readRegister(byte reg, EmuTime::param time) const
 			       : narrow_cast<byte>(borderX >> 8);
 		}
 	} else {
+		invalidRegisterReadCallback.execute(reg);
 		return 0xFF;
 	}
 }
@@ -569,6 +580,7 @@ void V9990::writeRegister(byte reg, byte val, EmuTime::param time)
 	assert(reg < 64);
 	if (!(regAccess[reg] & ALLOW_WRITE)) {
 		// register not writable
+		invalidRegisterWriteCallback.execute(reg, val);
 		return;
 	}
 	if (reg >= CMD_PARAM_SRC_ADDRESS_0) {
--- a/src/video/v9990/V9990.hh
+++ b/src/video/v9990/V9990.hh
@@ -3,6 +3,7 @@
 
 #include "MSXDevice.hh"
 #include "Schedulable.hh"
+#include "TclCallback.hh"
 #include "VideoSystemChangeListener.hh"
 #include "IRQHelper.hh"
 #include "V9990CmdEngine.hh"
@@ -528,6 +529,9 @@ private:
 	IRQHelper irq;
 
 	Display& display;
+
+	TclCallback   invalidRegisterReadCallback;
+	TclCallback   invalidRegisterWriteCallback;
 
 	/** VRAM
 	  */
--- a/src/commands/TclCallback.cc
+++ b/src/commands/TclCallback.cc
@@ -33,7 +33,7 @@ TclObject TclCallback::getValue() const
 	return getSetting().getValue();
 }
 
-TclObject TclCallback::execute()
+TclObject TclCallback::execute() const
 {
 	const auto& callback = getValue();
 	if (callback.empty()) return {};
@@ -42,7 +42,7 @@ TclObject TclCallback::execute()
 	return executeCommon(command);
 }
 
-TclObject TclCallback::execute(int arg1)
+TclObject TclCallback::execute(int arg1) const
 {
 	const auto& callback = getValue();
 	if (callback.empty()) return {};
@@ -51,7 +51,7 @@ TclObject TclCallback::execute(int arg1)
 	return executeCommon(command);
 }
 
-TclObject TclCallback::execute(int arg1, int arg2)
+TclObject TclCallback::execute(int arg1, int arg2) const
 {
 	const auto& callback = getValue();
 	if (callback.empty()) return {};
@@ -60,7 +60,7 @@ TclObject TclCallback::execute(int arg1, int arg2)
 	return executeCommon(command);
 }
 
-TclObject TclCallback::execute(int arg1, std::string_view arg2)
+TclObject TclCallback::execute(int arg1, std::string_view arg2) const
 {
 	const auto& callback = getValue();
 	if (callback.empty()) return {};
@@ -69,7 +69,7 @@ TclObject TclCallback::execute(int arg1, std::string_view arg2)
 	return executeCommon(command);
 }
 
-TclObject TclCallback::execute(std::string_view arg1, std::string_view arg2)
+TclObject TclCallback::execute(std::string_view arg1, std::string_view arg2) const
 {
 	const auto& callback = getValue();
 	if (callback.empty()) return {};
@@ -78,7 +78,7 @@ TclObject TclCallback::execute(std::string_view arg1, std::string_view arg2)
 	return executeCommon(command);
 }
 
-TclObject TclCallback::executeCommon(TclObject& command)
+TclObject TclCallback::executeCommon(TclObject& command) const
 {
 	try {
 		return command.executeCommand(callbackSetting.getInterpreter());
--- a/src/commands/TclCallback.hh
+++ b/src/commands/TclCallback.hh
@@ -22,17 +22,17 @@ public:
 	            bool useCliComm = true);
 	explicit TclCallback(StringSetting& setting);
 
-	TclObject execute();
-	TclObject execute(int arg1);
-	TclObject execute(int arg1, int arg2);
-	TclObject execute(int arg1, std::string_view arg2);
-	TclObject execute(std::string_view arg1, std::string_view arg2);
+	TclObject execute() const;
+	TclObject execute(int arg1) const;
+	TclObject execute(int arg1, int arg2) const;
+	TclObject execute(int arg1, std::string_view arg2) const;
+	TclObject execute(std::string_view arg1, std::string_view arg2) const;
 
 	[[nodiscard]] TclObject getValue() const;
 	[[nodiscard]] StringSetting& getSetting() const { return callbackSetting; }
 
 private:
-	TclObject executeCommon(TclObject& command);
+	TclObject executeCommon(TclObject& command) const;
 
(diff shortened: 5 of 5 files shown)


src/video/v9990/V9990.cc
```diff
@@ -166,9 +172,10 @@ byte V9990::readIO(word port, EmuTime::param time)
...
 		case KANJI_ROM_3:
-		case REGISTER_SELECT:
-		case SYSTEM_CONTROL:
 		default:
+			if (writeOnly) {
+				getCliComm().printWarning("Reading from the V9990 write-only port ", portName[port]);
+			}
 			return peekIO(port, time);
 		}
 	}();
```
m9710797: I would suggest to write this as:
```
byte result = [&] {
    switch (port) {
    case COMMAND_DATA:
        return ...;
    case SYSTEM_CONTROL:
    case REGISTER_SELECT:
    case KANJI_ROM_0:
    case KANJI_ROM_2:
        getCliComm().printWarning("Reading from the V9990 write-only port ", portName[port]);
        [[fallthrough]];
    ...
    default:
        return peekIO(port, time);
}();
```
In other words: instead of setting and later testing a boolean, remove that boolean and directly print the message in the only place where that boolean was set.

I think also the ports KANJI_ROM_0 and KANJI_ROM_2 are write-only, but please verify.

src/video/v9990/V9990.hh
```diff
@@ -454,6 +454,25 @@ private:
...
+		"KANJI_ROM_0",
+		"KANJI_ROM_1",
+		"KANJI_ROM_2",
+		"KANJI_ROM_3",
+		"RESERVED_0",
+		"RESERVED_1",
+		"RESERVED_2",
+		"RESERVED_3"
+	};
+
```
m9710797: This can be moved from the .hh to the .cc file (until we also need it in some other place than V9990.cc).
I'd also suggest to make the following tweaks:
* Use `std::string_view` instead of `std::string`. This is more light-weight. E.g. it allows for compile-time initialization rather than run-time initialization.
* Declare this array `static` because (once moved to the .cc file) it should only be visible in one file.
* Declare it `constexpr` to indicate that the table won't change (`constexpr` implies `const`) and that we want it to be initialized at compile-time.
* We prefer to use `std::array` over plain C-arrays (for reasons I won't go into now).

In other words, declare this table as:
```
static constexpr std::array<std::string_view, 16> portName = {
    "VRAM_DATA",
    ...
};
```
I also wonder if, for users of this feature, it's clearer to show the port name (e.g. KANJI_ROM_0), the local port number (e.g. 8), or the full port number (e.g. 0x68). Or maybe some combination?

I'm assuming that you don't intentionally read/write from/to the wrong port. It may not even have been the intention to access the V9990 at all. And in that case having the full port number might be more useful than having the name of a V9990 port. (And if the user was programming the V9990, he/she probably knows the port numbers by heart.)

src/video/v9990/V9990.cc
```diff
@@ -338,7 +345,7 @@ void V9990::writeIO(word port, byte val, EmuTime::param time)
 			break;
 
 		case STATUS:
-			// read-only, ignore writes
+			getCliComm().printWarning("Writing to the V9990 read-only port ", portName[port]);
 			break;
```
m9710797: This comment applies both to this `printWarning()` and the one above.

This is fine for a first version. But I'm wondering if we shouldn't replace `printWarning()` with the `TclCallback` mechanism. In the past we indeed always used `printWarning()`, but many (all?) such warnings that trigger from emulation behavior (stuff that gets triggered by the emulated CPU) have been replaced by `TclCallback`. For example accessing VDP registers too fast, setting invalid PPI mode, excuting "DI ; HALT", ...

These callbacks can have a default implementation that prints the same warning message. But the main advantage is that power users can still replace this behavior. E.g. disable it, or log something, ...

m9710797 (changes requested):
Thanks a lot for this PR!

Though I do have some suggestions for changes. See the comments in the code itself.

Feel free to ask for more info if anything is unclear. Or even for help to implement any of this.

MBilderbeek:
General remark: wouldn't it be a good idea to make a Tcl callback for these cases, instead of directly issuing a warning? With that, the warnings could also be easily disabled.

EDIT: oh, that's also what @m9710797 suggested :) Well, I agree :)

src/video/v9990/V9990.cc
pvmm: Should I put the function in `callbackprocs.tcl`? I am confused about the names of the callbacks. Some of them seem to use snake_case names (`invalid_psg_directions_callback`), some are squished together without separator (`vdpcmdinprogresscallback`). What name should I use? I think I will call them `v9990_invalid_port_read_callback` and `v9990_invalid_port_write_callback`.

m9710797: Yes, `share/scripts/callbackprocs.tcl` holds the stuff for our current callbacks. You should extend it for your new callbacks.

Unfortunately we haven't always been consistent with naming :(
But fixing the names isn't worth breaking backwards compatibility.
For new names we should try to do better. I like the style of the names you're proposing.

You've probably seen that the callbacks can take parameters. One parameter could be the port number. A second parameter could be an indication of "read or write". Though having separate callbacks for read and write is also fine (maybe it's even better .. I really haven't given it much thought yet .. I'll let you decide).

Yet another possibility is to allow for a future generic mechanism. I mean more generic than only V9990. Because e.g. also the V99x8 has write-only ports. Maybe pick generic names like `read_from_write_only_port_callback` and `write_to_read_only_port_callback` (no specific reference to V9990).

Note: I'm not expecting you to implement any of this future stuff.

src/video/v9990/V9990.hh
pvmm: Using port numbers only.

m9710797:
Hi pvmm,

I'm very sorry for only telling this now, after you've already done a lot of work on this PR. But I only thought about this myself just now.

I'll try to reconstruct my thought process...

I was wondering about extending this functionality to more devices in the future. For example the regular VDP (V99x8) or the PSG. Then I realized that for many devices we only register the actually accessible port numbers in the machine description file. For example for the PSG we have:
```
<io base="0xA0" num="2" type="O"/>
<io base="0xA2" num="1" type="I"/>
```
This means that for ports 0xA0 and 0xA1 only writes will be passed to the emulation code (the MSXPSG class) and for port 0xA2 only reads. In other words: we can't even extend the code in MSXPSG to flag invalid accesses, those accesses don't even reach that code!

I also thought that it might be useful to catch accesses to IO ports that don't have any devices connected at all. E.g. a typical MSX machine has nothing(?) on port 0x70. In a way this is similar to the situation in the previous paragraph: from an openMSX point of view there's nothing registered on OUTPUT port 0xA2 (only INPUT port 0xA2 goes to the PSG). And similarly there nothing registered on port 0x70 (neither input nor output).

My next thought was: "Should we really extend all IO port registrations in all machine configs only to be able to generate warnings on invalid access? That seems like a bad idea. And then that still won't cover completely unused ports."

I was also wondering: Is this something that should be enabled by default? It may trigger too many unwanted warnings. E.g. valid code that probes for the presence of a device by trying to access it (e.g. writing and reading IO ports, that might not be present on the current MSX machine).

Or there might be less valid MSX programs that accidentally write to the PSG read port (0xA2) or the V9990 status port (0x65), but otherwise work perfectly fine. As a _programmer_ of that particular MSX program I might be interested in such warnings. But as a _normal openMSX user_ such warnings would greatly annoy me (because the warning doesn't tell me anything I can do about, and it's totally harmless). So in any case these warnings should be DISABLED by default.

And then finally I realized that, given that the warnings must be opt-in (the developer of the MSX program must explicitly enable them in the openMSX console), we ALREADY have a mechanism that can accomplish this: IO watchpoints!!!

In other words: it can already be implemented via a Tcl script. No need to change the core openMSX code at all. For example: the V9990 stuff you're developing, can also be obtained via a script like this:
```
proc warn_write {port} { message [format "Write to read-only port 0x%x" $port] warning }
proc warn_read  {port} { message [format "Read from write-only port 0x%x" $port] warning }
debug set_watchpoint write_io 0x65 true "warn_write 0x65"
debug set_watchpoint read_io  0x64 true "warn_read  0x64"
debug set_watchpoint read_io  0x67 true "warn_read  0x67"
debug set_watchpoint read_io  0x68 true "warn_read  0x68"
debug set_watchpoint read_io  0x6A true "warn_read  0x6A"
```
Literally just 7 lines! Much simpler than the PR you're working on now.

It's also easy to customize/extend this script. It's even possible to warn on any (input) port that's not connected with a Tcl script like:
```
for {set i 0} {$i < 256} {incr i} {
    if {[machine_info input_port $i] eq ""} {
        debug set_watchpoint read_io $i true "warn_read $i"
    }
}
```

So pvmm, sorry again for all the work you've done already. But I think offering this functionality via a Tcl script is both much simpler and more flexible compared to the approach you're currently taking.

But do let me know if you have another opinion.

Wouter

pvmm:
How about instead of scrapping the whole thing, we change the code to detect invalid V9990 register access? That would be my next step and what interests me the most. And I think it's not easy to write using simple Tcl watchpoints.

src/video/v9990/V9990.cc
```diff
@@ -161,21 +170,18 @@ byte V9990::readIO(word port, EmuTime::param time)
 			return cmdEngine.getCmdData(time);
 		case SYSTEM_CONTROL:
 		case REGISTER_SELECT:
-			writeOnly = true;
+		case KANJI_ROM_0:
+		case KANJI_ROM_2:
+			invalidPortReadCallback.execute(0x60 + port);
```
m9710797: I've already wrote a (long) comment elsewhere where I explain that I changed my mind about this c++ approach. I now think the same can be done better/simpler via a Tcl script. (Sorry again).

Apart from that, reviewing this c++ code anyway:

We prefer to not make assumptions about the actual port addresses in the c++ code (0x60-0x6f). These ranges only appear in the xml description files. That way we can easily reconfigure the device. For example have a 2nd gfx9000 using ports 0x70-0x7f. (Never tested that it actually works, possibly there are other limitations elsewhere for gfx9000. But it was useful, and it works, for a 2nd PSG).

You can achieve this by passing the original value of the 'port' parameter to the callback. So before it was masked by the `port &= 0x0F;` statement.

pvmm: The advantage checking the use of registers instead of ports is that they are the same for any number of video cartridges.

m9710797 (commented):
TclCallback: const modifier on execute function 

This commit is fine. Thanks!

share/scripts/callbackprocs.tcl
```diff
@@ -107,6 +107,14 @@ proc sensorkidacquirecallback {port} {
...
 set sensor_kid_acquire_callback     sensorkidacquirecallback
 
+proc v9990invalidregisterreadcallback {reg} {
+	message [format "Reading from V9990 write-only register 0x%x" $reg] warning
+}
+proc v9990invalidregisterwritecallback {reg value} {
+	message [format "Writing 0x%x to V9990 read-only register 0x%x" $value $reg] warning
+}
+#set v9990_invalid_register_read_callback  v9990invalidregisterreadcallback
+#set v9990_invalid_register_write_callback v9990invalidregisterwritecallback
```
m9710797: Detail: also for `v9990invalidregisterreadcallback` it's better to use a consistent naming style. Maybe change it to `default_V9990_invalid_register_read_callback`?

src/video/v9990/V9990.cc
```diff
@@ -63,6 +64,16 @@ V9990::V9990(const DeviceConfig& config)
 	, v9990PalDebug(*this)
 	, irq(getMotherBoard(), getName() + ".IRQ")
 	, display(getReactor().getDisplay())
+	, invalidRegisterReadCallbackSetting(getCommandController(),
+		"v9990_invalid_register_read_callback",
+		"Tcl proc to call when a write-only register was read from. "
+		"Input: register number (0x0-0x63)", {})
```
m9710797: Could it be you mixed hexadecimal and decimal notation here? I think it should be either `0-63` or `0x00-0x3F`.
Same below for write.

src/video/v9990/V9990.cc
```diff
@@ -63,6 +64,16 @@ V9990::V9990(const DeviceConfig& config)
...
+	, invalidRegisterReadCallback(invalidRegisterReadCallbackSetting)
+	, invalidRegisterWriteCallbackSetting(getCommandController(),
+		"v9990_invalid_register_write_callback",
+		"Tcl proc to call when a read-only register was written to. "
+		"Input: register number (0x0-0x63)", {})
+	, invalidRegisterWriteCallback(invalidRegisterWriteCallbackSetting)
```
m9710797: While reviewing your previous commit I noticed that the `TclCallback` class offers the possibility to internally store a StringSetting. With that your code can be simplified to:
```
, invalidRegisterWriteCallback(getCommandController(),
		"v9990_invalid_register_write_callback",
		"Tcl proc to call when a read-only register was written to. "
		"Input: register number (0-63)",
                {}, Setting::SAVE)
```
(And same for read).
And then you can remove the two `StringSetting` objects from the V9990.hh file.

src/video/v9990/V9990.cc
```diff
@@ -1,5 +1,6 @@
 #include "V9990.hh"
 #include "Display.hh"
+#include "CliComm.hh"
```
m9710797: I assume this #include is no longer needed?

src/video/v9990/V9990.cc
```diff
@@ -569,6 +581,7 @@ void V9990::writeRegister(byte reg, byte val, EmuTime::param time)
 	assert(reg < 64);
 	if (!(regAccess[reg] & ALLOW_WRITE)) {
 		// register not writable
+		invalidRegisterWriteCallback.execute(reg, val);
```
m9710797, line 583: Ah, the write callback takes two parameters. Good idea.
But then the documentation above should reflect that (currently it only talks about the register parameter).

BTW now that I see the patch, I agree it's simpler in c++ than via a Tcl script.

m9710797 (commented):
V9990: warns about writes (reads) to read-only (write-only) registers 

OK, with some details that should be fixed/cleaned-up.

m9710797:
Thanks!
Also for persisting after I requested some major rework.

Issue #1478, opened 2023-01-17
diskmanipulator export hda bgm cause crash


denjhang:
[BGM.zip](https://github.com/openMSX/openMSX/files/10434545/BGM.zip)

`set power off`
`ext ide`
`diskmanipulator chdir hda /`
`diskmanipulator export hda bgm`

denjhang:
https://youtu.be/Mfxozkq2mng

This video shows the breakdown in action.

denjhang:
#1473

MBilderbeek:
@denjhang Does that last link mean the crash doesn't happen anymore with the fixes for #1473 ? Or does it still occur with the latest openMSX code?

denjhang:
> @denjhang Does that last link mean the crash doesn't happen anymore with the fixes for #1473 ? Or does it still occur with the latest openMSX code?

I tested with the latest build and it's still the same.
`openmsx-18.0-518-g8ba6e3875-windows-vc-x64-bin`

MBilderbeek:
I cannot reproduce it (on Linux, at least...) 

How did you start up openMSX exactly? I used:
```
$ openmsx -machine Panasonic_FS-A1GT -ext ide -hda BGM.DSK 
```

denjhang:
on windows10
`e:`
`cd E:\Program Files\openmsx-18.0-518-g8ba6e3875-windows-vc-x64-bin`
`openmsx.exe -machine Panasonic_FS-A1GT -ext ide -hda BGM.DSK -ext slotexpander -ext ram4mb -ext scc+`

denjhang:
There is no problem with the startup statement, but when I use diskexplorer to manually export, I find that diskexplorer will also crash, and some of the exported file names are garbled. I suspect an issue with the filename is causing the crash.

denjhang:
> I cannot reproduce it (on Linux, at least...)
> 
> How did you start up openMSX exactly? I used:
> 
> ```
> $ openmsx -machine Panasonic_FS-A1GT -ext ide -hda BGM.DSK 
> ```

I tried your statement, same situation.

denjhang:
https://user-images.githubusercontent.com/61811583/213071405-9f1e2541-850f-4389-ba09-d98a5540d848.mp4

diskexplorer also crashed.

MBilderbeek:
This is the list of dirs/files it outputs here:
```
$ ls -R1 /tmp/bgm
/tmp/bgm:
0
'~1.tra'
a
[...]

PR #1482, opened 2023-01-20
DiskManipulator: Transcode file names to Unicode.

grauw:
With this patch series DiskManipulator properly translates the encoding from the host system’s local character encoding to the specific MSX’s character encoding. This applies to `import`, `export`, `dir`, `chdir` and `mkdir`. See https://github.com/openMSX/openMSX/issues/1478.

@wouterv The MsxChar2Unicode is not easily accessible so it turns into this kind of mess. Any recommendations welcome.

Ideally I can retrieve it a bit easier, perhaps from the `Reactor` or the `MSXMotherBoard`, and assume it is always available (defaulting to a mapping of ASCII only).

I think the MSX character encoding is not a property of the keyboard, but a property of the BIOS. E.g. MSX-DOS (Nextor) also [relies on](https://github.com/Konamiman/Nextor/blob/v2.1/source/kernel/bank2/path.mac#L1418) the region specified in the BIOS at address `0x002B` for its upper-casing.

So maybe rather than to be located somewhere deep in the PPI / Keyboard systems, it would be better if it lived at a higher level. The encoding value is currently also not specified in the machine config, nor read from the BIOS ROM, but indirectly specified in the keyboard mapping config, reinforcing the dependency on it.

I also think a better name for the class would be `MSXCharacterEncoding` or something like that.

p.s. C++20 has `std::u8string` and `std::u8string_view`, which have an explicit UTF-8 encoding. Might be interesting to see if that can be used in the openMSX code base, though that would be a massive patch.

### DIFF
--- a/src/input/UnicodeKeymap.cc
+++ b/src/input/UnicodeKeymap.cc
@@ -88,6 +88,9 @@ UnicodeKeymap::UnicodeKeymap(string_view keyboardType)
 		auto buf = file.mmap();
 		parseUnicodeKeyMapFile(
 			string_view(reinterpret_cast<const char*>(buf.data()), buf.size()));
+		if (!msxChars.has_value()) {
+			msxChars.emplace("MSXVID.TXT");
+		}
 	} catch (FileException&) {
 		throw MSXException("Couldn't load unicode keymap file: ", filename);
 	}
--- a/src/MSXPPI.cc
+++ b/src/MSXPPI.cc
@@ -33,6 +33,10 @@ MSXPPI::~MSXPPI()
 	powerDown(EmuTime::dummy());
 }
 
+const Keyboard& MSXPPI::getKeyboard() const {
+	return keyboard;
+}
+
 void MSXPPI::reset(EmuTime::param time)
 {
 	i8255.reset(time);
--- a/src/MSXPPI.hh
+++ b/src/MSXPPI.hh
@@ -42,6 +42,8 @@ public:
 	explicit MSXPPI(const DeviceConfig& config);
 	~MSXPPI() override;
 
+	[[nodiscard]] const Keyboard& getKeyboard() const;
+
 	void reset(EmuTime::param time) override;
 	void powerDown(EmuTime::param time) override;
 	[[nodiscard]] byte readIO(word port, EmuTime::param time) override;
--- a/src/fdc/DiskManipulator.cc
+++ b/src/fdc/DiskManipulator.cc
@@ -12,6 +12,11 @@
 #include "FileException.hh"
 #include "FileOperations.hh"
 #include "SectorBasedDisk.hh"
+#include "MsxChar2Unicode.hh"
+#include "MSXDevice.hh"
+#include "MSXMotherBoard.hh"
+#include "MSXPPI.hh"
+#include "Keyboard.hh"
 #include "StringOp.hh"
 #include "TclObject.hh"
 #include "narrow.hh"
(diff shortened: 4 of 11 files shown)


m9710797 (commented):
MSXtar: Have a separate type for MSX file names. 

OK. Thanks, nice cleanups.

m9710797 (commented):
MSXtar: Improve filtering of unsupported file name characters. 

OK.

m9710797 (commented):
MsxChar2Unicode: Pass strings rather than uint8 vectors. 

I understand your reasoning. In general you're correct.
Though specifically in openMSX we have the convention that all(*1) std::string's(*2) are utf-8 encoded. And for that reason I picked a different type for these MSX-encoded strings.

I understand that in the last commit of this series (in the current implementation) it's convenient if the msx-encoded-strings use std::string. But I'd like to ask you to re-check if it isn't also possible to keep using a different type for msx-strings. If that makes the code too ugly then I'm fine with changing to std::string.


(*1)
Or at least as much as possible.

(*2)
It would be nicer if we could instead use std::u8string, that's very clearly a utf-8 string. Unfortunately u8string doesn't work very well yet (it doesn't integrate well with the rest of the library). There are already some changes in c++23, and likely more will follow in c++26.

m9710797 (commented):
MsxChar2Unicode: Add convenience functions. 

OK.

src/input/UnicodeKeymap.cc
```diff
@@ -88,6 +88,9 @@ UnicodeKeymap::UnicodeKeymap(string_view keyboardType)
 		auto buf = file.mmap();
 		parseUnicodeKeyMapFile(
 			string_view(reinterpret_cast<const char*>(buf.data()), buf.size()));
+		if (!msxChars.has_value()) {
+			msxChars.emplace("MSXVID.TXT");
+		}
```
m9710797, line 93: Fine for now. But in the (near?) future we'll require that all unicodemap files must specify the msx-encoding.
(Already done on the WIP-keyboard-stuff branch).

m9710797 (commented):
MSXtar: Convert MSX file names to UTF-8. 

100% agree that the way to retrieve the `MsxChar2Unicode` object is ugly. But don't worry about it. I'll fix it later. We already planned some changes here. See the branch "WIP-keyboard-stuff" (though work on that is progressing slowly).

This commit is fine. The only thing I'd like to repeat is: check if it's possible to keep using a different type than std::string for msx-encoded strings. But fine if that's too difficult.

grauw:
> MsxChar2Unicode: Pass strings rather than uint8 vectors.
> 
> I understand your reasoning. In general you're correct. Though specifically in openMSX we have the convention that all(*1) std::string's(*2) are utf-8 encoded. And for that reason I picked a different type for these MSX-encoded strings.
> 
> I understand that in the last commit of this series (in the current implementation) it's convenient if the msx-encoded-strings use std::string. But I'd like to ask you to re-check if it isn't also possible to keep using a different type for msx-strings. If that makes the code too ugly then I'm fine with changing to std::string.

I see. Well, it’s ugly, but it could be worse:

```diff
diff --git a/src/fdc/MSXtar.cc b/src/fdc/MSXtar.cc
index 29fb23f5c..7702467ab 100644
--- a/src/fdc/MSXtar.cc
+++ b/src/fdc/MSXtar.cc
@@ -470,8 +470,9 @@ static char toFileNameChar(char a)
 // dirEntries on an MSX
 FileName MSXtar::hostToMSXFileName(string_view hostName) const
 {
-       string hostNameMSXEncoding = msxChars.utf8ToMsx(hostName, '_');
-       auto [hostDir, hostFile] = StringOp::splitOnLast(hostNameMSXEncoding, '/');
+       std::vector<uint8_t> hostMSXName = msxChars.utf8ToMsx(hostName, '_');
+       std::string_view hostMSXNameView(reinterpret_cast<char*>(hostMSXName.data()), hostMSXName.size());
+       auto [hostDir, hostFile] = StringOp::splitOnLast(hostMSXNameView, '/');
 
        // handle special case '.' and '..' first
        FileName result;
@@ -803,7 +804,8 @@ string MSXtar::msxToHostFileName(const FileName& msxName) const
                        result += char(tolower(msxName[i]));
                }
        }
-       return msxChars.msxToUtf8(result, '_');
+       std::span<const uint8_t> resultSpan(reinterpret_cast<const uint8_t*>(result.data()), result.size());
+       return msxChars.msxToUtf8(resultSpan, '_');
 }
 
 
```

Every other alternative I tried (e.g. `vector<char>`, `basic_string<uint8_t>`) also turned out similarly awkward.

I just feel that as these conversion methods are used more, in the same vein as MSXtar does, there will probably be similar ugliness in those places as well. So it wouldn’t be my choice, but I can live with this.

PR #1483, opened 2023-01-20
DiskManipulator: Add partition command

grauw:
With this command you can (re)partition an existing disk image.

Unlike `create`, `partition` always creates a partition table, and also accepts a zero-length list of sizes in which case it will create an empty partition table.

If not all partitions can be created given the disk size, it will throw an exception to print out in the console after creating the partitions that it could. If the last partition is too large to fit on the disk, its size will be clamped to the available space, no exceptions are thrown.

See (very old) issue https://github.com/openMSX/openMSX/issues/170.

### DIFF
--- a/src/fdc/DiskImageUtils.cc
+++ b/src/fdc/DiskImageUtils.cc
@@ -401,8 +401,13 @@ static SetBootSectorResult setBootSector(
 
 void format(SectorAccessibleDisk& disk, MSXBootSectorType bootType)
 {
+	format(disk, bootType, disk.getNbSectors());
+}
+
+void format(SectorAccessibleDisk& disk, MSXBootSectorType bootType, size_t nbSectors)
+{
 	// first create a boot sector for given partition size
-	size_t nbSectors = disk.getNbSectors();
+	nbSectors = std::min(nbSectors, disk.getNbSectors());
 	SectorBuffer buf;
 	SetBootSectorResult result = setBootSector(buf.bootSector, bootType, nbSectors);
 	disk.writeSector(0, buf);
--- a/src/fdc/DiskImageUtils.cc
+++ b/src/fdc/DiskImageUtils.cc
@@ -585,13 +585,46 @@ static std::vector<unsigned> partitionSunrise(SectorAccessibleDisk& disk, std::s
 	return clampedSizes;
 }
 
+// Partition with standard master boot record (MBR) for Beer IDE 1.9.
+static std::vector<unsigned> partitionBeer(SectorAccessibleDisk& disk, std::span<const unsigned> sizes)
+{
+	SectorBuffer buf;
+	auto& pt = buf.ptNextor;
+
+	std::vector<unsigned> clampedSizes = clampPartitionSizes(sizes, disk.getNbSectors(), 1, 0);
+	if (clampedSizes.size() > pt.part.size()) {
+		clampedSizes.resize(pt.part.size());
+	}
+
+	ranges::fill(buf.raw, 0);
+	pt.header = NEXTOR_PARTITION_TABLE_HEADER; // TODO: Find out BEER IDE signature
+	pt.end = 0xAA55;
+
+	unsigned partitionOffset = 1;
+	for (auto [i, size] : enumerate(clampedSizes)) {
+		auto& p = pt.part[i];
+		p.boot_ind = (i == 0) ? 0x80 : 0x00; // boot flag
+		p.sys_ind = 0x01; // FAT12
+		p.start = partitionOffset;
+		p.size = size;
+		partitionOffset += size;
+	}
+
+	disk.writeSector(0, buf);
+	return clampedSizes;
+}
+
 unsigned partition(SectorAccessibleDisk& disk, std::span<const unsigned> sizes, MSXBootSectorType bootType)
 {
 	std::vector<unsigned> clampedSizes = [&] {
 		if (bootType == MSXBootSectorType::NEXTOR) {
 			return partitionNextor(disk, sizes);
-		} else {
+		} else if (bootType == MSXBootSectorType::DOS2) {
 			return partitionSunrise(disk, sizes);
+		} else if (bootType == MSXBootSectorType::DOS1) {
+			return partitionBeer(disk, sizes);
+		} else {
+			UNREACHABLE;
 		}
 	}();
 
--- a/src/fdc/DiskImageUtils.cc
+++ b/src/fdc/DiskImageUtils.cc
@@ -463,14 +463,56 @@ struct CHS {
 	return {cylinder, head, sector};
 }
 
+static std::vector<unsigned> clampPartitionSizes(std::span<const unsigned> sizes,
+	size_t diskSize, unsigned initialOffset, unsigned perPartitionOffset)
+{
+	std::vector<unsigned> clampedSizes;
+	size_t sizeRemaining = std::min(diskSize, size_t(std::numeric_limits<unsigned>::max()));
+
+	if (sizeRemaining >= initialOffset) {
+		sizeRemaining -= initialOffset;
+	} else {
+		return clampedSizes;
+	}
+
+	for (auto size : sizes) {
+		if (sizeRemaining <= perPartitionOffset) {
+			break;
+		}
+		sizeRemaining -= perPartitionOffset;
+		if (size <= perPartitionOffset) {
+			throw CommandException("Partition size too small: ", size);
+		}
+		size -= perPartitionOffset;
+		if (sizeRemaining > size) {
+			clampedSizes.push_back(size);
+			sizeRemaining -= size;
+		} else {
+			clampedSizes.push_back(narrow<unsigned>(sizeRemaining));
+			break;
+		}
+	}
+	return clampedSizes;
+}
+
 // Partition with standard master / extended boot record (MBR / EBR) for Nextor.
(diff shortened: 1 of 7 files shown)


src/fdc/DiskManipulator.cc
```diff
@@ -454,11 +441,27 @@ void DiskManipulator::create(std::span<const TclObject> tokens)
...
+		}
+	}
+	if (drive) {
+		auto& driveData = getDriveSettings(*drive);
+		auto partition = getPartition(driveData);
+		DiskImageUtils::format(partition, bootType);
+		driveData.workingDir[driveData.partition] = '/';
+	} else {
+		throw CommandException("Incorrect number of parameters");
+	}
```
m9710797, line 523: Very much a detail, coding style:
```
if (!drive) {
	throw CommandException("Incorrect number of parameters");
}
auto& driveData = getDriveSettings(*drive);
auto partition = getPartition(driveData);
DiskImageUtils::format(partition, bootType);
driveData.workingDir[driveData.partition] = '/';
```
This keep the normal code flow non-indented (and better separates it from the error path).
No problem at all if you like your version better.

m9710797 (commented):
DiskManipulator: Clean up format parameter parsing.

This and the previous commit are OK.

m9710797 (commented):
DiskManipulator: Optional size argument to format. 

OK. Thanks.

src/fdc/DiskManipulator.cc
```diff
@@ -403,17 +403,14 @@ void DiskManipulator::create(std::span<const TclObject> tokens)
...
 			bootType = *t;
-			continue;
-		}
-
-		if (sizes.size() >= MAX_PARTITIONS) {
-			throw CommandException(
-				"Maximum number of partitions is ", MAX_PARTITIONS);
+		} else if (sizes.size() >= MAX_PARTITIONS) {
+			throw CommandException("Maximum number of partitions is ", MAX_PARTITIONS);
+		} else if (size_t sectors = parseSectorSize(token_.getString()); sectors <= ~0u) {
```
m9710797: This is correct, and compact, though I prefer a more direct way (it took me a few moments to realize what was going on here):
```
} else if (size_t sectors = parseSectorSize(token_.getString());
           sectors <= std::numeric_limits<unsigned>::max()) {
```

src/fdc/DiskImageUtils.cc
```diff
@@ -463,13 +463,39 @@ struct CHS {
...
+			} else if (sizeRemaining > 0) {
+				clampedSizes.push_back(narrow<unsigned>(sizeRemaining));
+				break;
+			} else {
+				break;
+			}
+		}
+	}
+
+	if (clampedSizes.size() == 0) {
```
m9710797: Detail: `clampedSizes.empty()`

src/fdc/DiskImageUtils.cc
```diff
@@ -500,21 +526,42 @@ static void partitionNextor(SectorAccessibleDisk& disk, std::span<const unsigned
...
+				clampedSizes.push_back(size);
+				sizeRemaining -= size;
+			} else if (sizeRemaining > 0) {
+				clampedSizes.push_back(narrow<unsigned>(sizeRemaining));
+				break;
+			} else {
+				break;
+			}
+		}
+	}
```
m9710797, line 558: Could the calculation of `clampedSizes` be generalized and reused between `partitionNextor()` and `partitionSunrise()`?

src/fdc/DiskManipulator.cc
```diff
@@ -432,13 +441,49 @@ void DiskManipulator::create(std::span<const TclObject> tokens)
...
+		if (auto t = parseBootSectorType(token_.getString())) {
+			bootType = *t;
+		} else if (sizes.size() >= MAX_PARTITIONS) {
+			throw CommandException("Maximum number of partitions is ", MAX_PARTITIONS);
+		} else if (size_t sectors = parseSectorSize(token_.getString()); sectors <= ~0u) {
+			sizes.push_back(narrow<unsigned>(sectors));
+		} else {
+			throw CommandException("Partition size too large.");
+		}
+	}
```
m9710797: Is it worth trying to reuse this parsing code between `create()` and `partition()`?
(Same remark here about `sectors <= ~0u`.)

src/fdc/DiskManipulator.cc
```diff
@@ -370,7 +380,13 @@ void DiskManipulator::tabCompletion(std::vector<string>& tokens) const
 			// if it has partitions then we also add the partition
 			// numbers to the autocompletion
 			if (auto* disk = d.drive->getSectorAccessibleDisk()) {
-				for (unsigned i = 1; i <= MAX_PARTITIONS; ++i) {
+				for (unsigned i = 1; ; ++i) {
```
m9710797: Detail: in openMSX we usually fill-in all three fields in a for loop. In this case that's `true` instead of an empty condition  (and we use `/**/` if the other two fields would be empty).

src/fdc/DiskManipulator.cc
```diff
@@ -90,6 +87,19 @@ void DiskManipulator::unregisterDrive(DiskContainer& drive)
...
+std::string DiskManipulator::DriveSettings::getWorkingDir(unsigned p)
+{
+	return p < workingDir.size() ? workingDir[p] : "/";
+}
+
+void DiskManipulator::DriveSettings::setWorkingDir(unsigned p, std::string_view dir)
+{
+	while (p >= workingDir.size()) {
+		workingDir.emplace_back("/");
+	}
```
m9710797, line 99: Replace this loop with:
```
if (p >= workingDir.size()) {
    workingDir.resize(p + 1, "/");
}
```

src/fdc/DiskManipulator.cc
grauw: Heh, I had that at first, and then I thought to be clever to to fit it on one line :D. I’ll change it.

grauw:
I addressed your comments, except for the one coding style one that I preferred the way it is now.

Additionally in the last commit I added an MBR-only partitioning format under the `-dos1` option for Beer IDE 1.9. It’s also compatible with Nextor.

I intend to push to master tomorrow evening.

src/fdc/DiskManipulator.cc
```diff
@@ -410,24 +423,31 @@ void DiskManipulator::savedsk(const DriveSettings& driveData,
...
-	if (sizes.empty()) {
+
+	return {bootType, sizes};
+}
+
+void DiskManipulator::create(std::span<const TclObject> tokens)
+{
+	auto [bootType, sizes] = parsePartitionSizes(view::drop(tokens, 3));
+
+	size_t totalSectors = ranges::accumulate(sizes, size_t(0));
```
m9710797: Very much a detail. No need to change anything.
`ranges::accumulate(<range>, 0)` could be replaced with `sum(<range>)`

The former is closer to `std::accumulate`. So might be more familiar to c++ programmers who don't know the openMSX utility functions. OTOH there is no standard range version of accumulate.

The meaning of the later should also be clear. And it might be more future compatible with C++23/C++26: instead of adding `std::ranges::accumulate()` it seems more likely they will instead add `std::ranges::fold()` and `std::ranges::sum()` (https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2214r1.html#stdaccumulate-rangesfold)

m9710797 (commented):
DiskManipulator: Beer IDE 1.9 partitioning support. 

OK.
Thanks for your continued work on this.

src/fdc/DiskManipulator.cc
grauw: Good, if you’re looking for a fold and not so familiar with C++, accumulate is a confusing name with a confusing default operator, glad they’re changing that.

However with `sum` I can not express compactly that I want to sum unsigned types using a `size_t` accumulator. Unless I add a projection function which converts the type… I guess it’s more explicit that way, but also more verbose.

I could also implement it with `ranges::reduce`, that would potentially even be more efficient and proper, but accumulate has a longer pedigree and its name implies summing (which we happen to do here) so I chose that one.

m9710797: I overlooked the widening.

In any case, no need to change anything, current code is fine.
`sum()` with a projection would also be fine, you decide.

src/fdc/DiskImageUtils.cc
```diff
@@ -585,13 +585,45 @@ static std::vector<unsigned> partitionSunrise(SectorAccessibleDisk& disk, std::s
...
+}
+
 unsigned partition(SectorAccessibleDisk& disk, std::span<const unsigned> sizes, MSXBootSectorType bootType)
 {
 	std::vector<unsigned> clampedSizes = [&] {
-		if (bootType == MSXBootSectorType::NEXTOR) {
-			return partitionNextor(disk, sizes);
-		} else {
-			return partitionSunrise(disk, sizes);
+		switch (bootType) {
```
grauw: GCC complains:

```
src/fdc/DiskImageUtils.cc: In lambda function:
src/fdc/DiskImageUtils.cc:629:2: warning: control reaches end of non-void function [-Wreturn-type]
  629 |  }();
      |  ^
```

Turning it into an if/else with unreachable.

src/fdc/DiskManipulator.cc
grauw: Since it also makes the `accumulate` in DiskImageUtils’ `partitionNextor()` look better, and makes the widening more explicit in `DiskManipulator::create()`, I went for `sum`.

PR #1486, opened 2023-02-02
Expose register and palette latch status to Tcl (and hence debugger)

pvmm:
This will help debugging a lot.

### DIFF
--- a/src/video/VDP.cc
+++ b/src/video/VDP.cc
@@ -78,6 +78,9 @@ VDP::VDP(const DeviceConfig& config)
 	, vdpStatusRegDebug(*this)
 	, vdpPaletteDebug  (*this)
 	, vramPointerDebug (*this)
+	, registerLatchStatusDebug(*this)
+	, paletteLatchStatusDebug(*this)
+	, dataLatchDebug   (*this)
 	, frameCountInfo   (*this)
 	, cycleInFrameInfo (*this)
 	, lineInFrameInfo  (*this)
@@ -1682,6 +1685,47 @@ void VDP::VRAMPointerDebug::write(unsigned address, byte value, EmuTime::param /
 	}
 }
 
+// class RegisterLatchStatusDebug
+
+VDP::RegisterLatchStatusDebug::RegisterLatchStatusDebug(VDP &vdp_)
+	: SimpleDebuggable(vdp_.getMotherBoard(),
+			vdp_.getName() + " register latch status", "V99x8 register latch status (0 = expecting a value, 1 = expecting a register)", 1)
+{
+}
+
+byte VDP::RegisterLatchStatusDebug::read(unsigned /*address*/)
+{
+	auto& vdp = OUTER(VDP, registerLatchStatusDebug);
+	return byte(vdp.registerDataStored);
+}
+
+// class PaletteLatchStatusDebug
+
+VDP::PaletteLatchStatusDebug::PaletteLatchStatusDebug(VDP &vdp_)
+	: SimpleDebuggable(vdp_.getMotherBoard(),
+			vdp_.getName() + " palette latch status", "V99x8 palette latch status (0 = expecting red & blue, 1 = expecting green)", 1)
+{
+}
+
+byte VDP::PaletteLatchStatusDebug::read(unsigned /*address*/)
+{
+	auto& vdp = OUTER(VDP, paletteLatchStatusDebug);
+	return byte(vdp.paletteDataStored);
+}
+
+// class DataLatchDebug
+
+VDP::DataLatchDebug::DataLatchDebug(VDP &vdp_)
+	: SimpleDebuggable(vdp_.getMotherBoard(),
+			vdp_.getName() + " data latch value", "V99x8 data latch value (byte)", 1)
+{
+}
+
+byte VDP::DataLatchDebug::read(unsigned /*address*/)
+{
+	auto& vdp = OUTER(VDP, dataLatchDebug);
+	return vdp.dataLatch;
+}
 
 // class Info
 
--- a/src/video/VDP.hh
+++ b/src/video/VDP.hh
@@ -988,6 +988,21 @@ private:
 		void write(unsigned address, byte value, EmuTime::param time) override;
 	} vramPointerDebug;
 
+	struct RegisterLatchStatusDebug final : SimpleDebuggable {
+		explicit RegisterLatchStatusDebug(VDP& vdp);
+		[[nodiscard]] byte read(unsigned address) override;
+	} registerLatchStatusDebug;
+
+	struct PaletteLatchStatusDebug final : SimpleDebuggable {
+		explicit PaletteLatchStatusDebug(VDP& vdp);
+		[[nodiscard]] byte read(unsigned address) override;
+	} paletteLatchStatusDebug;
+
+	struct DataLatchDebug final : SimpleDebuggable {
+		explicit DataLatchDebug(VDP& vdp);
+		[[nodiscard]] byte read(unsigned address) override;
+	} dataLatchDebug;
+
 	class Info : public InfoTopic {
 	public:
 		void execute(std::span<const TclObject> tokens,


src/video/VDP.hh
```diff
@@ -1262,6 +1267,11 @@ private:
 	/** Cached CPU reference */
 	MSXCPU& cpu;
 	const byte fixedVDPIOdelayCycles;
+
+	enum {
+		VALUE,
+		ADDRESS,
+	} latchStatus;
```
m9710797: Any reason why you introduce a new variable for this? To me it seems this variable should "normally" be in sync with the existing variable `registerDataStored`.

In general, when you want to expose **existing** state as debug information you want to avoid introducing **new** variables for that state, and then hope to keep that in sync with what you want to expose (with the risk of getting it wrong). For example in this case you missed the fact that writing to port 0x98 (on line 658) resets the latched-state.

m9710797: In the current implementation the values `VALUE` and `ADDRESS` are exposed to Tcl. So these numerical values are public and must remain stable. Therefor I would suggest to make these numerical value explicit, with something like: `enum { VALUE = 0, ADDRESS = 1 }`.

Though because of my previous comment I would suggest to not have this enum at all, and just use `registerDataStored`.

src/video/VDP.cc
```diff
@@ -99,6 +100,7 @@ VDP::VDP(const DeviceConfig& config)
 		Setting::SaveSetting::SAVE)
 	, cpu(getCPU()) // used frequently, so cache it
 	, fixedVDPIOdelayCycles(getDelayCycles(getMotherBoard().getMachineConfig()->getConfig().getChild("devices")))
+	, latchStatus(VALUE)
```
m9710797: See previous comment: not needed.

src/video/VDP.cc
```diff
@@ -658,6 +660,7 @@ void VDP::writeIO(word port, byte value, EmuTime::param time_)
 		registerDataStored = false;
 		break;
 	case 1: // Register or address write
+		latchStatus = latchStatus == VALUE ? ADDRESS : VALUE;
```
m9710797: See previous comment: not needed and also wrong in corner cases.

src/video/VDP.cc
```diff
@@ -966,6 +969,7 @@ byte VDP::readIO(word port, EmuTime::param time)
 {
 	assert(isInsideFrame(time));
 
+	latchStatus = VALUE;
```
m9710797: See previous comment: not needed.

src/video/VDP.hh
```diff
@@ -988,6 +988,11 @@ private:
 		void write(unsigned address, byte value, EmuTime::param time) override;
 	} vramPointerDebug;
 
+	struct LatchStatusDebug final : SimpleDebuggable {
+		explicit LatchStatusDebug(VDP& vdp);
+		[[nodiscard]] byte read(unsigned address) override;
+	} latchStatusDebug;
```
m9710797: I support the intend of this patch. Though I'm not sure if this is the right approach.

This debuggable exposes a single bit of information: whether there currently is a byte latched or not.
But then the next question will be: what is the value of this latched byte?

We could make a second debuggable to query this value. Or we could extend the size of this debuggable to two and expose this byte as the second position. Though that seems arbitrary: does the 1-bit-latch or the 8-bit-value go in position 0 or 1?

Ah, and next to `registerDataStored` there is also `paletteDataStored` (*1). Maybe we should then also expose this in the debuggable? Maybe the latched value in position 0, register-latched in position 1 and palette-latched in position 2?

(*1) Now that I look at it this from this new perspective: could it be that the real HW only has one latch-status for both `registerDataStored` and `palletteDataStored`? We should verify this on real HW. And if it turns out the real VDP only has a single latched-status, then we have to break the interface of our debuggable. That's the danger of exposing detailed information, maybe too detailed. Is the MSX programmer really interested in this?

m9710797 (commented):
Good idea, it's useful information to expose to the debugger. But the implementation needs some rework.

src/video/VDP.hh
pvmm: That was my first version, but I must have messed something up while testing and I thought it was not really the same thing.

src/video/VDP.hh
sdsnatcher: Some context: The use case for a real MSX programmer is described in this feature request:

https://github.com/openMSX/debugger/issues/148


@pvmm is trying to implement that feature request, that will solve a huge nightmare most MSX programmers have when the Z80/VDP communication gets out of sync. (The OUTs to the port 99h mismatch the latch state).

grauw: The openMSX implementation for those flags seems to be correct, see attached test. Both openMSX and real MSX2 give me a yellow background with an # and a @ symbol at location 0,0.

[vdplatch.zip](https://github.com/openMSX/openMSX/files/10572603/vdplatch.zip)

sdsnatcher: @m9710797 , it might not be the optimal approach, but also it’s not that bad for a first implementation, is it?

At least the MSX programmer would have the feature.

Thinking in agile terms, it can be refined in more interactions for a more ideal approach.

m9710797: Sure, I agree it's a good idea.

But we should also be careful with breaking backwards compatibility in the debuggables. So if we can think/discuss a few days and then maybe have a better solution, than that's worth it IMHO.

Also because at the moment we don't have a good mechanism to detect the exact interface in the debugger and then adapt to it (if it would change between different openMSX versions).

But grauw's test has put me at easy. It's less likely we'll need to change it later.
Thanks grauw!

@pvmm: this discussion triggers another question/comment:
I assume one of your next patches will be to use this new debuggable in the debugger. Make sure that the debugger still works (e.g. doesn't crash) when combined with an older openMSX version that doesn't offer this new debuggable yet.

src/video/VDP.hh
pvmm: If I remove the enum how would the user know the meaning of `0` and `1` that they read from `debug read_block {VDP latch status} 0 1`? Maybe I should rename it to {register data stored} to interpret it as a boolean.

m9710797: I'm not sure I understand your question. I mean you have a valid point. An openMSX debuggable only exposes an array of bytes, it doesn't specify how that data should be interpreted (for stuff like RAM the interpretation is obvious, but not in this case).
But whether you use en enum **internally** in the openMSX code doesn't change that (unless you expect users to read the openMSX source code). The only solution is to document the interpretation somewhere.

pvmm: image
I've placed it in the description then.

m9710797: Good.

Can you also comment on how (and/or if) you plan to expose the actual value of the latched byte, and the palette-latch-status (`paletteDataStored`).

If you do want to extend this debuggable with new fields, then (later) also extend the description with the different addresses in the debuggable. And is `VDP latch status` then still a good name? Maybe it is.

If you do want to expose all three in the same debuggable then maybe the order 0->value, 1->register-latch, 2->palette-latch is the most logical? But that would mean the current exposed info (register-latch-status) shifts position.

Or if you want to expose them all as separate debuggables, then good names might be `VDP latch value`, `VDP register latch status` and `VDP palette latch status`? But that means the current debuggable must be renamed.

src/video/VDP.hh
pvmm: > We could make a second debuggable to query this value. Or we could extend the size of this debuggable to two and expose this byte as the second position. Though that seems arbitrary: does the 1-bit-latch or the 8-bit-value go in position 0 or 1?

I don't think that is a problem provided we define a standard for debuggables of latch statuses and stick to it. On that note, I think it makes more sense to put the status of the latch and then the value of the latch, since it is in the name of the debuggable itself, it should be the first (more important) piece of data.

On second thought, when debugging we know the value in the latch, since we can see the `out`s on the code view widget. I think it may be redundant.

> Ah, and next to `registerDataStored` there is also `paletteDataStored` (*1). Maybe we should then also expose this in the debuggable? Maybe the latched value in position 0, register-latched in position 1 and palette-latched in position 2?

Maybe a `{VDP latch statuses}` debuggable containing the two latches. In the two LSBs of the char?

> I assume one of your next patches will be to use this new debuggable in the debugger. Make sure that the debugger still works (e.g. doesn't crash) when combined with an older openMSX version that doesn't offer this new debuggable yet.

It starts by probing the contents of `debug list` for the existence of `{VDP latch status}`.

m9710797: > I don't think that is a problem provided we define a standard for debuggables of latch statuses and stick to it. On that note, I think it makes more sense to put the status of the latch and then the value of the latch, since it is in the name of the debuggable itself, it should be the first (more important) piece of data.

Ok. There are many designs that make sense. My main question was: "Have you already thought about this?" I'd like to avoid that later we extend this (e.g. with the palette stuff) but then it doesn't fit nicely anymore. For example if we picked the order 0->register-latch-status, 1->latch-value. Then later we can't cleanly extend it anymore with 2->palette-latch-status ... "Why is the order inconsistent: one 'status' in front and the other behind the 'value'?"

In other words, I'd like to see the larger design first before approving a small part of it. "See" either via code or via a textual description. And then it's fine if you don't yet fully implement it. But I want to avoid that we're painting ourselves into a corner. (And of course we can't predict everything we might want to change in the future, but we do already know about two possible extension on top of the current patch).
 
> On second thought, when debugging we know what's in the latch, since we can see the `out`s on the code view widget. I think it may be redundant.

Well, you could say the same about the latch-status.
Typically you're debugging exactly because something has gone wrong. And then maybe the OUT that wrote the latched-value is not easily visible (e.g. because the two OUTs are not cleanly paired anymore).

I don't feel strongly about whether this value should be shown in the debugger (by default) or not. But if we're extending openMSX with exposing this info, then IMHO we should try to be complete. In other words: it's fine for me that the debugger does not support visualizing everything that openMSX exposes.
 
> Maybe a `{VDP latch statuses}` debuggable containing the two latches. In the two LSBs of the char?

That's possible. And fine for me if you prefer that.
On the other hand, there's no advantage in cramming bits together (a larger debuggable doesn't cost anything more than a smaller debuggable). And using this information (e.g. from a Tcl script) might be more convenient if you can directly query (only) the stuff you're interested in, instead of also having to mask the desired bit. But I don't feel strongly about this, pick what you prefer.

> It starts by probing the contents of `debug list` for the existence of `{VDP latch status}`.

OK.

pvmm: There is a single `dataLatch` for all latch statuses. Would it be redundant to put the same data on the second byte of both `{VDP latch status}` and `{Palette latch status}`? I am leaning in favour of a third debuggable for the latch value alone actually.

m9710797: All fine for me. I don't mind repeating the same latched-value in both debuggables.

src/video/VDP.cc
```diff
@@ -1682,6 +1685,47 @@ void VDP::VRAMPointerDebug::write(unsigned address, byte value, EmuTime::param /
 	}
 }
 
+// class RegisterLatchStatusDebug
+
+VDP::RegisterLatchStatusDebug::RegisterLatchStatusDebug(VDP &vdp_)
+	: SimpleDebuggable(vdp_.getMotherBoard(),
+			vdp_.getName() + " register latch status", "V99x8 register latch status (0 = value, 1 = register)", 1)
```
m9710797: Detail: I understand what this means. But for occasional users it might be ambiguous whether value/register means the thing that was last written, or the thing that is expected to be written next.

Same for RB/G below.

m9710797 (commented):
Looks good.
Only the description should be tweaked a bit. 
Thanks!

src/video/VDP.cc
pvmm: Should I describe it more or less in this line: `v99x8 register latch status (0 = empty, 1 = stored value)`?

pvmm: And `v99x8 palette latch status (0 = expected red & blue components, 1 = expected green component)`

m9710797:
Thanks for keeping up with my many review comments.

PR #1489, opened 2023-02-06
Expose VRAM write access status to Tcl

pvmm:
Finishing the changes that started on c99fa3b

https://github.com/openMSX/debugger/issues/148

### DIFF
--- a/src/video/VDP.cc
+++ b/src/video/VDP.cc
@@ -79,6 +79,7 @@ VDP::VDP(const DeviceConfig& config)
 	, vdpPaletteDebug  (*this)
 	, vramPointerDebug (*this)
 	, registerLatchStatusDebug(*this)
+	, vramAccessStatusDebug(*this)
 	, paletteLatchStatusDebug(*this)
 	, dataLatchDebug   (*this)
 	, frameCountInfo   (*this)
@@ -262,6 +263,7 @@ void VDP::resetInit()
 	dataLatch = 0;
 	cpuExtendedVram = false;
 	registerDataStored = false;
+	writeAccess = false;
 	paletteDataStored = false;
 	blinkState = false;
 	blinkCount = 0;
@@ -685,6 +687,7 @@ void VDP::writeIO(word port, byte value, EmuTime::param time_)
 				}
 			} else {
 				// Set read/write address.
+				writeAccess = value & 0x40;
 				vramPointer = (value << 8 | dataLatch) & 0x3FFF;
 				if (!(value & 0x40)) {
 					// Read ahead.
@@ -1699,6 +1702,21 @@ byte VDP::RegisterLatchStatusDebug::read(unsigned /*address*/)
 	return byte(vdp.registerDataStored);
 }
 
+// class VramAccessStatusDebug
+
+VDP::VramAccessStatusDebug::VramAccessStatusDebug(VDP &vdp_)
+	: SimpleDebuggable(vdp_.getMotherBoard(), vdp_.getName() == "VDP" ?
+			"VRAM access status" : vdp_.getName() + " VRAM access status",
+			"VDP VRAM access status (0 = ready to read, 1 = ready to write)", 1)
+{
+}
+
+byte VDP::VramAccessStatusDebug::read(unsigned /*address*/)
+{
+	auto& vdp = OUTER(VDP, vramAccessStatusDebug);
+	return byte(vdp.writeAccess);
+}
+
 // class PaletteLatchStatusDebug
 
 VDP::PaletteLatchStatusDebug::PaletteLatchStatusDebug(VDP &vdp_)
@@ -1880,6 +1898,7 @@ int VDP::MsxX512PosInfo::calc(const EmuTime& time) const
 // version 7: removed cpuVramReqAddr again, fixed issue in a different way
 // version 8: removed 'userData' from Schedulable
 // version 9: update sprite-enabled-status only once per line
+// version 10: added writeAccess
 template<typename Archive>
 void VDP::serialize(Archive& ar, unsigned serVersion)
 {
@@ -1968,6 +1987,11 @@ void VDP::serialize(Archive& ar, unsigned serVersion)
 	} else {
 		assert(Archive::IS_LOADER);
 		spriteEnabled = (controlRegs[8] & 0x02) == 0;
+	}
+	if (ar.versionAtLeast(serVersion, 10)) {
+		ar.serialize("writeAccess", writeAccess);
+	} else {
+		writeAccess = !cpuVramReqIsRead; // best guess
 	}
 
 	// externalVideo does not need serializing. It is set on load by the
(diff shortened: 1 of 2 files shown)


m9710797:
Hi, thanks for working on this.

About the code itself:
The only missing thing is including the new `writeAccess` variable in the savestate (and taking care of backwards compatibility when loading an older version of the savestates that don't include this new info yet).

Without including this in the savestate, the queried info might be wrong until the first time a VRAM read/write address is setup. And this might be annoying when you're using savestates in combination with debugging: this happens a lot when you're using the "step-back" functionality during debugging (internally "step-back" uses the "reverse" subsystem, and that in turn is built on top of savestates).

--

HOWEVER:
It's always suspicious when a new debuggable requires introducing new state in the code. I suspect that also here there may be a better way.

Maybe it's not "officially supported" behavior in the V99x8, but setting up a VRAM read/write address and then doing the **opposite** IN/OUT instruction actually does "work" (for some definition of "work"). In particular there is existing MSX software that does setup a VRAM write address, and then performs a sequence of OUT and **IN** operations. The result of the IN instruction is not used, but it does advance the internal address-counter. So you can use this trick to change a block of VRAM, but also skip modifying some of the bytes in this block: for a byte you want to change you do an OUT, for a byte you want to skip you do an IN (and discard the result of that IN).

More in detail: when you write to port 0x99 to setup a new VRAM address the following happens:
* The address is 'remembered'. In openMSX this is the `vramPointer` variable, and this is already exposed in the "VRAM pointer" debuggable.
* If the setup was a read, then the VDP will "schedule" this VRAM read.
* "Scheduling" here means waiting for an available "access slot". Because VRAM access is shared (between rendering, command-engine, CPU-access, ..), the CPU-VRAM-read operation can not always start immediately.
* When this scheduled read finishes, the resulting byte is 'remembered' in some internal register. In openMSX we use the `cpuVramData` variable for this.
* An IN from port 0x98 returns the value stored in this register, and also schedules the next VRAM read.

Writing to VRAM works like this:
* Setting up the VRAM address, only remembers the address (it doesn't schedule anything yet).
* An OUT to port 0x98 stores the byte in the same internal register `cpuVramData` and then schedules a VRAM write. We again have to wait till there's a VRAM access slot available.

This design has some consequences:
* When you write port 0x98 too fast, that is: write a new byte before the previous byte was actually written to VRAM, then the `cpuVramData` register gets overwritten, and then only 1 byte is actually written to VRAM with the value of the last OUT.
* You can mix IN and OUT instructions (to port 0x98). Though the first IN after an OUT returns the value of that OUT (not the VRAM content from the corresponding address). But this is fine if you only want to skip some bytes during writing.
* You can setup a VRAM-read-address, but then do an OUT 0x98 (instead of IN). (Depending on the exact timing) this will then write to the address one higher than the setup-address.
* And similar for the opposite situation: setup a write address, but then do an IN: the first IN returns a dummy value, and then next IN returns the VRAM content from the setup address.

Ah, and when the VDP "schedules" such a read/write operation, it also has to remember whether it was a read or a write. In openMSX that information is stored in the `cpuVramIsRead` variable.

Maybe this `cpuVramIsRead` variable could be a good replacement for the `writeAccess` variable you introduced in this commit?
* This variable is set to true (=read) when a vram-read-address is setup (via port 0x99).
* It's set to false (=write) when a byte is written to port 0x98. (So not immediately when a VRAM-write-address is setup).

So it's not exactly the same as the `writeAccess` stuff you introduced. But it more closely resembles how the real VDP works internally.

Maybe I should have started by asking "Why did you introduce this new debuggable?" I mean did you have a concrete (debugging) use case in mind? Would that use case also be solved if we instead expose `cpuVramIsRead`?

pvmm:
I think that with VDP latch detection in place, we can easily replace this by a Tcl script declared by the debugger on connect.

m9710797:
Probably it's indeed possible with a Tcl script.

But it's not 100% clear to me what use case you're trying to cover. Can you explain it to me, from the MSX programmers point of view? What kind of problems will be easier to debug with this feature? A mismatch between setting up the vram address and the following IN vs OUT 0x98 instruction? (Such a mismatch might also be intentional.)

pvmm:
I guess the more information the user has before a runaway the better. Any spurious write to VRAM can be identified on the spot. In the case of a write to an address in read mode, only the next address will be written to. The user may not be expecting that because he has set the watch point to the wrong address.

m9710797:
BTW, I'm not against your original patch (if you complete the savestate part). But I want to make sure that it really covers the intended use case. And I was wondering if the existing variable `cpuVramIsRead` didn't also (or better?) cover the use case.

src/video/VDP.hh
```diff
@@ -1217,6 +1222,10 @@ private:
 	  */
 	byte dataLatch;
 
+	/** Direction of VRAM access for reading or writing
+	 */
+	bool writeAccess;
+
```
m9710797, line 1232: Please add in this comment something like:
```
Note: this variable is _only_ used for the 'VRAM access status' debuggable.
The real VDP allows to setup a read/write VRAM address and then do the opposite out/in operation.
See the variables 'cpuVramData' and 'cpuVramReqIsRead' for more details.
```
Otherwise in a few year we may wonder why we're having both `writeAccess` and `cpuVramReqIsRead`.

src/video/VDP.cc
```diff
@@ -1931,7 +1949,8 @@ void VDP::serialize(Archive& ar, unsigned serVersion)
 	             "blinkState",           blinkState,
 	             "dataLatch",            dataLatch,
 	             "registerDataStored",   registerDataStored,
-	             "paletteDataStored",    paletteDataStored);
+	             "paletteDataStored",    paletteDataStored,
+	             "writeAccess",          writeAccess);
```
m9710797: Unfortunately this is insufficient:
As-is this change will break loading existing savestate files: it will fail with an error like `"missing tag <writeAccess>"`.
To properly implement this you must:
* Around line 1882: add a comment like `// version 10: added writeAccess`.
* In VDP.hh around line 1281: increase the version from 9 to 10.
* Around line 1972: add a block like:
```
if (ar.versionAtLeast(setVersion, 10)) {
    ar.serialize("writeAccess", writeAccess);
} else {
    writeAccess = !cpuVramReqIsRead; // best guess
}
```

The comment describes how the savestate format has evolved in the different versions. So that we don't have to go digging in git-history for this information. We then increase the version number from 9 to 10. These version numbers are local to the VDP object itself (so the full savestate has many sub-version-numbers for each of the different components). And then we have a code block that deals with loading version 10 (and higher) and versions 9 (and lower). For the lower versions we can't 100% correctly initialize `writeAccess`, so we make a best-effort attempt. (Note that at this point `cpuVramReqIsRead` must already be restored from the savestate, the order of these things is important).

For backwards compatibility, this "version 10" of the VDP savestate format is something we need to keep supporting forever. That's also a reason why I would have preferred to not introduce extra member variables only for a debuggable.

m9710797 (changes requested):
Thanks!

The savestate code needs some fixes (to deal with backwards compatibility).

src/video/VDP.cc
pvmm: I've tested in on my machine with VDP version 9 and 10. It's working fine now. 👍

m9710797:
Thanks.

PR #1492, opened 2023-02-15
getting GLEW to work on Wayland

pvmm:
GLEW fails to initialise because there is no GLX on wayland and the one provided by distros was built to use GLX instead of EGL. This PR ignores the GLEW_ERROR_NO_GLX_DISPLAY error and it is a temporary fix until this is fixed by GLEW upstream and released by the major distros. See nigels-com/glew#172

Without this patch, Wayland users will get an `unknown error` when starting OpenMSX and fall back to SDL renderer.


m9710797:
Thanks!

PR #1499, opened 2023-04-12
Add YM2151 recording to _vgmrecorder.tcl

qiqitori:
Note: the memory write watch point doesn't have any slot conditions. Ideally we should (at init time) first figure out which slot the SFG-01 is inserted into, and (at watch point time) discard any writes targeting the same address in other slots.

This could perhaps be done by iterating though:
machine_info 0..3 0 0
For example, without any other ROMs, machine_info 1 0 0 should return: {Yamaha SFG-01}
In such a case we should only record the event if ps==1.

The code to handle SCC does something like this.
However, this seems like a somewhat theoretical problem, as there is hardly any YM2151 software out there.

### DIFF
--- a/share/scripts/_vgmrecorder.tcl
+++ b/share/scripts/_vgmrecorder.tcl
@@ -18,6 +18,7 @@ variable directory [file normalize $::env(OPENMSX_USER_DATA)/../vgm_recordings]
 
 variable psg_logged       false
 variable fm_logged        false
+variable y2151_logged     false
 variable y8950_logged     false
 variable moonsound_logged false
 variable scc_logged       false
@@ -35,14 +36,14 @@ variable mbwave_title_hack       false
 variable mbwave_loop_hack	 false
 variable mbwave_basic_title_hack false
 
-variable supported_chips [list MSX-Music PSG MoonSound MSX-Audio SCC]
+variable supported_chips [list MSX-Music PSG MoonSound MSX-Audio SCC SFG-01]
 
 set_help_proc vgm_rec [namespace code vgm_rec_help]
 proc vgm_rec_help {args} {
         switch -- [lindex $args 1] {
                 "start"    {return {VGM recording will be initialised, specify one or more soundchips to record.
 
-Syntax: vgm_rec start <MSX-Audio|MSX-Music|MoonSound|PSG|SCC>
+Syntax: vgm_rec start <MSX-Audio|MSX-Music|Moonsound|PSG|SCC|SFG-01>
 
 Actual recording will start when audio is detected to avoid silence at the beginning of the recording. This mechanism will only work if the MSX and/or playback routine does not send data to the soundchip when not playing, recording will start immediately in those cases.
 }}
@@ -139,6 +140,7 @@ proc vgm_rec {args} {
 
 	variable psg_logged
 	variable fm_logged
+	variable y2151_logged
 	variable y8950_logged
 	variable moonsound_logged
 	variable scc_logged
@@ -211,6 +213,7 @@ proc vgm_rec {args} {
 		}
 		set psg_logged       false
 		set fm_logged        false
+		set y2151_logged     false
 		set y8950_logged     false
 		set moonsound_logged false
 		set scc_logged       false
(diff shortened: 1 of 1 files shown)


qiqitori:
Hi, this is a tiny feature; doubt there is a lot of interest.
1) There's a big caveat, please see the commit message for details. If you'd prefer, I can try to implement some code to search for the SFG-01.
2) Furthermore, I haven't managed to test if this breaks VGM recording for other chips. (I don't have any software to test VGM recording for other chips.) Is there something like a test case for that?

MBilderbeek:
Hi,

Thank you for making this patch!

Some remarks:
- perhaps SFG-01 is a bit confusing, as there is also the SFG-05, which contains also the OPM (or actually OPP, which is practically the same)
  - To find the proper device(s) with the correct slot, you could do the following:
  -  `machine_info sounddevice` lists all sound devices
  - For each of these devices, use `machine_info device <sounddevice>` and check whether the type is `YamahaSFG`. 
  - For each sound device of that type, you have to find the slot in which it is active. For all primary and secondary slots (`ps` and `ss`), check where you find the sound device you're looking for with `machine_info slot ps ss 1` (probably page 1 is good enough).
-  You can simply run a Konami SCC game for instance (like Salamander, Space Manbow, etc.) and record PSG and SCC into a VGM. I have never used VGM recording myself, so perhaps you can try it.

m9710797 (changes requested):
Hi, thanks a lot for working on this!

I reviewed your patch, with the following caveats:
* I didn't check the details about the Y2151 registers (e.g. are the address correct).
* I didn't check the vgm file format (e.g. is the value 0x54 correct)
But I assume you tested your script, and the resulting .vgm file results in a correct Y2151 recording? Then for sure the above details will be correct.

So purely from an openMSX and a Tcl point of view, your patch looks correct. Thanks!

My preference is also that you do still extend your patch to only react to writes to the correct (sub)slot. The strategy suggested by Manuel to look for the correct slot looks OK to me. But feel free to ask if you need any help implementing that.

Thanks,

Wouter

qiqitori:
Thank you for the reviews! Sorry for taking two weeks to get back to this.

I copied and pasted most of the required code from the find_all_scc function. I tested that recording still works with my own software (both with the SFG-01 in slot 2 and in slot 3) and Konami's Space Manbow. Using 'debug list_watchpoints' I also verified that only the expected watchpoints were set:

With SFG-01 in slot 2:
```
> debug list_watchpoints
wp#5 write_mem 0x3ff0 {[watch_in_slot 2 0]} vgm::write_y2151_address
wp#6 write_mem 0x3ff1 {[watch_in_slot 2 0]} vgm::write_y2151_data
```
With SFG-01 in slot 3:
```
> debug list_watchpoints
wp#15 write_mem 0x3ff0 {[watch_in_slot 3 0]} vgm::write_y2151_address
wp#16 write_mem 0x3ff1 {[watch_in_slot 3 0]} vgm::write_y2151_data
```

> perhaps SFG-01 is a bit confusing, as there is also the SFG-05, which contains also the OPM (or actually OPP, which is practically the same)

I think the addresses are probably the same on the SFG-05, but I don't have one to test with. Therefore I don't 100% know if this all works with software targeting the SFG-05. The VGM spec (https://github.com/vgmrips/vgmplay/blob/master/VGMPlay/vgmspec171.txt) mentions that the YM2164 used on the SFG-05 expects bit 31 to be set in a certain field in the header; I don't have support for that either. (It doesn't really look like vgmplay actually cares about that though)

> I didn't check the details about the Y2151 registers (e.g. are the address correct).

Yes, worked on real hardware using these addresses. Here's some documentation showing the addresses: http://map.grauw.nl/resources/midi/ym2148.php

> I didn't check the vgm file format (e.g. is the value 0x54 correct)

Yes, 0x54 is the value for YM2151 commands, here's the relevant documentation: https://vgmrips.net/wiki/VGM_Specification#Commands

The header should be correct too. I tested my recorded audio using vgmplay and it sounds exactly the same, and the header dump looks good too.

If you want I can add a link to a cartridge ROM file that plays the first 20 seconds of a VGM using the SFG-01. However, the VGM track is copyrighted and I don't have any rights to it.

m9710797:
Thanks!

MBilderbeek:
@qiqitori Just a question: as far as I understand the code, you are now properly searching for any YamahaSFG device. But that will also find the SFG-05 based device, which is the YM2164. Was that intentional?
If yes, the code has a bit misleading names (everywhere SFG-01) and you did mention there are some differences perhaps in the VGM format for it. 
I'd personally like the SFG-05 also supported, even if it's not perfect (better than nothing, I guess).

qiqitori:
> @qiqitori Just a question: as far as I understand the code, you are now properly searching for any YamahaSFG device. But that will also find the SFG-05 based device, which is the YM2164. Was that intentional? If yes, the code has a bit misleading names (everywhere SFG-01) and you did mention there are some differences perhaps in the VGM format for it. I'd personally like the SFG-05 also supported, even if it's not perfect (better than nothing, I guess).

Hi, sorry for again taking ages to get back to this. This wasn't really intentional, though I did have a feeling that the code would probably work with the SFG-05. I just performed a quick test and recording stuff played back on the SFG-05 seems to work just fine! However, I don't have access to a real SFG-05 and don't really know if the program I wrote (which targets the SFG-01) works with a real hardware SFG-05 too.

Therefore it might make sense to change the command name to "SFG".

MBilderbeek:
OK, did that, please confirm by testing it again :)

PR #1561, opened 2023-11-21
build/flavour-unittest: space between CXXFLAGS assignment operator

ajakk:
Without this, GNU Make (at least) seems to treat the variable as `CXXFLAGS+` as the variable when CXXFLAGS is already defined, resulting in the unittest CXXFLAGS being effectively ignored rather than appending to the already existing CXXFLAGS.

### DIFF
--- a/build/flavour-unittest.mk
+++ b/build/flavour-unittest.mk
@@ -2,7 +2,7 @@
 # Build executable that runs unit tests.
 
 # Debug flags.
-CXXFLAGS+=-O3 -g -DUNITTEST -IContrib/catch2 -fsanitize=address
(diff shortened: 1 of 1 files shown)


m9710797:
Hi ajakk,

I don't think this patch solves the problem, with or without spaces around '+=' makes no difference.
I tested with this dummy Makefile:

```
A+=bar

all:
        echo $A
```

Executing via just `make`, prints `bar`. Executing via `make A=foo` prints `foo`, not `foo bar`. And adding spaces doesn't make any difference.

@mthuurne Do you know how to fix this?

ajakk:
Ah, no, you're right, I've confused myself here and this isn't the real cause.

PR #1569, opened 2023-12-11
YMF262: Don’t mirror control registers to 2nd bank

grauw:
The old code originating back to the initial submit mirrored control registers `0x001`, `0x002`, `0x003`, `0x008` to `0x101`, `0x102`, `0x103`, `0x108` in the 2nd bank (but *not* `0x104` and `0x105` which early return), whilst this does not match the [OPL3 datasheet](http://map.grauw.nl/resources/sound/yamaha_ymf262.pdf).

This does not occur on a real OPL3, as evidenced by a write to `0x102` affecting timer register `0x002` in openMSX, but not on real hardware.

I found this because Omega reported to me that a homebrew OPL3 cartridge he recently built was not playing at the correct speed in VGMPlay. This turned out to be the fault of the \INT line not being connected on the cartridge, however in the meanwhile with one of his two test files (Doom track 2) I was able to reproduce similar behaviour in openMSX. This VGM would write `0x00` to register `0x102`, which VGMPlay does not mask out, and due to the implemented mirroring it affected the speed of timer-1 and thus the playback speed.

After fixing the \INT line on his cartridge, Omega reported that the track now plays fine with VGMPlay, which indicates that the mirroring as implemented in openMSX is incorrect and probably an oversight rather than intentional.

Unfortunately I don’t have an OPL3 cartridge to perform any additional tests with myself, but I believe Omega’s test to be sufficient and do not have reason to believe this mirroring exists in real hardware. Nuked-OPL3 also [does not implement](https://github.com/nukeykt/Nuked-OPL3/blob/master/opl3.c#L1368) this.

Test case: [opl3-no-control-mirror.zip](https://github.com/openMSX/openMSX/files/13640665/opl3-no-control-mirror.zip) (run on MSX2 with DOS2 and sufficient RAM)

### DIFF
--- a/src/sound/YMF262.cc
+++ b/src/sound/YMF262.cc
@@ -1046,6 +1046,35 @@ void YMF262::writeRegDirect(unsigned r, uint8_t v, EmuTime::param time)
 	reg[r] = v;
 
 	switch (r) {
+	case 0x001: // test register
+		break;
+
+	case 0x002: // Timer 1
+		timer1->setValue(v);
+		break;
+
+	case 0x003: // Timer 2
+		timer2->setValue(v);
+		break;
+
+	case 0x004: // IRQ clear / mask and Timer enable
+		if (v & 0x80) {
+			// IRQ flags clear
+			resetStatus(0x60);
+		} else {
+			changeStatusMask((~v) & 0x60);
+			timer1->setStart((v & R04_ST1) != 0, time);
+			timer2->setStart((v & R04_ST2) != 0, time);
+		}
+		break;
+
+	case 0x008: // x,NTS,x,x, x,x,x,x
+		nts = (v & 0x40) != 0;
+		break;
+
+	case 0x101: // test register
+		break;
+
 	case 0x104:
 		// 6 channels enable
 		channel[ 0].extended = (v & 0x01) != 0;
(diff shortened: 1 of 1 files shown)


src/sound/YMF262.cc
```diff
@@ -1045,75 +1045,81 @@ void YMF262::writeRegDirect(unsigned r, uint8_t v, EmuTime::param time)
...
-		case 0x04: // IRQ clear / mask and Timer enable
-			if (v & 0x80) {
-				// IRQ flags clear
-				resetStatus(0x60);
-			} else {
-				changeStatusMask((~v) & 0x60);
-				timer1->setStart((v & R04_ST1) != 0, time);
-				timer2->setStart((v & R04_ST2) != 0, time);
+		if ((r & 0x100) == 0) {
+			switch (r & 0x1F) {
```
m9710797: Thanks a lot for your PR.
May I suggest one small tweak:
* Remove the test for `((r & 0x100) == 0)`.
* And instead do a `switch (r)` (instead of `switch (r & 0x1f)`), which then handles both 0x000-0x01f and 0x100-0x11f.

The main advantage is a reduction in indentation level.

m9710797:
Thanks for making the change.
I merged your PR but with some additional edits (because I didn't explain well enough what I meant).
Please review.

PR #1576, opened 2024-01-16
updated some game names to match softwaredb

oksoelsha:
As discussed on msx.org, this is to update some game names to match what's in softwaredb. The _trainerdefs.tcl file seems to be generated though. In this case feel free to reject this PR and maybe make the same changes in the original source.

### DIFF
--- a/share/scripts/_trainerdefs.tcl
+++ b/share/scripts/_trainerdefs.tcl
@@ -131,7 +131,7 @@ create_trainer "A Life M36 Planet" {time 1} {
 	"Life" {dpoke 0xc527 99;poke 0xd3c2 99}
 }
 
-create_trainer "A-Na-Za - Kaleidoscope Special" {time 1} {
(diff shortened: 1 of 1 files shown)


Vampier:
that file was auto generated using a database (part of the romdb)

m9710797:
@Vampier Could you use this patch to fix some mistakes (or at least inconsistencies) in the database(s)?

Vampier:
the trainer filenames are disconnected from the romdb names - I stopped
maintaining the trainer definitions and the linkage between them quite some
time ago (since the last update of the trainer file)

I also decided to go with the naming convention like 'The Maze of Galious'
instead of 'Maze Of Galious, The' but I'm wondering if this was a smart
thing to do.

Anyway - I'm ok with merging this trainer definition as I'm no longer
updating it

m9710797:
Merged.
Thanks oksoelsha.

PR #1585, opened 2024-02-21
Add RS232Net pluggable with IP232 support

retrocomputacion:
This adds a RS232 <-> Socket network pluggable, with optional user selectable IP232 support (for in conjunction with a virtual modem software such as TCP2SER).

Two new settings added:

rs232net-address: Network address in the format `IP:PORT` or `URL:PORT`
rs232net-ip232: Turn ip232 protocol `on` or `off`


m9710797:
Hi retrocomputacion,

Thanks a lot for this PR!

Before doing a more detailed code review I'd like to ask for some background information. Because I'm not familiar with IP232 nor TCP2SER. A pointer to some (external) documentation would be useful.

Could you explain how this could be used? What MSX software to use? And how/what to setup on the host machine(s)?

I only had a _very_ brief look at the code. You seem to be using BSD sockets, right? AFAIK that doesn't compile on Windows. That's fine, but then we need to make sure we have proper #ifdef guards to prevent this code from (trying to) compile on windows. Alternatively we could try to rework this PR to use the "Socket" abstraction we have in "src/events/Socket.hh". That one does also work on windows. But probably it doesn't yet support everything you need (e.g. for sure ipv6 support is missing).

Wouter

retrocomputacion:
> Hi retrocomputacion,
> 
> Thanks a lot for this PR!
> 
> Before doing a more detailed code review I'd like to ask for some background information. Because I'm not familiar with IP232 nor TCP2SER. A pointer to some (external) documentation would be useful.
> 
> Could you explain how this could be used? What MSX software to use? And how/what to setup on the host machine(s)?
> 
> I only had a _very_ brief look at the code. You seem to be using BSD sockets, right? AFAIK that doesn't compile on Windows. That's fine, but then we need to make sure we have proper #ifdef guards to prevent this code from (trying to) compile on windows. Alternatively we could try to rework this PR to use the "Socket" abstraction we have in "src/events/Socket.hh". That one does also work on windows. But probably it doesn't yet support everything you need (e.g. for sure ipv6 support is missing).
> 
> Wouter

Hi, TCPSer is a modem emulator, is used to bridge either a real RS-232 or local IP:port with an external IP connection.
It emulates the basic AT commands used in smart modems.
The official repository is https://github.com/go4retro/tcpser/
The IP232 protocol was first implemented as a hack in the VICE emulator, and later officially in both VICE and TCPSer, it uses the byte 0xFF as a prefix to send or receive the status of the RS232 port pins (currently DCD, DTR and RI)

For example (adapted from VICE's wiki):
```
tcpser->OpenMSX

0xff nn ->
 nn = 0      DCD = false     
 nn = 1      DCD = true
 nn = 255    literal 0xff
other   ->   unchanged

OpenMSX->tcpser

0xff nn ->
 nn = 0      DTR = false
 nn = 1      DTR = true
 nn = 255    literal 0xff
other   ->   unchanged
```
This means that the terminal software on the emulated machine can detect an active connection, and also 'hang up' the line to terminate it.

Usage of OpenMSX with TCPSer would be as follow:

- Start tcpser (at least before enabling the rs232net pluggable):
```
tcpser -p 6400 -v 25232 
```
This is the minimum setup for tcpser
`-p 6400` is the listening local port for incoming connections 
`-v 25232` is the local port for the IP232 connection with the emulator
(TCPSer has more settings, for logging and debugging, and phone book presets)

- In OpenMSX select a machine with built-in RS-232 or add one of the RS-232 extensions
- If it hasn't been done before, set the address for the rs232net pluggable in the console, for this example will be the localhost and the port passed to TCPSer: `set rs232-net-address 127.0.0.1:25232`
- Turn on IP232 support: `set rs232-net-ip232 on`
- Now select `rs232-socket` for the RS232 port connector.
- A reset may be necessary (for best results pause the emulation at or before the startup screen and complete these first steps before unpausing.
- To test from BASIC: `CALL COMTERM` will  start the built-in ROM terminal. Typing `at<ENTER>` should return `OK`.
- Try connecting to a external server: `atd"sotanomsxbbs.org:23"<ENTER>` if the server is online TCPSer should respond with `CONNECT` and start receiving data
- To manually terminate a connection first we need to go back to command mode by typing `+++` in rapid succession
(with no caracters preceding of following), once TCPSer responds with `OK` type `ath<ENTER>` to "hang up" the line.

Picking up an incoming connection:
For this example a native terminal software is needed, I'll use PuTTY
- Starting with TCPSer, OpenMSX already setup correctly and COMTERM running in the emulated machien
- Set up you session connection type as `Other->RAW` and set the host/ip address to `127.0.0.1:6400`
- Open the connection, in OpenMSX a series of `RING` will appear on screen
- To accept the connection, type `ata<ENTER>`, `CONNECT` will be shown on the screen
- Now whatever you type on OpenMSX will appear in PuTTY, depending on your settings what your type in PuTTY will appear on OpenMSX as you type or after pressing <Enter>

Other terminal software to test with is COMS6 which supports ANSI color text at 80 columns.

Direct connection without using TCPSer is also possible, in these case `rs232-net-ip232` must be set to OFF and rs232-net-address` must be set to the server you want to connect to. But the pluggable must not be enabled until a terminal program is running in OpenMSX.

I think this covers pretty much all there is to it. I don't know of many communications software for MSX (I come from the Commodore 64 where there's a lot of those type of programs). I know there's a terminal for SymbOS but I haven't tried it.

I'll take a look at the sockets issue. It didn't even occurred to me that it didn't work under windows, sorry.

retrocomputacion:
I checked the code again and I'm actually already using the OpenMSX's socket abstraction. Only problem I find is that I'm not calling `sock_startup()`, I see that it is called in src/events/CliServer.cc, but I don't know if that's called on every session.

retrocomputacion:
I tried compiliing under windows (win7) last night, made more advances, but there's still problems with some macros that don't exist on windows. I'll do a commit once I got it all sorted out

m9710797:
[I started typing this message before I saw your last message]
[Sorry, it seems that some the layout of the code snippets is messed up when formatted as a github comment. I hope it's still understandable.]

Hi,

Thanks for the explanation. Also thanks for your updated PR.

Unfortunately it still doesn't compile in windows. I haven't investigated myself yet (and I'm not a windows user), but based on the build logs from the github actions it seems to be a link error for your new R2232Net class. That probably means we should add it in the msvc build project: "build/msvc/openmsx.vcxproj" and "build/msvc/openmsx.vcxproj.filters" (typically I just manually edit those files, I don't know of a better way).

I started reviewing the code itself. I propose to do this in two steps:
* First I look at the code, but only locally. And I give feedback about small local details.
* Second I actually try to understand the code, and then maybe I can give feedback on the full design.

Today I'll only do step 1. I didn't have time yet for step 2.
Also note that I may have written much of the openMSX code in src/serial. But that was literally over 20 years ago. It may also take a bit of time for me to understand my own code again :-)

So today only some details. Some about code style. I'm happy to change some of these details myself after pulling the PR. But I'll mention them anyway.


--- a/src/serial/I8251.hh
+++ b/src/serial/I8251.hh
@@ -18,6 +18,8 @@ public:
 	virtual void setRxRDY(bool status, EmuTime::param time) = 0;
 	virtual void setDTR(bool status, EmuTime::param time) = 0;
 	virtual void setRTS(bool status, EmuTime::param time) = 0;
+
+	[[nodiscard]] virtual bool getDCD(EmuTime::param time) = 0; // This is not an i8251 signal but I dont know where else to declare it
 	[[nodiscard]] virtual bool getDSR(EmuTime::param time) = 0;
 	[[nodiscard]] virtual bool getCTS(EmuTime::param time) = 0; // TODO use this
 	virtual void signal(EmuTime::param time) = 0;

Adding getDCD() here indeed seems like a good place. Retain the comment that it's not a I8251 signal, but omit the part about "not knowing where to put it".

The following files were changed (only) because you added this new virtual method:
    MSXFacMidiInterface.{hh,cc} MSXMidi.{hh,cc} RS232Device.{hh,cc}
These classes all only implement this method as "return true;"
An alternative could be to implement that as a default implementation only once here.
That would make the PR smaller, but I'm not sure it makes the code clearer. Either alternative is fine for me.


--- a/src/serial/MSXRS232.cc
+++ b/src/serial/MSXRS232.cc
@@ -239,6 +239,10 @@ byte MSXRS232::readStatus(EmuTime::param time)
  	// TODO bit 0: carrier detect
+	if (!interface.getDCD(time)) {
+		result |= 0x01;
+	}
+
 	if (!rxrdyIRQenabled && switchSetting && switchSetting->getBoolean()) {
 		result |= 0x08;
 	}

This change implements the TODO that's 2 lines above it, right? Then please remove that comment.




diff --git a/src/serial/I8254.cc b/src/serial/I8254.cc
index b1ed95c96..964cc2497 100644
--- a/src/serial/I8254.cc
+++ b/src/serial/I8254.cc
@@ -251,8 +251,8 @@ void Counter::writeLoad(uint16_t value, EmuTime::param time)
 		if (clock.isPeriodic()) {
 			counter = counterLoad;
 			EmuDuration::param high = clock.getTotalDuration();
-			EmuDuration total = high * counter;
-			output.setPeriodicState(total, high, time);
+			EmuDuration::param total = high * counter;
+			output.setPeriodicState(total, total / 2, time);
 		} else {
 			// TODO ??
 		}

I'm curious why you made this change. Or better said: how you discovered that this needed to be changed.

Like I said: it's +20 years ago since I wrote this code. I re-read the I8254 datasheet for just a few minutes (so it's very possible that I'm mis-interpreting it), and I think we actually need both the original and the new version of this code:
* The old version is needed for mode 2: this mode generates a pulse: short duration high, long duration low.
* The new version is needed for mode 3: this mode generates a square wave: high and low are active for "halve" the time.

Actually it's not always exactly "halve" the time. When 'counter' is odd, then high and low are not active for the same amount of time. It's more like:
* low-duration:  counter / 2                 (round down)
* high-duration: counter - (counter / 2)     (round up)
But please double-check that I didn't swap round-up and round-down.

Detail: you use the type 'EmuDuration::param', but it should just be 'EmuDuration'.
'EmuDuration::param' is only used when passing this type as a parameter to a function.

Background info: 'EmuDuration::param' is the same as EmuDuration' on 64-bit machines. But on 32-bit machines it's 'EmuDuration&'. So this is a relic from the time that 32-bit machines were still the norm, and passing a (32-bit) pointer was cheaper than passing a 64-bit integer (in some locations in our code this difference really mattered for performance). Maybe now we should just remove this?



diff --git a/src/serial/RS232Net.hh b/src/serial/RS232Net.hh
new file mode 100644
index 000000000..f6304a05f
--- /dev/null
+++ b/src/serial/RS232Net.hh
@@ -0,0 +1,158 @@
+#ifndef RS232SOCKET_HH
+#define RS232SOCKET_HH

The name of the include-guard is not in sync with the filename. I assume that's a left over from an older version where 'RS232Net.hh' used to be called 'RS232Socket.hh'?

+#include "RS232Device.hh"
+#include "EventListener.hh"
+// #include "FilenameSetting.hh"

In the final version of the PR there shouldn't be any commented out code blocks or commented out #ifdefs
I'm using this line as an example, but there are several more lines like this in the PR.

+#include <cstdio>
+
+namespace openmsx {
+
+#ifndef INVALID_SOCKET
+# define INVALID_SOCKET -1
+#endif

We don't use #define to define constants, Instead write this like:
	static constexpr int INVALID_SOCKET = -1;
The main advantage is that this constant now has a clear type: "int" in this case.

This constant isn't used in RS232Net.hh, only in R2232Net.cc. So better move this constant to that file.

Though, when keeping portability between unix/windows in mind, shouldn't the type be 'SOCKET'? As defined in "src/events/Socket.hh". And I see that that Socket.hh defines a constant 'OPENMSX_INVALID_SOCKET', again to abstract the difference between unix/windows. So maybe just use that one?


+private:
+	void run();
+
+	typedef struct timeval TIMEVAL;

We prefer to use 'using' instead of 'typedef'. So that would become:
   using TIMEVAL = struct timeval;
But 'TIMEVAL' isn't used in this .hh file, only in the .cc file. So better move it there.
And then it's only used once in RS232Net.cc, so I would propose to just use 'struct timeval' instead.


+	bool DTR;	//Data Terminal Ready output status
+	bool RTS;	//Request To Send output status
+	bool DCD;	//Data Carrier Detect input status
+	bool RI;	//Ring Indicator input status
+	bool IP232;

Very much a detail. We prefer to put a space after the "//".


+	union socket_addresses_u {
+		struct sockaddr generic;     /*!< the generic type needed for calling the socket API */

Again a detail:  use "//" instead of "/* ... */" for single line comments.
I only mention this once, but it's applicable to many more lines in your PR.


+	// #ifdef HAVE_UNIX_DOMAIN_SOCKETS
+	#ifndef _WIN32
+		struct sockaddr_un local;    /*!< an Unix Domain Socket (file system) socket address */
+	#endif
+
+		struct sockaddr_in ipv4;     /*!< an IPv4 socket address */
+
+	// #ifdef HAVE_IPV6
+		struct sockaddr_in6 ipv6;    /*!< an IPv6 socket address */
+	// #endif
+	};

Here are examples of commented out #ifdef's. In the final version we should resolve this.
* Either we need them and we should make sure that macro's like 'HAVE_IPV6' exists.
* Or we assume that IPV6 is always available, and then we should remove the commented out #ifdef.


+	#else
+	   int len;
+	#endif

I'm using this line as an example. But it's applicable to many more lines in the PR.
The openMSX code is formatted with the assumption that tabs expand to 8 spaces.
I think your editor is configured to show tabs as only 4 spaces, right?
Some of your changes use a mix of space and tabs. I assume that looks fine for you, but for me much of your code is visually wrongly indented.
No big problem, I can also 'fix' this afterwards, but it does make reviewing the code a bit more difficult.


+	typedef struct rs232_network_socket_address_s rs232_network_socket_address_t;

I assume this code started originally as C code (not C++)?
In C++ typedefs like these aren't needed. Just name this struct "rs232_network_socket_address", so without "_s" or "_t" suffix. And don't have any typedef.



--- /dev/null
+++ b/src/serial/RS232Net.cc
+void RS232Net::plugHelper(Connector& connector_, EmuTime::param /*time*/)
+{
...
+    // Open socket client
+    network_client();
+    if (sockfd < 0) {
+        throw PlugException("Can't open connection");
+    }

Maybe this should be
	if (sockfd == OPENMSX_INVALID_SOCKET) {
To be portable between unix/windows and not make any assumption about the value of 'INVALID_SOCKET'.
And similar on several other places (e.g. in unplugHelper() and run(), and probably more).

+void RS232Net::run()
+{
+    ptrdiff_t ret = -1;
+    char b;

Prefer to declare variables as close to their initialization as possible.
Both 'ret' and 'b'. Don't need to keep their value over different loop-iterations. In other words: they can also be local variables _inside_ the loop ...


+    bool ipmagic = false;
+    while (1){
+		if ((poller.aborted()) || (sockfd < 0)) {
+			break;
+		}
+        // if(!RTS) continue;
+        ret = net_getc(&b);

... then this line becomes:
	char b;
	auto ret = net_getc(&b);

Detail: I'm using 'auto' instead of 'ptrdiff_t' because I don't really care about the type. I mean I care that it is the same type as what the net_getc() function returns. But I don't want to explicitly spell it out (ans risk getting it wrong). But I know some people don't like using 'auto'.

Later when we come to the net_getc() function itself, I actually want to propose changing this return type.


+            } else {
+                ipmagic = false;
+                if (b != IP232MAGIC) {
+                    DCD = (b & IP232DCDMASK) == IP232DCDHI ? true : false;
+                    // RI implemented in TCPSer 1.1.5 (not yet released) 
+                    // RI present at least on Sony HBI-232 and HB-G900AP (bit 1 of &H82/&HBFFA status register)
+                    //    missing on SVI-738
+                    RI = (b & IP232RIMASK) == IP232RIHI ? true : false;

Detail: this can be simplified to:
	DCD = (b & IP232DCDMASK) == IP232DCDHI;
and
	RI = (b & IP232RIMASK) == IP232RIHI;

+                    continue;
+                }
+            }
+        }
+
+        if (ret != 1) continue;
Can this test be move before the "if ((IP232) && (ret > 0)) { ..."?
Then the code simplifies to
	if (ret != 1) continue;
	if (IP232) { ...


+        // conn->recvByte(t, time);

Leftover. Should be removed in the final version.


+// input
+void RS232Net::signal(EmuTime::param time)
+{
+	auto* conn = checked_cast<RS232Connector*>(getConnector());
+    char b;

Don't forward declare variables ...

+
+	if (!conn->acceptsData()) {
+		std::lock_guard<std::mutex> lock(mutex);
+		queue.clear();
+		return;
+	}
+	if ((!conn->ready())||(!RTS)) return;
+	std::lock_guard<std::mutex> lock(mutex);
+	if (queue.empty()) return;
+    b = queue.pop_front();

... instead write this line as:
	char b = queue.pop_front();

+	conn->recvByte(b, time);
+}



+// Socket routines below based on VICE emulator socket.c

Could it be that these routines were originally C instead of C++, and maybe even an old (ancient) version of C (prior to C99?).

I have the following comments on all these functions:
* Don't forward declare variable, instead only declare the variable at the place where it's initialized.
* Some functions use a "do { ... } while (0);" construction. I assume that is to write the function in the single-exit-style. There's no reason for that anymore with modern compilers.
* Many functions use an 'int' return type to indicate success(=0)/failure(!=0). In c++ it's more idiomatic to return bool (true=success, false=failure   note: this does change the if-condition that tests success/failure).

Below I'll give a few examples for how some functions could be improved/simplified.
Disclaimer: I have _not_ actually tested my versions.


+int RS232Net::net_putc(uint8_t b)
+{
+    ptrdiff_t n;
+    if (sockfd < 0)
+    {
+        return -1;
+    }
+    n = sock_send(sockfd, reinterpret_cast<char *>(&b), 1);
+    if (n < 0) {
+        sock_close(sockfd);
+        sockfd = INVALID_SOCKET;
+        return -1;
+    }
+    return 0;
+}

	bool RS232Net::net_putc(uint8_t b)
	{
		if (sockfd == OPENMSX_INVALID_SOCKET) return false;
		if (auto n = sock_send(sockfd, reinterpret_cast<char*>(&b), 1); n < 0) {
			sock_close(sockfd);
			sockfd = INVALID_SOCKET;
			return false;
		}
		return true;
	}

+ptrdiff_t RS232Net::net_getc(char * b)
+{
+    int ret;
+    ptrdiff_t no_of_read_byte = -1;
+    if (sockfd < 0) {
+        return no_of_read_byte;
+    }
+    do {
+        /* from now on, assume everything is ok,
+            but we have not received any bytes */
+        no_of_read_byte = 0;
+        ret = selectPoll(sockfd);
+        if (ret > 0) {
+            no_of_read_byte = sock_recv(sockfd, b, 1);
+            if ( no_of_read_byte != 1 ) {
+                sock_close(sockfd);
+                sockfd = INVALID_SOCKET;
+                // if ( no_of_read_byte < 0 ) {
+                    // fprintf(stderr,"Error reading socket\n");
+                // } else {
+                    // fprintf(stderr,"EOF\n");
+                // }
+            }
+        }
+    } while (0);
+    return no_of_read_byte;
+}

This function actually wants to return two things:
* success or failure
* the byte that was received
The current implementation returns the 1st via the return value and the 2nd via an output-parameter.
Output-parameters are discouraged in modern c++. Instead we can return a "std::optional<char>"

I also noticed that:
* net_putc() works with 'uint8_t'
* net_getc() works with 'char'
Either is fine for me, but it should be consistent, both the same type.

	std::optional<char> RS232Net::net_getc()
	{
		if (sockfd == OPENMSX_INVALID_SOCKET) return {};
		if (auto ret = selectPoll(sockfd); ret <= 0) return {};
		char b;
		if (auto no_of_read_byte = sock_recv(sockfd, &b, 1);
		    no_of_read_byte != 1) {
			sock_close(sockfd);
			sockfd = INVALID_SOCKET;
		}
		return b;
	}

Using this function (which returns std::optional<>) can then be done like this:
	auto b = net_getc();
	if (!b) continue;      <-- we got an empty optional, this indicates an error
	...
	use(*b)                <-- using a non-empty optional is done via 'operator*'



+// Check if the socket has incoming data to receive
+// Returns: 1 if the specified socket has data; 0 if it does not contain
+// any data, and -1 in case of an error.
+int RS232Net::selectPoll(SOCKET readsock)
+{
+	TIMEVAL timeout = { 0, 0 };
+	fd_set sdsocket;
+	FD_ZERO(&sdsocket);
+	FD_SET(readsock, &sdsocket);
+	return select( readsock + 1, &sdsocket, NULL, NULL, &timeout);
+}

Detail: use "nullptr" instead of "NULL".



+// Open a socket and initialise it for client operation
+//
+//      The socket_address variable determines the type of
+//      socket to be used (IPv4, IPv6, Unix Domain Socket, ...)
+
+void RS232Net::network_client()
+{
+    sockfd = INVALID_SOCKET;
+    int error = 1;
+    do {
+        sockfd = socket(socket_address.domain, SOCK_STREAM, socket_address.protocol);
+        if (sockfd == INVALID_SOCKET) {
+            break;
+        }
+// #if defined(TCP_NODELAY)
+        setsockopt(sockfd, SOL_TCP, TCP_NODELAY, static_cast<const void*>(&error), sizeof(error)); /* just an integer with 1, not really an error */
+// #endif
+        if (connect(sockfd, &socket_address.address.generic, socket_address.len) < 0) {
+            break;
+        }
+        error = 0;
+    } while (0);
+    if (error) {
+        if (sockfd != INVALID_SOCKET) {
+            sock_close(sockfd);
+        }
+        sockfd = INVALID_SOCKET;
+    }
+}

I think this simpler version is 100% equivalent, no?

	void RS232Net::network_client()
	{
		sockfd = socket(socket_address.domain, SOCK_STREAM, socket_address.protocol);
		if (sockfd == INVALID_SOCKET) return;

		int one = 1;
		setsockopt(sockfd, SOL_TCP, TCP_NODELAY, static_cast<const void*>(&one), sizeof(one));

		if (connect(sockfd, &socket_address.address.generic, socket_address.len) < 0) {
			sock_close(sockfd);
			sockfd = INVALID_SOCKET;
		}
	}



+// Generate a socket address
+int RS232Net::network_address_generate()
+{
+	auto address_string = rs232NetAddressSetting.getString();
+    int error = 1;
+    do {
+        if (address_string.data() && strncmp("ip6://", address_string.data(), sizeof("ip6://" - 1)) == 0) {
+            if (network_address_generate_ipv6(&address_string.data()[sizeof("ip6://" - 1)])) {
+                break;
+            }
+        } else if (address_string.data() && strncmp("ip4://", address_string.data(), sizeof("ip4://" - 1)) == 0) {
+            if (network_address_generate_ipv4(&address_string.data()[sizeof("ip4://" - 1)])) {
+                break;
+            }
+        } else {
+            /* the user did not specify the type of the address, try to guess it by trying IPv6, then IPv4 */
+// #ifdef HAVE_IPV6
+            if (network_address_generate_ipv6(address_string.data()))
+// #endif /* #ifdef HAVE_IPV6 */
+            {
+                if (network_address_generate_ipv4(address_string.data())) {
+                    break;
+                }
+            }
+        }
+        error = 0;
+    } while (0);
+    return error;
+}

I changed the return type from 0=success, 1=error to true=success, false=error. (Both for this and the network_address_generate_ipv{4,6}() functions).
I also use c++ (instead of c) string manipulations.
I use strlen("..") instead of "sizeof("..") - 1". It may seem that the latter is more efficient, but in practice compilers optimize the former also to a constant.
And then the code simplifies to:

	bool RS232Net::network_address_generate()
	{
		auto address = rs232NetAddressSetting.getString();

		if (address.starts_with("ip6://")) {
			return network_address_generate_ipv6(address.substr(strlen("ip6://")));
		}
		if (address.starts_with("ip4://")) {
			return network_address_generate_ipv4(address.substr(strlen("ip4://")));
		}
		// the user did not specify the type of the address, try to guess it by trying IPv6, then IPv4
		return network_address_generate_ipv6(address) ||
		       network_address_generate_ipv4(address);
	}

+int RS232Net::network_address_generate_ipv6(const char * address_string)
+{
	...
+}

I'll first discuss the _ipv6 version:
I change the return type as discussed above.
I changed the "address_string" type from "const char*" to "zstring_view", and then renamed it to just "address" (because the "_string" suffix is redundant).
If I didn't make any mistakes, the function then simplifies to the following.

	bool RS232Net::network_address_generate_ipv6(zstring_view address)
	{
		// preset the socket address
		memset(&socket_address.address, 0, sizeof socket_address.address);
		socket_address.domain = PF_INET6;
		socket_address.protocol = IPPROTO_TCP;
		socket_address.len = sizeof socket_address.address.ipv6;
		socket_address.address.ipv6.sin6_family = AF_INET6;
		socket_address.address.ipv6.sin6_port = 0;
		socket_address.address.ipv6.sin6_addr = in6addr_any;

		if (address.empty()) {
			// there was no address given, do not try to process it.
			return true;
		}

		struct hostent* host_entry = gethostbyname2(address.c_str(), AF_INET6);
		if (!host_entry) {
			return false;
		}

		memcpy(&socket_address.address.ipv6.sin6_addr, host_entry->h_addr, host_entry->h_length);
		return true;
	}

Note 1: apparently it's fine for this function to pass an empty string. But is that also fine in the context of the RS232Net class?

Note 2: I am a bit concerned about the "gethostbyname2()" function. I didn't know this function, so I looked up the documentation, and my documentation says this function is deprecated??



+int RS232Net::network_address_generate_ipv4(const char *address_string)
+{
	...
+}

The _ipv4 versions is a lot more complicated than the _ipv6 version. The main reason for that is the the v4 version also handles port numbers (and then it has to do a lot more string manipulations). Why this asymmetry? Should the v6 version also handle port numbers? Or should this be removed from the v4 version?
Anyway, I think the v4 version can also be simplified a lot. But before doing that I first wanted to see how the port number situation should be resolved.

retrocomputacion:
@m9710797  Sorry I didnt notice your reply until now.
I'll read it and do any other improvement, style changes needed

retrocomputacion:
I think I've implemented all your suggested changes.
 `network_address_generate` could be merged with `network_address_generate_sockaddr` but I chose to keep them separate just in case.

I dont quite remember how I found out the 8254 timer was not right. That was one of the first things I modified, even before adding the rs232net pluggable (I was using the rs232tester mapped to a couple of PTY virtual devices IIRC.
But I think the receiving speed was obviously wrong (half of what it should).

src/events/Socket.hh
```diff
@@ -25,6 +26,7 @@ using SOCKET = int;
 // INVALID_SOCKET is #defined as  (SOCKET)(~0)
 // but that gives a old-style-cast warning
 static const SOCKET OPENMSX_INVALID_SOCKET = static_cast<SOCKET>(~0);
+typedef UINT32 in_addr_t;
```
m9710797: Detail:
In modern c++ you should replace "typedef" with "using". Then this line becomes:

    using in_addr_t = UINT32;

Reason:
* Everything that can be done with "typedef" can also be done with "using".
* And "using" can also do some things that are not possible with "typedef" (stuff related to templates).
* The order in "using" is more logical compared to the order in "typedef".

However, I don't actually see `in_addr_t` being used anywhere in the code. And then maybe this line isn't needed at all?

src/serial/I8254.cc
```diff
@@ -247,16 +247,27 @@ void Counter::writeLoad(uint16_t value, EmuTime::param time)
...
 		counter = counterLoad;
 	}
-	if (!active && (mode == one_of(CNTR_M2, CNTR_M2_, CNTR_M3, CNTR_M3_))) {
+	if (!active && (mode == one_of(CNTR_M3, CNTR_M3_))) {
 		if (clock.isPeriodic()) {
 			counter = counterLoad;
-			EmuDuration::param high = clock.getTotalDuration();
+			EmuDuration high = clock.getTotalDuration();
+			EmuDuration total = high * counter;
+			output.setPeriodicState(total, total / 2, time);
```
m9710797: This is wrong when `counter` is odd. Then the duration high/low isn't exactly half the total duration.
I think it should be the following (but please verify):

    int half = (counter + 1) / 2; // round up
    EmuDuration tick = clock.getTotalDuration();
    EmuDuration total = tick * counter;
    EmuDuration high  = tick * half;
    output.setPeriodicState(total, high, time);

src/serial/MSXRS232.hh
```diff
@@ -4,6 +4,7 @@
 #include "MSXDevice.hh"
 #include "IRQHelper.hh"
 #include "RS232Connector.hh"
+#include "RS232Device.hh"
```
m9710797: This #include isn't needed.

src/serial/RS232Net.hh
```diff
@@ -0,0 +1,149 @@
+#ifndef RS232NET_HH
+#define RS232NET_HH
+
+#include "RS232Device.hh"
+#include "EventListener.hh"
+// #include "FilenameSetting.hh"
```
m9710797: Detail (for the final version):

Don't add any commented out code. Same on line 99 and 125.

src/serial/RS232Net.hh
```diff
@@ -0,0 +1,149 @@
...
+	virtual void setDTR(bool status, EmuTime::param time);
+	virtual void setRTS(bool status, EmuTime::param time);
+
+	template<typename Archive>
+	void serialize(Archive& ar, unsigned version);
+
+private:
+	void run();
+
+	typedef struct timeval TIMEVAL;
```
m9710797: Detail:
Similar as an earlier review comment: in modern c++ "typedef" should be replaced with "using".
But in this specific case, there's only exactly one user of `TIMEVAL`. Then it's better to just use `struct timeval` in RS232Net.cc and remove this line.

src/serial/RS232Net.hh
```diff
@@ -0,0 +1,149 @@
...
+	//
+	// However, in practice, this approach works.
+	//
+	union socket_addresses_u {
+		struct sockaddr generic;     // the generic type needed for calling the socket API
+
+	// #ifdef HAVE_UNIX_DOMAIN_SOCKETS
+	#ifndef _WIN32
+		struct sockaddr_un local;    // an Unix Domain Socket (file system) socket address
+	#endif
```
m9710797: The `local` member of this union isn't used anywhere in the RS232Net code. (I assume you copied this from other code that did use it).
To simplify the code I propose to just remove these 3 lines. We can always add it back in the future when we do need it (if ever).

src/serial/RS232Net.hh
```diff
@@ -0,0 +1,149 @@
...
+					* This is used for debugging the buffer
+					* allocation strategy.
+					*/
+		int domain;             // the address family (AF_INET, ...) of this address
+		int protocol;           // the protocol of this address. This can be used to distinguish between different types of an address family.
+	#ifndef _WIN32
+		socklen_t len;          // the length of the socket address
+	#else
+		int len;
+	#endif
```
m9710797: You could add a line like

    using socklen_t = int;

In src/events/Socket.hh inside the windows-specific #ifdef block. Then these 5 lines simplify to just one line.

src/serial/RS232Net.cc
```diff
@@ -0,0 +1,446 @@
...
+	while (1){
+
+		if ((poller.aborted()) || (sockfd == OPENMSX_INVALID_SOCKET)) {
+			break;
+		}
+
+		// char b;
+		auto b = net_getc();
+
+		if (!b.has_value()) continue;
```
m9710797: Detail:
Using `b.has_value()` is perfectly fine. And some people prefer it.
In openMSX we instead prefer to write this like:

    if (!b) continue;

Similarly, a bit later in the code you use `b.value()`. Again that's perfectly fine. But in openMSX we instead prefer:

    *b

src/serial/RS232Net.cc
```diff
@@ -0,0 +1,446 @@
...
+        		if (!ipmagic) {
+				if (b.value() == IP232MAGIC) {
+					ipmagic = true;
+                    			/* literal 0xff */
+                    			continue;
+                		}
+            		} else {
+                		ipmagic = false;
+				if (b.value() != IP232MAGIC) {
+        				DCD = (b.value() & IP232DCDMASK) == IP232DCDHI ? true : false;
```
m9710797: This construction is redundant:

    bool myBool = <some-condition> ? true : false

because `some-condition` already produces true/false.
And then this line can be simplified to:

    DCD = (*b & IP232DCDMASK) == IP232DCDHI;

And similar on line 139.

src/serial/RS232Net.cc
```diff
@@ -0,0 +1,446 @@
...
+}
+
+std::optional<char> RS232Net::net_getc()
+{
+	// int ret;
+	// ptrdiff_t no_of_read_byte = -1;
+
+	if (sockfd == OPENMSX_INVALID_SOCKET)  return {};
+
+	char b;
```
m9710797: Uninitialized variables should be avoided. But unfortunately in cases like this, when calling a C-function like `sock_recv()` it's required.
But still, the variable `b` is only used inside the `if (selectPoll(...)) {` branch. Then move that variable inside.
In other words: swap lines 253 and 255.

src/serial/RS232Net.cc
```diff
@@ -0,0 +1,446 @@
...
+
+	if ( getaddrinfo(address_part,port_part,&hints,&res) == 0) {
+		memcpy(&socket_address.address, res->ai_addr, res->ai_addrlen);
+		result = true;
+	}
+	freeaddrinfo(res);
+
+	free(address_part);
+	return result;
+}
```
m9710797, line 413: String manipulation isn't easily, especially in C.
Using C++ combined with some extra openMSX utility functions allows to simplify this.
I think the following should be equivalent. This replaces `network_address_generate()` and `network_address_generate_sockaddr()`.
Disclaimer: I have NOT tested this.

    bool RS232Net::network_address_generate()
    {
            std::string_view address = rs232NetAddressSetting.getString();
            if (address.empty()) {
                    // there was no address given, do not try to process it.
                    return false;
            }
    
            // preset the socket address
            memset(&socket_address.address, 0, sizeof(socket_address.address));
            socket_address.protocol = IPPROTO_TCP;
    
            struct addrinfo hints;
            memset(&hints, 0, sizeof(hints));
            hints.ai_socktype = SOCK_STREAM;
            hints.ai_protocol = IPPROTO_TCP;
    
            auto setIPv4 = [&] {
                    hints.ai_family = AF_INET;
                    socket_address.domain = PF_INET;
                    socket_address.len = sizeof(socket_address.address.ipv4);
                    socket_address.address.ipv4.sin_family = AF_INET;
                    socket_address.address.ipv4.sin_port = 0;
                    socket_address.address.ipv4.sin_addr.s_addr = INADDR_ANY;
            };
            auto setIPv6 = [&] {
                    hints.ai_family = AF_INET6;
                    socket_address.domain = PF_INET6;
                    socket_address.len = sizeof(socket_address.address.ipv6);
                    socket_address.address.ipv6.sin6_family = AF_INET6;
                    socket_address.address.ipv6.sin6_port = 0;
                    socket_address.address.ipv6.sin6_addr = in6addr_any;
            };
    
            // Parse 'address', fills in 'addressPart' and 'portPart'
            std::string addressPart;
            std::string portPart;
            auto [ipv6_address_part, ipv6_port_part] = StringOp::splitOnFirst(address, ']');
            if (!ipv6_port_part.empty()) { // there was a ']' character (and it's already dropped)
                    // Try to parse as: "[IPv6]:port"
                    if (!ipv6_address_part.starts_with('[') || !ipv6_port_part.starts_with(':')) {
                            // malformed address, do not try to process it.
                            return false;
                    }
                    addressPart = std::string(ipv6_address_part.substr(1)); // drop '['
                    portPart    = std::string(ipv6_port_part   .substr(1)); // drop ':'
                    setIPv6();
            } else {
                    auto numColons = ranges::count(address, ':');
                    if (numColons == 0) {
                            // either IPv4 or IPv6
                            addressPart = std::string(address);
                    } else if (numColons == 1) {
                            // ipv4:port
                            auto [ipv4_address_part, ipv4_port_part] = StringOp::splitOnFirst(address, ':');
                            addressPart = std::string(ipv4_address_part);
                            portPart    = std::string(ipv4_port_part);
                            setIPv4();
                    } else {
                            // ipv6 address
                            addressPart = std::string(address);
                            setIPv6();
                    }
            }
    
            // Interpret 'addressPart' and 'portPart' (possibly the latter is empty)
            struct addrinfo* res;
            if (getaddrinfo(addressPart.c_str(), portPart.c_str(), &hints, &res) != 0) {
                    return false;
            }
    
            memcpy(&socket_address.address, res->ai_addr, res->ai_addrlen);
            freeaddrinfo(res);
            return true;
    }

I hope I didn't make any mistakes while rewriting this code.

Note 1: this requires these #includes at the top of the .cc file:

    #include "ranges.hh"
    #include "StringOp.hh"

Note 2: I believe calling `freeaddrinfo()` is only needed when getaddrinfo() returns success. At least that would be logical, but I'm not 100% sure about it (and the documentation isn't clear).

m9710797 (commented):
Hi,

Thanks again for your work on this PR and for your updates so far.

I did a 2nd pass of review. Hopefully this time it's more readable (I messed up the layout last time).

I repeated some of my remarks from the 1st review. Note that many of my remarks start with "detail". Feel free to fix those. But I'm also happy to fix those myself after the PR has been merged into the main openMSX branch.

Now I also tried to better understand whet the code is doing. Is my following  understanding correct?
* The code is largely based on `RS232Tester`, but instead of read/writing from/to a file, you send/receive data to/from a socket.
* The code is largely controller via two settings:
    * "rs232-net-address": and paring this setting is actually the most complicated part of this PR :-)
    * "rs232-net-ip232": this makes the code operate in two modes:
        * false: msx bytes are directly connected to the socket, no extra translation
        * true: bytes other than 0xff are send/received without translation. A 0xff byte is translated to two 0xff bytes (two 0xff are send, on receive two 0xff are truncated to a single 0xff again). And 0xff followed by a non 0xff byte has special meaning, in particular, it used to communicate changes in the 'DCD' and 'RI' status.

After you've addressed these review comments (with or without the ones marked as "detail"), I think this PR will be ready to be merged.

Ah, and of course please retest the code after the changes.

Wouter

src/serial/RS232Net.cc
retrocomputacion: I did it this way because I was getting typecasting errors yesterday. Today it compiled correctly, I dont know what changed

src/serial/RS232Net.hh
retrocomputacion: Yes, this is not used in the original code either, the code that uses it is disabled by an #if 0

src/serial/RS232Net.cc
retrocomputacion: Yeah I forgot to change it after I had problems with the previous point

src/serial/RS232Net.cc
retrocomputacion: Yes, this works!

m9710797:
Thanks. Later today (hopefully) I'll do a final review and then I'll probably merge your PR.

After the merge I'll still do some minor cleanups. (But explaining what to change (and why) is more work than quickly doing it myself.)

And one extra thing we still need to do is call `sock_startup()` and `sock_cleanup()` (needed on windows). But do that in a way that we don't call these twice when also the `CliServer` code is active. I'll handle this: I have some idea's for this but I still need to test them out.

I would like to ask you to still run some tests after I've pulled your PR and made the tweaks. I'll report my progress in this ticket.

m9710797:
I've merged this PR. And I did some small tweaks.
I plan to make more tweaks later (don't have time anymore today).

@retrocomputacion Could you test my tweaks? (Did I break anything?)

retrocomputacion:
> I've merged this PR. And I did some small tweaks. I plan to make more tweaks later (don't have time anymore today).
> 
> @retrocomputacion Could you test my tweaks? (Did I break anything?)

Everything seems to be working correctly.
Thanks for your patience!

m9710797:
Hi. I plan to make more tweaks. Maybe today, maybe later. Please continue monitoring this ticket and help testing.
For now, I'm still studying the code in more detail. And I have a question:

At the moment the `RI` (ring indicator) variable is write-only. It tracks whether a `IP232_RI_HI` magic marker has been received, but nothing further depends on it. So basically it has no effect.
Do you still plan to improve the code so that it is used? Maybe the `RS232Device` needs to be extended with a `getRI()` method? And then maybe some MSX device code needs to use that extra method?

retrocomputacion:
> Hi. I plan to make more tweaks. Maybe today, maybe later. Please continue monitoring this ticket and help testing. For now, I'm still studying the code in more detail. And I have a question:
> 
> At the moment the `RI` (ring indicator) variable is write-only. It tracks whether a `IP232_RI_HI` magic marker has been received, but nothing further depends on it. So basically it has no effect. Do you still plan to improve the code so that it is used? Maybe the `RS232Device` needs to be extended with a `getRI()` method? And then maybe some MSX device code needs to use that extra method?

Yes, for completeness a `getRI()` is needed, except for the SVI738, RI is present on bit 1 of port 82H.
At least version 1 of the RS232 ROM doesn't test that bit at all. I don't know about version 2.

In practical terms, RI would be only really needed for a BBS server program, and I'm not aware of any existing for MSX.

m9710797:
Ok, do you plan to implement this? I assume it should be added in src/serial/MSXRS232.cc:242 in the function readStatus().
And we may want to only conditionally expose this bit (because e.g. on SVI738 it's apparently missing).

I pushed 3 more changes for the RS232Net code. If you're interested please review my changes.
Please also help me test this. I tried to be careful, but it's always possible I accidentally broke something.

After this I still have one more thing I'd like to investigate further (it's related to the 'Poller' stuff). But I won't have time to finish that today.

retrocomputacion:
> Ok, do you plan to implement this? I assume it should be added in src/serial/MSXRS232.cc:242 in the function readStatus(). And we may want to only conditionally expose this bit (because e.g. on SVI738 it's apparently missing).

Yes, in fact I just implemented it and is working, I'm just missing the part that hides it for the SVI738. How should I go about it?

> I pushed 3 more changes for the RS232Net code. If you're interested please review my changes. Please also help me test this. I tried to be careful, but it's always possible I accidentally broke something.

Everything looks good here, tested it on windows too.

> After this I still have one more thing I'd like to investigate further (it's related to the 'Poller' stuff). But I won't have time to finish that today.

Is it possible that this is related with the problem where this pluggable no longer receives data after unplugging and plugging it back again? (but it still sends out data normally)

retrocomputacion:
> > After this I still have one more thing I'd like to investigate further (it's related to the 'Poller' stuff). But I won't have time to finish that today.
> 
> Is it possible that this is related with the problem where this pluggable no longer receives data after unplugging and plugging it back again? (but it still sends out data normally)

I think I fixed this, at least is working as I expected it.

I changed line 211 of RS232Net.cc to

`		if (sockfd == OPENMSX_INVALID_SOCKET) {`

m9710797:
So remove the `poller.aborted()` part, right?

I've investigated (only a little bit yet), it seems like the current `Poller` implementation is "single use": after being aborted it remains in that state. (This is sufficient for the `CliConnection` stuff for which we originally introduced this class).

I think the same bug is present in some other places of our code (e.g. `RS232Tester` and `MidiInReader`). Apparently we never tested to to unplug and replug. And in 20 years nobody reported this bug before. So thanks a lot for reporting it now.

I'll also try to fix this.

m9710797:
I pushed two more commits for RS232Net.
(If you have time) please read the commit message, review the code, and especially help test my changes.

From my side this finishes the RS232Net code.

Thanks.

m9710797:
> > Ok, do you plan to implement this? I assume it should be added in src/serial/MSXRS232.cc:242 in the function readStatus(). And we may want to only conditionally expose this bit (because e.g. on SVI738 it's apparently missing).
> 
> Yes, in fact I just implemented it and is working, I'm just missing the part that hides it for the SVI738. How should I go about it?

I'm sorry. I read this question. But then got interrupted, and I forgot to answer.

To make devices "configurable", and I mean configurable by editing the XML description (e.g. in `share/extensions/rs232.xml` or `share/machines/Sony_HB-G900P.xml`), you can query properties from the `DeviceConfig` object that gets passed to the `MSXRS232` constructor.

See for example src/serial/MSXRS232.cc line 35:

    hasMemoryBasedIo(config.getChildDataAsBool("memorybasedio", false))

This initializes the (const bool) member variable `hasMemoryBasedIo`. The `getChildDataAsBool` method searches for the "memorybasedio" (the 1st parameter) tag in the XML file, it takes the data that's inside that XML tag and parses it to a boolean. When that tag is missing we instead use the default value "false" (the 2nd parameter).

If you look at `share/machines/Sony_HB-G900P.xml` on line 91, you see

     <memorybasedio>true</memorybasedio>

Other configuration files that use the "MSX-RS232" device may specify the value "false", or may not have this tag at all.

So to make the ring-indicator status bit configurable, you could do something similar.

retrocomputacion:
> I pushed two more commits for RS232Net. (If you have time) please read the commit message, review the code, and especially help test my changes.
> 
> From my side this finishes the RS232Net code.
> 
> Thanks.

All is working correctly!
CPU usage is also much lower, as you noted in the commit message.

I have updated my fork, I'll recheck the default logic level of the RI line and will send a new PR for that.

PR #1586, opened 2024-03-01
Add RS232 Ring Indicator handling, fix levels for unused bits in port 0x82 and active level for I8254 counter output pins.

retrocomputacion:
This adds Ring Indicator (RI) line handling to the RS232net pluggable (with IP232 protocol enabled).
Current release of TCPSer (1.1.4) doesn't support RI emulation, but TCPSer v1.1.6_beta from the official repository (https://github.com/go4retro/tcpser) does.

Also added 'ripinpresent' configuration tag, which defaults to true.
Set to false for machines with no RI pin present in the RS232 port (SVI-738 and its variants).

I don't know of any software that reads the status of the RI line, but a simple BASIC program can be made to test it:

```
10 CALLCOMINI("0:8n1nnnnn",1200,1200,0)
20 OPEN"com0:"AS#1
30 P=0:R=3
40 A=INP(&H82)AND2:IFA<>PTHENP=A:IFA=0THENPRINT"RING!":R=R-1
50 IF R>0THEN40
60 PRINT#1,"ATA";CHR$(13):CLS:CLOSE1
70 CALLCOMTERM
```

Running that program in OpenMSX (with the R232 pluggable correctly setup and TCPser running), and connecting to TCPSer's listening port from another terminal program (ie PuTTY) should result in three 'rings' and then the program answering the 'call' and calling the ROM terminal.

### DIFF
--- a/src/serial/I8251.hh
+++ b/src/serial/I8251.hh
@@ -20,6 +20,7 @@ public:
 	virtual void setRTS(bool status, EmuTime::param time) = 0;
 
 	[[nodiscard]] virtual bool getDCD(EmuTime::param time) = 0; // This is not an I8251 signal
+	[[nodiscard]] virtual bool getRI(EmuTime::param time) = 0; // This is not an I8251 signal
 	[[nodiscard]] virtual bool getDSR(EmuTime::param time) = 0;
 	[[nodiscard]] virtual bool getCTS(EmuTime::param time) = 0; // TODO use this
 	virtual void signal(EmuTime::param time) = 0;
--- a/src/serial/I8251.hh
+++ b/src/serial/I8251.hh
@@ -19,8 +19,6 @@ public:
 	virtual void setDTR(bool status, EmuTime::param time) = 0;
 	virtual void setRTS(bool status, EmuTime::param time) = 0;
 
-	[[nodiscard]] virtual bool getDCD(EmuTime::param time) = 0; // This is not an I8251 signal
-	[[nodiscard]] virtual bool getRI(EmuTime::param time) = 0; // This is not an I8251 signal
 	[[nodiscard]] virtual bool getDSR(EmuTime::param time) = 0;
 	[[nodiscard]] virtual bool getCTS(EmuTime::param time) = 0; // TODO use this
 	virtual void signal(EmuTime::param time) = 0;
--- a/src/serial/MSXRS232.cc
+++ b/src/serial/MSXRS232.cc
@@ -33,7 +33,7 @@ MSXRS232::MSXRS232(const DeviceConfig& config)
 		: nullptr)
 	, rxrdyIRQ(getMotherBoard(), MSXDevice::getName() + ".IRQrxrdy")
 	, hasMemoryBasedIo(config.getChildDataAsBool("memorybasedio", false))
-	, hasRIPin(config.getChildDataAsBool("ripinpresent",true))
+	, hasRIPin(config.getChildDataAsBool("has_ri_pin",true))
 	, ioAccessEnabled(!hasMemoryBasedIo)
 	, switchSetting(config.getChildDataAsBool("toshiba_rs232c_switch",
 		false) ? std::make_unique<BooleanSetting>(getCommandController(),
@@ -225,7 +225,8 @@ byte MSXRS232::readStatus(EmuTime::param time)
 	//
 	//  Bit Name  Expl.
 	//  0   CD    Carrier Detect   (0=Active, 1=Not active)
-	//  1   RI    Ring Indicator   (0=Active, 1=Not active) (N/C in MSX)
+	//  1   RI    Ring Indicator   (0=Active, 1=Not active)
+	//            (Not connected on many RS232 BIOS v1 implementations)
 	//  6         Timer Output from i8253 Counter 2
 	//  7   CTS   Clear to Send    (0=Active, 1=Not active)
 	//
@@ -236,26 +237,30 @@ byte MSXRS232::readStatus(EmuTime::param time)
 	//   on this I/O port, if CN1 is open. If CN1 is closed, it always
 	//   reads back as "0". ...
 
-	byte result = 0; // TODO check unused bits
+	byte result = 0xFF;	// Start with 0xFF, open lines on the data bus pull to 1
+	auto& dev = getPluggedRS232Dev();
 
-	if (!interface.getDCD(time)) {
-		result |= 0x01;
+	// Mask out (active low) bits
+	if (dev.getDCD(time)) {
+		result &= 0xFE;
 	}
 
-	if (!interface.getRI(time) && hasRIPin) {
-		result |=0x02;
+	if (hasRIPin && dev.getRI(time)) {
+		result &= 0xFD;
 	}
 
-	if (!rxrdyIRQenabled && switchSetting && switchSetting->getBoolean()) {
-		result |= 0x08;
+	if (rxrdyIRQenabled && switchSetting && switchSetting->getBoolean()) {
+		result &= 0xF7;
 	}
 
-	if (!interface.getCTS(time)) {
-		result |= 0x80;
+	if (interface.getCTS(time)) {
+		result &= 0x7F;
 	}
+
 	if (i8254.getOutputPin(2).getState(time)) {
-		result |= 0x40;
+		result &= 0xBF;
 	}
+
 	return result;
 }
 
@@ -307,18 +312,6 @@ void MSXRS232::Interface::setRTS(bool status, EmuTime::param time)
 {
 	auto& rs232 = OUTER(MSXRS232, interface);
 	rs232.getPluggedRS232Dev().setRTS(status, time);
-}
-
-bool MSXRS232::Interface::getDCD(EmuTime::param time)
-{
-	auto& rs232 = OUTER(MSXRS232, interface);
-	return rs232.getPluggedRS232Dev().getDCD(time);
-}
-
-bool MSXRS232::Interface::getRI(EmuTime::param time)
-{
-	auto& rs232 = OUTER(MSXRS232, interface);
-	return rs232.getPluggedRS232Dev().getRI(time);
 }
 
 bool MSXRS232::Interface::getDSR(EmuTime::param time)
--- a/src/serial/MSXRS232.cc
+++ b/src/serial/MSXRS232.cc
@@ -33,6 +33,7 @@ MSXRS232::MSXRS232(const DeviceConfig& config)
 		: nullptr)
 	, rxrdyIRQ(getMotherBoard(), MSXDevice::getName() + ".IRQrxrdy")
 	, hasMemoryBasedIo(config.getChildDataAsBool("memorybasedio", false))
+	, hasRIPin(config.getChildDataAsBool("ripinpresent",true))
 	, ioAccessEnabled(!hasMemoryBasedIo)
 	, switchSetting(config.getChildDataAsBool("toshiba_rs232c_switch",
 		false) ? std::make_unique<BooleanSetting>(getCommandController(),
@@ -241,6 +242,10 @@ byte MSXRS232::readStatus(EmuTime::param time)
 		result |= 0x01;
 	}
 
+	if (!interface.getRI(time) && hasRIPin) {
+		result |=0x02;
+	}
+
 	if (!rxrdyIRQenabled && switchSetting && switchSetting->getBoolean()) {
 		result |= 0x08;
 	}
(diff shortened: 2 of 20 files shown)


src/serial/MSXRS232.cc
```diff
@@ -33,6 +33,7 @@ MSXRS232::MSXRS232(const DeviceConfig& config)
 		: nullptr)
 	, rxrdyIRQ(getMotherBoard(), MSXDevice::getName() + ".IRQrxrdy")
 	, hasMemoryBasedIo(config.getChildDataAsBool("memorybasedio", false))
+	, hasRIPin(config.getChildDataAsBool("ripinpresent",true))
```
m9710797: When reviewing the changed in the XML files, it took me a moment to understand the name "ripinpresent". Maybe it's clearer to name this tag something like: "has_ri_pin".

I know that other tags in this device also do not have underscores in their name (like "memorybasedio", but that's also not very readable). Though those names were chosen +20 years ago, and for backwards compatibility it's not worth changing them now. But for new names we should try to do better.

src/serial/I8251.hh
```diff
@@ -20,6 +20,7 @@ public:
 	virtual void setRTS(bool status, EmuTime::param time) = 0;
 
 	[[nodiscard]] virtual bool getDCD(EmuTime::param time) = 0; // This is not an I8251 signal
+	[[nodiscard]] virtual bool getRI(EmuTime::param time) = 0; // This is not an I8251 signal
```
m9710797: This is basically the same change as when you added `getDCD()`. And indeed also the new `getRI()` method does not really belong here: it's not a I8251 signal. Now that I'm reviewing this type of code for the 2nd time I think I know how to resolve this (thus both for `getDCD()` and `getRI()`).

We have two interface classes that define this method (and there's no inheritance relation between these two):
* `I8251Interface`: this represents the communication between some device (e.g. `MSXRS232`) and the `I8251` chip.
* `RS232Device`: the represents the communication over a RS232 serial connector.

The DCD and RI signals are present on the RS232 connector, but they do NOT go to the I8251 chip. So the `getDCD()` and 'getRI()` method should be removed from `I8251Interface`.

After removing these, we need to make a change in MSXRS232, see below.

src/serial/MSXRS232.cc
```diff
@@ -310,6 +315,12 @@ bool MSXRS232::Interface::getDCD(EmuTime::param time)
 	return rs232.getPluggedRS232Dev().getDCD(time);
 }
 
+bool MSXRS232::Interface::getRI(EmuTime::param time)
+{
+	auto& rs232 = OUTER(MSXRS232, interface);
+	return rs232.getPluggedRS232Dev().getRI(time);
+}
```
m9710797: After removing `getRI()` (and `getDCD()`) from `I8251Interface`, this method should also be removed.

And similar in `MSXFacMidiInterface` and `MSXMidi`  (I assume a MIDI connector does not carry a RI or DCD signal).

src/serial/MSXRS232.cc
```diff
@@ -241,6 +242,10 @@ byte MSXRS232::readStatus(EmuTime::param time)
 		result |= 0x01;
 	}
 
+	if (!interface.getRI(time) && hasRIPin) {
+		result |=0x02;
+	}
+
```
m9710797, line 251: This should become:

    auto& dev = getPluggedRS232Dev();
    if (!dev.getDCD(time)) result |= 0x01;
    if (hasPiPin && !dev.getRI(time)) result |= 0x02;

So we don't get the RI and DCD status from the I8251Interface anymore (because that one does not have those signals). Instead we get them directly from the RS232Connector.

m9710797, line 251: Detail:
Note that I swapped the order of `hasPiPin` and `!dev.getRI(time)` compared to your version.
In general it's a good idea to put cheaper sub-expression (testing a boolean) before the more expensive sub-expressions (calling a virtual function).

src/serial/RS232Device.cc
```diff
@@ -37,6 +37,11 @@ bool RS232Device::getDCD(EmuTime::param /*time*/) const
 	return true; // TODO check
 }
 
+bool RS232Device::getRI(EmuTime::param /*time*/) const
+{
+	return true; // TODO check
+}
+
```
m9710797, line 56: (Not required before merging this PR.)

It would be nice if you could resolve this TODO.  And then ideally for all 4 of `getRI()`, `getDCD()`, `getDSR()` and `getCTS()`.

I think these implement the signal value that's read when there's no device plugged in the RS232 connector. So it would be nice if someone could test what a real MSX with RS232 returns when reading these signals in this scenario.

m9710797 (commented):
Hi,

3 remarks (see my comments in the code for more details), from more to less important:
* `I8251Interface` does not need `getRI()` (nor `getDCD()`).
* A suggestion to rename "ripinpresent" to "has_ri_pin".
* (not required for the merge) A request to investigate the behavior on a real MSX.

m9710797:
Ah one more thing. Thanks for the detailed description of this PR as a github comment.
It would be very useful to include the same information in your commit message.

src/serial/MSXRS232.cc
retrocomputacion: Yes I was hesitant to use that name, but wanted to keep the same format than the other settings.

src/serial/MSXRS232.cc
retrocomputacion: MIDI only has Rx and Tx

src/serial/RS232Device.cc
retrocomputacion: That can be deduced from the schematics, all three I've found (Sony HB-G900, Sony HBI-232 and SVI-738) use the same line driver for the RS-232 input lines.
The Sony devices leave the inputs floating when there's no device plugged, resulting in a logical 0 being read from the inputs, the opposite of what it's implemented right now in OpenMSX. (The bit should be set to1 when reading the port)
The SVI on the other hand pulls them to 12v which reads as a logical 1, which is consistent with the current implementation (bit is set to 0 when reading the port).

retrocomputacion: More on this, I tested on the SVI738 I have on loan.
inp(&h82) returns 0x7E
That matches my previous comment, the bits for DCD and CTS both are set to 0.
Also noticed something that I had missed before, the I8254 counter outputs are active low, but is implemented as active high right now.
Open data lines are set to 1.
On openMSX inp(&h82) with nothing plugged returns 0.

I propose invert the logic in readStatus(): start with `result = 0xFF;` and then mask out the active bits.

Oh also asked the Sysop of Sotanomsx to test this in his HX-22 which has a built-in RS232 port.

m9710797: Good idea (start from 0xff, and then clear specific bits).

It would also be great if you could add some comments about these experiments/results in the code. So that next year we still know which parts of the implementation are best guesses and which parts are verified on real MSX machines.

Thanks.

retrocomputacion: Got tests back for the HX-22:
With the RS-232 switch in the back set to ON:
0xFF

Set to OFF:
0xBF

I'll write my full findings as of today here, will include a shorter version in the code.

Looking at this msx.org thread https://www.msx.org/forum/msx-talk/hardware/toshiba-hx-22?page=3, this machine doesn't have the RI pin but the inputs are left floating like on the Sony devices.

From the PCB pictures on msx.org:
ASCII MSXSerial-232: RI present, floating input lines
Sanyo MRS-001: No RI, floating input lines
Toshiba HX-R700: No RI, floating input lines

From the service manual:
SVI-737 Modem/RS-232 interface: RI present, pull-down lines (read the same as if floating on the port)

From the user manual:
SVI-757 RS-232 interface: RI present. This device is probably the same as the 737 without the modem parts populated, there's a good chance the input lines are also pull-down.

src/serial/MSXRS232.cc
```diff
@@ -236,26 +237,30 @@ byte MSXRS232::readStatus(EmuTime::param time)
...
-	byte result = 0; // TODO check unused bits
+	byte result = 0xFF;	// Start with 0xFF, open lines on the data bus pull to 1
+	auto& dev = getPluggedRS232Dev();
 
-	if (!interface.getDCD(time)) {
-		result |= 0x01;
+	// Mask out (active low) bits
+	if (dev.getDCD(time)) {
+		result &= 0xFE;
 	}
```
m9710797, line 246: Detail:
My personal preference for this is to write it like:
    `result &= ~0x01;`
For 0xFE <-> 0x01 is still easy to see what's going on, but for 0xBF <-> 0x40 it's already more difficult (at least for me).

Not my preference, but equally valid. Other people prefer to use binary notation with digit separators:
    `result &= ~0b0100'0000;`

src/serial/RS232Device.cc
```diff
@@ -22,24 +22,36 @@ void RS232Device::setParityBit(bool /*enable*/, ParityBit /*parity*/)
...
+//                                   the same as left floating.
+//
+// This has been tested to be true for the SVI-738 and Toshiba HX-22,
+// for the other devices the info comes from manuals, schematics, or
+// circuit board pictures.
+
 bool RS232Device::getCTS(EmuTime::param /*time*/) const
 {
-	return true; // TODO check
+	return false; // TODO return the inverse of config tag 'rs232_pullup'
```
m9710797, line 39: Thanks a lot for your very detailed investigation.
In openMSX we don't often take pull-up, pull-down (or really 'floating') into account. But you're absolutely right that we should.

I have a suggestion for the implementation:
* You added a `<rs232_pullup>` tag in the MSX RS232 device. IMHO this is correct
* In the commit message you talk about passing a 'DeviceConfig` object to RS232Device. IMHO that's not correct.

My reasoning is the following:
* Suppose you have a hypothetical MSX machine with two RS232 interfaces, one with pull-up resistors, and the other with pull-down.
* You have a RS232 peripheral with some unconnected pins.
* If you plug this peripheral in one interface the corresponding status bit will read as '1', in the other interface it will read as '0'.
* So the same (unchanging) peripheral behaves different depending on in which interface it gets plugged. That's an indication that the `<rs232_pullup>` tag indeed belongs in the interface.

But still, from a practical point of view, in a method like `bool getCTS()` we must either return `true` or `false`. And which is the correct value depends on whether pull-up/down resistors are present. And I assume that's why you suggested to pass this information (via a `DeviceConfig` object) to `RS232Device`.

I would propose a different approach:
* A real pin in a RS232 connector can be in 3 states: it can be driven high, driven low, or it can be unconnected (=floating).
* But in the current implementation we model the value of a pin with a `bool`, which can only have 2 states.

Therefor to resolve this I would propose to not return a `bool` but something else which _can_ have 3 states. The first (simplest?) thing that comes to mind is a `std::optional<bool>`:
* An empty optional then represents an unconnected pin.
* An optional with a value represents a driven high/low pin.

--

I 100% agree with you that we should implement the difference between pull-up/pull-down. And I'd like to encourage you to continue working on this. But I propose to do that in a separate PR.

So for this PR that `TODO` comment is fine.

m9710797 (commented):
Thanks again!

I only had time for a quick review. At first sight this PR is looking very good.
* One minor detail (personal preference) for how to reset bits.
* One suggestion for how to implement `rs232_pullup` in `RS232Device`. But I propose to do that in a different PR.

Maybe this evening or tomorrow I'll do a final review and then merge this PR.

src/serial/RS232Net.hh
```diff
@@ -37,7 +37,6 @@ public:
 		} address;
 	};
 
-public:
```
m9710797, line 40: Very much a detail, personal coding style (i know other coding styles choose to do this differently).

I like to group public types, public methods, private methods, private member variables (and typically there shouldn't be any public member variables) together, separated by public/private keywords. Even if that means repeating `public` or `private` when the previous section had already the same access specifier.

m9710797:
I merged this PR  (and followed up with some very minor tweaks).

@retrocomputacion Thanks for your work on this stuff. I hope you're still motivated to finish the pull-up/down stuff in RS232Device (in a new PR).

src/serial/RS232Device.cc
retrocomputacion: > I would propose a different approach:
> 
> * A real pin in a RS232 connector can be in 3 states: it can be driven high, driven low, or it can be unconnected (=floating).
> * But in the current implementation we model the value of a pin with a `bool`, which can only have 2 states.
> 
> Therefor to resolve this I would propose to not return a `bool` but something else which _can_ have 3 states. The first (simplest?) thing that comes to mind is a `std::optional<bool>`:
> 
> * An empty optional then represents an unconnected pin.
> * An optional with a value represents a driven high/low pin.
> 

Yes I think this would be a simple and effective way to implement it.
My first idea was to implement it in RS232device the way I described in the commit, then I figured to implement it in MSXRS232 would be better but couldn't find a way to distinguish when something was plugged in. 

> --
> 
> I 100% agree with you that we should implement the difference between pull-up/pull-down. And I'd like to encourage you to continue working on this. But I propose to do that in a separate PR.
> 
> So for this PR that `TODO` comment is fine.

PR #1588, opened 2024-03-04
Simulate unplugged RS-232 input control lines with configurable 'pull-up'/'pull-down' behaviour

retrocomputacion:
RS232Device.cc input control lines methods now return a boolean optional.
An empty return value means the input is unplugged, and it should default to the value of the configuration tag 'rs232_pullup'.
This configuration tag defaults to false and reflects the hardware implementation of the RS-232 port.
As of now only the SVI-738 is known to need this tag set to true.

### DIFF
--- a/src/serial/MSXRS232.cc
+++ b/src/serial/MSXRS232.cc
@@ -242,24 +242,17 @@ byte MSXRS232::readStatus(EmuTime::param time)
 	auto& dev = getPluggedRS232Dev();
 
 	// Mask out (active low) bits
-	auto dcd = dev.getDCD(time);
-	auto ri = dev.getRI(time);
 
-	if (!dcd ? inputsPullup : *dcd) {
-		result &= ~0x01;
-	}
-	if (hasRIPin && (!ri ? inputsPullup : *ri)) {
-		result &= ~0x02;
-	}
-	if (rxrdyIRQenabled && switchSetting && switchSetting->getBoolean()) {
-		result &= ~0x08;
-	}
-	if (i8254.getOutputPin(2).getState(time)) {
-		result &= ~0x40;
-	}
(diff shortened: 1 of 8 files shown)


src/serial/MSXRS232.cc
```diff
@@ -241,10 +242,13 @@ byte MSXRS232::readStatus(EmuTime::param time)
 	auto& dev = getPluggedRS232Dev();
 
 	// Mask out (active low) bits
-	if (dev.getDCD(time)) {
+	auto dcd = dev.getDCD(time);
+	auto ri = dev.getRI(time);
+
+	if (!dcd ? inputsPullup : *dcd) {
```
m9710797: This code is correct.

It could be simplified a bit by using some of the convenience functions offered by `std::optional`, in particular:
    https://en.cppreference.com/w/cpp/utility/optional/value_or

`value_or(<default>)` returns the value contained in the optional, or a default value if the optional was empty.

And with that the code simplifies to:

    if (dev.getDCD(time).value_or(inputsPullup)) result &= ~0x01;
    ...
    if (hasRIPin && dev.getRI(time).value_or(inputsPullup)) result &= ~0x02;

And similar below:

    return rs232.getPluggedRS232Dev().getDSR(time).value_or(rs232.inputsPullup);
    ....
    return rs232.getPluggedRS232Dev().getCTS(time).value_or(rs232.inputsPullup);

m9710797 (commented):

I reviewed the patch: only one suggestion for a simplification. But your version is also correct.

Feel free to tweak. Or else I can do it after merging your PR.

src/serial/MSXRS232.cc
retrocomputacion: Ah yes, that's more readable, I'll make the change.

PR #1591, opened 2024-03-06
more simple and efficient segment size calculation

pvmm:


src/imgui/ImGuiDebugger.cc
```diff
@@ -628,14 +628,12 @@ void ImGuiDebugger::drawSlots(MSXCPUInterface& cpuInterface, Debugger& debugger)
...
-							} else { // 8kB
-								ImGui::StrCat('R', segments[0], '/', segments[2]);
-							}
-						} else { // 4kB
+						if (segments[0] != segments[1]) { // 4kB
 							ImGui::StrCat('R', segments[0], '/', segments[1], '/', segments[2], '/', segments[3]);
+						} else if (segments[0] != segments[2]) { // 8kB
+							ImGui::StrCat('R', segments[0], '/', segments[2]);
+						} else { // 16kB
+							ImGui::StrCat('R', segments[0]);
```
m9710797: Thanks for your PR. But unfortunately it's wrong. Suppose `segments = {0, 0, 1, 2}`, then your code will print it as a 8kB, but it should be 4kB.

The following will work:

    if ((segments[0] != segments[1]) || (segments[2] != segments[3]) {
        ... handle 4kB
    } else if (segments[0] != segments[2]) {
        ... handle 8kB
    } else {
        ... handle 16kB
    }

This is equivalent to the original code (the same tests, just inverted the condition and swapped the then/else branches), so it's equally efficient. But the code looks nicer because it's less deeply indented.

Can you fix your PR? Or else I can make this tweak this evening.

pvmm: I don't think I'm wrong. At line 631 we are looping through 16k pages and checking which ROM blocks each 4k blocks belongs to. A 16k page cannot be divided into different block sizes at the same time, right? For instance, 2 lower 4k blocks and a single 8k block, totalling 16k. If this is the case and the lower 4k block (segments[0]) is different from the next 4k block (segments[1]), all remaining blocks must be 4k blocks too.

pvmm: Hmm... I missed your example above.

pvmm: I was trying to write a different solution at first, allowing the slots widget to remember the block size so even if it changes to {0, 0, 0, 0} temporarily, it still knows that the block size is 4k (because it was {0, 1, 2, 3} before).

m9710797: When you say "remember" I don't think I'll agree with the implementation. Because then there should also be a part to "forget" (e.g. we switch to a different slot, or the user inserts a different mapper in the same slot (while the MSX is running)). And such "cache-invalidation-code" is often a source of bugs.

The ImGui-idiom is that the GUI should be stateless (as much as possible). And "remembering" (=keeping state) is thus the wrong approach.

If you insist on always displaying according to the bank-size, then you should query what the bank-size is. I believe there is a way to do this. But I'm not 100% sure, and I cannot look it up right now.

Personally I don't mind seeing something like `R2` for a 8kB mapper. It means segment 2 is selected in the whole 16kB page, so twice the same segment in case of a 8kB mapper (or 4 times the same for a 4kB mapper).

When you're debugging your code, you typically already have a very good understanding about which mapper you're using. Therefor I propose to close this PR.

pvmm: But the meaning of R2/2 and R2 are different, since R2/2 mirrors the memory from 0x0000 to 0x2000 and R2 does not. Besides, I can plug two different rom type mappers at the same time, right?

I like the idea of quering the bank size.

m9710797: > But the meaning of R2/2 and R2 are different ...

I tried (but failed?) to explain that `R2` and `R2/2` have the same meaning, which is: "segment 2 is selected in the whole page (the whole 16kB region)". I do agree that with this notation you cannot tell the bank-size. But IMHO when you're debugging your program you probably already know what the bank-size is. And therefor, personally, I didn't mind that that information isn't also reflected in the notation. (Motivated by the fact that it simplifies the implementation).

> Besides, I can plug two different rom type mappers at the same time, right?
 
Euhm, you can, but not in the same slot. And thus a single page cannot have a mix of different bank-sizes.
I'm not sure I understand what you meant with that remark.

So sure, if you have a mix of mapper sizes (in different slots), you also have to look at the slot number to know the bank-size of the mapper. But this scenario is rare (maybe not worth complicating the code for?).

> I like the idea of quering the bank size.

Here's something to get you started:
* `src/memory/RomInfo.hh:19` `struct RomTypeInfo` this contains the `blockSize`.
* `src/memory/RomInfo.hh:68` `getBlockSize(RomType)` returns the block size for a given `RomType`.
* But I don't yet see how to get from a `MSXDevice` (like you have in `src/imgui/ImGuiDebugger.cc:621` to the correct `RomType`. Maybe that link is missing?

mrparn: I don't know if I'm completely missing the point here, but I think some notation where you can easily and quickly distinguish an 8kB mapper from a 16kB mapper could be useful when you aren't sure about how OpenMSX detected a mapper. Also useful if you're trying to debug a situation where two cartridges with different mappers are interacting in an unexpected way (I know, it's a long shot), since you can just glance at the widget and know what's active.

m9710797: To be clear: I 100% agree with you that the notation `R2/2` is better than `R2`. So if it were _easy_ to implement we would do it for sure. Since this morning I looked at it a bit more, and I believe implementing this will require some extensions.

In my _opinion_, the extra information is often not critical. So therefor I initially choose to not implement it.

The above is of course only _my_ opinion. _You_'re free to have a different opinion. And if you feel strongly about this, and if you're willing to spend time on this, I can support you with extending openMSX so that it will become possible to implement this.

mrparn: I'll try to be clearer. I'm from an UX background, so that colors my thinking. If you have the same notation mean two different things, this adds an unnecessary difficulty for the user to parse the info, which is what's important, right? Since we have 8kB mappers, and we already have Rx/y to mean two different segments selected in the same page (albeit in two different "subpages", if you will), it's weird to have just Rx when for some reason x=y.

To me, it feels like inconsistent behavior. And I'd have to remind myself that, when that happens, and if I'm using an 8kB mapper, the second "subpage" has the same selected segment as the first one. Which would be easily and directly communicated by Rx/y.

I propose all segments in a given page are always explicitly shown when using an 8kB mapper, even if they are both the same. Plus it would look neater. ;)

pvmm: > I tried (but failed?) to explain that `R2` and `R2/2` have the same meaning, which is: "segment 2 is selected in the whole page (the whole 16kB region)".

I understood fine, but they have completely different outcomes when using different bank sizes.

m9710797: So we agree. We only have a different opinion about how sever this "inconsistent" notation is w.r.t. the extra implementation effort required to make it "consistent".

And if someone is willing to do that effort I'll help.

m9710797:
@pvmm: I applied your simplification.

Let me know if you want to work on improving this further, then we can discuss (either here or on IRC). I have some rough ideas on how this could be done (it are a _lot_ of small changes).

pvmm:
Yes, let's continue this.

m9710797:
Ok, I've pushed 9223ec4d750f57bf. This makes it possible to query the `RomType` from a `MSXRom`.
Yesterday I was afraid that this step would require a lot of changes, but that fear turned out to be unfounded. (Although the current implementation does a few string lookups/conversions, this could be optimized later (if needed at all)).

So the next steps would be the following (I'm hoping you can work on this):
* In `ImGuiDebugger.cc`, in `drawSlots()`, we now check for `!dynamic_cast<RomPlain*>(device)`. That should be replaced with `dynamic_cast<MSXRom*>`.
* If the device turns out to be a `MSXRom`, we can call the new function `MSXRom::getRomType()`.
* From that `RomType` we can get the bank-size via RomInfo::getBlockSize(RomType)`.
* This block-size can be 0x4000, 0x2000, 0x1000 but also 0. The latter means it's a mapper without bank-switching.
* The next steps could be similar to the current code. Though I'd suggest to only read the minimum required values from the `romblocks` debuggable. I mean today we always read 4 values, but in case of a 16kB bank-size, only one of these values is actually used. So I'd suggest to then only read that 1 value.
* You could even choose to make the code fully generic, so that it'll immediately work for a hypothetical future 2kB mapper. (So write a loop instead of having 3 hard-coded cases for 16kB, 8kB and 4kB).

I hope this is clear. If not then please ask for more details.



BTW I checked: the old debugger also used the notation `R2` instead of `R2/2`. But it's good if you want to work on improving this for the new debugger.

src/imgui/ImGuiDebugger.cc
```diff
@@ -620,21 +621,30 @@ void ImGuiDebugger::drawSlots(MSXCPUInterface& cpuInterface, Debugger& debugger)
...
+							ImGui::StrCat('R', segments[0], '/', segments[1]);
+							break;
+						case 4: // 4kB
+							ImGui::StrCat('R', segments[0], '/', segments[1], '/', segments[2], '/', segments[3]);
+							break;
+						default:
+							ImGui::TextUnformatted("-"sv);
 						}
 					} else {
 						ImGui::TextUnformatted("-"sv);
```
m9710797, line 641: Looks good. (If I have time) I'll probably merge this PR this evening.

But may I suggest one change. Replace the code inside the `if ((rom = dynamic_cast...` with the following.
It's both shorter and more general (handles all block sizes).
Caveat: I have NOT tested this.

    if (unsigned blockSize = RomInfo::getBlockSize(rom->getRomType())) {
        std::string text;
        char separator = 'R';
        for (int offset = 0; offset < 0x4000; offset += blockSize) {
            strAppend(text, separator, romBlocks->read(addr + offset));
            separator = '/';
        }
        ImGui::TextUnformatted(text);
    } else {
        ImGui::TextUnformatted("-"sv);
    }

pvmm: I've tested it. It's OK.

m9710797:
I got an email with the following comment. Though I can't find it in the github UI (did you remove that comment?).
> As per [3102312](https://github.com/openMSX/openMSX/commit/31023120966c5ff89d0cb9b8adfa3bb5cd51044f) there is this NEO-mapper format that supposedly returns a 16-bit segment, so maybe we need to change uint8_t to uint16_t (line 632) and read segment twice (from addr and addr+1 in line 634) if we detect that rom type is ROM_NEO8 or ROM_NEO16.

Anyway, at the moment the "romblocks" debuggable does NOT (yet?) support 16-bit segments. Changing this would break backwards compatibility with existing Tcl scripts.
At the moment scripts can just read this debuggable at any offset (including odd offsets) and expect to get the segment number. For example the script `share/scripts/_slot.tcl` does exactly that: it reads the debuggable at the position of the program-counter. We could adjust that script, but that would still break user scripts that maybe copied this approach.

pvmm:
It was "pending", so I guess you can't see it.

PR #1593, opened 2024-03-06
Reorder key event priority so ImGui can capture them

pvmm:
This seems to fix duplicate keyboard events affecting both emulation and ImGui widgets. For instance: pressing PgDown inside memory window debuggable and also rewinding emulation (go_back_one_step).

reference: https://github.com/ocornut/imgui/blob/master/docs/FAQ.md#q-how-can-i-tell-whether-to-dispatch-mousekeyboard-to-dear-imgui-or-my-application

TODO: need to check if it works for mouse events.

### DIFF
--- a/src/events/HotKey.hh
+++ b/src/events/HotKey.hh
@@ -155,6 +155,7 @@ private:
 	GlobalCommandController& commandController;
 	EventDistributor& eventDistributor;
 	std::optional<Event> lastEvent;
+	std::optional<std::tuple<HotKeyInfo, Event>> postponedEvent;
 };
 
 } // namespace openmsx
--- a/src/events/HotKey.hh
+++ b/src/events/HotKey.hh
@@ -24,15 +24,17 @@ class HotKey final : public RTSchedulable, public EventListener
 {
 public:
 	struct HotKeyInfo {
-		HotKeyInfo(Event event_, std::string command_,
-		           bool repeat_ = false, bool passEvent_ = false)
+		HotKeyInfo(Event event_, std::string command_, bool repeat_ = false,
+			   bool passEvent_ = false, bool global_ = false)
 			: event(std::move(event_)), command(std::move(command_))
 			, repeat(repeat_)
-			, passEvent(passEvent_) {}
+			, passEvent(passEvent_)
+			, global(global_) {}
 		Event event;
 		std::string command;
 		bool repeat;
 		bool passEvent; // whether to pass event with args back to command
+		bool global; // whether event has highest priority
 	};
 	using BindMap = std::vector<HotKeyInfo>; // unsorted
 	using KeySet  = std::vector<Event>;   // unsorted
@@ -43,7 +45,7 @@ public:
 	~HotKey();
 
 	void loadInit();
-	void loadBind(std::string_view key, std::string_view cmd, bool repeat, bool event);
+	void loadBind(std::string_view key, std::string_view cmd, bool repeat, bool event, bool global);
 	void loadUnbind(std::string_view key);
 
 	template<typename XmlStream>
@@ -60,6 +62,9 @@ public:
 			}
 			if (info.passEvent) {
 				xml.attribute("event", "true");
+			}
+			if (info.global) {
+				xml.attribute("global", "true");
 			}
 			xml.data(info.command);
 			xml.end("bind");
--- a/share/scripts/keybindings.tcl
+++ b/share/scripts/keybindings.tcl
@@ -23,8 +23,8 @@ if {$is_dingux} {
 # reverse
 #  note: you can't use bindings with modifiers like SHIFT, because they
 #        will already stop the replay, as they are MSX keys as well
-bind_default PAGEUP   -repeat "go_back_one_step"
-bind_default PAGEDOWN -repeat "go_forward_one_step"
+bind_default PAGEUP   -global "go_back_one_step"
+bind_default PAGEDOWN -global "go_forward_one_step"
 
 # savestate
 if {$tcl_platform(os) eq "Darwin"} {
--- a/src/config/SettingsConfig.cc
+++ b/src/config/SettingsConfig.cc
@@ -39,6 +39,7 @@ struct SettingsParser : rapidsax::NullHandler
 		std::string_view cmd;
 		bool repeat = false;
 		bool event = false;
+		bool global = false;
 	};
 	std::vector<Bind> binds;
 	std::vector<std::string_view> unbinds;
@@ -116,9 +117,9 @@ void SettingsConfig::loadSetting(const FileContext& context, std::string_view fi
 	}
 
 	hotKey.loadInit();
-	for (const auto& [key, cmd, repeat, event] : parser.binds) {
+	for (const auto& [key, cmd, repeat, event, global] : parser.binds) {
 		try {
-			hotKey.loadBind(key, cmd, repeat, event);
+			hotKey.loadBind(key, cmd, repeat, event, global);
 		} catch (MSXException& e) {
 			commandController.getCliComm().printWarning(
 				"Couldn't restore key-binding: ", e.getMessage());
--- a/src/events/EventDistributor.cc
+++ b/src/events/EventDistributor.cc
@@ -22,8 +22,6 @@ void EventDistributor::registerEventListener(
 {
 	std::lock_guard<std::mutex> lock(mutex);
 	auto& priorityMap = listeners[size_t(type)];
-	// a listener may only be registered once for each type
-	assert(!contains(priorityMap, &listener, &Entry::listener));
 	// insert at highest position that keeps listeners sorted on priority
 	auto it = ranges::upper_bound(priorityMap, priority, {}, &Entry::priority);
 	priorityMap.emplace(it, Entry{priority, &listener});
@@ -100,7 +98,6 @@ void EventDistributor::deliverEvents()
 
 				// This might throw, e.g. when failing to initialize video system
 				if (int block = e.listener->signalEvent(event)) {
-					assert(block > e.priority);
 					blockPriority = block;
 				}
 			}
--- a/src/events/EventDistributor.hh
+++ b/src/events/EventDistributor.hh
@@ -20,8 +20,9 @@ public:
 	  */
 	enum Priority {
 		OTHER,
-		HOTKEY,
+		HOTKEY, // global hot keys are resolved here
 		IMGUI,
+		POSTPONED, // normal hot keys are resolved after ImGui
 		MSX,
 		LOWEST, // should only be used internally in EventDistributor
 	};
--- a/src/events/HotKey.cc
+++ b/src/events/HotKey.cc
@@ -52,6 +52,9 @@ HotKey::HotKey(RTScheduler& rtScheduler,
 
 	eventDistributor.registerEventListener(
 		EventType::KEY_DOWN, *this, EventDistributor::HOTKEY);
+	// process event after IMGUI.
+	eventDistributor.registerEventListener(
+		EventType::KEY_DOWN, *this, EventDistributor::POSTPONED);
 	eventDistributor.registerEventListener(
 		EventType::KEY_UP, *this, EventDistributor::HOTKEY);
 	eventDistributor.registerEventListener(
@@ -331,12 +334,26 @@ int HotKey::signalEvent(const Event& event)
 
 int HotKey::executeEvent(const Event& event)
 {
+	if (postponedEvent && getType(std::get<1>(*postponedEvent)) == getType(event)) {
+		// Postponed event runs in POSTPONED priority.
+		executeBinding(std::get<1>(*postponedEvent), std::get<0>(*postponedEvent));
+		postponedEvent.reset();
+		return EventDistributor::MSX; // deny event to the MSX
+	}
+	postponedEvent.reset();
+
 	// First search in active layers (from back to front)
 	bool blocking = false;
 	for (auto& info : view::reverse(activeLayers)) {
 		auto& cmap = layerMap[info.layer]; // ok, if this entry doesn't exist yet
 		if (auto it = findMatch(cmap, event); it != end(cmap)) {
-			executeBinding(event, *it);
+			if (!it->global) {
+				postponedEvent = std::make_tuple(*it, event);
+			} else {
+				postponedEvent.reset();
+				executeBinding(event, *it);
+				return EventDistributor::IMGUI; // deny event to ImGui
+			}
 			// Deny event to MSX listeners, also don't pass event
 			// to other layers (including the default layer).
 			return EventDistributor::MSX;
@@ -347,7 +364,13 @@ int HotKey::executeEvent(const Event& event)
 
 	// If the event was not yet handled, try the default layer.
 	if (auto it = findMatch(cmdMap, event); it != end(cmdMap)) {
-		executeBinding(event, *it);
+		if (!it->global) {
+			postponedEvent = std::make_tuple(*it, event);
+		} else {
+			postponedEvent.reset();
+			executeBinding(event, *it);
+			return EventDistributor::IMGUI; // deny event to ImGui
+		}
 		return EventDistributor::MSX; // deny event to MSX listeners
 	}
 
@@ -510,6 +533,7 @@ void HotKey::BindCmd::execute(std::span<const TclObject> tokens, TclObject& resu
 	}
 	}
 }
+
 string HotKey::BindCmd::help(std::span<const TclObject> /*tokens*/) const
 {
 	auto cmd = getBindCmdName(defaultCmd);
--- a/src/events/HotKey.cc
+++ b/src/events/HotKey.cc
@@ -168,10 +168,10 @@ void HotKey::loadInit()
 	cmdMap = defaultMap;
 }
 
-void HotKey::loadBind(std::string_view key, std::string_view cmd, bool repeat, bool event)
+void HotKey::loadBind(std::string_view key, std::string_view cmd, bool repeat, bool event, bool global)
 {
 	bind(HotKeyInfo(createEvent(key, commandController.getInterpreter()),
-			std::string(cmd), repeat, event));
+			std::string(cmd), repeat, event, global));
 }
 
 void HotKey::loadUnbind(std::string_view key)
@@ -441,11 +441,13 @@ void HotKey::BindCmd::execute(std::span<const TclObject> tokens, TclObject& resu
 	bool layers = false;
 	bool repeat = false;
 	bool passEvent = false;
+	bool global = false;
 	std::array parserInfo = {
 		valueArg("-layer", layer),
 		flagArg("-layers", layers),
 		flagArg("-repeat", repeat),
 		flagArg("-event", passEvent),
+		flagArg("-global", global),
 	};
 	auto arguments = parseTclArgs(getInterpreter(), tokens.subspan<1>(), parserInfo);
 	if (defaultCmd && !layer.empty()) {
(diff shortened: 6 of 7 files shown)


m9710797:
Caveat: I'm at work so I couldn't actually test your PR yet. The following are my first impressions, but I could be mistaken.

Thanks for working on this. But unfortunately I think this is wrong.
In particular it breaks global hotkey handling. For example hotkeys like: F10 -> toggle the console, F11 -> fullscreen, F12 ->mute. These hotkeys should always work, even if ImGui has focus (`WantCaptureMouse` or `WantCaptureKeyboard` are true).

Some background info:
At the moment there are three priority levels for handling events (see `src/events/EventDistributor.hh:21`). From high to low priority these are:
* Hot keys    (e.g. stuff like F12: mute)
* ImGui (all interaction with Dear ImGui)
* MSX emulation (events that control the MSX itself, emulated MSX keyboard, mouse, joystick)

And between these lowest two levels (ImGui and MSX) we actually already have the same logic which you add in this PR. See `src/imgui/ImGuiManager.cc:355). When ImGui has focus we do already block events from the MSX emulation.

So the mismatch is that currently ALL hotkeys have higher priority than ImGui. We clearly want that for some (most?) hotkey bindings (e.g. F12, mute), but not for all (e.g. PgUp).

The real solution would be to _somehow_ be able to specify the _level_ for the hotkey.
* Global: for stuff like F12: (toggle mute) that should always be active.
* Bound to a specific window (or group of windows?). For example (function keys for various step-functions in the debugger).
* Only active when MSX emulation has focus (=when Imgui has no focus). For example PgUp for reverse.

I have no clear idea yet for how to achieve the above.
One possibility which I have NOT yet looked in to is the new, and still experimental/undocumented, ImGui::Shortcut() functionality. But I have no idea if this will (partially/fully) solve the problem or not.

pvmm:
> I have no clear idea yet for how to achieve the above.

I have an idea, but it's a bit clumsy:
1) record predefined keys each ImGui widget consumes.
2) check which ImGui widget has focus.
3) if the pressed keys in the event are contained in the predefined keys for that widget, consume the keys.

pvmm:
> One possibility which I have NOT yet looked in to is the new, and still experimental/undocumented, ImGui::Shortcut() functionality. But I have no idea if this will (partially/fully) solve the problem or not.

I suspect this requires OpenMSX to use ImGui::Shortcut exclusively to dispatch all keyboard related events.

pvmm:
> Some background info: At the moment there are three priority levels for handling events (see `src/events/EventDistributor.hh:21`). From high to low priority these are:
> 
>     * Hot keys    (e.g. stuff like F12: mute)
>     * ImGui (all interaction with Dear ImGui)
>     * MSX emulation (events that control the MSX itself, emulated MSX keyboard, mouse, joystick)

I think this needs to change. If we want ImGui widgets to have priority when they have focus, they need to have the highest priority. Only when they don't consume key events, we send them to the usual HotKey signalEvent to be consumed. EventDispatcher already does this, we just need to extend it to ImGui widgets.

I have some ideas if you wish to proceed.

pvmm:
> I think this needs to change. If we want ImGui widgets to have priority when they have focus, they need to have the highest priority. Only when they don't consume key events, we send them to the usual HotKey signalEvent to be consumed. EventDispatcher already does this, we just need to extend it to ImGui widgets.

I decided to use this strategy.

m9710797:
Hi, thanks for working this!

I saw you put the status back to "draft". I hope that's not because of the "sonarcloud" thing? The sonarcloud thing is highly experimental. I only enabled it yesterday evening, and I'm still evaluating whether it's useful for us or not. Feel free to ignore it for now.

I also have some remarks about the patch itself. First some high level things, and then some implementation specific things. I also understand it's still work in progress, and that we have to do some experiments if we want to make progress on this front.

This patch gives ImGui events the highest priority. I'm not sure I like that. I think it should remain possible for some hotkeys to have higher priority than ImGui. For example F10 to open/close the console, F12 to (un)mute, etc. Such hotkeys should also work even when e.g. we're entering text in an ImGui input field (and thus when ImGui eats all keyboard input).

I don't like that you're adding special rules specifically for the DebuggableEditor windows. E.g. it would also be useful to use PgUp/PgDown in the disassmbly view, and many other windows.  I understand that it's work in progress, and that the patch could be extended to include those windows. But I'm not sure that's the best idea maintenance-wise. (Wouldn't almost all windows then need this special treatment?)

Currently you've hardcoded PgUp/PgDown in the patch. Sure that fixes the current problem of PgUp/PgDown not working in the DebuggableEditor. But the problem we should be solving is not limited to that. E.g. maybe some users have bound the reverse functionality from PgUp/PgDown to some other keys (maybe "[" and "]", or "CTRL+LEFT" and "CTRL+"RIGHT"). Or maybe some users have bound "TAB" to some functionality while some ImGui windows also want to use that key. (And it's not a good solution to extend the list of hardcoded keys in your current approach).

I admit that I don't have a full solution myself. But maybe one idea to explore is the following:
* Instead of having a single priority level for hotkey events, have two levels. One higher and one lower priority than ImGui events.
* Then upon creating a new hotkey binding (via the "bind" command), the user must "somehow" indicate this level.
* For keys like F10, F12 the level may be higher than ImGui. For reverse (PgUp/PgDown) the level could be lower.
* This means F10, F12 are always active, even if an ImGui window has focus. But PgUp/PgDown are only active when the main openMSX window has focus.
* Note that this solution makes no distinction between which ImGui window is active (and whether that window actually does something with PgUp/PgDown or not). Maybe this is good enough?? Or maybe we still need to refine this in the future?? I don't know yet.


Now some implementation details. (Again I understand this patch is experimental, and that the focus was on getting something working rather than efficiency).

On each window creation you call "FindWindowByName()" that seems expensive. I have not checked, but I think the function "GetItemID()" will return the same result. At least when called between "ImGui::Begin()" and "ImGui::End()" (the same location where you currently call "FindWindowByName()".

You're using "ImGui::IsWindowChildOf(NavWindow, ...)", that's interesting. Is my understand correct that this checks that the current window is that one that has (keyboard?) navigation focus? Unfortunately it uses stuff from "imgui_internal.h", (which is the unstable API) so we should first look for an alternative approach. Did you try "ImGui::IsWindowFocused()"? That might not give the exact same result, but maybe close enough? Also, maybe in the first version we don't care about which ImGui window has focus? (So simply: any ImGui window has focus, or the main openMSX window has focus).


Again thanks a lot for working on this. Please don't be discouraged by my comments. At least you're trying things (while I've kept this hanging for months now).

pvmm:
> This patch gives ImGui events the highest priority. I'm not sure I like that. I think it should remain possible for some hotkeys to have higher priority than ImGui. For example F10 to open/close the console, F12 to (un)mute, etc. Such hotkeys should also work even when e.g. we're entering text in an ImGui input field (and thus when ImGui eats all keyboard input).

With this change, Hotkeys like F10 still work if memory debuggable has focus, as should F12 and any other key unused by the memory debuggable. This change puts ImGui at high priority but it also creates a filter in executeEvent that blocks only key events that the memory debuggable knows it will use (currently, there is only PageUp and PageDown, but I will improve this to be airtight). So keys like F10 and F12 are returned to the Hotkey event handler.

pvmm:
> I don't like that you're adding special rules specifically for the DebuggableEditor windows. E.g. it would also be useful to use PgUp/PgDown in the disassmbly view, and many other windows. I understand that it's work in progress, and that the patch could be extended to include those windows. But I'm not sure that's the best idea maintenance-wise. (Wouldn't almost all windows then need this special treatment?)

In the next change to this PR (when I have time), I intend to move all the key bindings associated with debuggable widgets to their specific classes so they can filter the key events themselves. It will be their responsibility, which is desirable because it's more modular.

pvmm:
> On each window creation you call "FindWindowByName()" that seems expensive. I have not checked, but I think the function "GetItemID()" will return the same result. At least when called between "ImGui::Begin()" and "ImGui::End()" (the same location where you currently call "FindWindowByName()".

I thought about it, but I used the same solution imgui_widgets.c uses to get the recently created ImGuiWindow in it's BeginXXXXXX() functions. I will experiment a little more on this issue.

m9710797:
About the filtering in executeEvent(), I get that. But it's still the case that the event is FIRST ALSO processed by ImGui. It's no longer possible for a hotkey to block the event from ImGui.

The problem is that a typical PC keyboard has not enough keys for all the functionality that we'd like to offer. And this is made worse by the many keys that are already taken for emulating the MSX keyboard. So there will always be conflicts. And different users have different needs. We offer a default set of keybinding, but it's important that users can customize this.

For example the HOME and END keys are in principle already in use for the MSX keyboard. But often you don't actually need those in specific MSX programs. So some users may want to bind those to other functionality (and then maybe loose the corresponding MSX keys).

But the HOME and END keys also already have a meaning in ImGui (e.g. when editing a text field). Depending on what functionality the user bound to HOME/END it may be desirable to loose the ImGui functionality (the loss may be acceptable for some users) or you may want to keep the ImGui functionality and only have the HOME/END bindings active when no ImGui window has focus.

Another example. The current PR tries to give PgUp/PgDown functionality to the DebuggableEditor window instead of to hotkeys (for reverse back/forward). And I assume a later version of this PR will do the same for the disassembly view. I agree that's a good default behavior. HOWEVER, I can imagine that some user DO prefer to have PgUp/PgDown (with reverse functionality) active even when the disassembly view has focus. For example when you're debugging a difficult assembly problem, and you quickly want to step back and re-run till a certain breakpoint.

My points are:
* Almost all keys can be used by ImGui (e.g. in a text input field), so filtering out specific keys (like you do now for PgUp/PgDown) won't work.
* Sometimes you want a binding to have higher priority than ImGui, sometimes lower priority. And different users may have different preferences, even for the SAME keys. So we cannot hard code this decision in the c++ code. In the end we do need to give the user control over this decision. (We can only try to establish a reasonable set of default bindings).

pvmm:
> On each window creation you call "FindWindowByName()" that seems expensive. I have not checked, but I think the function "GetItemID()" will return the same result. At least when called between "ImGui::Begin()" and "ImGui::End()" (the same location where you currently call "FindWindowByName()".

This is fixed for now by using `ImGui::GetCurrentFocusScope()` since `GetItemID()` doesn't return what I need. I still need to include "imgui_internal.h". But you have control over the version of Dear ImGui that is bundled with OpenMSX, so at least there are no surprises by the user providing his own.

src/imgui/ImGuiCpp.hh
```diff
@@ -60,20 +61,24 @@
 namespace im {
 
 // im::Window(): wrapper around ImGui::Begin() / ImGui::End()
-inline void Window(const char* name, bool* p_open, ImGuiWindowFlags flags, std::invocable<> auto next)
+inline ImGuiID Window(const char* name, bool* p_open, ImGuiWindowFlags flags, std::invocable<> auto next)
 {
+	ImGuiID id = 0;
 	if (ImGui::Begin(name, p_open, flags)) {
+		id = ImGui::GetCurrentFocusScope();
```
pvmm: This function requires imgui_internal.h

src/imgui/ImGuiCpp.hh
```diff
@@ -82,9 +87,12 @@ struct WindowStatus {
 	                       //         gets automatically reset when done
 	void raise() { do_raise = open = true; }
 };
-inline void Window(const char* name, WindowStatus& status, ImGuiWindowFlags flags, std::invocable<> auto next)
+inline ImGuiID Window(const char* name, WindowStatus& status, ImGuiWindowFlags flags, std::invocable<> auto next)
 {
+	ImGuiID id = 0;
 	if (ImGui::Begin(name, &status.open, flags)) {
+		id = ImGui::GetCurrentFocusScope();
```
pvmm: This function requires imgui_internal.h

pvmm:
I thought about what you said and here is my idea for next changes to this PR:
1) Separate HotKey class into a `EventListener` class (basically just `signalEvent` method) and a Management class (hotkey initialization, loading and saving);
2) After separation, we create a new event listener: HotKey-To-ImGui-Bridge `EventListener` that only listens to key bindings in ImGui root window;
3) The Management class knows how to initialize, save and load both key bindings (high priority Hotkey and ImGui) because they use a common interface.

m9710797:
Hi,

Thanks again work working on this! Though I think you're moving too fast. You're already trying to implement a solution, while (at least to me) it's not clear yet what kind of solution we actually want. Also because I have the feeling that your current implementation approach doesn't fit well with my personal (partial, still incomplete) understanding of the problem.

So let's try to fix this. Let's start by listing some observations and requirements. Very likely this list is incomplete, and it may even contain some contradictions. So absolutely feel free to discuss, correct and append to this list.

=Requirements=
I'm repeating a lot of what I've said earlier in this thread, but it might be good to have all info in a single place.
* The host keyboard only has so many keys. The MSX keyboard has many keys. Preferably every MSX key is mapped to a host key. This only leaves so many unused host keys. Not enough for all the functionality that we want to offer (as keybindings). To make it even more complicated, we recently integrated ImGui, and that also wants to use keyboard input.
* This means we'll have to make some compromises. We can only (try to) offer a reasonable set of default keybindings. But for sure this default set won't be suited for everyone or for every use case. So we should also offer a flexible way to customize/override the default bindings.
* Part of this flexibility is that sometimes we want a keybinding to be always active (so take priority over both ImGui and the MSX keyboard). Or sometimes we want the binding to first go to ImGui, and only if ImGui did not consume it (e.g. when no ImGui window has focus) then process it as a hotkey (but then still block it from the MSX keyboard).
* A good example could be the HOME/END keys. These are mapped to the MSX keyboard and also ImGui uses these keys (e.g. in text input fields). Though for some users it might be acceptable to hide the corresponding MSX keys (e.g. because the MSX program they are using doesn't use those keys). For other users it may even be acceptable to also hide the ImGui functionality (e.g. because they don't (often or at all) use HOME/END in text input fields). But important: this choice will be different for different users. The solution we implement must be flexible enough to support both (and leave this choice up to the user).
* [I have not given this part much thought yet. Especially on this part input would be very welcome.] We would also like to create keybindings for specific ImGui functionality. For example CTRL+G to focus (and select all text in) the address input bar in the DebuggableEditors. But CTRL+G should only be active when a DebuggableEditor window is active (and it's possible that multiple DebuggableEditor windows are open, only the one that has focus should react). So these bindings are specific to certain windows (and possibly even which tab in a window is selected (e.g. the "breakpoint" vs "watchpoint" tabs in the Breakpoints window)).
* Our current key-binding mechanism (via the "bind" command) always maps a key to a Tcl command. So far that worked fine. Though I'm not sure this is also a good fit for ImGui bindings. E.g. it would not be a good idea to create a Tcl command for "give the address input bar focus in 'some' DebuggableEditor".

=User interface=
In this section I'll talk more about how the above requirements could look from a user point of view. At the moment this is not yet fully clear to me. So this section still has many questions, and some things that I say here are certainly still open to change.

We currently have the "bind" command. This allows to connect some event (but for simplicity let's limit that to keyboard events in this discussion) to arbitrary Tcl commands. At the moment such bindings are global (take priority over e.g. ImGui). Global bindings should remain supported. And then for backwards compatibility I propose that the existing bind commands keep working unchanged (so the current syntax will still create global bindings).

We also want to be able to create bindings with lower priority than ImGui. One possibility could be a syntax like:
`bind -msx <event> <command>`
Note the "-msx" flag. This means: this binding is only active when the MSX window has focus. For symmetry we could also allow a flag like "-global" (but for backwards compatibility omitting this flags has the same meaning).

Yesterday I had a discussion about this with Quibus on IRC. There we came up with the syntax:
`bind -window <window-name> <event> <command>`
And "window-name=msx" means the same as the above "-msx" flag. But this syntax could also allow other window-names, e.g. to create a binding for a specific ImGui window.

HOWEVER I no longer believe binding to an "ImGui window" will be sufficient. I think (in the future) we may also want to have specific bindings for different tabs in the same window. And certain bindings maybe not be active when e.g. a radio-button is not pressed, or when certain sections are collapsed, etc ...

My current thinking is that for ImGui bindings we should investigate using the ImGui::Shortcut() function. And then make those customizable (I haven't studied ImGui::Shortcut() in detail yet, but it seem very flexible: support tabs, conditionally enabled, not active when collapsed, ...). We should 'somehow' give each of these a unique name, and then provide a mechanism for the user to associate those Shortcut-names with key-combinations. Maybe this could be via the "bind" command. Or it could be via a completely new mechanism. (I mean the internal implementation for hotkeys and for ImGui-shortcuts will be completely separate, but we could still present both to the user via the "bind" command).

I believe this is also how other applications typically allow to customize GUI shortcuts: you get a (long) list of possible actions, and then you can select a key-combination (or no key-combination) for each element in this list.

=Implementation=
This section is even more unclear. There are only my best guesses at the moment. But I could be completely wrong.

I think the HotKeys class should remain limited to binding events to Tcl commands. It should not be extended to ImGui shortcuts. Instead for ImGui shortcuts we need a new mechanism.

Hotkeys should be extended with an extra boolean, this indicates the priority (higher or lower than ImGui). So the changes to the current code should be minimal. Adding a filter in the ImGui code seems like the wrong approach to me. And certainly trying to enumerate the keys that are used per ImGui window seems wrong to me.

For ImGui shortcuts we should experiment with using the ImGui::Shortcut() API (at the moment this API is experimental, so part of "imgui_internal.h"). Initially we could hard-code the key-combination in the code. Later we could add a mechanism to associate a name with each Shortcut() call in the code, and then let the user associate a custom key-combination (or none) with these names.

pvmm:
I thought about a rather simple approach and implemented a solution based on it. It could be a temporary solution before we think about this complex layered widget event handling approach since it is quite simple.

src/events/HotKey.hh
```diff
@@ -150,6 +155,7 @@ private:
 	GlobalCommandController& commandController;
 	EventDistributor& eventDistributor;
 	std::optional<Event> lastEvent;
+	std::optional<std::tuple<HotKeyInfo, Event>> postponedEvent;
```
pvmm, line 158: I guess a better solution is to not store events, but to keep track of the priority level it is currently in (change `signalEvent()` to include this as a 2nd parameter?).

m9710797: That's possible.

My idea was to have two separate EventListeners (two inner objects) in HotKey. So that we keep the simplicity of exactly one priority per listener object (and so that you don't have to remove the asserts from EventDistributor.cc).

But I've no idea how easy that is. And I don't have much time today to investigate.
I also didn't have time yet to read your full PR in detail. I hope to do that soon (I hope to make time for it this weekend).

Thanks!

m9710797:
I pushed 836528a08a9.
That's basically a rework of this PR (very much inspired by it, but some aspects implemented slightly different).
Thanks a lot pvmm for starting this work!

Next step could be to experiment with ImGui::Shortcut().   (But best in a different PR).

pvmm:
> I pushed [836528a](https://github.com/openMSX/openMSX/commit/836528a08a9c5eb2606a8535db46836eb9e10e35). That's basically a rework of this PR (very much inspired by it, but some aspects implemented slightly different).

So now with this change what is the status of the ImGui interface? Is it good enough for an official release?

m9710797:
Personally I think we should still try to create keybindings for ImGui specific things. Because that the most frequently requested feature and also a regression in functionality compared to the old debugger. The following bindings are most often requested:
* a binding for goto-address. both in disassembly view and hexviewers
* step-in, step-over, run, break, ...
* (less frequent) set a breakpoint.

I *think* the ImGui::Shortcut() function seems like a good fit. But I still haven't tried it.
If I understand correctly these bindings will only have effect when the corresponding window has focus.
Initially we could just hardcode some keys (e.g. CTRL-G for goto address). Later we can extend this to make the keys configurable. And then at that point we can decide whether to expose this functionality to the user via the "bind" command or via something different (new). (But the internal implementation will likely not reuse the bind/hotkey functionality).

pvmm:
Closing this PR since the quest has evolved: https://github.com/openMSX/openMSX/pull/1601

pvmm:
Closing this PR since the quest has evolved to https://github.com/openMSX/openMSX/pull/1601

PR #1596, opened 2024-03-19
digit separators improves legibility

pvmm:


src/fdc/DiskImageUtils.cc
```diff
@@ -523,7 +523,7 @@ static std::vector<unsigned> partitionNextor(SectorAccessibleDisk& disk, std::sp
 		// Add partition entry
 		auto& p = pt.part[0];
 		p.boot_ind = (i == 0) ? 0x80 : 0x00; // boot flag
-		p.sys_ind = size > 0x10000 ? 0x0E : 0x01; // FAT16B (LBA), or FAT12
+		p.sys_ind = size > 0x10'000 ? 0x0E : 0x01; // FAT16B (LBA), or FAT12
```
m9710797: I generally agree with this PR.
Except for this single change. I think you intended to write `0x1'0000` (so group per 4 digits)?

I'll pull this PR after this fix.

PR #1600, opened 2024-03-31
Completing slotviewer changes from #1591

pvmm:
Completing changes introduced in #1591, mirrored ROMs should not display segment information since they are not mappers of any kind, just plain and simple ROM.

### DIFF
--- a/src/imgui/ImGuiDebugger.cc
+++ b/src/imgui/ImGuiDebugger.cc
@@ -679,7 +679,9 @@ void ImGuiDebugger::drawSlots(MSXCPUInterface& cpuInterface, Debugger& debugger)
 						ImGui::StrCat(mapper->getSelectedSegment(page));
 					} else if ((rom = dynamic_cast<MSXRom*>(device)) &&
 						(romBlocks = debugger.findDebuggable(device->getName() + " romblocks"))) {
-						if (unsigned blockSize = RomInfo::getBlockSize(rom->getRomType())) {
+						unsigned blockSize = RomInfo::getBlockSize(rom->getRomType());
+						// Mirrored ROMs are not mappers, so displaying segments doesn't make sense.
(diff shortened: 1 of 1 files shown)


MBilderbeek:
Perhaps we could somehow indicate in RomInfo.cc that a ROM type is one that has segments or one without segments. 
I'm not sure what that block size column is used for exactly, it is set to 8kB for many non-mapper ROM types and it's set to 0 for several others. So, not sure if this field could be used, but perhaps we should add another?
Then we don't have to hardcode the type without segments in any other place. @m9710797 what do you think?

pvmm:
> I'm not sure what that block size column is used for exactly, it is set to 8kB for many non-mapper ROM types and it's set to 0 for several others.

It is used when running command `openmsx_info romtype <romtype name>` and it returns the block size in the description. Other than that, it is used in the slot viewer and that's it.

MBilderbeek:
> > I'm not sure what that block size column is used for exactly, it is set to 8kB for many non-mapper ROM types and it's set to 0 for several others.
> 
> It is used when running command `openmsx_info romtype <romtype name>` and it returns the block size in the description. Other than that, it is used in the slot viewer and that's it.

But then it makes no sense to have it on 8kB for non-mapper ROM types... if this is the only use, let's just correct it and set them to the proper values, e.g. 0 for non-mapped ROMs and check on the block size to know whether to display segment info.

m9710797:
I don't have much time today (I only very quickly checked the source code).
MBilderbeek's comment makes sense. And we should most likely do it for more than only ROM_MIRRORED.
At first sight for all mappers in RomInfo.cc lines 88-107.

m9710797:
I pushed 041233232dd27b727cd04ff0c9807e3faaded357 which implements MBilderbeek's solution.

PR #1604, opened 2024-04-02
null Platform_SetWindowFocus called when trying to raise window

pvmm:
Seems like almost all `Platform_XXXXXXX()` functions are not implemented by the backend. So calling `Platform_SetWindowFocus()` crashes when a window is already visible and we try to raise it above the others. Fortunately there is another function that fits the bill.

```
(gdb) print ImGui::GetPlatformIO() 
$1 = (ImGuiPlatformIO &) @0x20f20e8: {Platform_CreateWindow = 0x0, Platform_DestroyWindow = 0x0, Platform_ShowWindow = 0x0, 
  Platform_SetWindowPos = 0x0, Platform_GetWindowPos = 0x0, Platform_SetWindowSize = 0x0, Platform_GetWindowSize = 0x0, 
  Platform_SetWindowFocus = 0x0, Platform_GetWindowFocus = 0x0, Platform_GetWindowMinimized = 0x0, 
  Platform_SetWindowTitle = 0x0, Platform_SetWindowAlpha = 0x0, Platform_UpdateWindow = 0x0, Platform_RenderWindow = 0x0, 
  Platform_SwapBuffers = 0x0, Platform_GetWindowDpiScale = 0x0, Platform_OnChangedViewport = 0x0, 
  Platform_CreateVkSurface = 0x0, Renderer_CreateWindow = 0x0, Renderer_DestroyWindow = 0x0, Renderer_SetWindowSize = 0x0, 
  Renderer_RenderWindow = 0x738240 <ImGui_ImplOpenGL3_RenderWindow(ImGuiViewport*, void*)>, Renderer_SwapBuffers = 0x0, 
  Monitors = {Size = 2, Capacity = 8, Data = 0x2329af0}, Viewports = {Size = 1, Capacity = 8, Data = 0x20265a0}}

```

### DIFF
--- a/src/imgui/ImGuiCpp.hh
+++ b/src/imgui/ImGuiCpp.hh
@@ -88,7 +88,11 @@ inline void Window(const char* name, WindowStatus& status, ImGuiWindowFlags flag
 		if (status.do_raise) {
 			status.do_raise = false;
 			if (!ImGui::IsWindowAppearing()) { // otherwise crash, viewport not yet initialized???
+			// depending on the backend this could be uninitialized
+			if (ImGui::GetPlatformIO().Platform_SetWindowFocus != nullptr)
 				ImGui::GetPlatformIO().Platform_SetWindowFocus(ImGui::GetWindowViewport());
+			else
+				ImGui::SetWindowFocus();
 			}
 		}
 		next();


m9710797:
Hi pvmm,

Thanks for reporting/investigating, but I'm not sure I fully understand ...

You say '... almost all `Platform_XX() functions are not implemented by the backend ...`. That's not what I see in `src/3rdparty/imgui/imgui_impl_sdl2.cc` (lines 1105-1117). But that contradicts what you showed in the gdb output. Were you trying to use a different backend? (but openMSX currently requires the sdl2 backend). Or could it be that the backend was not yet initialized at the time you did the query in gdb?

Your patch proposes to replace `Platform_SetWindowFocus()` with `ImGui::SetWindowFocus()`. Despite the similar name, both functions don't do the same time thing. A better name for the former would be `Platform_RaiseWindow()`. It both raises the window and gives it focus, while the latter only gives the window focus. But depending on how your window manager is configured, giving a window focus might also raise it. I think it's the default behavior on windows (and hence maybe the bad function name?). But it's not how I've configured my window manager (KDE).

See commit 9b56a839e4 (and especially the commit message) for more details. I did initially try using `ImGui::SetWindowFocus()`, but that was insufficient.

What was the trigger to open this PR? You talk about a crash, but it doesn't crash for me, nor did I see crash reports from any other people (about this). This code should trigger very often, so I'm surprised you only see a crash now.

pvmm:
> You say '... almost all `Platform_XX() functions are not implemented by the backend ...`. That's not what I see in `src/3rdparty/imgui/imgui_impl_sdl2.cc` (lines 1105-1117). But that contradicts what you showed in the gdb output. Were you trying to use a different backend? (but openMSX currently requires the sdl2 backend). Or could it be that the backend was not yet initialized at the time you did the query in gdb?

In my Fedora/Wayland/GNOME machine, Dear ImGui seems to run with different settings. I believe this could be the default now that GNOME is ditching Xorg for good and KDE will probably follow. The backend should be fully initialized by the time of the crash and it is easy to reproduce here:

1) Open menu Settings > Configure messages... and leave it open
2) Bind a bogus command to a hot key that causes error, for instance `bind Ctrl+F9 bogus`
3) Pressing Ctrl+F9 will open an error dialog with the message _Warning: Error executing hot key command: invalid command name "bogus"_
4) Pressing "Configure..." will try to raise the "Configure messages..." window by calling that specific Platform_SetWindowFocus.

If I start GNOME in Xorg (which option still exists in GNOME 45), it behaves exactly like you described. I guess we could test if `Platform_SetWindowFocus` is null and just call `SetWindowFocus` if it is.

pvmm:
Or maybe Dear ImGui is the one that should be fixed since SDL2 supports Wayland since version 2.0.2

pvmm:
This is what "About Dear ImGui" reports when I run it inside Wayland/GNOME:
```
Dear ImGui 1.90.5 WIP (19044)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=202002
define: IMGUI_DISABLE_OBSOLETE_FUNCTIONS
define: IMGUI_DISABLE_OBSOLETE_KEYIO
define: __linux__
define: __GNUC__=4
define: __clang_version__=17.0.6 (Fedora 17.0.6-2.fc39)
define: IMGUI_HAS_VIEWPORT
define: IMGUI_HAS_DOCK
--------------------------------
io.BackendPlatformName: imgui_impl_sdl2
io.BackendRendererName: imgui_impl_opengl3
io.ConfigFlags: 0x00000041
 NavEnableKeyboard
 DockingEnable
io.ConfigViewportsNoDecoration
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x0000100E
 HasMouseCursors
 HasSetMousePos
 RendererHasVtxOffset
 RendererHasViewports
--------------------------------
io.Fonts: 2 fonts, Flags: 0x00000000, TexSize: 512,256
io.DisplaySize: 1280.00,960.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------
style.WindowPadding: 8.00,8.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
style.ItemInnerSpacing: 4.00,4.00
```

pvmm:
Here is the diff of Dear Imgui Config/Build information running on Xorg against running on Wayland:
```
[pedro@fedora openMSX]$ diff -u xorg.txt wayland.txt 
--- xorg.txt	2024-04-03 10:44:49.370773174 -0300
+++ wayland.txt	2024-04-03 10:38:15.613819167 -0300
@@ -13,19 +13,16 @@
 --------------------------------
 io.BackendPlatformName: imgui_impl_sdl2
 io.BackendRendererName: imgui_impl_opengl3
-io.ConfigFlags: 0x00000441
+io.ConfigFlags: 0x00000041
  NavEnableKeyboard
  DockingEnable
- ViewportsEnable
 io.ConfigViewportsNoDecoration
 io.ConfigInputTextCursorBlink
 io.ConfigWindowsResizeFromEdges
 io.ConfigMemoryCompactTimer = 60.0
-io.BackendFlags: 0x00001C0E
+io.BackendFlags: 0x0000100E
  HasMouseCursors
  HasSetMousePos
- PlatformHasViewports
- HasMouseHoveredViewport
  RendererHasVtxOffset
  RendererHasViewports
 --------------------------------
```

pvmm:
I know what it is now. SDL2 requires function Wayland_GetGlobalMouseState() to get the global mouse position, but Wayland doesn't provide that for security reasons and then Dear ImGui won't set `PlatformHasViewports` without that information. There is a workaround to calculate absolute mouse position in SDL2, but it was included last year and will only be available in version 3 of the SDL2 API (SDL3?).

Either way, calling Platform_SetWindowFocus should be tested against uninitialized function pointer.

m9710797:
[EDIT: I wrote this comment before I read your last comment. But we're basically saying the same.]

I think you're right.  The function pointer `Platform_SetWindowFocus()` (and many others) gets initialized inside the `ImGui_ImplSDL2_InitPlatformInterface()` function (in `imgui/imgui_impl_sdl2.cc`), but that function is only called when the `ImGuiConfigFlags_ViewportsEnable` and `ImGuiBackendFlags_PlatformHasViewports` flags are set. But unfortunately the latter is NOT set when using Wayland. I searched in the "Dear ImGui" github issues, and unfortunately there seem to be many issues related to Wayland. And it doesn't seem they'll get fixed soon :(

I think your suggestion makes sense: we should first check whether `Platform_SetWindowFocus` is `nullptr`. And if it is we can indeed fallback to `ImGui::FocusWindow()`.

All my testing (so far) was done with X11 instead of Wayland. I'll try to test Wayland later.  But no viewport support with Wayland might indeed be a big miss. And now I also understand better some of the choices you made earlier. E.g. with only a single viewport it may indeed make sense to center new configuration windows.

I wonder if there's something we can do to help Dear ImGui get better Wayland support. Because with only a single viewport the openMSX debugger is probably a LOT less usable (I'll try it out soon).

pvmm:
I will try compiling OpenMSX with SDL3 (or SDL2-compat, the SDL2 compatibility layer) if it is possible, then maybe viewports will be supported again.

In the meantime, can we merge this?

m9710797:
Merged.

BTW I tested with Wayland/KDE version 5.27.8, Intel UHD Graphics 630.
And for me viewports in Dear ImGui did work fine. So I couldn't reproduce the crash you saw.

Here's my output for "About Dear Imgui". I double checked that I was using Wayland.
```
Dear ImGui 1.90.5 WIP (19044)
--------------------------------
sizeof(size_t): 8, sizeof(ImDrawIdx): 2, sizeof(ImDrawVert): 20
define: __cplusplus=202002
define: IMGUI_DISABLE_OBSOLETE_FUNCTIONS
define: IMGUI_DISABLE_OBSOLETE_KEYIO
define: __linux__
define: __GNUC__=13
define: IMGUI_HAS_VIEWPORT
define: IMGUI_HAS_DOCK
--------------------------------
io.BackendPlatformName: imgui_impl_sdl2
io.BackendRendererName: imgui_impl_opengl3
io.ConfigFlags: 0x00000441
 NavEnableKeyboard
 DockingEnable
 ViewportsEnable
io.ConfigViewportsNoDecoration
io.ConfigInputTextCursorBlink
io.ConfigWindowsResizeFromEdges
io.ConfigMemoryCompactTimer = 60.0
io.BackendFlags: 0x00001C0E
 HasMouseCursors
 HasSetMousePos
 PlatformHasViewports
 HasMouseHoveredViewport
 RendererHasVtxOffset
 RendererHasViewports
--------------------------------
io.Fonts: 2 fonts, Flags: 0x00000000, TexSize: 512,256
io.DisplaySize: 1280.00,960.00
io.DisplayFramebufferScale: 1.00,1.00
--------------------------------
style.WindowPadding: 8.00,8.00
style.WindowBorderSize: 1.00
style.FramePadding: 4.00,3.00
style.FrameRounding: 0.00
style.FrameBorderSize: 0.00
style.ItemSpacing: 8.00,4.00
style.ItemInnerSpacing: 4.00,4.00
```

m9710797:
Ok, when I set the environment variable SDL_VIDEODRIVER=wayland, then I do get only one viewport :(
By default my version of SDL (2.28.3) seems to use the "x11" backend.

PR #1606, opened 2024-04-04
fix window not raising above others in main viewport in some cases

pvmm:
XWayland+GNOME: when window is inside the main viewport, `ImGui::SetWindowFocus()` is needed to raise focus even if `Platform_SetWindowFocus()` exists.

### DIFF
--- a/src/imgui/ImGuiCpp.hh
+++ b/src/imgui/ImGuiCpp.hh
@@ -91,9 +91,10 @@ inline void Window(const char* name, WindowStatus& status, ImGuiWindowFlags flag
 				// depending on the backend (e.g. Wayland) this could be nullptr
 				if (auto* swf = ImGui::GetPlatformIO().Platform_SetWindowFocus) {
 					swf(ImGui::GetWindowViewport());
-				} else {
-					ImGui::SetWindowFocus();
 				}
+				// XWayland/GNOME: when window is inside the main viewport, this function is
+				// needed to raise focus even if Platform_SetWindowFocus exists
+				ImGui::SetWindowFocus();
 			}
 		}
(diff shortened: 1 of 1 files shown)


m9710797:
Thanks. I'll merge this PR, this evening when I'm back at home (I first want to run some small experiments myself).

About your XWatland/GNOME" remark: I think it's independent. For ImGui windows that are fully inside the main window, Dear ImGui does not create a OS window, it's merely a rectangle that's drawn inside the content of the main window. The OS does not see this as a window, so functions like Platform_SetWindowFocus() make no sense. And this is true for all platforms, not only XWayland/GNOME. So I'll remove the "Xwaylang/GNOME" part from the comment in your PR.

--

Slightly off-topic:
I checked the SDL source code:
* SDL2 should always prefer the "x11" video driver over the "wayland" video driver.  Only in SDL version 2.0.21 this order was reversed, but then quickly reverted in version 2.0.22.
* SDL3 does prefer "wayland" over "x11", but there are discussions to also revert the order there (before the official SDL3 release). See: https://www.phoronix.com/news/SDL-3.0-Wayland-Default-For-Now

@pvmm: so I wonder why in your setup you seem to have the wayland sdl video driver by default.

pvmm:
> @pvmm: so I wonder why in your setup you seem to have the wayland sdl video driver by default.

Yes, according to the Fedora build routine at https://koji.fedoraproject.org/koji/rpminfo?rpmID=36728907 they apply the self explanatory SDL2-2.0.22-prefer-wayland.patch patch.

m9710797:
Oops, I forgot about this. Merged now.

Issue #1612, opened 2024-04-12
hex editor mappings of HOME and END keys


pvmm:
Pressing the END key on the large window of the hex editor will rewrite the value under the cursor and move it forward one byte, which is a non expected side effect of the END key, since it is supposed to be a navigation key not an editing key.

But if you plug a generic megarom cartridge and write to its memory blocks, it will change the segment address since, on generic 8kb ROM, every memory position is a switching address. So pressing the END key on the hex editor while the large window of the hex editor has focus will cause it to change segment (to the value under the cursor % number of memory banks) which may crash the MSX. To fix this, the END key should change from an editing key to a proper navigation key. And while we are at it, map the HOME and END keys to jump to the beginning and ending of the hex editor memory.

m9710797:
That's unexpected. But I can make a **guess** about what is happening:

In the current implementation of the hex-editor (based on this code: https://github.com/ocornut/imgui_club) the cell under the cursor is replaced with a `ImGui::InputText()` field. A typical behavior for a text input field is to accept the new value after the enter key was pressed. but that's not desired for a hex-editor. The alternative behavior is to accept the new value on every edit action, but that's also not desired here (we want 2 hex digits before writing a new value to memory). So the solution that's currently implemented is: on every input check whether the cursor has reached position = 2. If that's the case then (the assumption is that) 2 hex digits must have been typed, and then we write to memory.

But my **guess** is that this also triggers when the cursor reaches postion = 2 because the END key was pressed???

(If this guess is correct then) I have no idea yet for how to solve this. Does setting an `ImGui::Shortcut()` for `END` override the `END` key in `ImGui::InputText()`?? 
Another possibility could be to not use `ImGui::InputText()` but implement our own widget. For hex-digit input that might be doable, but I'd prefer to not have to deal with ASCII input myself (I haven't looked yet at the implementation of `ImGui::InputText()`, maybe it is doable??).

--

Orthogonal to this problem:
Navigating to the top/bottom is more usually done via Ctrl+Home/Ctrl+End. And plain Home/End navigate to the start/end of a line.

Currently the shortcuts for move-up/down/left/right can be customized. That was a good experiment while working on the new shortcut-infrastructure. But now that the infrastructure is in good shape, I think we should remove these customization points again. I can't imagine why anyone would want to have something else than the cursor keys for this, or why someone would want to unbind these shortcuts). Offering these customization points can only cause confusion.
[...]

PR #1613, opened 2024-04-12
fix issue #1612

pvmm:
This should fix issue #1612

### DIFF
--- a/src/imgui/DebuggableEditor.cc
+++ b/src/imgui/DebuggableEditor.cc
@@ -272,6 +272,8 @@ void DebuggableEditor::drawContents(const Sizes& s, Debuggable& debuggable, unsi
 	}
 
 	auto handleInput = [&](unsigned addr, int width, auto formatData, auto parseData, int extraFlags = 0) {
+		// Setting End key to do nothing in TextInput##data
+		if (ImGui::IsKeyPressed(ImGuiKey_End)) return;
 		// Display text input on current byte
 		if (dataEditingTakeFocus) {
 			ImGui::SetKeyboardFocusHere();


m9710797:
I must say I'm surprised this works. My expectation was that shortcuts would work strictly hierarchical. And then that the InputText() widget (which uses keys like END) would take priority over the shortcuts in the higher hierarchical level (=the window itself).
I can't test myself right now. But I'll trust your patch. Thanks for trying this!

... Reading the patch in more detail ... I understand that this is only a proof-of-concept patch, but was it the intention to fully skip the call to `ImGui::inputText()' when END is pressed? I think this implies that the text is not drawn on the frame that the END key is pressed.

--

May I ask you to try to extend this PR to not only ignore the END key for text input, but actually make it do some useful stuff? For example:
* Home: move cursor to the start of the line.
* End: move cursor to the end of the line.
* Ctrl+Home: move cursor to the very start.
* Ctrl+End: move cursor to the very end.

I think it's OK if these shortcuts are fixed (not needed to be able to reconfigure them). Also because if it were possible to use something else than END, the problem of issue 1612 would re-appear.

--

Another idea (could be part of this PR, or it could be a separate one):
Maybe we should do something with PgUp and PgDown. I believe currently those control the vertical scroll bar. But maybe it's more logical that they also move the cursor one page up/down.

pvmm:
Trying to fix InputText externally proved neither easy nor consistent. Sometimes the repeated End key escapes the InputText focus and is captured by the scrollbar, since it's not a proper shortcut. This means that sometimes the End key moves to the very end of the scrollbar and sometimes it doesn't. I think this requires a substitute widget for InputText. So it will take more time to fix it properly.

MBilderbeek:
@pvmm the original issue has been closed. Do you want to keep this PR open? If not feel free to close it.

PR #1615, opened 2024-04-15
repeat property tooltip

pvmm:


m9710797:
Merged. Though we may remove this again later (if we decide that repeat should not be configurable).

PR #1620, opened 2024-05-12
FlashROM improvements and ASCII16-X mapper support

grauw:
I’ve been working on an ASCII16-based mapper with 8 MB of FlashROM called “ASCII16-X Mapper XL”. This contains an implementation of that mapper, as well as a large number of FlashROM emulation improvements done along the way.

Testing has been done on real hardware with the S29GL064S and M29W640GB chips, I also have a M29W800DB laying about but I still need to build the board. I can’t currently test with AM29F040 and AM29F016 so those are based on their datasheets and the comments in the openMSX source code. Anyway, if there are any open questions on hardware behaviour I can answer them with that in mind.

I still want to do some tests before pushing, so marked as draft.

I’d welcome a code review though.


grauw:
Test flash program + ROM: [ascii16x-test.zip](https://github.com/openMSX/openMSX/files/15287863/ascii16x-test.zip)

Invoke with e.g.:

    openmsx -machine Panasonic_FS-A1GT -ext ide -ext ASCII16-X_Mapper_XL

And then on the MSX-DOS2 command line:

    ascii16x /s2 /i /e /p /v FMPCM.ROM

To query the mapper in slot 2 for CFI information, do a continuity check, erase, program and verify. (Also works when changing the flash chip to M29W640GB btw.)

grauw:
I notice GCC and MSVC are having spurious complaints that Clang doesn’t fuss over:

GCC:

```
src/memory/AmdFlash.hh:231:50: error: either all initializer clauses should be designated or none of them should be
  231 |                 .autoSelect{ManufacturerID::AMD, .device{0xA4}, .extraCode = 0x01},
      |                                                  ^
src/memory/AmdFlash.hh:232:48: error: either all initializer clauses should be designated or none of them should be
  232 |                 .geometry{DeviceInterface::x8, .regions{{{8, 64 * 1024}}}},
      |                                                ^
```

These are `std::array`s where the initialiser list is supposed to default initialise the remaining fields. Not sure how to best address that, perhaps adding `= {}` would avoid the warning, or I could make the instantiation uglier by adding `{}` for the omitted values, or I could make a wrapper struct with a constructor with defaults.

MSVC:

```
     2>D:\a\openMSX\openMSX\src\memory\AmdFlash.hh(231,31): error C7556: cannot mix designated-initializers with non-designated-initializers [D:\a\openMSX\openMSX\build\msvc\openmsx.vcxproj]
         (compiling source file '../../src/memory/RomFactory.cc')
         
     2>D:\a\openMSX\openMSX\src\memory\AmdFlash.hh(232,30): error C7556: cannot mix designated-initializers with non-designated-initializers [D:\a\openMSX\openMSX\build\msvc\openmsx.vcxproj]
         (compiling source file '../../src/memory/RomFactory.cc')
         
     2>D:\a\openMSX\openMSX\src\memory\AmdFlash.hh(230,24): error C2737: 'openmsx::AmdFlashChip::AM29F040': constexpr object must be initialized [D:\a\openMSX\openMSX\build\msvc\openmsx.vcxproj]
         (compiling source file '../../src/memory/RomFactory.cc')
```

I wonder if the un-fun complaints about mixing initializers is an MSVC limitation, or maybe these errors (warnings?) could be silenced? Alternatively I could be explicit about the initializers, though it’d be uglier since it duplicates like `.manufacturer{ManufacturerID::STM}`, could shorten it with `using enum ManufacturerID` to `.manufacturer{STM}` I guess.

[Edit: SonarCloud says “mixture of designated and non-designated initializers in the same initializer list is a C99 extension”.]

Not sure why it complains about constexpr object initialisation, maybe as a result of the preceding errors.

MacOS build fails as usual for pull requests because the signing key is absent (seems unnecessary / fixable).

m9710797:
In c++ it's indeed not allowed to mix designated initializers with unnamed (order-based) initialization. Additionally in c++ (not in C), even when you use designated initializers, the fields must be enumerated in declaration-order (but you can skip fields). In C you can enumerate them in an arbitrary order (and even do crazy stuff like initialize the same field multiple times (only the last assignment will have an effect)).

So the fix is indeed to also use designated initializer syntax for the first field (even if that might be a bit redundant).

m9710797:
Hi Grauw,

Thanks a lot for this PR. I'll try to review later today. What kind of review do you expect (e.g. how detailed .. how close to 'finished' is the code)?

So far I only did a _very_ quick read of the commit messages. Seems like all but the last commit make improvements to the flash-emulation code, only the last commit adds a new mapper.

But I am curious about that last one. What's the context of that new mapper? Is it new hardware you're developing? For some new project? Why a new mapper type? E.g. wasn't a mapper like NEO-8 or NEO-16 suitable? (Not saying you should change anything, just curious).

grauw:
Hi Wouter, thanks,

> I'll try to review later today. What kind of review do you expect (e.g. how detailed .. how close to 'finished' is the code)?

In principle this code is finished, so a full review would be great. I may add additional functions later but in principle this is as much as needed for me to be able to develop and test the flashing and game software in openMSX.

> So far I only did a _very_ quick read of the commit messages. Seems like all but the last commit make improvements to the flash-emulation code, only the last commit adds a new mapper.

Correct!

> But I am curious about that last one. What's the context of that new mapper? Is it new hardware you're developing? For some new project? Why a new mapper type?

Yes it’s hardware I’ve developed since the beginning of this year. It will be used for Usas 2 (announced @ GOTO40) and possibly another project as well. The extra capacity is the headline feature, because we wanted to use a ROM larger than 4 MB. It will be open source hardware. I will announce the mapper on msx.org soon with a number of them available for developers, and people will be able to order production batches pre-assembled from PCBWay.



> E.g. wasn't a mapper like NEO-8 or NEO-16 suitable? (Not saying you should change anything, just curious).

I’ve been discussing with Aoineko whether NEO could be implemented with plain logic ICs, or what kind of modifications it would need for that, and I made a rough schematic outlining what would be possible and what not. However we concluded that NEO could only be practically implemented using FPGA, and it would be better to not limit it by what could be done with plain logic. I continued to develop this design as the ASCII-X mapper, through several prototypes, and now it’s about ready.

> Additionally in c++ (not in C), even when you use designated initializers, the fields must be enumerated in declaration-order (but you can skip fields). In C you can enumerate them in an arbitrary order.

Yeah the order thing is also a bit unfortunate since struct definition order for the most efficient packing may not be the most intuitive order to group things for reading, so there’s a bit of conflicting interest.

m9710797:
> I’ve been discussing with Aoineko whether NEO could be implemented with plain logic ICs, or what kind of modifications it would need for that, and I made a rough schematic outlining what would be possible and what not. However we concluded that NEO could only be practically implemented using FPGA, and it would be better to not limit it by what could be done with plain logic. I continued to develop this design as the ASCII-X mapper, through several prototypes, and now it’s about ready.

Interesting! After your discussions, do you happen to know the status of those NEO-x mappers?
IIRC we added those mappers in openMSX to allow for software to be developed before the hardware was finished. But does this make it unlikely that real NEO-x mappers will ever be produced?
 
> > Additionally in c++ (not in C), even when you use designated initializers, the fields must be enumerated in declaration-order (but you can skip fields). In C you can enumerate them in an arbitrary order.
> 
> Yeah the order thing is also a bit unfortunate since struct definition order for the most efficient packing may not be the most intuitive order to group things for reading, so there’s a bit of conflicting interest.

I agree with you. C++ does not have a good solution to decouple the human-logical-reading-order from the in-memory-order for struct-fields. On the other hand I do also understand the reasoning behind this limitation. The main difference between C and C++ is that C does not have constructors/destructors ...

In general there may be dependencies between the different fields of a struct: construction of one field may already make use of the previous (already constructed) other fields in the same struct. Also in general, for the same reason, it's a good idea to destruct the fields in reverse order of construction. 

A class/struct can have many different constructors. For a moment suppose it would be allowed to construct the fields in an arbitrary order in each constructor. An object does not remember via which constructor it was constructed, so the (only) destructor can't know this (arbitrary) construction order, and then it can't reverse this order to destruct the fields.  To get around this problem c++ always constructs the fields in declaration order, and destructs in reverse declaration order.

The above paragraphs talked about constructors, but in a way initializing a struct via designated initializers can have the same problems.  But I agree that in practice this is (more) rare. In many cases the fields in a struct are independent, and then it would be useful if they could be initialized in a arbitrary order.

aoineko-fr:
> Interesting! After your discussions, do you happen to know the status of those NEO-x mappers?
> IIRC we added those mappers in openMSX to allow for software to be developed before the hardware was finished. But does this make it unlikely that real NEO-x mappers will ever be produced?

Hello,

The status of the NEO mapper is still the same:
- For now, it's only real through the MSX Pico flash cartridge (NEO sample programs run on the real MSX through this device).
- There are a number of projects underway (hardware and software), but it's hard to be sure they'll see the light of day.

I think it's a pity that we haven't managed to create a format that suits everyone needs (there's even a Japanese guy working on a 3rd 16-bit mapper format), but Grauw's project has a better chance of success: he can create the hardware himself (whereas I can't), and USAS 2 has a much better chance of seeing the light of day than any of the current NEO mapper projects.

So if openMSX has to choose one, I wouldn't be shocked if it were ASCII16-X. :)

m9710797:
> ...
> The status of the NEO mapper is still the same:
> ...

Thanks for the update.

> So if openMSX has to choose one, I wouldn't be shocked if it were ASCII16-X. :)

It's not that we have to choose only one. If having the NEO-x mappers in openMSX helps then there's no problem to keep them (next to ASCII 16-X).

src/memory/AmdFlash.cc
```diff
@@ -52,6 +53,9 @@ AmdFlash::AmdFlash(const std::string& name, std::span<const SectorInfo> sectorIn
 void AmdFlash::init(const std::string& name, const DeviceConfig& config, Load load, const Rom* rom)
 {
 	assert(std::has_single_bit(size()));
+	assert(std::popcount(to_underlying(chip.autoSelect.manufacturer)) & 1); // check parity
+	assert(chip.autoSelect.device[0] == 0x7E ||
+		(chip.autoSelect.device[1] == 0x00 && chip.autoSelect.device[2] == 0x00)); // single or triple
```
m9710797: I'll try to review commit per commit, but sometimes also looking ahead to the final code. E.g. in this case I saw that several more asserts on `chip` were added.

Detail:
Would it be possible (and easy) to move these kind of checks to compile-time? Then these things are caught at compile-time instead of only at-run-time and then only when the chip is actually being used? Maybe add a `constexpr Chip` constructor??

src/memory/AmdFlash.cc
```diff
@@ -211,32 +215,73 @@ void AmdFlash::setState(State newState)
...
-			// TODO what is this? According to this it reads as '1'
-			//  http://www.msx.org/forumtopicl8329.html
-			return 1;
 		}
+	case 0x3:
+		// On AM29F040 it indicates "Autoselect Device Unprotect Code".
+		// Datasheet does not elaborate. Value is 0x01 according to
+		//  https://www.msx.org/forum/semi-msx-talk/emulation/matra-ink-emulated
+		//
+		// On M29W640G it indicates “Extended Block Verify Code”.
```
m9710797: Detail:
I'll only mention this once, but I saw it in several locations, including in commit messages.
Depending on in what editor/terminal I view your patch, these "double quotes" are shown either as
* Regular double quotes (as-if ASCII char 34).
* As question marks or as `<E2><80<9C>`.
* Or as the actual unicode character \u201C left double quotation mark .

I agree that typographically \u201C may be better. But IMHO for comments in source code or for commit message, this does not bring enough advantages over the disadvantages (that not all tools show this correctly). Why not use the regular ASCII double quote character?

So far I've seen this problem for:
* \u2019 right single quotation mark
* \u201C left double quotation mark
* \u201D right double quotation mark

m9710797 (commented):
The reviewed this first patch. Looks fine.
I also reviewed the behavioral changes in `AmdFlash.cc`, it all looks plausible, but I did not check in the datasheets that this is indeed the correct behavior. I'm going to trust you on this.

src/utils/power_of_two.hh
```diff
@@ -0,0 +1,28 @@
+#ifndef POWER_OF_TWO_HH
+#define POWER_OF_TWO_HH
+
+#include "narrow.hh"
+#include <bit>
+#include <cassert>
```
m9710797, line 6: This code did not compile for me.
I had to add `#include <cstdint>`.

src/utils/power_of_two.hh
```diff
@@ -0,0 +1,28 @@
...
+
+namespace openmsx {
+
+template<typename T>
```
m9710797: Just a suggestion: maybe instead of accepting any type `T` you could constrain this template to:
`template<std::unsigned_integral T>`
(and then also `#include <concepts>`).

src/utils/power_of_two.hh
```diff
@@ -0,0 +1,28 @@
...
+struct power_of_two {
+	constexpr power_of_two(T size)
+		: exponent(narrow_cast<uint8_t>(std::bit_width(size - 1)))
+	{
+		assert(std::has_single_bit(size));
+	};
+
+	[[nodiscard]] constexpr T operator()() const
```
m9710797: I don't know if this is a good idea or not, but have you considered instead of using `operator()`, to use an (implicit) conversion operator?
`[[nodiscard]] constexpr operator T() const { ...`

src/memory/AmdFlash.cc
```diff
@@ -26,7 +26,7 @@ AmdFlash::AmdFlash(const Rom& rom, Chip chip_,
 	: motherBoard(config.getMotherBoard())
 	, chip(chip_)
 	, sectorInfo(sectorInfo_)
-	, sz(sum(sectorInfo, &SectorInfo::size))
+	, sz(sum(chip.geometry.regions, [](Region r) { return r.count << r.size.exponent; }))
```
m9710797: Just an idea (I don't know yet if it's a good idea).
In `power_of_two` we could add something like this (not tested):
```
template<std::unsigned_integral T2> friend T2 operator*(T2 x, power_of_two<T> y) { return x << y; }
template<std::unsigned_integral T2> friend T2 operator*(power_of_two<T> x, T2 y) { return y << x; }
```
and then this usage can be simplified to just `return r.count * r.size`.

src/memory/AmdFlash.cc
```diff
@@ -38,7 +38,7 @@ AmdFlash::AmdFlash(const std::string& name, Chip chip_,
...
```
m9710797: This is a repeat of the above expression. Should we make this a member function of `Chip` instead?
(And then we can still cache the result of that expression here in `AmdFlash`).

src/memory/AmdFlash.cc
```diff
@@ -26,7 +26,7 @@ AmdFlash::AmdFlash(const Rom& rom, Chip chip_,
 	: motherBoard(config.getMotherBoard())
 	, chip(chip_)
 	, sectorInfo(sectorInfo_)
```
m9710797: If I read the code correctly, then `sectorInfo` is no longer needed after `init()`. Then maybe it can be passed as a parameter to `init()` and we don't need to store it as a member variable.

I think now `sectorInfo` only stores whether a sector is write protected or not. Maybe then we could give this parameter a more descriptive name?

src/memory/AmdFlash.cc
```diff
@@ -186,17 +191,15 @@ AmdFlash::~AmdFlash() = default;
...
+		address -= it->count << it->size.exponent;
+		sector += it->count;
 		++it;
-		assert(it != sectorInfo.end());
+		assert(it != chip.geometry.regions.end());
 	}
-	auto sectorSize = it->size;
-	auto offset = address;
-	return {sector, sectorSize, offset};
+	return {sector + (address >> it->size.exponent), it->size(), address & (it->size() - 1)};
```
m9710797: Similar as above. For `>>` we could add an `operator/` in `power_of_two`.

src/memory/AmdFlash.hh
```diff
@@ -33,12 +33,19 @@ public:
 		bool oddZero : 1 = false;      // odd bytes are zero, not mirrored
 		size_t readMask = 0x03;        // read address mask
 	};
+	struct Region {
+		size_t count = 0;
+		power_of_two<size_t> size = 1;
+	};
+	struct Geometry {
+		std::array<Region, 4> regions;
```
m9710797: Initially I was wondering what this '4' means. But it's arbitrary, right? So far for (M29W800DB) 4 is the maximum number of regions we need.

src/memory/AmdFlash.hh
```diff
@@ -147,21 +154,25 @@ namespace AmdFlashChip
...
 	// AMD AM29F016
 	static constexpr Chip AM29F016 = {
 		.autoSelect{ManufacturerID::AMD, .device{0xAD}},
+		.geometry{.regions{{{32, 64 * 1024}}}},
 	};
 
 	// Numonyx M29W800DB
 	static constexpr Chip M29W800DB = {
 		.autoSelect{ManufacturerID::STM, .device{0x5B}},
+		.geometry{.regions{{{1, 16 * 1024}, {2, 8 * 1024}, {1, 32 * 1024}, {15, 64 * 1024}}}},
```
m9710797: I'm not sure. Does the code become more clear with a `kB()` helper function, or with a `_kB` user defined literal?
Here's a small experiment, I don't know if it's worth the trouble:
https://godbolt.org/z/o8vdGad6G

src/memory/RomManbow2.cc
```diff
@@ -31,45 +31,29 @@ namespace openmsx {
...
 	using enum RomType;
 	case MANBOW2:
-	case MANBOW2_2:             return config1;
-	case HAMARAJANIGHT:         return config2;
-	case MEGAFLASHROMSCC:       return config3;
-	case RBSC_FLASH_KONAMI_SCC: return config4;
-	default: UNREACHABLE;
+	case MANBOW2_2:
+		{
+			static const std::array<const AmdFlash::SectorInfo, 8> sectorInfo =
```
m9710797: Detail: probably doesn't make a difference here. But maybe it's better to make this `constexpr` instead of `const`.

src/memory/RomManbow2.cc
```diff
@@ -31,45 +31,29 @@ namespace openmsx {
...
+				{{{true}, {true}, {true}, {true}, {true}, {true}, {true}, {false}}};
+			return sectorInfo;  // only last 64kB writable
+		}
+	case HAMARAJANIGHT:
+		{
+			static const std::array<const AmdFlash::SectorInfo, 8> sectorInfo =
+				{{{true}, {true}, {true}, {true}, {false}, {false}, {true}, {true}}};
```
m9710797: Now that I see it written like this ... is it worth keeping the `SectorInfo` struct? It only contains one `bool`. Maybe just have an array/span of `bool`. And if we rename `sectorInfo` to something like `sectorsWriteProtected`, that's then equally clear?

src/memory/RomMatraInk.cc
```diff
@@ -5,17 +5,9 @@
...
-	using Info = AmdFlash::SectorInfo;
-	std::array<Info, 2> result = {};
-	ranges::fill(result, Info{64 * 1024, false});
-	return result;
-}();
-
 RomMatraInk::RomMatraInk(const DeviceConfig& config, Rom&& rom_)
         : MSXRom(config, std::move(rom_))
-        , flash(rom, AmdFlashChip::AM29F040, sectorInfo,
+        , flash(rom, AmdFlashChip::AM29F040, {},
```
m9710797: This does introduce a functional difference. But possibly/probably this patch is an improvement (closer to real hardware)??

Before we only had 2 * 64kB. Now (via the AM29F040 chip) we have 8 * 64kB.

m9710797 (commented):
Thanks. This is a nice code cleanup.

m9710797 (commented):
Ok.
Weird that there's both a short and a long reset command.

src/memory/AmdFlash.cc
```diff
@@ -55,6 +55,7 @@ void AmdFlash::init(const std::string& name, const DeviceConfig& config, Load lo
 	assert(std::popcount(to_underlying(chip.autoSelect.manufacturer)) & 1); // check parity
 	assert(chip.autoSelect.device[0] == 0x7E ||
 		(chip.autoSelect.device[1] == 0x00 && chip.autoSelect.device[2] == 0x00)); // single or triple
+	assert(!chip.program.fastCommand || chip.program.pageSize.exponent > 0); // fast program requires page size > 1
```
m9710797: Detail:  why `pageSize.exponent > 0` instead of `pageSize() > 1`   (or without `()` if the implicit conversion operator idea works fine). Then you don't need the comment explaining what's going on.

m9710797 (commented):
Commit: AmdFlash: Improve fast program implementation. 

I knew I took some shortcuts here (implement all commands for all chips).
So thanks for cleaning this up!

src/memory/MegaFlashRomSCCPlus.cc
```diff
@@ -162,6 +162,23 @@ Main features:
...
+ In order to be compatible with OPFX which sends command sequences like
+ (0x555) = 0xAA, (0xAAA) = 0x55, whilst dual mode 8/16-bit chips like the
+ M29W800DB expect (0xAAA) = 0xAA, (0x555) = 0x55 when operating in byte mode,
+ CPU line A12 is connected to Flash line A-1.
+
+  CPU line | Flash line
+ ----------+---------------------------
+  A0-A11   |  A0-A11
+  A12      |  A-1
+  A13-A15  |  A12-A18 (through mapper)
```
m9710797: Is there an error in this comment? At least I don't immediately see how it maps on the explanation (the 2nd possibility) in the commit message. (The way I understand this comment is that you also loose half the capacity).

... Ah the code below (on line 256) makes it clear. I do think this comment is wrong.

m9710797 (commented):
MegaFlashRomSCCPlus: Implement address line swizzle. 

It's indeed unfortunate that read caching is no longer possible. But that's acceptable.

src/memory/AmdFlash.hh
```diff
@@ -33,11 +33,17 @@ public:
 		bool oddZero : 1 = false;      // odd bytes are zero, not mirrored
 		size_t readMask = 0x03;        // read address mask
 	};
+	enum class DeviceInterface : uint16_t {
+		x8 = 0x0000,
+		x16 = 0x0001,
+		x8x16 = 0x0002,
+	};
```
m9710797, line 49: Is my understand of this enum correct?
* x8: chip has an 8-bit data bus (addresses are interpreted as 8-bit addresses)
* x16: chip has a 16-bit data bus.  (addresses are interpreted as 16-bit addresses)
* x8x16: chip can operate in either 8 or 16 bit mode.  And depending on the mode the address should be interpreted differently

While reading the code I was expecting to see special cases for x16, but I saw none. Later I realized that's maybe because there currently are no chips yet that specify x16 (and maybe because of that handling of this case is missing?)

Do chips with x16 make sense in the context of MSX (an 8-bit system).  Maybe, but then you loose half the capacity of the ROM?

m9710797 (commented):
AmdFlash: Support CFI query command. 

Implementation looks sane, but I can't really review in detail (i haven't checked it against the datasheet).

I was only wondering: many of the values returned from `peekCFI()` are 8-bit constants while the method itself returns a 16-bit result. E.g. `supply.minVcc` is `uint8_t`. But then currently none of the current chip definitions specifies a non-default (non zero) value for these constants.

src/memory/AmdFlash.hh
```diff
@@ -198,10 +201,11 @@ private:
 	const std::span<const SectorInfo> sectorInfo;
 	const power_of_two<size_t> sz;
 
-	static constexpr unsigned MAX_CMD_SIZE = 8;
+	static constexpr unsigned MAX_CMD_SIZE = 37;
 	std::array<AmdCmd, MAX_CMD_SIZE> cmd;
```
m9710797: Hmm, I'm not sure what this does with savestates. More specifically I mean loading an old savestate which has `cmd` an array of size 8, in a newer openMSX version which expects `cmd` to be an array of size 37.

Probably this should be handled via bw-compat code in `serialize()`.

src/memory/AmdFlash.cc
```diff
@@ -659,7 +724,8 @@ void AmdFlash::serialize(Archive& ar, unsigned version)
 	ar.serialize("ram",    *ram,
 	             "cmd",    cmd,
 	             "cmdIdx", cmdIdx,
-	             "state",  state);
+	             "state",  state,
+	             "status", status);
```
m9710797: And this should also be handled with bw-compat code.
Older savestates won't have `status` yet.

Search for `versionAtLeast` for examples of how to do this.

m9710797 (commented):
AmdFlash: Support buffered program command. 

I've only partially reviewed this commit yet. But I've run out of time. I'll continue reviewing tomorrow (or possibly in a few hours).

src/memory/AmdFlash.cc
```diff
@@ -613,6 +636,47 @@ bool AmdFlash::checkCommandQuadrupleByteProgram()
...
+					const unsigned confirmIndex = 4 + cmd[3].value + 1;
+					if (cmdIdx >= 5 && cmdIdx <= confirmIndex) {
+						status = (status & 0x7F) | (~cmd[cmdIdx - 1].value & 0x80);
+					}
+					if (cmdIdx <= confirmIndex) return true;
+					if (cmdIdx == confirmIndex + 1 && cmd[confirmIndex].value == 0x29 && getSectorInfo(cmd[confirmIndex].addr).sector == sector) {
+						if (isSectorWritable(sector)) {
+							// TODO de-duplicate same-address writes to the last one
+							for (auto i : xrange(size_t(4), confirmIndex)) {
+								auto [sector_, sectorSize_, offset_] = getSectorInfo(cmd[i].addr);
```
m9710797: I understand why (to suppress a warning about shadowing), but I don't like the names `sector, ...` on line 646 and the names `sector_, ...` on this line. These names are too similar, too easy to accidentally forget the `_` suffix and overlook it during review.

Maybe name them `sector1, ...` and `sector2, ...`?  (Too force making an explicit choice).
An alternative is to _not_ use structured bindings, so something like `auto info1 = getSectorInfo(..); ... use(info1.sector) ...`.

src/memory/AmdFlash.hh
```diff
@@ -198,10 +201,11 @@ private:
 	const std::span<const SectorInfo> sectorInfo;
 	const power_of_two<size_t> sz;
 
-	static constexpr unsigned MAX_CMD_SIZE = 8;
+	static constexpr unsigned MAX_CMD_SIZE = 37;
```
m9710797: It took me a while to figure out where this '37' comes from. I only figured it out after reading the `assert(... pageSize + 5)`. Maybe it would help to write this as `5 + 32` (and `5 + 256` later)? Adding a command like `// largest command is BufferProgram` could also help.

m9710797 (commented):
AmdFlash: Support buffered program command.   (part 2)

I can't claim to fully understand `checkCommandBufferProgram()`. But at least the structure of the code looks sane to me.

However the `serialize()` code (previous review) is a bug and should be fixed.

m9710797 (commented):
AmdFlash: Initialise status register on hard reset. 

Ok.

src/memory/AmdFlash.cc
```diff
@@ -251,7 +257,7 @@ uint8_t AmdFlash::peek(size_t address) const
 		} else {
 			return narrow_cast<uint8_t>(peekCFI(address));
 		}
-	} else if (state == State::PRGERR) {
+	} else if (state == State::STATUS || state == State::PRGERR) {
```
m9710797: This is perfectly fine. No need to change it. But if you prefer (personal preference) this could be an alternative:
`} else if (state == one_of(State::STATUS, State::PRGERR)) {`

src/memory/AmdFlash.cc
```diff
@@ -555,6 +573,35 @@ bool AmdFlash::checkCommandLongReset()
 	return false;
 }
 
+bool AmdFlash::checkCommandStatusRead()
+{
+	static constexpr std::array<uint8_t, 3> cmdSeq = {0x70};
```
m9710797: I only noticed this now, maybe I overlooked this in previous patches:
The size of the array is specified as 3, but the initializer list only contains 1 element. Is that intentional? I guess not? Or if it is, then it's probably clearer to explicitly write `0x70, 0x00, 0x00`.

Same remark in the next method.

src/memory/AmdFlash.cc
```diff
@@ -555,6 +573,35 @@ bool AmdFlash::checkCommandLongReset()
...
+		softReset();
+	}
+	return false;
+}
+
+bool AmdFlash::checkCommandContinuityCheck()
+{
+	if (chip.misc.continuityCommand && cmd[0].addr == 0x5554AB && cmd[0].value == 0xFF) {
+		if (cmdIdx < 2) return true;
+		if (cmdIdx == 2 && cmd[1].addr == 0x2AAB54 && cmd[1].value == 0x00) {
```
m9710797: Detail. I only notice this now. I'm not sure, but is it clearer to write this like:
`if (cmdIdx == 2 && cmd[1] == AmdCmd{0x2AAB54, 0x00}) {`
But maybe `AmdCmd` isn't the best name. Maybe something like `AddrAndValue`?

2 lines above something similar could be done. And possibly in various other places in the code (I didn't check).

src/memory/AmdFlash.cc
```diff
@@ -214,6 +215,11 @@ void AmdFlash::softReset()
 {
 	cmdIdx = 0;
 	setState(State::IDLE);
+	clearStatus();
+}
+
+void AmdFlash::clearStatus()
+{
 	status &= 0xC4;
```
m9710797: In general I agree with small helper functions. Even if only called once.
But maybe this is a bit extreme ;-)     (a comment `// clear status` could also work)
But if you like it, that's fine for me.

src/memory/RomAscii16X.cc
```diff
@@ -0,0 +1,135 @@
...
+	: MSXRom(config, std::move(rom_))
+	, debuggable(getMotherBoard(), getName())
+	, flash(rom, AmdFlashChip::S29GL064S70TFI040, {}, config)
+{
+	powerUp(getCurrentTime());
+}
+
+RomAscii16X::~RomAscii16X()
+{
+}
```
m9710797: In some cases it's useful to have an empty destructor (e.g. when you have `unique_ptr` member variables). But that's not the case here, I think. Better remove this destructor.

Minor:  maybe not applicable here, but in general there are some technical advantages in letting the compiler generate the destructor.

src/memory/RomAscii16X.cc
```diff
@@ -0,0 +1,135 @@
...
+//  first  16kB: 0x6000 - 0x6FFF, 0xA000 - 0xAFFF, 0xE000 - 0xEFFF, 0x2000 - 0x2FFF
+//  second 16kB: 0x7000 - 0x7FFF, 0xB000 - 0xBFFF, 0xF000 - 0xFFFF, 0x3000 - 0x3FFF
+//
+// 12-bit bank number is composed by address bits 8-11 and the data bits.
+//
+// Backed by FlashROM.
+
+#include "RomAscii16X.hh"
+#include "DummyAY8910Periphery.hh"
+#include "MSXCPUInterface.hh"
```
m9710797: I think these 2 #includes are left-overs from your start version?
The `#include <array>` below is not wrong but redundant (.hh file already includes it).

src/memory/RomAscii16X.cc
```diff
@@ -0,0 +1,135 @@
...
+
+namespace openmsx {
+
+RomAscii16X::RomAscii16X(const DeviceConfig& config)
+	: MSXRom(config, Rom(std::string(config.getAttributeValue("id")), "ASCII16-X", config))
+	, debuggable(getMotherBoard(), getName())
+	, flash(rom, AmdFlashChip::S29GL064S70TFI040, {}, config)
+{
+	powerUp(getCurrentTime());
+}
```
m9710797, line 25: I wonder if it's possible to use a delegating constructor here?
Something like (not tested):
```
RomAscii16X::RomAscii16X(const DeviceConfig& config)
	: RomAscii16X(config, Rom(std::string(config.getAttributeValue("id")), "ASCII16-X", config)) {}
```

src/memory/RomAscii16X.cc
```diff
@@ -0,0 +1,135 @@
...
+
+RomAscii16X::~RomAscii16X()
+{
+}
+
+void RomAscii16X::powerUp(EmuTime::param time)
+{
+	reset(time);
+}
```
m9710797, line 32: The base class already implements `powerUp()` in exactly this way. This is not wrong, but redundant.
(Very few devices need an explicit `powerUp()` or `powerDown()` implementation.)

src/memory/RomAscii16X.cc
```diff
@@ -0,0 +1,135 @@
...
+	ranges::iota(bankRegs, word(0));
+
+	flash.reset();
+
+	invalidateDeviceRCache(); // flush all to be sure
+}
+
+unsigned RomAscii16X::getFlashAddr(word addr) const
+{
+	word bank = bankRegs[((addr >> 14) & 1) ^ 1];
```
m9710797, line 45: I was wondering if it was a good idea to omit that `^ 1` (so basically swap the order of the 2 elements in the array).
It's a tiny optimization, but probably not worth it for code clarity. (And then in the Debuggable below you do need to add it again).

src/memory/RomAscii16X.cc
```diff
@@ -0,0 +1,135 @@
...
+
+void RomAscii16X::writeMem(word addr, byte value, EmuTime::param /* time */)
+{
+	flash.write(getFlashAddr(addr), value);
+
+	if ((addr & 0x3FFF) >= 0x2000) {
+		bankRegs[(addr >> 12) & 1] = (addr & 0x0F00) | value;
+	}
+
+	invalidateDeviceRCache();
```
m9710797: This is fine. But if you want you could only invalidate the region that's actually swapped.
Ah, or maybe you do need to invalidate everything because writing to flash could influence how reads behave everywhere.

src/memory/RomAscii16X.cc
```diff
@@ -0,0 +1,135 @@
...
+}
+
+byte* RomAscii16X::getWriteCacheLine(word /* addr */)
+{
+	return nullptr; // not cacheable
+}
```
m9710797, line 79: This is the default implementation in the base class.
However I think it's good to make this very explicit. So I'm fine with reimplementing this here.

src/memory/RomAscii16X.cc
```diff
@@ -0,0 +1,135 @@
...
+//
+// Banks:
+//  first  16kB: 0x4000 - 0x7FFF and 0xC000 - 0xFFFF
+//  second 16kB: 0x8000 - 0xBFFF and 0x0000 - 0x3FFF
+//
+// Address to change banks:
+//  first  16kB: 0x6000 - 0x6FFF, 0xA000 - 0xAFFF, 0xE000 - 0xEFFF, 0x2000 - 0x2FFF
+//  second 16kB: 0x7000 - 0x7FFF, 0xB000 - 0xBFFF, 0xF000 - 0xFFFF, 0x3000 - 0x3FFF
+//
+// 12-bit bank number is composed by address bits 8-11 and the data bits.
```
m9710797, line 11: Not a comment on this PR, but a comment on the mapper design.

Using 4 address bits to "extend" the data-bus to 12-bits is a bit unusual. And unexpected (to me). From a hardware point of view I understand how this could indeed lead to a simpler design. From a software point of view I have mixed feelings:
* Switching to a fixed bank is only two Z80 instructions: `ld a,<n> ; ld (<nn>),a` (or equivalent).
* But switching to a dynamic bank (e.g. a calculated bank number for a large lookup table) is more complex. Maybe something like this:
```
Input: 12-bit bank number in HL
    ld a,h
    or 0x60   ; 0x70 for 2nd bank
    ld h,a
    ld (hl),l
```
So actually not too bad, and maybe that `or 0x60` could be pre-calculated.

Is it a hard requirement that the switch regions are compatible with the ASCII16 mapper? I often like it that the switch region for a bank is located in the bank itself. That allows to work with the mapper when it's only partially selected in the MSX slots.

Anyway, just my first impression from reading this documentation.

m9710797 (commented):
This finishes the review.

src/memory/RomManbow2.cc
grauw: Apparently [it does](https://stackoverflow.com/a/13867690). I’ll change it.

src/memory/RomMatraInk.cc
grauw: Nice spot, I realised this when I implemented it but then forgot about it. I just double-checked, and functionally there’s no difference since it only peeks, reads and writes with 16-bit addresses.

It indeed now matches the real hardware, also it is practical implementation-wise since otherwise there’d need to be a special “fake” chip definition.

src/memory/AmdFlash.cc
grauw: Ah, a matter of habit, to me typing these is muscle memory at this point. I’ll change the CPP comments for sure, but hrmrm if I change the commit messages today some’ll sneak back in later I fear…

src/utils/power_of_two.hh
grauw: Oh, I didn’t know that was possible. I’ll change that for sure.

src/memory/AmdFlash.cc
grauw: I considered this, but thought I was maybe being too fancy and it was better to be explicit, so that you can see it’s a quick operation at the place where it’s used.

However now that you bring it up as well I will add these methods after all, because it _is_ pretty neat to do this and get an automatic optimisation even if you don’t think about it, and is more intuitive to read.

src/memory/AmdFlash.cc
grauw: Will do, also operator% for the `& size-1`.

src/utils/power_of_two.hh
grauw: Will do.

src/utils/power_of_two.hh
grauw: Will add.

src/memory/AmdFlash.hh
grauw: In the CFI query format there are up to four region entries, so this maximum is assured.

src/memory/AmdFlash.hh
grauw: Of the two I prefer the suffix. Or `_k` since the suffix can apply to more than just bytes. Or actually `_ki` since they’re not thousands.

But honestly not sure if it’s worth it, maybe too fancy / specific. Either way that’s more of an openMSX-wide change I think, should also consider in the context of what it does to the code base as a whole, whether it’s friendly or hostile to new contributors, etc.

grauw:
> Ok. Weird that there's both a short and a long reset command.

Perhaps very early flash chips only had the long version, and later the short one was introduced but the long one was kept for compatibility.

Practically speaking though, the S29GL064S only accepts the long reset command to exit out of the PRGERR state.

src/memory/AmdFlash.cc
grauw: Will do.

src/memory/MegaFlashRomSCCPlus.cc
grauw: I get where the confusion is coming from; on x8x16 devices in their native x16 mode it uses address lines A0-A18, however in their x8 mode it additionally uses D15 as A-1 (as-in bit minus one).

Anyway, I’ll just renumber the flash lines column to be more intuitive to the casual reader. This detail is not so important.

src/memory/AmdFlash.hh
grauw: At some point I was considering adding `read16()`, `peek16()`, `write16()` and similar methods. Indeed x16 chips could be used on MSX if you add a multiplexer between the flash ROM and the bus, controlled by A0. The x8x16 chips also internally are more x16 than they are x8, so implementation-wise it could potentially make the code a bit more internally consistent / logical.

However since there is no current hardware using x16 chips, I decided to not add extra complexity for that at the moment. This x16 entry here is a leftover of that consideration, I’ll remove it.

grauw:
> I was only wondering: many of the values returned from `peekCFI()` are 8-bit constants while the method itself returns a 16-bit result. E.g. supply.minVcc is uint8_t. But then currently none of the current chip definitions specifies a non-default (non zero) value for these constants.

This is because internally the x8x16 devices are 16-bits, and this affects the values read from the CFI query on the odd addresses. According to the CFI specification they are supposed to read the same as the even addresses, however both the M29W640GB and the S29GL064S instead return the high byte of the 16-bit values.

All the CFI values themselves are defined as 8-bit with the upper bits being zero, however the undefined values return 0xFFFF, and the (undocumented) autoselect data visible in the region < 0x10 also includes the MSBs from their 16-bit values.

In the end it was more logical to have `peekAutoSelect()` and `peekCFI()` take a “native mode” address, meaning word address with bit 0 shifted out on x8x16 devices, and return the 16-bit result, with bits 8-15 being irrelevant for x8 devices. This also relates to what I mentioned earlier about considering adding `peek16()`, etc. methods.

src/memory/AmdFlash.cc
grauw: Ok, I’ll make it nicer.

src/memory/AmdFlash.hh
grauw: Will do.

src/memory/AmdFlash.cc
grauw: Will do.

src/memory/AmdFlash.cc
grauw: Whoops, no definitely not intentional, a copy/pasta.

Now it makes more sense why GCC would [complain about this](https://github.com/openMSX/openMSX/pull/1620#issuecomment-2106403301).

Will fix.

src/memory/AmdFlash.cc
grauw: Yes, nice one.

src/memory/AmdFlash.cc
grauw: Consider it gone.

src/memory/RomAscii16X.cc
grauw: Yes I started from one of the Pazos Flash cartridges, will remove ’em.

src/memory/AmdFlash.cc
m9710797: I did a small experiment: https://godbolt.org/z/TMdchqvrY
When we add the (implicit) conversion operator, there's no need to add our own _fancy_ `operator* / %`, the compiler already performs these optimizations for us.

(Also on modern CPUs multiplication is not that expensive anymore.)

src/memory/RomAscii16X.cc
grauw: I’ll keep it the way it is.

src/memory/AmdFlash.hh
grauw: Replaced with hex values `0x2000`, `0x4000`, `0x8000` and `0x10000`. To me that’s easier to read without getting too fancy.

src/memory/AmdFlash.hh
grauw: I’m changing it to `static_vector<Region, 4> regions;`, this will improve the logic in `peekCFI()` and should get rid of GCC’s “either all initializer clauses should be designated or none of them should be” complaints.

grauw: And same for `AutoSelect::device`, changing it to `static_vector<uint8_t, 2> device;`.

src/memory/AmdFlash.cc
grauw: Ah, I didn’t think too much of it, thought it would be tolerant of added / removed properties like I believe Unreal Engine’s serialisation is. Will change.

src/memory/AmdFlash.hh
grauw: I’m planning to change it to:

    static_vector<AddressValue, 5 + 32> cmd;  // longest command is BufferProgram

After I make `static_vector` work with the serialisation, going forward it should no longer require a new serialisation version when the capacity is increased.

Also we won’t need to track a separate `cmdIdx` (which was a misnomer for `cmdSize`).

src/memory/AmdFlash.cc
grauw: If I add a constructor I can’t use designated initializers anymore. I think I’ll wrap it in a `constexpr Chip validate(Chip&& chip)` function or something like that.

src/memory/AmdFlash.hh
m9710797: To serialize a `static_vector` you probably need to create a specialization `serialize_as_collection<static_vector<T, N>>` in `serialize_stl.hh`.

src/memory/RomAscii16X.cc
grauw: > Using 4 address bits to "extend" the data-bus to 12-bits is a bit unusual. And unexpected (to me). From a hardware point of view I understand how this could indeed lead to a simpler design. From a software point of view I have mixed feelings: [...] `ld (hl),l` [...] So actually not too bad, and maybe that or 0x60 could be pre-calculated.

Yes, indeed if you want to do a 16-bit bank switch you’d do `ld (hl),l` (that’s why it uses address bits 8-11), which is only 7 cycles for a 16-bit switch. You’ll typically want to embed the register address in the bank number.

Another approach (which Usas2 will take probably) is to divide the game data into two 4 MB sections, e.g. level data in the lower half and music data in the upper half. Then continue to use 8-bit bank numbers within them and you’d write to `0x6000` and `0x6100` depending on the type of data.

From a hardware perspective I receive 24 bits (address + data) in a single operation, so I can use them all in one go rather than having two operations. Two operations means they need to be stored in separate registers. This would require two additional 74LVC823 chips (relatively expensive) as well as an extra decoder chip.

So the mapper’s design involves both hardware considerations in terms of cost (minimal chip count, 2-layer board, and components on a single side), as well as software considerations (being fast and versatile).

> Is it a hard requirement that the switch regions are compatible with the ASCII16 mapper?

Not per se, I started out with ASCII16 because it is the largest capacity mapper, also fairly simple, and I thought it would add some value if it were compatible. Both in terms of making it easier to use or adapt existing developer tooling and code, as well as to make it useful as a platform to run or release ASCII16 games on.

Usas2 code currectly built for ASCII16 runs on it verbatim, so does Aleste, Penguin Wars, Pointless Fighting, etc.

Also note that ASCII mappers are flash-able because you can write to memory without affecting its bank register. Konami mapper bank registers overlap with the memory they switch, so they can’t be flashed normally.

> I often like it that the switch region for a bank is located in the bank itself. That allows to work with the mapper when it's only partially selected in the MSX slots.

The bank registers are mirrored, so you can access them from any page.

The only downside of this mirroring is that they’re also mirrored to page 3, which means the BIOS’s expanded slot detection messes up the initial bank of page 2 (same issue for NEO btw). Unfortunately not something I can solve without adding extra chips or removing the mirroring entirely. It bugs me from a theoretical point of view that it’s no longer _fully_ ASCII16 compatible, but in practice there’s no real problem, even for existing ASCII16 games that I tried, so I kept it as-is.

src/memory/RomAscii16X.cc
grauw: `AmdFlash::setState(…)` already does:

    motherBoard.getCPU().invalidateAllSlotsRWCache(0x0000, 0x10000);

So I guess this `invalidateDeviceRCache();` here can be made more specific.

grauw:
There’s one more thing I’d like to do…

When used as a ROM mapper with initial content (via `-carta x.rom -romtype ASCII16-X`), once the SRAM is created it doen not matter if the ROM changes next time openMSX is opened. This is not ideal, since e.g. during development you wouldn’t want to have to delete the SRAM file every time you got a new build, and also for players, if there’s a new version of the ROM out it needs additional instructions. Normally most sectors aren’t written to during testing, only save games are.

So it would be good if the SRAM only stored the modifications. In case of Manbow2 it happens to have the non-savegame sectors hardware write protected, so conveniently `writeProtectSectors` could be utilised to sparsely store only the writable savegame sectors as SRAM. But this is not generally applicable to ASCII16-X (or KonamiUltimateCollection, etc.).

So I want to keep track of modified sectors by having an array of `Sector { bool modified; }` on `AmdFlash`, which is `false` initially and gets set to `true` once the sector is erased or programmed. When `false`, it will read from the ROM, when `true` it will read from the SRAM. So far so good.

My biggest question though is how to best implement the persistence part. The SRAM class has the ability to have a string header (not currently used anywhere I think), I could use this right now to store that modified-array. The extensibility of this scheme is limited though, since `SRAM::load()` bails out if the header is not equal in size, so if I’d want to store more later, I can’t extend the storage without making the SRAM files incompatible.

I could change the SRAM code to have a more flexible header option, perhaps using the existing serialisation framework which has built in versioning. For back compat, possibly putting it after the data part instead of in front (since the data size should be static), then I could check its presence based on the file size. Or have it support a variable-length ASCIIZ header without bailing out. But maybe also worth considering something even more complex where it doesn’t even store the sectors which haven’t been modified (though I’m not leaning towards that).

Do you have any thoughts / ideas? The simpler the better…

m9710797:
The SRAM header functionality is used in e.g. src/memory/MSXPac.cc to have the .sram files in a compatible(?) format with on-disk MSX-Pac backup files. But I'm not sure if this was really a useful feature.

Though I don't think reusing this header functionality is the way to go. My current thinking is:
* Always store the full (writable part) of the flash rom in the .sram file.
* Have a _separate_ file next to it to indicate which regions of this .sram file are relevant.

That new file could maybe be an ASCII file with lines like:
```
modified-region: 0x30000-0x3ffff
modified-region: 0x60000-0x9ffff
```
I have NOT fully worked out this idea. Here are some open questions:
* What should the name of this separate file be? Same base-name, but different extension? What extension?
* What if that separate file is missing? For bw-compat interpret this as if the whole content is modified?
* We could say that the irrelevant parts in the .sram file have unspecified content. That would allow to use unix sparse files (reading these parts returns 0, but they don't take up any disk space). I don't know if windows supports sparse files. And maybe it's an irrelevant (disk space) optimization.
* Maybe ".sram" is not the best file extension for flash-content. We could pick a different extension (maybe .flash? and maybe .flashinfo for the meta-info?). That would also allow to differentiate between the old and the new mechanism.
* If we go for a new format (I mean with a new extension, easily distinguishable from the old format), and if that would make the implementation simpler, then we could consider also storing the read-only parts in the .flash file (possibly as zeros, to allow for sparse-file-optimization), and possibly annotated with a line like `read-only-region: 0x00000-0x7ffff` in the .flashinfo file. At least that could simplify all the file-offset-calculations.

EDIT:
* I have not thought much about the format of the .flashinfo file. I think using ASCII files is flexible enough to deal with future extensions. But I'm not sure.

grauw:
> The SRAM header functionality is used in e.g. src/memory/MSXPac.cc to have the .sram files in a compatible(?) format with on-disk MSX-Pac backup files. But I'm not sure if this was really a useful feature.

Ah I see, just an ID string then, that explains why the header works the way it does.

> Though I don't think reusing this header functionality is the way to go. My current thinking is:
> 
> * Always store the full (writable part) of the flash rom in the .sram file.
> * Have a _separate_ file next to it to indicate which regions of this .sram file are relevant.

Also a good option. There is some merit to the `.sram` file be a plain memory dump and keeping metadata separate. And the persistent directory structure allows for it.

> That new file could maybe be an ASCII file with lines like:
> 
> ```
> modified-region: 0x30000-0x3ffff
> modified-region: 0x60000-0x9ffff
> ```
> [...]
> * I have not thought much about the format of the .flashinfo file. I think using ASCII files is flexible enough to deal with future extensions. But I'm not sure.

What about relying on the existing serialisation mechanism? That seems quicker to implement and easier to code for… I don’t think there’s a binary on-disk format, so that would mean XML, as used by savestates and replays.

> I have NOT fully worked out this idea. Here are some open questions:
> 
> * What should the name of this separate file be? Same base-name, but different extension? What extension?

Same base name sounds fair, extension can be bikeshed over in review :).

> * What if that separate file is missing? For bw-compat interpret this as if the whole content is modified?

That makes the most sense to me. Don’t want to lose any existing savegames that people may have.

> * We could say that the irrelevant parts in the .sram file have unspecified content. That would allow to use unix sparse files (reading these parts returns 0, but they don't take up any disk space). I don't know if windows supports sparse files. And maybe it's an irrelevant (disk space) optimization.

Aside from OS support, I think this also relies on the file system capabilities? An alternative is to store an offset in the metadata file, and have the memory file be a noncontiguous block of sectors that gets appended to when a new sector gets modified and needs to be persisted.

It’s not a feature I particularly care for at the moment though. Both of these options could be added later in a compatible way I think.

> * Maybe ".sram" is not the best file extension for flash-content. We could pick a different extension (maybe .flash? and maybe .flashinfo for the meta-info?). That would also allow to differentiate between the old and the new mechanism.

In principle I don’t care much about whether `.sram` accurately describes the media type, it’s a legacy choice that doesn’t necessarily need to change just because it’s awkward. If anything, I would then name it something media type agnostic like `.memory` or something; it’s just a memory data blob. I would only rename it if the format actually changes and we need to be able to differentiate.

> * If we go for a new format (I mean with a new extension, easily distinguishable from the old format), and if that would make the implementation simpler, then we could consider also storing the read-only parts in the .flash file (possibly as zeros, to allow for sparse-file-optimization), and possibly annotated with a line like `read-only-region: 0x00000-0x7ffff` in the .flashinfo file. At least that could simplify all the file-offset-calculations.

Yeah, I think the new thing should replace the current sparse persistence mechanism based on `writeProtectSectors`. There’s no need to have two different mechanisms for the same thing. And although write protecting sectors is not typically accessible to the CPU, we could in theory make an interface similar to diskmanipulator but for flashROMs which would allow you to toggle these bits like an EPROM programmer would be able to, so hardcoding it isn’t ideal.

src/memory/AmdFlash.cc
grauw: I will cache `size` on `Geometry`, together with `sectorCount`.

grauw:
Updated, with review comments addressed.

I’ll do the sparse persistence thing in a separate PR later, to not conflate this one.

src/memory/AmdFlash.hh
```diff
@@ -107,17 +124,39 @@ public:
...
+			program.validate();
+			cfi.validate();
+			misc.validate();
+		}
+	};
+	struct ValidatedChip {
+		constexpr ValidatedChip(Chip&& chip_) : chip(std::move(chip_)) {
+			chip.validate();
+		}
+		const Chip chip;
```
m9710797, line 168: Detail, background info, no need to change anything.
I understand why you want to store `chip` as a `const` member: after validation it may not be changed anymore.

On the other hand some c++ coding guidelines recommend to not use const member variables (*1). The reason for this is that it disables move-operations. E.g. see this experiment: https://godbolt.org/z/zoxfrPoef

In case of `{Validated,}Chip` this isn't relevant (moving and copying is the same). But I noticed anyway because you explicitly *move* `Chip` into `ValidatedChip`, (and then later *copy* it out into `AmdFlash`).


(*1) We don't always follow this guideline in openMSX. Partly because I only recently (in the last few years) learned about this. And partly because the tradeoff "allowing move" vs "more descriptive code" tilts towards the latter.

src/memory/AmdFlash.hh
```diff
@@ -225,35 +266,35 @@ SERIALIZE_CLASS_VERSION(AmdFlash, 3);
 
 namespace AmdFlashChip
 {
-	using Chip = AmdFlash::Chip;
+	using Chip = AmdFlash::ValidatedChip;
```
m9710797: This confused me. Not this line itself, but the lines below which use `Chip`, which is then actually a `ValidatedChip`. Maybe change this to `using ValidatedChip = ...`?

m9710797 (commented):
I reviewed (only) your last 3 commits.
Only two minor remarks on the last commit   (AmdFlash: Compile-time Chip validation).

src/memory/AmdFlash.hh
grauw: Given the typical compile-time context it doesn’t really matter indeed. I’ll change the constructor to take `const Chip& chip_` though, so that it doesn’t give the false impression that it’s being moved.

PR #1622, opened 2024-05-15
Static vector

grauw:
Hi Wouter,

Here’s a spin-off from pull request #1620 where I expand the implementation of static_vector, add serialization support and use it to store the AmdFlash command list. Also already adds a dummy “status” field for use later.

### DIFF
--- a/src/memory/AmdFlash.cc
+++ b/src/memory/AmdFlash.cc
@@ -11,6 +11,7 @@
 #include "one_of.hh"
 #include "ranges.hh"
 #include "serialize.hh"
+#include "serialize_stl.hh"
 #include "xrange.hh"
 #include <bit>
 #include <cassert>
@@ -198,7 +199,7 @@ AmdFlash::GetSectorInfoResult AmdFlash::getSectorInfo(size_t address) const
 
 void AmdFlash::reset()
 {
-	cmdIdx = 0;
+	cmd.clear();
 	setState(State::IDLE);
 }
 
@@ -264,10 +265,8 @@ const uint8_t* AmdFlash::getReadCacheLine(size_t address) const
 
 void AmdFlash::write(size_t address, uint8_t value)
 {
-	assert(cmdIdx < MAX_CMD_SIZE);
-	cmd[cmdIdx].addr = address;
-	cmd[cmdIdx].value = value;
-	++cmdIdx;
+	cmd.push_back({address, value});
+
 	if (checkCommandManufacturer() ||
 	    checkCommandEraseSector() ||
 	    checkCommandProgram() ||
@@ -298,7 +297,7 @@ bool AmdFlash::checkCommandEraseSector()
 {
 	static constexpr std::array<uint8_t, 5> cmdSeq = {0xaa, 0x55, 0x80, 0xaa, 0x55};
 	if (partialMatch(cmdSeq)) {
-		if (cmdIdx < 6) return true;
+		if (cmd.size() < 6) return true;
 		if (cmd[5].value == 0x30) {
 			auto addr = cmd[5].addr;
 			auto [sector, sectorSize, offset] = getSectorInfo(addr);
@@ -315,7 +314,7 @@ bool AmdFlash::checkCommandEraseChip()
 {
 	static constexpr std::array<uint8_t, 5> cmdSeq = {0xaa, 0x55, 0x80, 0xaa, 0x55};
 	if (partialMatch(cmdSeq)) {
-		if (cmdIdx < 6) return true;
+		if (cmd.size() < 6) return true;
 		if (cmd[5].value == 0x10) {
 			if (ram) ram->memset(0, 0xff, ram->size());
 		}
@@ -326,7 +325,7 @@ bool AmdFlash::checkCommandEraseChip()
 bool AmdFlash::checkCommandProgramHelper(size_t numBytes, std::span<const uint8_t> cmdSeq)
 {
 	if (partialMatch(cmdSeq)) {
-		if (cmdIdx < (cmdSeq.size() + numBytes)) return true;
+		if (cmd.size() < (cmdSeq.size() + numBytes)) return true;
 		for (auto i : xrange(cmdSeq.size(), cmdSeq.size() + numBytes)) {
 			auto addr = cmd[i].addr;
 			auto [sector, sectorSize, offset] = getSectorInfo(addr);
@@ -361,10 +360,10 @@ bool AmdFlash::checkCommandManufacturer()
 {
 	static constexpr std::array<uint8_t, 3> cmdSeq = {0xaa, 0x55, 0x90};
 	if (partialMatch(cmdSeq)) {
-		if (cmdIdx == 3) {
+		if (cmd.size() == 3) {
 			setState(State::IDENT);
 		}
-		if (cmdIdx < 4) return true;
+		if (cmd.size() < 4) return true;
 	}
 	return false;
 }
@@ -376,7 +375,7 @@ bool AmdFlash::partialMatch(std::span<const uint8_t> dataSeq) const
 	(void)addrSeq; (void)cmdAddr; // suppress (invalid) gcc warning
 
 	assert(dataSeq.size() <= 5);
-	return ranges::all_of(xrange(std::min(unsigned(dataSeq.size()), cmdIdx)), [&](auto i) {
+	return ranges::all_of(xrange(std::min(dataSeq.size(), cmd.size())), [&](auto i) {
 		// convert the address to the '11 bit case'
 		auto addr = (addressing == Addressing::BITS_12) ? cmd[i].addr >> 1 : cmd[i].addr;
 		return ((addr & 0x7FF) == cmdAddr[addrSeq[i]]) &&
@@ -398,13 +397,25 @@ void AmdFlash::AmdCmd::serialize(Archive& ar, unsigned /*version*/)
 	             "value",   value);
 }
 
+// version 1: Initial version.
+// version 2: Added vppWpPinLow.
+// version 3: Changed cmd to static_vector, added status.
 template<typename Archive>
 void AmdFlash::serialize(Archive& ar, unsigned version)
 {
-	ar.serialize("ram",    *ram,
-	             "cmd",    cmd,
-	             "cmdIdx", cmdIdx,
-	             "state",  state);
+	ar.serialize("ram", *ram);
+	if (ar.versionAtLeast(version, 3)) {
+		uint8_t status = 0x80;
+		ar.serialize("cmd",    cmd,
+		             "status", status);
+	} else {
+		std::array<AmdCmd, 8> cmdArray;
+		unsigned cmdSize = 0;
+		ar.serialize("cmd",    cmdArray,
+		             "cmdIdx", cmdSize);
+		cmd = {from_range, subspan(cmdArray, 0, cmdSize)};
+	}
+	ar.serialize("state", state);
 	if (ar.versionAtLeast(version, 2)) {
 		ar.serialize("vppWpPinLow", vppWpPinLow);
 	}
--- a/src/serialize_stl.hh
+++ b/src/serialize_stl.hh
@@ -3,6 +3,7 @@
 
 #include "serialize_core.hh"
 #include "circular_buffer.hh"
+#include "static_vector.hh"
 #include <deque>
 #include <iterator>
 #include <vector>
@@ -62,6 +63,9 @@ template<typename T> struct serialize_as_collection<cb_queue<T>>
 		return std::back_inserter(q.getBuffer());
 	}
 };
+
+template<typename T, size_t N> struct serialize_as_collection<static_vector<T, N>>
+	: serialize_as_stl_collection<static_vector<T, N>> {};
 
 } // namespace openmsx
 
--- a/src/utils/static_vector.hh
+++ b/src/utils/static_vector.hh
@@ -1,6 +1,7 @@
 #ifndef STATIC_VECTOR_HH
 #define STATIC_VECTOR_HH
 
+#include "ranges.hh"
 #include <algorithm>
 #include <array>
 #include <cassert>
@@ -10,6 +11,7 @@
 #include <limits>
 #include <span>
 #include <type_traits>
+#include <utility>
 
 // This is a _very_ minimal implementation of the following (we can easily
 // extend this when the need arises).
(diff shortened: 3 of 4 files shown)


grauw:
One open question; can I forward `elem` here or not?

```diff
	template<typename Range>
	constexpr static_vector(from_range_t, Range&& range) {
		for (auto&& elem : range) {
-			push_back(elem);
+			push_back(std::forward<T>(elem));
		}
	}
```

Originally I thought I could `std::move(elem)` directly, since `&&` variables are always an rvalue (I looked that up to double check since I wasn’t sure). But then Sonarcloud [complained](https://sonarcloud.io/project/issues?resolutions=FIXED&pullRequest=1622&id=openMSX_openMSX&open=AY9-nhHhKXKyw1ZFtjQC) and I found that _actually_ if `Type` is a template parameter or auto variable then it’s _not_ always an rvalue (thank you C++) so I should use `std::forward`. But Sonarcloud also tells me that if `T` is a class template parameter it is _not_ forwardable, and the type of `elem` would be `T`, so now I’m immensely confused.

Then I thought maybe I should do `std::forward<Range>(elem)` since it depends on whether `Range` is an rvalue, but that fails to compile. Maybe I can still do `std::forward<T>(elem)`, but maybe that’s a no-op? I gave up and went back to the original implementation which just copies the damn thing. Maybe your C++-fu is stronger than mine 🙂.

m9710797:
> One open question; can I forward `elem` here or not?

This is indeed a bit tricky. And what makes this even more difficult is that Sonarcloud generates some false-positive warnings in this area. Usually I just follow some rule-of-thumbs and then stuff just works. But since you ask I'll try to reconstruct the underlying C++ rules (and I learned some new things myself while doing this).

There are two separate issues. How to pass the `range` function parameter. How to pass the `elem` variable to `push_back()`.

* Rule of thumb: pass ranges by r-value.

Unfortunately Sonarcloud does not agree with this rule. But I double checked this with the advice given in some c++ book I read (I think it was https://cppstd20.com/ the section about ranges and views).
The reason is that in general a range can caches internal state, and then passing by const-reference won't compile. For example calling `begin()` on a views::filter may do some non-trivial amount of work, that's cached for the next time `begin()` is called. Passing by value would be an option for most views (views are typically cheap to copy), but that's not a good option when the range is a full container. Passing as an r-value works for both cases.

* How to forward `elem`?

Using a variable by-name always gives an l-value. With `std::move()` you can unconditionally cast this to an r-value. With `std::forward()` you cast either to an l- or an r-value, depending on the "original" value-category (l/r-value) of the type. But in order to do this `std::forward()` needs extra information, namely the original type of that variable. This information is passed as a template parameter to `std::forward()`.

So in this specific example we need to know the type of the `elem` variable. In the code that type 's written as `auto&&` (you can think of this as a new unnamed template type). So we cannot directly name that type. Instead we can obtain it via `decltype(elem)`. The full statement then becomes:
```
push_back(std::forward<decltype(elem)>(elem));
```
Feel free to make this change in the code.
This pattern `std::forward<decltype(X)>(X)` is a common idiom, and some people write a macro to shorten it (but personally I don't like that).


Here is the experiment I did to verify some of the above information: https://godbolt.org/z/az1zxY1vr
 I found out that "moving" ranges is still a bit more complicated than moving "other" types. (E.g. sometimes you have to use `std::views::as_rvalue` to _really_ move the elements out of a range).

m9710797 (commented):
static_vector: Expand implementation. 

Looks good.

I assume you need (some of) this in a later patch? Possible because you use it in combination with `view::take()`?

Background info:
The intention is that in the future we replace our `static_vector` with a c++ standard implementation. The proposal is being discussed here: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p0843r11.html  (the original name was also `static_vector`, later it was renamed to `inplace_vector`).

grauw:
> I assume you need (some of) this in a later patch? Possible because you use it in combination with `view::take()`?

The serialisation needed `using value_type = T` and `data()`. In the other patch series I make use of `max_size()` and `back()`. I also wanted to use `emplace_back()` but couldn’t fully (see comment in 3rd patch). The rest is me just padding out the implementation.

> Background info: The intention is that in the future we replace our `static_vector` with a c++ standard implementation. The proposal is being discussed here: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p0843r11.html (the original name was also `static_vector`, later it was renamed to `inplace_vector`).

Would be great to have in the standard library indeed.

src/serialize_stl.hh
```diff
@@ -63,6 +64,19 @@ template<typename T> struct serialize_as_collection<cb_queue<T>>
...
+	: serialize_as_stl_collection<static_vector<T, N>>
+{
+	static constexpr bool loadInPlace = true;
+	static void prepare(static_vector<T, N>& v, int n) {
+		assert(n > 0 && size_t(n) <= N);
+		v.clear();
+	}
+	static T* output(static_vector<T, N>& v) {
+		return v.data();
+	}
```
m9710797: I think this is wrong (saving is fine, but loading is not).
`prepare()` calls `clear()`, making the `static_vector` have size 0. But there's nothing in the code that later set the size to the correct value.

For the other containers `serialize_as_collection::output()` returns a `std::back_inserter()`, that one will call `push_back()`, and thus the size gets increased incrementally.

I see also that you've set `loadInPlace = true`. When possible that's a good idea. But I'm not sure that's compatible with `std::back_inserter()` (or maybe it is, I'm just not sure). (Detail: the effect of `loadInPlace` is implemented in `serialize_core.hh` lines 719-742).

I see two possible solutions:
* The easiest solution is to not use `loadInPlace` (I mean keep the default value `false`). And then you can also keep the default implementation for `prepare()` and `output()`. (Maybe you could still do the `assert()` you have now in `prepare()`).
* If you want to use `loadInPlace=true` you could in `prepare()` replace the current `clear()` with `resize(n)`. Though that requires you also implement `static_vector::resize()`. And that method should must default-initialize the new elements (but these elements will be overwritten again so that's useless work).

Other alternatives could be to extend `static_vector` with a non-standard method to force setting the size, without changing the content of the internal array. But I'd prefer not to do that because that will make it more difficult to in the future switch to `std::inplace_vector`.

Because this code is not performance critical I'd suggest to go for the easiest solution (which is not using `loadInPlace`). That's also how we current serialize a regular `std::vector`.

m9710797 (commented):
serialize_stl: Support static_vector serialisation. 

Loading has a bug, see below.

src/memory/AmdFlash.cc
```diff
@@ -398,13 +399,25 @@ void AmdFlash::AmdCmd::serialize(Archive& ar, unsigned /*version*/)
...
+// version 2: Added vppWpPinLow.
+// version 3: Changed cmd to static_vector, added status.
 template<typename Archive>
 void AmdFlash::serialize(Archive& ar, unsigned version)
 {
-	ar.serialize("ram",    *ram,
-	             "cmd",    cmd,
-	             "cmdIdx", cmdIdx,
-	             "state",  state);
+	ar.serialize("ram", *ram);
```
m9710797, line 406: Detail:
In general we prefer to have fewer/larger blocks of `serialize()` calls. So maybe merge this line with line 420 below.

The underlying technical reason is that we have some optimization on "simple" types (e.g. integers or arrays of integers), (optimizations like: only reserve size in the output buffer only once instead of separately for each individual field). But these optimizations only trigger when all info is passed via a single `serialize()` call.

src/memory/AmdFlash.cc
```diff
@@ -398,13 +399,25 @@ void AmdFlash::AmdCmd::serialize(Archive& ar, unsigned /*version*/)
...
+	if (ar.versionAtLeast(version, 3)) {
+		uint8_t status = 0x80;
```
m9710797, line 408: I assume that `status` is a mockup for the corresponding field in your other PR?  (And that you won't push it like this?)

In the else-branch below you should still fill-in `status`. Possibly the initial value for `status` set by the constructor is already fine (and we only de-serialize into freshly-constructed objects), but for clarity it might still be a good idea to explicitly state how we de-serialize `status` when that info is missing in an old savestate.

src/memory/AmdFlash.cc
```diff
@@ -398,13 +399,25 @@ void AmdFlash::AmdCmd::serialize(Archive& ar, unsigned /*version*/)
...
+		ar.serialize("cmd",    cmd,
+		             "status", status);
+	} else {
+		std::array<AmdCmd, 8> cmdArray;
+		unsigned cmdSize = 0;
+		ar.serialize("cmd",    cmdArray,
+		             "cmdIdx", cmdSize);
+		cmd = {from_range, view::take(cmdArray, cmdSize)};
```
m9710797: This is "good enough", this code will only rarely trigger (when loading an old savestate). But anyway here are some details (feel free to ignore):
* Instead of constructing a new `static_vector` object (via from_range) and then copy-assigning that to the existing `cmd` object, it might be better/clearer to use the `assign()` method (https://en.cppreference.com/w/cpp/container/vector/assign). This directly assigns the new values to `cmd`. Or even better: assign_range() (https://en.cppreference.com/w/cpp/container/vector/assign_range), that's a c++23 feature, but we have full control over our `static_vector`.
* `view::take()` is our own approximation of `std::views::take` (but the latter is not yet (well enough) supported in clang-15). I would assume that the standard version is well optimized, but our version is pretty bare-bones. E.g. our version always returns a forward-range, but I assume that (when given a random-access-range) the standard std::views::take() returns a random-access-range. For the former the size can only be calculated in O(n), for the latter it can be calculated in O(1). That can make a difference when the range is later passed to a `from_range` constructor or a `assign_range()` method (maybe not in our `static_vector` implementation, but I expect the future `std::inplace_vector` will exploit this knowledge).
In this specific case we could replace `view::take()` with something like `std::span(cmdArray).subspan(0, cmdSize)`. And a std::span is random-access with constant-time `size()`.
EDIT: or using our own `subspan()` utility:     `subspan(cmdArray, 0, cmdSize)`

m9710797 (commented):
AmdFlash: Use static_vector to store command list. 

Only some comments about small tweaks or suggestions (feel free to ignore).
And one comment about loading `status` in case of an old savestate.

src/serialize_stl.hh
grauw: Yeah you’re right there’s a bug.

I’ll just remove the load in place as you say, then the implementation becomes trivial.

grauw: > * (Maybe you could still do the `assert()` you have now in `prepare()`).

`push_back()` already has an assert in it so not needed.

src/memory/AmdFlash.cc
grauw: > So maybe merge this line with line 420 below.

Do you mean to do something like this?

	if (ar.versionAtLeast(version, 3)) {
		uint8_t status = 0x80;
		ar.serialize("ram",    *ram,
		             "cmd",    cmd,
		             "state",  state,
		             "status", status,
		             "vppWpPinLow", vppWpPinLow);
	} else {
		std::array<AmdCmd, 8> cmdArray;
		unsigned cmdSize = 0;
		ar.serialize("ram",    *ram,
		             "cmd",    cmdArray,
		             "cmdIdx", cmdSize,
		             "state",  state);
		if (ar.versionAtLeast(version, 2)) {
			ar.serialize("vppWpPinLow", vppWpPinLow);
		}
		cmd = {from_range, view::take(cmdArray, cmdSize)};
	}

(Since `ram` and `state` are not consecutive.)

src/memory/AmdFlash.cc
grauw: > I assume that status is a mockup for the corresponding field in your other PR? (And that you won't push it like this?)

It’s a placeholder, since I wanted to only up the version number once. I do want to push it like this, but in a commit in the other PR I will replace this local variable with a class member variable. Like:

```diff
diff --git a/src/memory/AmdFlash.cc b/src/memory/AmdFlash.cc
index 164116976..5bc69b2cc 100644
--- a/src/memory/AmdFlash.cc
+++ b/src/memory/AmdFlash.cc
@@ -407,7 +407,6 @@ void AmdFlash::serialize(Archive& ar, unsigned version)
 {
        ar.serialize("ram", *ram);
        if (ar.versionAtLeast(version, 3)) {
-               uint8_t status = 0x80;
                ar.serialize("cmd",    cmd,
                             "status", status);
        } else {
diff --git a/src/memory/AmdFlash.hh b/src/memory/AmdFlash.hh
index 0d9a40990..745faa905 100644
--- a/src/memory/AmdFlash.hh
+++ b/src/memory/AmdFlash.hh
@@ -116,6 +116,7 @@ private:
        static constexpr unsigned MAX_CMD_SIZE = 8;
        static_vector<AmdCmd, MAX_CMD_SIZE> cmd;
        State state = State::IDLE;
+       uint8_t status = 0x80;
        bool vppWpPinLow = false; // true = protection on
 };
 SERIALIZE_CLASS_VERSION(AmdFlash, 3);
```

It has no meaning before that commit so currently it’s just discarded when deserialized. It just is there to make sure that until then, when an AmdFlash object with version 3 is serialised, the serialised data has a status field containing the default value.

Though I guess I might as well just add the member variable to the object right there to begin with, and skip the temp variable here, even if it’s not being used yet by the object.

src/memory/AmdFlash.cc
m9710797: This was very much a detail.
I meant keep all the same except for moving `ram` and `state` together. But again: not super important.

src/memory/AmdFlash.cc
m9710797: That's OK. 
It's also OK if you spread the changes over multiple commits if they are 'close' together (e.g. within the same push, or even a few days apart).

src/memory/AmdFlash.cc
grauw: Replacing `view::take()` with `subspan()`, saves us an include even though the FP style is pretty cool. I took a look at the `assign()` and `assign_range()` implementations but it was getting a bit too hairy so skipping that. Indeed performance doesn’t really matter for this case.

grauw:
Review comments addressed, and issue fixed in `back()` found during testing.

src/utils/static_vector.hh
```diff
@@ -30,45 +33,78 @@ template<typename T, size_t N>
...
+	[[nodiscard]] constexpr const T* data()  const noexcept { return array.data(); }
 
-	constexpr void clear() { sz = 0; }
+	template<typename... Args>
+	constexpr T& emplace_back(Args&&... args) {
+		assert(sz < N);
+		std::construct_at(std::addressof(array[sz]), std::forward<Args>(args)...);
+		++sz;
+		return back();
+	}
```
m9710797: This is wrong in general (but works in our current use case).

`std::construct_at()` constructs an object in _uninitialized_ memory (memory where no other object is currently living). So it does NOT first destruct the object at the given address.

We should do this destruct manually via something like `array[sz].~T()`.
In our current use case in AmdFlash the type in `static_array` has a trivial destructor (=destructor does nothing), so this doesn't matter.

More in general: the current implementation stores the data in an `std::array<T, N>`. But that means that for non-trivial types at construction we construct `N` times  a `T` object. And similar in the destructor. A better `static_vector` implementation could be to not use a `std::array` but just an array of uninitialized memory (with proper size and alignment) and then manually manage the lifetime of all stored objects.

We could try that, or maybe simpler for now only support types that have a trivial destructor. So maybe add something like: `static_assert(is_trivially_destructible_v<T>);`  (Given that we don't currently need this and that in the future we want to switch to `std::inplace_vector` which does handle this all correctly).

m9710797 (commented):
Since you mentioned a bug fix in this class (which I overlooked in my first review), I did a 2nd review. I found another bug, but maybe we shouldn't fix it? (instead we can accept a limitation).

src/utils/static_vector.hh
grauw: Another option is to get rid of `emplace_back()` I guess? And rely on the assignment operator in `push_back()`. But it’d be nicer to have it construct as needed, and if `clear()` would destruct, etc.

I’ll see what I can do.

m9710797: I'm not sure what you mean with "_...and if `clear()` would destruct ..._". In the current version it's wrong for `clear()` to destruct, because then later when `static_vector` itself gets destroyed, the elements get destroyed twice. So I assume you meant proper lifetime management for all operations (and this implies we can no longer use a `std::array<T, N>` as the internal storage).

I see 3 possible solutions:
* Do a minimum local fix in emplace_back(). This could be adding a call to `~T()`. Or it could be: first construct a temporary object and then move-assign that over the existing element in the array.
* Document and enforce the limitation that our `static_vector` is only meant to be used with trivially destructible types. So add that `static_assert` I mentioned above.
* Implement proper lifetime management. Maybe we can get inspiration from this [reference implementation](https://godbolt.org/z/5P78aG5xE). Unfortunately we can't use this implementation directly because it uses some c++23 features (but no doubt it's possible to adapt this).

The last option would be technically the most interesting (a good learning experience), BUT at the moment we don't need the extra functionality, and the intention remains to eventually replace our implementation with a standard version (c++26?).

Personally I'd go for the 2nd option. But I'd be happy to help you with any other option as well.

grauw: What I mean by “_...and if `clear()` would destruct ..._” is if only the actual elements in use were constructed, so constructing the vector would not default-construct, push / emplace would construct, and pop / clear would destruct.

> I see 3 possible solutions:

The fourth possible solution is to remove `emplace_back()`, right? Like in the new change I just pushed.

This fourth option would be my choice atm as the most trivial and safe choice. This commit in the pull request would then just add some functionality and not (fail to) attempt to optimise the construction. Optimizing the construction and lifetime can be done separately, later.

Even with a `static_assert` that restricts it to trivial types I don’t necessarily have full confidence that it would do the correct thing in all possible situations. I feel there are nuances that I don’t necessarily have the deep C++ language knowledge for to properly anticipate. I lost some confidence here ;). So at the moment it feels safer to me to steer clear of memory and lifetime management entirely, I don’t need it to achieve my goals with this PR.

Is this ok with you?

m9710797: All these options are fine for me. I did a very quick review of the latest version of this PR, and I'm OK with pulling it as it is now.  Do you permission to do this? (I think you have.) Or should I do it?

A fifth alternative could be to implement `emplace_back()` in terms of `push_back()`. That way users of `static_vector` can already write `emplace_back()`, but it only becomes an advantage in the future, e.g. when we switch to `std::inplace_vector`.
Maybe something like this (untested):
```
template<typename... Args>
T& emplace_back(Args&& ...args) {
    push_back(T(std::forward<Args>(args)...));
    return back();
}
```
But feel free to ignore this suggestion.

PR #1629, opened 2024-05-21
FlashROM sparse persistence

### LINKED ISSUE #2083: Allow easy updates of ASCII-X ROMs
When developing (i.e. getting new roms produced all the time), I need a start-script for openmsx and it must contain:

`del /Q C:\Users\palha\OneDrive\Documents\openMSX\persistent\roms\gofigure.rom`

I'm about to release this rom/game. At present, most people are not in possession of physical 8MB ASCII-X Flashroms and will use openMSX. And I'd say there is a good chance that I need to update the ROM at some point down the line. Right now, openMSX silently ignores any new ROM if it has the same name, and users will likely not understand. Asking users to delete the file, as described above, seems undesirable.

I know this is a [known issue](https://www.grauw.nl/projects/ascii-x/ascii16-x/#emulator-support), but I found no open ticket on this, hence this.

For simplicity, I suggest that openmsx detects a new file, and overwrites the existing part of the persistent ROM. 

Other ways could also be considered, like asking the user what to do upon detection (keep or overwrite ROM, keep or erase other flash-data), but others have to decide, I just want people to get their supplied ROM to be executed.

grauw:
When a FlashROM is used by a ROM mapper with initial content (e.g. via `-carta my.rom -romtype ASCII16-X`), when the program writes to the FlashROM, currently the entire memory is persisted and the initial content passed on the command line is henceforth ignored.

This is not ideal. Let’s say a game just modifies a single sector to write a savegame, if the user gets a new version of the ROM file they either can’t use it, or have to delete their persistent file and the savegame along with it.

This patch makes it so that only the modified sectors are persisted. It stores a metadata file in addition to the data, which indicates modified sectors, and the data contains only those. If the metadata file is not found but the data is, it will default to marking all non-write-protected sectors as modified (for backwards compatibility). If neither are found then it will default to having all sectors marked unmodified, and it won’t write the persistent files until it modified.

This replaces the previous scheme which relied on `writeProtectSectors` for cases like Manbow2, or a Load::DONT flag for Ink (actually not needed anymore after f1bbc93). Additionally, it removes the need for `writeAddress[]` and `readAddress[]`, the `TrackedRAM` buffer can now always be accessed directly. The initialisation logic is also greatly simplified.

The first three commits in this pull request are some mostly unrelated clean-ups, they should be fine to push.

The last commit is the most relevant one. It’s a draft, e.g. the serialisation hasn’t been updated, but before I finish it up I’d particularly like to get some feedback on this one.

Prior relevant discussion occurred [here](https://github.com/openMSX/openMSX/pull/1620#issuecomment-2118458438).

Closes #2083.

### DIFF
--- a/src/memory/AmdFlash.cc
+++ b/src/memory/AmdFlash.cc
@@ -1,14 +1,20 @@
 #include "AmdFlash.hh"
-#include "Rom.hh"
-#include "SRAM.hh"
-#include "MSXMotherBoard.hh"
-#include "MSXCPU.hh"
-#include "MSXDevice.hh"
-#include "MSXCliComm.hh"
+#include "File.hh"
+#include "FileContext.hh"
+#include "FileException.hh"
+#include "FileNotFoundException.hh"
+#include "FileOperations.hh"
 #include "HardwareConfig.hh"
 #include "MSXException.hh"
+#include "MSXCliComm.hh"
+#include "MSXCPU.hh"
+#include "MSXMotherBoard.hh"
+#include "Reactor.hh"
+#include "Rom.hh"
+#include "enumerate.hh"
 #include "narrow.hh"
 #include "one_of.hh"
+#include "outer.hh"
 #include "ranges.hh"
 #include "serialize.hh"
 #include "serialize_stl.hh"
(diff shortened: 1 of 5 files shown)


src/memory/AmdFlash.cc
```diff
@@ -675,32 +675,37 @@ bool AmdFlash::checkCommandBufferProgram()
 {
 	static constexpr std::array<uint8_t, 2> cmdSeq = {0xaa, 0x55};
 	if (chip.program.bufferCommand && partialMatch(cmdSeq)) {
-		if (cmd.size() <= 2) return true;
-		if (cmd.size() >= 3 && cmd[2].value == 0x25) {
-			if (cmd.size() <= 3) return true;
+		if (cmd.size() <= 2) {
+			return true;
+		}
+		if (cmd[2].value == 0x25) {
```
m9710797, line 645: Very much a detail about (personal) coding style preferences.
No need to change if you like this better.

Personally I prefer code with an early return over code with an extra indentation level:
```
if (condtion) {
    stuff;
}
```
vs
```
if (!condition) return;
stuff;
```
Various code metrics/heuristics that try to measure code complexity assign a higher cost to code that's more deeply indented. Of course this is only an heuristic, but in general I tend to agree with it. And according to this metric, the above transformation is a good way to reduce the (perceived) complexity of the code (again I tend to agree).

So specifically for this line, my preference would be to write it like `if (cmd[2].value != 0x25) return false;`.

And my (personal) preference for such simple test-and-return statements is to write them on a single line (more compact, and equally readable).

m9710797 (commented):
AmdFlash: Improve checkCommandBufferProgram().

I like the cleanup. Thanks.
One small remark about code style. But that's a personal preference thing, so feel free to ignore if you disagree.

m9710797 (commented):
AmdFlash: Use delegating constructor instead of init(). 

OK

m9710797 (commented):
AmdFlash: Remove don’t-load hack for Ink. 

Thanks. Nice cleanup.

src/memory/AmdFlash.cc
```diff
@@ -35,152 +41,125 @@ AmdFlash::AmdFlash(const std::string& name, const ValidatedChip& validatedChip,
...
-			// only if an actual sha1sum was given, tell the user we failed to use it
-			if (initialContentSpecified) {
-				config.getCliComm().printWarning(
-					"Could not load specified initial content "
-					"for flash ROM: ", e.getMessage());
-			}
+			file = File(path, File::OpenMode::LOAD_PERSISTENT);
+		} catch (FileException& e) {
+			config.getCliComm().printWarning("Couldn't load Flash data ",
+				filename, " (", e.getMessage(), ").");
```
m9710797, line 84: In a previous patch you removed the warning "SRAM file not found, assuming blank content" (and I agree with that change). But doesn't this re-introduce a very similar warning?

Also I think the `catch` in the line below (line 86) doesn't do anything ATM.

src/memory/AmdFlash.cc
```diff
@@ -35,152 +41,125 @@ AmdFlash::AmdFlash(const std::string& name, const ValidatedChip& validatedChip,
...
-				assert(rom); // must have rom constructor parameter
-				if ((offset + sectorSize) <= romSize) {
-					readAddress[sector] = &(*rom)[offset];
-				} else {
-					readAddress[sector] = nullptr;
-				}
+				const size_t size = std::min(address + region.size, romData.size()) - std::min(address, romData.size());
+				ranges::copy(romData.subspan(address, size), ramData.subspan(address, size));
+				ranges::fill(ramData.subspan(address + size, region.size - size), 0xFF);
 			}
```
m9710797, line 104: In each iteration you fill an extra part of `ramData`, and the `then` and the `else` branch fill the same part. This is one of the things I verified during my review. So it might be worth making this explicit, by moving the sub-expression `ramData.subspan(address, size)` before the `if`. Also both branches fill the remainder of this part with 0xff, could that be moved after the `if`?

src/memory/AmdFlash.cc
```diff
@@ -35,152 +41,125 @@ AmdFlash::AmdFlash(const std::string& name, const ValidatedChip& validatedChip,
...
-			offset += sectorSize;
 		}
 	}
-	assert(offset == size());
```
m9710797, line 176: If I understand correctly:
* The data-file only contains those sectors that have been modified.
* The meta-file indicates which sectors exactly that are.
* And if this meta-file is missing we assume all writable sectors have been modified. This is the scenario when upgrading from an older openMSX release.

Right?

--

In general we don't try to be super robust when loading savestates which have been manually edited by the users (so savestates which contain invalid data).

However in this case it's possible the user specified the wrong `-romtype` command line parameter, and then we may have a mismatch between the current `chip.geometry` and `std::vector sectors`. E.g. `sectors` may contain more or less number of elements than the number of sectors in `geometry`.

Maybe we should add a check (around line 69, where we do `in.serialize(...)`) that the loaded number of sectors matches the expected amount. (I'm not sure what to do in case of a mismatch).

Also if the number of sectors does match, the geometry may still be different. I think the loading code itself is robust (e.g. it won't read beyond the end of the file). But the data may still end up in the wrong places.

Maybe we should extend the meta info with a field indicating the chip ID?

src/memory/AmdFlash.cc
```diff
@@ -35,152 +41,125 @@ AmdFlash::AmdFlash(const std::string& name, const ValidatedChip& validatedChip,
...
+					file.write(std::span{ram}.subspan(address, region.size));
+				}
+			}
+		}
+		XmlOutputArchive out(path + ".meta");
+		out.serialize("sectors", sectors);
+	} catch (FileException& e) {
+		config.getCliComm().printWarning(
+			"Couldn't save SRAM ", filename, " (", e.getMessage(), ").");
+	}
```
m9710797, line 146: Because we've split the info over two files we've introduced some extra failure modes (maybe not likely, but possible):
* It's possible that only one of the two files gets written (e.g. some error during writing the second one).
* Then either that 2nd file will be missing (the bw-compat scenario). And it will be wrongly interpreted as-if all sectors have been written).
* Or the 2nd file is not written and a previous version of the file is kept. This may also lead to wrong conclusions.

Maybe this is an unlikely scenario? Should we do something about it? How much effort should we spend on this? Here are some possibilities:
* Think about the order in which we write these two files. Is one order better than the other (in case of a partial failure I mean)?
* Should we first remove the existing files before writing the new ones?
* We could write both to a temporary files (same directory, but different filename. e.g. with extra `.partial` suffix). And only when both writes were successful, rename the files (so possibly removing the old files).

--

An alternative could be to go back to a single file.

I proposed the two-file solution, assuming that the data-file would be a full-copy of the flash rom content. If that's not the case (when it can not be meaningfully interpreted without also knowing the meta-data), then I see less value in having it two separate files.

Since you're now using the serialization stuff it becomes relatively easy to embed binary data in the same file  (via `serialize_as_blob`). Maybe a bonus: this binary data will also be stored compressed (and base64 encoded).

--

I don't have a strong preference, I on;y wanted to point out some advantages/disadvantages of both approaches.

src/memory/AmdFlash.cc
```diff
@@ -200,18 +179,13 @@ void AmdFlash::setState(State newState)
...
 uint8_t AmdFlash::peek(size_t address) const
 {
 	if (state == State::IDLE) {
-		auto [sector, sectorSize, offset] = getSectorInfo(address);
-		if (const uint8_t* addr = readAddress[sector]) {
-			return addr[offset];
-		} else {
-			return 0xFF;
-		}
+		return ram[address % size()];
```
m9710797, line 188: Here and in a few other places you added `% size()`. Why is that? Isn't it fine/better to assert that `ram` won't be accessed out-of-bounds?

Detail: (orthogonal) the return-type of `size()` is a `size_t`. Maybe it should be a `power_of_two`? So retaining this knowledge for longer, increasing the likelihood that the compiler optimizes such `%` operations to an `&`. 
A smart enough compiler may still see through this, so maybe this doesn't matter. On the other hand, also for human readers of the code, retaining the `power_of_two` knowledge might be useful.

src/memory/AmdFlash.hh
```diff
@@ -1,7 +1,10 @@
 #ifndef AMDFLASH_HH
 #define AMDFLASH_HH
 
+#include "DeviceConfig.hh"
```
m9710797, line 4: Very much a detail: after adding this #include you can remove the forward declaration on line 23.

m9710797 (commented):
AmdFlash: Sparse persistence of modified sectors only. 

I understand this is still work in progress. So I focused more on design stuff than on implementation stuff.

src/memory/AmdFlash.cc
grauw: The call to `config.getFileContext().resolveSavePaths(filename)` will throw if the file doesn’t exist, which the catch on line 86 below it will receive, so it behaves the same way as in the previous patch.

Though aside, not a big fan of how these resolve functions communicate failure by throwing, it leads to confusions like this because you can’t tell by just looking at it that the function can fail (and with which exception). Would perhaps be nicer if they returned an optional or variant.

m9710797: Aha, I understand now.
I agree with you. In _this_ case a non-throwing version would be much cleaner.

In the majority of the cases where we use `resolve()` it is a hard error if the file cannot be found. And then it's nice that the error handling (via an exception) is automatic. But here we **do** want to handle that case. I wouldn't mind offering two versions: a `resolve()` like we have now, and an extra `tryResolve()` method which returns an `std::optional`. (Suggestions for better names are welcome ofcourse).

In the XMLElement class we have something similar:  the `getChild()` (and variations) methods throw an exception (it's a hard error if the tag is missing). But the `findChild()` methods may return a `nullptr`, these are used to query optionally present tags.

(And in our defense, this code is +20 years old, at the time `std::optional` and `std::variant` didn't exist yet. But that doesn't mean we shouldn't try to improve it now).

src/memory/AmdFlash.cc
grauw: An address greater than the capacity is legit input IMO. Mirroring will be the hardware implementation for unused address lines, so I think it is correct to allow it. It’s better than adding `& (flash.size() - 1)` in all places that interact with AmdFlash.

For example take an ASCII16-X device with 8 MB Flash ROM, where the mapper can address up to 64 MB. The software can poke the mapper to a bank >= `0x200`, thus providing an address greater than the capacity. Currently it does not mirror, while it should, and I think it’s best to do that in AmdFlash itself.

It’s also defensive programming to make sure we don’t access out of bounds; a mistake would mean a security bug (asserts are compiled out of release builds). AmdFlash has nontrivial code that has been written with in-range addresses in mind. It’s easy to miss handling unexpected input, and hard to make guarantees. A class of bugs can be eliminated by making sure in the public interface that the input is never unexpected.

I’ll pull this out into a separate commit, including `power_of_two<size_t> size()`.

src/memory/AmdFlash.cc
grauw: I wrote a long story with my opinion on single-line ifs, braces, guard patterns, indentation, but then did other stuff and accidentally closed the tab… Oh well, never mind 😄.

I will make the code flow / read a bit nicer, around how far it received the command. And also performs fewer checks; after the command is locked in (`cmd.size() >= 3`) we can make assumptions about each byte being fed in one by one, e.g. we only need to check `cmd[3].value < chip.program.pageSize` when `cmd.size() == 4` and not every time afterwards.

src/memory/AmdFlash.cc
grauw: > However in this case it's possible the user specified the wrong `-romtype` command line parameter, and then we may have a mismatch between the current `chip.geometry` and `std::vector sectors`. E.g. `sectors` may contain more or less number of elements than the number of sectors in `geometry`.

Hmm very good point, I hadn’t thought of that.

Perhaps rather than store the modified sectors, it’d be better to store the ranges of memory that were modified (rounded to whole sectors), and when loading it would just mark the relevant sectors as modified. That wouldn’t be too difficult to implement, and it’d be more robust since it relates the the physical location in memory which is related to the ROM rather than the mapper.

That might be nicer than adding chip ID to the metadata, because then I don’t know what a good fallback behaviour would be (as you remark as well). Whereas with this even if the FlashROM geometry is different it will just remap the modified regions to that and mark the appropriate sectors modified.

m9710797: A good goal might be:
* If a flash-rom has already been written to (e.g. a savegame), and the user accidentally uses this data in combination with a different (wrong?) rom-type, then this shouldn't result in removing/corrupting the existing savegame.

I think your suggestion (using absolute offsets rather than sector numbers) already comes closer. But it might still loose data. For example in the scenario where the 2nd (wrong) mapper-type is a smaller flash rom, and the modified data is located past the size of the smaller mapper.

A second scenario that we may want to consider is:
* You initially start the game with the wrong mapper-type.
* Later you correct this mistake (= switch to the correct mapper).
* It should be possible to recover from this. So end up in the state that is identical to as-if you immediately used to correct type.

I'm not sure how to fully achieve this second goal.

Maybe we could:
* Do store the flashrom-id in the meta-data.
* Compare this ID when we use the data.
* If there's a mismatch, make the full flash read-only And print a warning with the full path of the file that should be deleted to recover (if the new mapper-type is actually the correct one).

When the data is stored in a single file, it's easier to tell the user what file to delete.

src/memory/AmdFlash.cc
grauw: > * Think about the order in which we write these two files. Is one order better than the other (in case of a partial failure I mean)?

Since the metadata file is meaningless without the data file, it’s probably best to write that one first, and if it fails, not write the data file.

Thinking about when this could fail, I think file creation permissions are the most likely rather than disk space. If you can create / modify the metadata file, then it is unlikely you can’t also create / modify the data file. Whereas if a data file already exists without metadata, it could be that it was created under different permission conditions and you could modify that but then fail to create the metadata.

> * Should we first remove the existing files before writing the new ones?

That can lead to unrecoverable data loss, so rather not…

> * We could write both to a temporary files (same directory, but different filename. e.g. with extra `.partial` suffix). And only when both writes were successful, rename the files (so possibly removing the old files).

I think that’s a bit over the top, that’s not done anywhere else in openMSX afaik even if there could very well be situations where you could in theory e.g. write an incomplete file when running out of disk space.

A bad alternative that also came to mind: write one file per sector :).

> An alternative could be to go back to a single file.
>
> I proposed the two-file solution, assuming that the data-file would be a full-copy of the flash rom content. If that's not the case (when it can not be meaningfully interpreted without also knowing the meta-data), then I see less value in having it two separate files.

The reason why I went for the sparse implementation immediately is because it integrated a bit more neatly with the existing method, back compat was easy to do, and it’s also more space-efficient with the savegame sectors use case in mind. But, I’m not necessarily against storing a full copy.

If it’d store a full copy, the data in the non-modified sectors would be meaningless.

  * If it stored ROM data (straight dump of the tracked RAM) it could be outdated compared to the ROM passed in. But it’s kinda cool that’d then be a usable .rom actually.
  * If it filled those unmodified parts with FFH or something, fine I guess, feels a bit unsatisfying tho.

> Since you're now using the serialization stuff it becomes relatively easy to embed binary data in the same file (via `serialize_as_blob`). Maybe a bonus: this binary data will also be stored compressed (and base64 encoded).

Downside is that this does introduce a new file format, so it’d mean a more elaborate back compat code path, a new file extension, etc. Also, maybe I needn’t worry about this, but wouldn’t it be slower due to the encoding and compression? These can be fairly large (8 MB) files… Then again, they won’t change often, so probably ok.

Something also to consider here; a FlashROM can persist more data than just the ROM data. Some Flash ROMs like the M29W640G and S29GL064S have a 256-byte “extended memory block” or “secure silicon region”, currently it’s not implemented, but let’s say it would be sometime in the future, how do we store that? If we have separate files I don’t think it makes sense to append to the plain data, I would want to either write it as a separate file or include it in the metadata.

Anyway not really sure what’s best here either. Let me push some updates and then see if we can make up our minds. The serialisation details are relatively easy to change.

src/memory/AmdFlash.cc
m9710797: Also if you store ranges of modified data (rounded to sector boundaries). But if you then load this data in a different flash-rom. These old boundaries may not map to new boundaries. And what will we do then? Extend again to sector boundaries? But that will result in problems when we correct our mistake and switch back to the original mapper (now we unintentionally extended the number of written-to sectors).

src/memory/AmdFlash.cc
grauw: I’ve had some more time to think about it, and I think a new single-file `.flash` format is the way to go. It’s just much more easily extensible this way. Future extensions of FlashROM support may include persisting the extended memory block, and also the S29GL064S supports ECC information, etc. It would make no sense to save these as separate files, so why is the main Flash ROM data special. Just extra hassle.

src/memory/AmdFlash.cc
grauw: Another option;

I could encode the manufacturer / device ID in the filename, like `man-01-dev-1000.flash`. Then if the device is different for some reason you will get a separate persistent file. But if the user is certain they are compatible and wants to carry over the data, they can simply rename the file.

grauw: I notice by the way that the manufacturer / device ID combination is not a perfectly unique identifier, and does not guarantee that they support the same command set;

* The AM29F040 and AM29F040B both are identified as 01 / A4, but the latter supports status DQ2 and programming in Erase Suspend mode which the former does not.
* The AM29F016, AM29F016B and AM29F016D are all identified as 01 / AD, but the D version supports CFI and Unlock Bypass which the other two do not.

They do have the same geometry though, and I think it can be assumed to be the case. These are all iterations on the same design, coinciding with process node upgrades from 0.5 µm to 0.32 µm (B) to 0.23 µm (D).

m9710797: Yes, in this context (different filenames for different mappers) the geometry should be sufficient, I think.

PR #1631, opened 2024-05-29
[[maybe_unused]] warnings

pvmm:


m9710797:
We should indeed fix this warning. So thanks for your PR.

In hundreds (maybe more) of other places in the current code we instead use something like `void(my_variable)` instead of `[[maybe_unused]] my_variable` to suppress the warning. But the reason for this is historical, at the time this (c++17) attribute was not yet available or not yet supported by the compilers we were using. But maybe now is a good time to start using `[[maybe_unused]]`.

However in this specific case I think it's even better to rewrite this code as:
```
assert(ranges::all_of(regions, [](const auto& region) { return region.count > 0; }));
```

m9710797:
I've fixed this on master.
Thanks for pointing out the warning.

PR #1632, opened 2024-05-30
Create breakpoints with slot/subslot info in symbol manager

pvmm:
* [x] display slot and subslot in symbol manager;
sample image
 * [ ] ~change slot and subslot in a popup dialog;~
 * idea: after clicking with right mouse button on slots field?
image
* [ ] ~save session (omds file);~
* [ ] ~load session (omds file);~

### DIFF
--- a/src/debugger/SymbolManager.cc
+++ b/src/debugger/SymbolManager.cc
@@ -114,38 +114,51 @@ SymbolManager::SymbolManager(CommandController& commandController_)
 	return result;
 }
 
-[[nodiscard]] std::optional<uint16_t> SymbolManager::parseValue(std::string_view str)
+template<typename T>
+[[nodiscard]] std::optional<T> SymbolManager::parseValue(std::string_view str)
 {
 	if (str.ends_with('h') || str.ends_with('H')) { // hex
 		str.remove_suffix(1);
-		return StringOp::stringToBase<16, uint16_t>(str);
+		return StringOp::stringToBase<16, T>(str);
 	}
 	if (str.starts_with('$') || str.starts_with('#')) { // hex
 		str.remove_prefix(1);
-		return StringOp::stringToBase<16, uint16_t>(str);
+		return StringOp::stringToBase<16, T>(str);
 	}
 	if (str.starts_with('%')) { // bin
 		str.remove_prefix(1);
-		return StringOp::stringToBase<2, uint16_t>(str);
+		return StringOp::stringToBase<2, T>(str);
 	}
 	// this recognizes the prefixes "0x" or "0X" (for hexadecimal)
 	//                          and "0b" or "0B" (for binary)
 	// no prefix in interpreted as decimal
 	// "0" as a prefix for octal is intentionally NOT supported
-	return StringOp::stringTo<uint16_t>(str);
+	return StringOp::stringTo<T>(str);
 }
 
-[[nodiscard]] std::optional<Symbol> SymbolManager::checkLabel(std::string_view label, uint16_t value)
+// explicitly instantiate for uint16_t and uint32_t (needed for unittest)
+template std::optional<uint16_t> SymbolManager::parseValue<uint16_t>(std::string_view);
+template std::optional<uint32_t> SymbolManager::parseValue<uint32_t>(std::string_view);
+
+[[nodiscard]] std::optional<Symbol> SymbolManager::checkLabel(std::string_view label, uint32_t value)
 {
 	if (label.ends_with(':')) label.remove_suffix(1);
 	if (label.empty()) return {};
 
-	return Symbol{std::string(label), value};
+	return Symbol{std::string(label), static_cast<uint16_t>(value), {}, static_cast<uint16_t>(value >> 16)};
 }
 
 [[nodiscard]] std::optional<Symbol> SymbolManager::checkLabelAndValue(std::string_view label, std::string_view value)
 {
-	if (auto num = parseValue(value)) {
+	if (auto num = parseValue<uint16_t>(value)) {
+		return checkLabel(label, *num);
+	}
+	return {};
+}
+
+[[nodiscard]] std::optional<Symbol> SymbolManager::checkLabelSegmentAndValue(std::string_view label, std::string_view value)
+{
+	if (auto num = parseValue<uint32_t>(value)) {
 		return checkLabel(label, *num);
 	}
 	return {};
@@ -168,13 +181,14 @@ SymbolManager::SymbolManager(CommandController& commandController_)
 
 [[nodiscard]] SymbolFile SymbolManager::loadNoICE(std::string_view filename, std::string_view buffer)
 {
-	auto parseLine = [](std::span<std::string_view> tokens) -> std::optional<Symbol> {
+	auto parseLine = [&](std::span<std::string_view> tokens) -> std::optional<Symbol> {
 		if (tokens.size() != 3) return {};
 		auto def   = tokens[0];
 		auto label = tokens[1];
 		auto value = tokens[2];
 		if (StringOp::casecmp cmp; !cmp(def, "def")) return {};
-		return checkLabelAndValue(label, value);
+		// detecting segment information above 16bits
+		return checkLabelSegmentAndValue(label, value);
 	};
 	return loadLines(filename, buffer, SymbolFile::Type::NOICE, parseLine);
 }
@@ -324,7 +338,7 @@ SymbolManager::SymbolManager(CommandController& commandController_)
 			if (it == et) break;
 			auto value = *it++; // this could either be the psect or the value column
 			if (auto val = is4DigitHex(value)) {
-				result.symbols.emplace_back(std::string(label), *val);
+				result.symbols.emplace_back(std::string(label), *val, std::nullopt, std::nullopt);
 				continue;
 			}
 
@@ -332,14 +346,14 @@ SymbolManager::SymbolManager(CommandController& commandController_)
 			value = *it++; // try again with 3rd column
 			auto val = is4DigitHex(value);
 			if (!val) break; // if this also doesn't work there's something wrong, skip this line
-			result.symbols.emplace_back(std::string(label), *val);
+			result.symbols.emplace_back(std::string(label), *val, std::nullopt, std::nullopt);
 		}
 	}
 
 	return result;
 }
 
-[[nodiscard]] SymbolFile SymbolManager::loadSymbolFile(const std::string& filename, SymbolFile::Type type)
+[[nodiscard]] SymbolFile SymbolManager::loadSymbolFile(const std::string& filename, SymbolFile::Type type, std::optional<uint8_t> slot)
 {
 	File file(filename);
 	auto buf = file.mmap();
@@ -351,21 +365,31 @@ SymbolManager::SymbolManager(CommandController& commandController_)
 	}
 	assert(type != AUTO_DETECT);
 
-	switch (type) {
-		case ASMSX:
-			return loadASMSX(filename, buffer);
-		case GENERIC:
-			return loadGeneric(filename, buffer);
-		case HTC:
-			return loadHTC(filename, buffer);
-		case LINKMAP:
-			return loadLinkMap(filename, buffer);
-		case NOICE:
-			return loadNoICE(filename, buffer);
-		case VASM:
-			return loadVASM(filename, buffer);
-		default: UNREACHABLE;
+	auto symbolFile = [&]{
+		switch (type) {
+			case ASMSX:
+				return loadASMSX(filename, buffer);
+			case GENERIC:
+				return loadGeneric(filename, buffer);
+			case HTC:
+				return loadHTC(filename, buffer);
+			case LINKMAP:
+				return loadLinkMap(filename, buffer);
+			case NOICE:
+				return loadNoICE(filename, buffer);
+			case VASM:
+				return loadVASM(filename, buffer);
+			default: UNREACHABLE;
+		}
+	}();
+
+	// Update slot info for the file and each of its symbol
+	symbolFile.slot = slot;
+	for (auto& symbol: symbolFile.getSymbols()) {
+		symbol.slot = slot;
 	}
+
+	return symbolFile;
 }
 
 void SymbolManager::refresh()
@@ -386,9 +410,9 @@ void SymbolManager::refresh()
 	if (observer) observer->notifySymbolsChanged();
 }
 
-bool SymbolManager::reloadFile(const std::string& filename, LoadEmpty loadEmpty, SymbolFile::Type type)
+bool SymbolManager::reloadFile(const std::string& filename, LoadEmpty loadEmpty, SymbolFile::Type type, std::optional<uint8_t> slot)
 {
-	auto file = loadSymbolFile(filename, type); // might throw
+	auto file = loadSymbolFile(filename, type, slot); // might throw
 	if (file.symbols.empty() && loadEmpty == LoadEmpty::NOT_ALLOWED) return false;
 
 	if (auto it = ranges::find(files, filename, &SymbolFile::filename);
@@ -434,7 +458,7 @@ std::optional<uint16_t> SymbolManager::parseSymbolOrValue(std::string_view str) 
 		}
 	}
 	// also not found, then try to parse as a numerical value
-	return parseValue(str);
+	return parseValue<uint16_t>(str);
 }
 
 std::span<Symbol const * const> SymbolManager::lookupValue(uint16_t value)
@@ -451,6 +475,15 @@ std::span<Symbol const * const> SymbolManager::lookupValue(uint16_t value)
 		return *sym;
 	}
 	return {};
+}
+
+SymbolFile* SymbolManager::findFile(std::string filename)
+{
+	if (auto it = ranges::find(files, filename, &SymbolFile::filename); it == files.end()) {
+		return nullptr;
+	} else {
+		return &(*it);
+	}
 }
 
 std::string SymbolManager::getFileFilters()
--- a/src/debugger/SymbolManager.cc
+++ b/src/debugger/SymbolManager.cc
@@ -145,7 +145,8 @@ template std::optional<uint32_t> SymbolManager::parseValue<uint32_t>(std::string
 	if (label.ends_with(':')) label.remove_suffix(1);
 	if (label.empty()) return {};
 
-	return Symbol{std::string(label), static_cast<uint16_t>(value), {}, static_cast<uint16_t>(value >> 16)};
+	auto tmp{value > 0xFFFF ? std::optional<uint16_t>(static_cast<uint16_t>(value >> 16)) : std::nullopt};
+	return Symbol{std::string(label), static_cast<uint16_t>(value), {}, tmp};
 }
 
 [[nodiscard]] std::optional<Symbol> SymbolManager::checkLabelAndValue(std::string_view label, std::string_view value)
@@ -181,6 +182,7 @@ template std::optional<uint32_t> SymbolManager::parseValue<uint32_t>(std::string
 
 [[nodiscard]] SymbolFile SymbolManager::loadNoICE(std::string_view filename, std::string_view buffer)
 {
+	bool anySegment = false;
 	auto parseLine = [&](std::span<std::string_view> tokens) -> std::optional<Symbol> {
 		if (tokens.size() != 3) return {};
 		auto def   = tokens[0];
@@ -188,9 +190,19 @@ template std::optional<uint32_t> SymbolManager::parseValue<uint32_t>(std::string
 		auto value = tokens[2];
 		if (StringOp::casecmp cmp; !cmp(def, "def")) return {};
 		// detecting segment information above 16bits
-		return checkLabelSegmentAndValue(label, value);
+		auto symbol = checkLabelSegmentAndValue(label, value);
+		anySegment |= symbol->segment.has_value();
+		return symbol;
 	};
-	return loadLines(filename, buffer, SymbolFile::Type::NOICE, parseLine);
+	auto file = loadLines(filename, buffer, SymbolFile::Type::NOICE, parseLine);
+	 // Heuristic: if all segments in the symbol file are 0,
+	 // then assume the file contains no segment information.
+	if (anySegment) {
+		for (auto& symbol: file.getSymbols()) {
+			if (!symbol.segment) symbol.segment = 0;
+		}
+	}
+	return file;
 }
 
 [[nodiscard]] SymbolFile SymbolManager::loadHTC(std::string_view filename, std::string_view buffer)
--- a/src/debugger/SymbolManager.hh
+++ b/src/debugger/SymbolManager.hh
@@ -19,11 +19,13 @@ class CommandController;
 
 struct Symbol
 {
-	Symbol(std::string n, uint16_t v)
-		: name(std::move(n)), value(v) {} // clang-15 workaround
+	Symbol(std::string n, uint16_t v, std::optional<uint8_t> s1, std::optional<uint16_t> s2)
+		: name(std::move(n)), value(v), slot(s1), segment(s2) {} // clang-15 workaround
 
 	std::string name;
 	uint16_t value;
+	std::optional<uint8_t> slot;
+	std::optional<uint16_t> segment;
 
 	auto operator<=>(const Symbol&) const = default;
 };
@@ -44,7 +46,9 @@ struct SymbolFile
 	};
 	[[nodiscard]] static zstring_view toString(Type type);
 	[[nodiscard]] static std::optional<Type> parseType(std::string_view str);
+	[[nodiscard]] auto& getSymbols() { return symbols; }
 
+	std::optional<uint8_t> slot;
 	std::string filename;
 	std::vector<Symbol> symbols;
 	Type type;
@@ -72,12 +76,13 @@ public:
 	//   existing file is not replaced and this method returns 'false'.
 	//   Otherwise it return 'true'.
 	enum class LoadEmpty { ALLOWED, NOT_ALLOWED };
-	bool reloadFile(const std::string& filename, LoadEmpty loadEmpty, SymbolFile::Type type);
+	bool reloadFile(const std::string& filename, LoadEmpty loadEmpty, SymbolFile::Type type, std::optional<uint8_t> slot = {});
 
 	void removeFile(std::string_view filename);
 	void removeAllFiles();
 
 	[[nodiscard]] const auto& getFiles() const { return files; }
+	[[nodiscard]] SymbolFile* findFile(std::string filename);
 	[[nodiscard]] std::span<Symbol const * const> lookupValue(uint16_t value);
 	[[nodiscard]] std::optional<uint16_t> parseSymbolOrValue(std::string_view s) const;
 
@@ -88,9 +93,11 @@ public:
 	// These should not be called directly, except by the unittest
 	[[nodiscard]] static std::optional<unsigned> isHexDigit(char c);
 	[[nodiscard]] static std::optional<uint16_t> is4DigitHex(std::string_view s);
-	[[nodiscard]] static std::optional<uint16_t> parseValue(std::string_view str);
-	[[nodiscard]] static std::optional<Symbol> checkLabel(std::string_view label, uint16_t value);
+	template <typename T>
+	[[nodiscard]] static std::optional<T> parseValue(std::string_view str);
+	[[nodiscard]] static std::optional<Symbol> checkLabel(std::string_view label, uint32_t value);
 	[[nodiscard]] static std::optional<Symbol> checkLabelAndValue(std::string_view label, std::string_view value);
+	[[nodiscard]] static std::optional<Symbol> checkLabelSegmentAndValue(std::string_view label, std::string_view value);
 	[[nodiscard]] static SymbolFile::Type detectType(std::string_view filename, std::string_view buffer);
 	[[nodiscard]] static SymbolFile loadLines(
 		std::string_view filename, std::string_view buffer, SymbolFile::Type type,
@@ -101,7 +108,7 @@ public:
 	[[nodiscard]] static SymbolFile loadVASM(std::string_view filename, std::string_view buffer);
 	[[nodiscard]] static SymbolFile loadASMSX(std::string_view filename, std::string_view buffer);
 	[[nodiscard]] static SymbolFile loadLinkMap(std::string_view filename, std::string_view buffer);
-	[[nodiscard]] static SymbolFile loadSymbolFile(const std::string& filename, SymbolFile::Type type);
+	[[nodiscard]] static SymbolFile loadSymbolFile(const std::string& filename, SymbolFile::Type type, std::optional<uint8_t> slot = {});
 
 private:
 	void refresh();
--- a/src/imgui/ImGuiSymbols.cc
+++ b/src/imgui/ImGuiSymbols.cc
@@ -169,7 +169,7 @@ static void checkSort(const SymbolManager& manager, std::vector<SymbolRef>& symb
 
 void ImGuiSymbols::drawContext(MSXMotherBoard* motherBoard, const SymbolRef& sym)
 {
-	if (ImGui::MenuItem("Set breakpoint")) {
+	if (ImGui::MenuItem("Set breakpoint", nullptr, nullptr, motherBoard != nullptr)) {
 		std::string cond;
 		if (auto slot = sym.slot(symbolManager)) {
 			strAppend(cond, "[pc_in_slot ", *slot & 3, ' ', (*slot >> 2) & 3);
@@ -201,8 +201,8 @@ void ImGuiSymbols::drawTable(MSXMotherBoard* motherBoard, const std::string& fil
 	            (FILTER_FILE ? ImGuiTableFlags_ScrollY : 0);
 	im::Table(file.c_str(), (FILTER_FILE ? 4 : 5), flags, {0, 100}, [&]{
 		ImGui::TableSetupScrollFreeze(0, 1); // Make top row always visible
-		ImGui::TableSetupColumn("name");
-		ImGui::TableSetupColumn("value", ImGuiTableColumnFlags_DefaultSort | ImGuiTableColumnFlags_WidthFixed);
+		ImGui::TableSetupColumn("name", ImGuiTableColumnFlags_NoHide);
+		ImGui::TableSetupColumn("value", ImGuiTableColumnFlags_NoHide | ImGuiTableColumnFlags_DefaultSort | ImGuiTableColumnFlags_WidthFixed);
 		ImGui::TableSetupColumn("slot", ImGuiTableColumnFlags_DefaultHide);
 		ImGui::TableSetupColumn("segment", ImGuiTableColumnFlags_DefaultHide);
 		if (!FILTER_FILE) {
--- a/src/imgui/ImGuiSymbols.cc
+++ b/src/imgui/ImGuiSymbols.cc
@@ -3,6 +3,7 @@
 #include "ImGuiBreakPoints.hh"
 #include "ImGuiCpp.hh"
 #include "ImGuiManager.hh"
+#include "ImGuiDebugger.hh"
 #include "ImGuiOpenFile.hh"
 #include "ImGuiUtils.hh"
 #include "ImGuiWatchExpr.hh"
@@ -10,7 +11,10 @@
 #include "CliComm.hh"
 #include "File.hh"
 #include "Interpreter.hh"
+#include "MSXCliComm.hh"
 #include "MSXException.hh"
+#include "MSXMotherBoard.hh"
+#include "MSXCPUInterface.hh"
 #include "Reactor.hh"
 
 #include "enumerate.hh"
@@ -45,10 +49,16 @@ void ImGuiSymbols::save(ImGuiTextBuffer& buf)
 	for (const auto& file : symbolManager.getFiles()) {
 		buf.appendf("symbolfile=%s\n", file.filename.c_str());
 		buf.appendf("symbolfiletype=%s\n", SymbolFile::toString(file.type).c_str());
+		if (file.slot) {
+			buf.appendf("slotsubslot=%d\n", *file.slot);
+		}
 	}
-	for (const auto& [file, error, type] : fileError) {
+	for (const auto& [file, error, type, slot] : fileError) {
 		buf.appendf("symbolfile=%s\n", file.c_str());
 		buf.appendf("symbolfiletype=%s\n", SymbolFile::toString(type).c_str());
+		if (slot) {
+			buf.appendf("slotsubslot=%d\n", *slot);
+		}
 	}
 }
 
@@ -65,11 +75,18 @@ void ImGuiSymbols::loadLine(std::string_view name, zstring_view value)
 	} else if (name == "symbolfile") {
 		fileError.emplace_back(std::string{value}, // filename
 		                       std::string{}, // error
-		                       SymbolFile::Type::AUTO_DETECT); // type
+		                       SymbolFile::Type::AUTO_DETECT, // type
+				       std::nullopt); // slot-subslot
 	} else if (name == "symbolfiletype") {
 		if (!fileError.empty()) {
 			fileError.back().type =
 				SymbolFile::parseType(value).value_or(SymbolFile::Type::AUTO_DETECT);
+		}
+	} else if (name == "slotsubslot") {
+		if (!fileError.empty()) {
+			if (auto slot = StringOp::stringTo<int>(value)) {
+				fileError.back().slot = slot;
+			}
 		}
 	}
 }
@@ -79,16 +96,16 @@ void ImGuiSymbols::loadEnd()
 	std::vector<FileInfo> tmp;
 	std::swap(tmp, fileError);
 	for (const auto& info : tmp) {
-		loadFile(info.filename, SymbolManager::LoadEmpty::ALLOWED, info.type);
+		loadFile(info.filename, SymbolManager::LoadEmpty::ALLOWED, info.type, info.slot);
 	}
 }
 
-void ImGuiSymbols::loadFile(const std::string& filename, SymbolManager::LoadEmpty loadEmpty, SymbolFile::Type type)
+void ImGuiSymbols::loadFile(const std::string& filename, SymbolManager::LoadEmpty loadEmpty, SymbolFile::Type type, std::optional<uint8_t> slot)
 {
 	auto& cliComm = manager.getCliComm();
 	auto it = ranges::find(fileError, filename, &FileInfo::filename);
 	try {
-		if (!symbolManager.reloadFile(filename, loadEmpty, type)) {
+		if (!symbolManager.reloadFile(filename, loadEmpty, type, slot)) {
 			cliComm.printWarning("Symbol file \"", filename,
 			                     "\" doesn't contain any symbols");
 		}
@@ -100,9 +117,23 @@ void ImGuiSymbols::loadFile(const std::string& filename, SymbolManager::LoadEmpt
 			it->error = e.getMessage(); // overwrite previous error
 			it->type = type;
 		} else {
-			fileError.emplace_back(filename, e.getMessage(), type); // set error
+			fileError.emplace_back(filename, e.getMessage(), type, it->slot); // set error
 		}
 	}
+}
+
+static bool isSlotExpanded(MSXMotherBoard* motherBoard, int ps)
+{
+	return motherBoard != nullptr ? motherBoard->getCPUInterface().isExpanded(ps) : true;
+}
+
+static std::string formatSlot(std::optional<uint8_t> slot, MSXMotherBoard* motherBoard = nullptr)
+{
+	if (!slot) return "-";
+	int ps = (*slot >> 0) & 3;
+	int ss = (*slot >> 2) & 3;
+	return isSlotExpanded(motherBoard, ps) ? strCat(ps, '-', ss)
+	                                       : strCat(ps);
 }
 
 static void checkSort(const SymbolManager& manager, std::vector<SymbolRef>& symbols)
@@ -122,15 +153,39 @@ static void checkSort(const SymbolManager& manager, std::vector<SymbolRef>& symb
 	case 1: // value
 		sortUpDown_T(symbols, sortSpecs, [&](const auto& sym) { return sym.value(manager); });
 		break;
-	case 2: // file
+	case 2: // slot
+		sortUpDown_String(symbols, sortSpecs, [&](const auto& sym) { return formatSlot(sym.slot(manager)); });
+		break;
+	case 3: // segment
+		sortUpDown_T(symbols, sortSpecs, [&](const auto& sym) { return sym.segment(manager); });
+		break;
+	case 4: // file (all symbols)
 		sortUpDown_String(symbols, sortSpecs, [&](const auto& sym) { return sym.file(manager); });
 		break;
 	default:
 		UNREACHABLE;
 	}
 }
+
+void ImGuiSymbols::drawContext(MSXMotherBoard* motherBoard, const SymbolRef& sym)
+{
+	if (ImGui::MenuItem("Set breakpoint")) {
+		std::string cond;
+		if (auto slot = sym.slot(symbolManager)) {
+			strAppend(cond, "[pc_in_slot ", *slot & 3, ' ', (*slot >> 2) & 3);
+			if (auto segment = sym.segment(symbolManager)) {
+				strAppend(cond, ' ', *segment);
+			}
+			strAppend(cond, ']');
+		}
+		BreakPoint newBp(sym.value(symbolManager), TclObject("debug break"), TclObject(cond), false);
+		auto& cpuInterface = motherBoard->getCPUInterface();
+		cpuInterface.insertBreakPoint(std::move(newBp));
+	}
+}
+
 template<bool FILTER_FILE>
-static void drawTable(ImGuiManager& manager, const SymbolManager& symbolManager, std::vector<SymbolRef>& symbols, const std::string& file = {})
+void ImGuiSymbols::drawTable(MSXMotherBoard* motherBoard, const std::string& file)
 {
 	assert(FILTER_FILE == !file.empty());
 
@@ -141,12 +196,15 @@ static void drawTable(ImGuiManager& manager, const SymbolManager& symbolManager,
 	            ImGuiTableFlags_Resizable |
 	            ImGuiTableFlags_Reorderable |
 	            ImGuiTableFlags_Sortable |
+	            ImGuiTableFlags_Hideable |
 	            ImGuiTableFlags_SizingStretchProp |
 	            (FILTER_FILE ? ImGuiTableFlags_ScrollY : 0);
-	im::Table(file.c_str(), FILTER_FILE ? 2 : 3, flags, {0, 100}, [&]{
+	im::Table(file.c_str(), (FILTER_FILE ? 4 : 5), flags, {0, 100}, [&]{
 		ImGui::TableSetupScrollFreeze(0, 1); // Make top row always visible
 		ImGui::TableSetupColumn("name");
 		ImGui::TableSetupColumn("value", ImGuiTableColumnFlags_DefaultSort | ImGuiTableColumnFlags_WidthFixed);
+		ImGui::TableSetupColumn("slot", ImGuiTableColumnFlags_DefaultHide);
+		ImGui::TableSetupColumn("segment", ImGuiTableColumnFlags_DefaultHide);
 		if (!FILTER_FILE) {
 			ImGui::TableSetupColumn("file");
 		}
@@ -159,10 +217,23 @@ static void drawTable(ImGuiManager& manager, const SymbolManager& symbolManager,
 			if (ImGui::TableNextColumn()) { // name
 				im::ScopedFont sf(manager.fontMono);
 				ImGui::TextUnformatted(sym.name(symbolManager));
+				auto symNameMenu = tmpStrCat("symbol-manager##", sym.name(symbolManager)).data();
+				if (ImGui::IsItemClicked(ImGuiMouseButton_Right)) {
+					ImGui::OpenPopup(symNameMenu);
+				}
+				im::Popup(symNameMenu, [&]{ drawContext(motherBoard, sym); });
 			}
 			if (ImGui::TableNextColumn()) { // value
 				im::ScopedFont sf(manager.fontMono);
 				ImGui::StrCat(hex_string<4>(sym.value(symbolManager)));
+			}
+			if (ImGui::TableNextColumn()) { // slot
+				im::ScopedFont sf(manager.fontMono);
+				ImGui::TextUnformatted(formatSlot(sym.slot(symbolManager), motherBoard));
+			}
+			if (ImGui::TableNextColumn()) { // segment
+				im::ScopedFont sf(manager.fontMono);
+				ImGui::TextUnformatted(sym.segment(symbolManager) ? tmpStrCat(*sym.segment(symbolManager)).c_str() : "-");
 			}
 			if (!FILTER_FILE && ImGui::TableNextColumn()) { // file
 				ImGui::TextUnformatted(sym.file(symbolManager));
@@ -171,10 +242,11 @@ static void drawTable(ImGuiManager& manager, const SymbolManager& symbolManager,
 	});
 }
 
-void ImGuiSymbols::paint(MSXMotherBoard* /*motherBoard*/)
+void ImGuiSymbols::paint(MSXMotherBoard* motherBoard)
 {
 	if (!show) return;
 
+	const auto& style = ImGui::GetStyle();
 	ImGui::SetNextWindowSize(gl::vec2{24, 18} * ImGui::GetFontSize(), ImGuiCond_FirstUseEver);
 	im::Window("Symbol Manager", &show, [&]{
 		if (ImGui::Button("Load symbol file...")) {
@@ -189,6 +261,7 @@ void ImGuiSymbols::paint(MSXMotherBoard* /*motherBoard*/)
 
 		im::TreeNode("Symbols per file", ImGuiTreeNodeFlags_DefaultOpen, [&]{
 			auto drawFile = [&](const FileInfo& info) {
+				auto* file = symbolManager.findFile(info.filename);
 				im::StyleColor(!info.error.empty(), ImGuiCol_Text, getColor(imColor::ERROR), [&]{
 					auto title = strCat("File: ", info.filename);
 					im::TreeNode(title.c_str(), [&]{
@@ -196,8 +269,45 @@ void ImGuiSymbols::paint(MSXMotherBoard* /*motherBoard*/)
 							ImGui::TextUnformatted(info.error);
 						}
 						im::StyleColor(ImGuiCol_Text, getColor(imColor::TEXT), [&]{
+							auto arrowSize = ImGui::GetFrameHeight();
+							auto extra = arrowSize + 2.0f * style.FramePadding.x;
+							ImGui::SetNextItemWidth(ImGui::CalcTextSize("3-3").x + extra);
+						        ImGui::AlignTextToFramePadding();
+
+							auto preview = formatSlot(file->slot, motherBoard);
+				                        im::Combo(tmpStrCat("Slot##", info.filename).data(), preview.c_str(), [&]{
+								// Set slot and all the symbols in it
+								auto setSlot = [&](std::optional<uint8_t> newSlot) {
+ 									file->slot = newSlot;
+									for (auto& symbol : file->getSymbols()) {
+										symbol.slot = newSlot;
+									}
+								};
+								// initial state
+								if (ImGui::Selectable("-", !file->slot)) {
+									setSlot({});
+								}
+
+								for (uint8_t ps = 0; ps < 4; ++ps) {
+									if (isSlotExpanded(motherBoard, ps)) {
+										for (uint8_t ss = 0; ss < 4; ++ss) {
+											auto slotInfo = tmpStrCat(ps, "-", ss);
+											int psSs = (ss << 2) + ps;
+											if (ImGui::Selectable(slotInfo.c_str(), psSs == file->slot)) {
+												setSlot(psSs);
+											}
+										}
+									} else {
+										if (ImGui::Selectable(tmpStrCat(ps).c_str(), file->slot && ps == file->slot)) {
+											setSlot(ps);
+										}
+									}
+								}
+							});
+							ImGui::SameLine();
+
 							if (ImGui::Button("Reload")) {
-								loadFile(info.filename, SymbolManager::LoadEmpty::NOT_ALLOWED, info.type);
+								loadFile(info.filename, SymbolManager::LoadEmpty::NOT_ALLOWED, info.type, info.slot);
 							}
 							ImGui::SameLine();
 							if (ImGui::Button("Remove")) {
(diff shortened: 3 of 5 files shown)


src/imgui/ImGuiSymbols.cc
```diff
@@ -158,11 +160,70 @@ static void drawTable(ImGuiManager& manager, const SymbolManager& symbolManager,
...
+			}
+			if (ImGui::TableNextColumn()) { // slots
+				if (sym.slots(symbolManager) == -1) {
+					im::ScopedFont sf(manager.fontMono);
+					ImGui::Selectable("all slots");
+				} else {
+					zstring_view slots{};
+					std::string_view sep{};
+					for (auto sl = 0; sl < 4; ++sl) {
+						slots = tmpStrCat(slots, sep, sl);
```
m9710797: This is a bug:
* A `string_view` or `zstring_view` is only a view into another string that exists somewhere in memory.
* The function `tmpStrCat()` creates a `TemporaryString` object. A `(z)string_view` can indeed point to such an `TemporaryString` object. But that object is not kept alive (e.g. it's not kept alive in some local variable) so when this statement ends that object gets out of scope and gets deleted. And then the `(z)string_view` points to an invalid (deleted) object. It may only appear to work fine because in this specific case (with this version of the compiler) the memory that hold the `TemporaryString` happens to be not yet reused to hold something else.

One solution could be to change the type of `slots` from `zstring_view` to `std::string`. That type really "owns" the string-content (it does not merely refer to some existing string).

Instead of using `tmpStrCat()` (which returns a `TemporaryString`) it might be better to use plain `strCat()` which returns a `std::string`. And also instead of doing `x = strCat(x, ...)` it's more efficient to do `strAppend(x, ...)`.

src/debugger/SymbolManager.hh
```diff
@@ -19,11 +19,12 @@ class CommandController;
...
 struct Symbol
 {
-	Symbol(std::string n, uint16_t v)
-		: name(std::move(n)), value(v) {} // clang-15 workaround
+	Symbol(std::string n, uint16_t v, int16_t s)
+		: name(std::move(n)), value(v), slots(s) {} // clang-15 workaround
 
 	std::string name;
 	uint16_t value;
+	int16_t slots;
```
m9710797: I know the old debugger worked like this. But is it really useful to be able to select an arbitrary set of (sub)slots? Wouldn't it be sufficient to be able to select:
* no slots (this symbol is not tied to memory).
* all slots (this symbol is valid in all slots)  (and actually do we need both no slots and all slots?).
* **one** specific slot.

I can't immediately think of a situation where one symbol applies to more than one slot, but not to all slots.

m9710797 (commented):
Hi.

First of all I think it's great that you're willing to work on this. Thanks a lot!

But I do have some concerns. So before spending too much time on fine tuning this PR, I think you should clarify how this stuff will or could be used in practice.

I do already understand that e.g. when loading a symbol file we can add an option to specify the slot for all symbols in that file (because none of the currently supported symbol file formats carry slot-information). So I'm no longer concerned about how the symbols will be annotated with slot info.

I am concerned about how that slot information will be used. For example in the old debugger it was possible to annotate symbols with slots (and I think even with a "kind" of symbol (data or code)). But that info wasn't used at all! So we had a lot of code for no purpose. In the past I had already given this some thought, and I didn't see how that approach could be made to work well (and probably that's the main reason why nobody finished that code?).

So far, in this very early(?) version of this PR, I'm seeing you take a very similar approach as was taken in the old debugger. Hence my concern, and hence my question for clarification.

I've given this some thought, and currently I only see these meaningful uses:
* The "address" column in the "disassembly" window. We can only replace the numeric address with a symbol if it matches the currently visible slot. But in particular I do NOT see how we could do the same in the "mnemonic" column. There are just too many cases where you can't decide whether or not to replace the symbol.
* When setting a breakpoint from within the symbol window (that functionality doesn't exist at the moment) we could immediately fill in the correct slot info.

(On IRC) you already mentioned the possibility to enable/disable symbols. But I don't believe in that. Whether a symbol makes sense in some context depends too much on the details of the context (e.g. the "address" vs  the "mnemonic" column in the disassembly view). And in many contexts it never makes sense to disable a symbol (e.g. in an input field). So I don't think that a single (global) enable/disable status can work.

So before you continue, I think it would be good if you could first clarify your vision. Otherwise there's a good chance large parts of this PR will have to be reworked or it might be rejected completely (e.g. because it needs a totally different approach).

Thanks.

pvmm:
> I've given this some thought, and currently I only see these meaningful uses:
> 
>     * The "address" column in the "disassembly" window. We can only replace the numeric address with a symbol if it matches the currently visible slot. But in particular I do NOT see how we could do the same in the "mnemonic" column. There are just too many cases where you can't decide whether or not to replace the symbol.
> 
>     * When setting a breakpoint from within the symbol window (that functionality doesn't exist at the moment) we could immediately fill in the correct slot info.

Things that could be easier to do, but  I don't think have ever been proposed:
* setting breakpoints or watchpoints from symbols automatically from a single button click in the symbol manager because of the previously defined slot/subslot.
* creating a new hex view in the right slot/subslot and address by clicking on a symbol that exists only in a specific slot/subslot.

Edit:
* with some tool integration, the two examples above can make debugging code for the MSX easier since the tools can set slot/subslot automatically for the developer.

pvmm:
> I am concerned about how that slot information will be used. For example in the old debugger it was possible to annotate symbols with slots (and I think even with a "kind" of symbol (data or code)). But that info wasn't used at all! So we had a lot of code for no purpose. In the past I had already given this some thought, and I didn't see how that approach could be made to work well (and probably that's the main reason why nobody finished that code?).

We can use dynamic code analysis scripts in TCL to distinguish code from data, use special annotation in the symbol manager to display, for instance, "db &lt;data&gt;" instead of code in the disassembler widget. For instance:

```
db "Hello world"
```

instead of

```
ld c,b
ld h,l
ld l,h
ld l,h
ld l,a
jr nz,#c07e
ld l,a
ld (hl),d
ld l,h
ld h,h
```

m9710797:
Such a tool would be great. BUT the current disassembly view is not the right place for it:
* An analysis for figuring out what is data and what is code is pretty complex/expensive.
* Such an analysis is hard to get correct. And depending on how "close" you want to get to "correct" is becomes still _much_ harder.
* And it' not even clear what "correct" means in case of self-modifying code. Or relocating code, Or a disk-loading or decompression routine (at one point in time the same memory locations are data, and at another time they become code).
* The whole point of the current ImGui is that it's statelless. That would imply the analysis needs to be redone at every frame (60 fps).
* Doing a reasonably good (not perfect) off-line analysis is already hard. I don't believe in doing a decent real-time analysis. 
* For sure people will complain if the tool makes a wrong decision between code and data. I personally think it's better to go for simple/predictable behavior: the disassembly view always interprets the memory as code. The memory view always interprets the memory content as data.

I don't mind adding a _separate_ viewer(*1) that works like you describe. But I think it will be much much complex than the current disassembly view. (And I think you underestimate it to be honest). 


(*1) Or maybe it's better to have this as a separate off-line tool. Then it's OK to let the tool run for a few minutes to do a whole program analysis

pvmm:
> (*1) Or maybe it's better to have this as a separate off-line tool. Then it's OK to let the tool run for a few minutes to do a whole program analysis

That's the approach I was thinking: a script tool that goes in share/scripts and the developer can use and modify to their liking. But even if it is less useful for code in RAM that can modify itself, it should still prove useful for games and programs that run exclusively on ROM which are more common anyway. Even if it is not very precise, it could be used to kick-start an annotated version of the code (that can be saved as an omds file). But for this to work, the debugger needs some kind of memory annotation support. I would use the symbol manager for this.

m9710797:
OK. That wasn't clear (to me) from your original description.

But then I'd propose to go a step further.
* It should also be possible to show comments.
* And unexpanded macros.
* In other words: you want to step trough the (original) source code.

In fact that's how many "traditional" debuggers work (e.g. gdb). They can operate in two modes:
* Either you step though a disassembly view. This always shows disassembled CPU instructions, no extra stuff.
* Or you step through the original source code.

The latter only works when there is extra debug information. This extra information "somehow"(*1) contains a mapping from every PC-value to a source file and line-number (in some cases also column number). And with accurate debug info, you can skip over comments, and can map a range of PC-values to the same unexpanded macro.
(*1) Because this can be A LOT of data, there are various different ways to encode this information is a more compact format.

Most compilers can produce such debug information. I don't know if there are any MSX assemblers that can produce this info. Maybe we can get inspiration from some MSX C compilers, most likely they can already produce some kind of debug info.

The script you're talking about could then be the combination of:
* Disassemble the full program to a "source code" file.
* Run an analysis pass to figure out what parts are data or code and update the source file accordingly.
* At the same time produce enough debug information to be able to map PC values to line numbers in the source file (or possibly multiple files).

It would be great if afterwards the user could still edit the source file (e.g. add comments, add or rename labels, or fix code/data mistakes). While preserving/updating the debug information.

In other words: these script(s) become a full reverse-engineering tool.

But I believe this should be a separate tool from openMSX. I also believe the source-code view and the disassembly view should be two completely separate viewers in the openMSX debugger. Just like in most (all?) other debuggers.

pvmm:
Based on what Artrag reported about SDCC segments in NOI files, I fixed this PR to display only the segment a symbol belongs to:
image

What if I created a first column on the left with a checkbox that allows the user to activate a breakpoint in that address/segment combination?

m9710797:
> What if I created a first column on the left with a checkbox that allows the user to activate a breakpoint in that address/segment combination?

I'm not sure I like an extra column(*1) for that functionality. Maybe right-click context menu is better suited? We actually already had a [feature request](https://github.com/openMSX/openMSX/issues/1648) for this (though it doesn't mention the desired mechanism).

This patch adds a (16-bit) segment number. But there's no indication of whether this number is meaningful or not. E.g. if I load symbols for the MSX-BIOS, there are no segment numbers. And my preference would then be to not show a dummy segment number 0. And when setting a breakpoint on such a symbol, don't add a check for this meaningless segment.

Also this patch removed the slot information again. Isn't that also required to be able to set the breakpoint?  We already discussed that the existing symbol-file formats don't carry slot-information, but this could e.g. be specified when loading the file. (But as I mentioned before, I would only allow zero or one specific slot, not a combination of multiple slots).

(*1) This is not my personal preference, but if this were to be implemented via an extra column, then I'd draw it with a small red circle, like in the disassembly view.
But also if this was an extra column, then I'd expect it to always correctly show whether there's a breakpoint set on this symbol. But the question "Is there a breakpoint on this symbol?" does not have an easy answer. For example: what if I set a breakpoint on the numerical address 0x005f. Does that then automatically result in a breakpoint on the symbol CHMOD? And what if there are multiple symbols with the same numeric value? Does setting a breakpoint on one such symbol automatically set a breakpoint on all those symbols?
If instead we only offer the option "Set a breakpoint on this symbol" in the context menu, then we don't necessarily need to indicate whether there already is such a breakpoint (thus avoiding the difficult question).

m9710797:
Thanks again for working on this. But I'd still like to ask you to explain:
* First. _What_ you're trying to achieve. Not in terms of the implementation, but from a user point of view. What functionality is missing today? My *guess* is that you want to make it easier to set breakpoints on _precise_ memory locations (thus including slot/segment) via symbols? Correct? Is this the only goal or are there more?
* Second. _How_ will this work from the _user point of view_. It's OK if this is still vague at this point. E.g. it could be stuff like "an extra column in the symbol manager with slot and segment information. And how the _user_ provide this extra information (is it somehow obtained from symbol file, does to user manually edit it, ...). This extra info is used to "..." (e.g. to set an extra condition when you create a breakpoint via the symbol manager).
* And only as the last step, explain your ideas for how this could actually be implemented in the code.

I have the impression that you're starting with this last step, without having a clear idea yet _what_ it is you're trying to achieve. (My reason for this impression is: you first add detailed slot stuff, including edit functionality, but then replace that with segment info, and no indication at all of how this info will be used).

pvmm:
> I'm not sure I like an extra column(*1) for that functionality. Maybe right-click context menu is better suited? We actually already had a [feature request](https://github.com/openMSX/openMSX/issues/1648) for this (though it doesn't mention the desired mechanism).
>
> This patch adds a (16-bit) segment number. But there's no indication of whether this number is meaningful or not. E.g. if I load symbols for the MSX-BIOS, there are no segment numbers. And my preference would then be to not show a dummy segment number 0. And when setting a breakpoint on such a symbol, don't add a check for this meaningless segment.

I though about printing every name that is not in segment 0 in italic and remove the segment column. I think that is a good way to tell the user that the symbol is different without causing visual clutter. But the user could also configure this to be visible in the symbol table and it should be turned off by default.

> Also this patch removed the slot information again. Isn't that also required to be able to set the breakpoint? We already discussed that the existing symbol-file formats don't carry slot-information, but this could e.g. be specified when loading the file. (But as I mentioned before, I would only allow zero or one specific slot, not a combination of multiple slots).

You wanted the segment column gone, but not the slot and subslot which are part of the same information? Yes, I removed it because it should be specified when loading the file. Or maybe it could stay in a hidden column that is invisible by default in the symbol table.

> (*1) This is not my personal preference, but if this were to be implemented via an extra column, then I'd draw it with a small red circle, like in the disassembly view. But also if this was an extra column, then I'd expect it to always correctly show whether there's a breakpoint set on this symbol. But the question "Is there a breakpoint on this symbol?" does not have an easy answer. For example: what if I set a breakpoint on the numerical address 0x005f. Does that then automatically result in a breakpoint on the symbol CHMOD? And what if there are multiple symbols with the same numeric value? Does setting a breakpoint on one such symbol automatically set a breakpoint on all those symbols? If instead we only offer the option "Set a breakpoint on this symbol" in the context menu, then we don't necessarily need to indicate whether there already is such a breakpoint (thus avoiding the difficult question).

As long as the symbols are useful for debugging and inspecting code in different slots, subslots and segments I have no objection.

pvmm:
> * First. _What_ you're trying to achieve. Not in terms of the implementation, but from a user point of view. What functionality is missing today? My _guess_ is that you want to make it easier to set breakpoints on _precise_ memory locations (thus including slot/segment) via symbols? Correct? Is this the only goal or are there more?

Correct, but there is more. As a game developer writing programs for the MSX, there is no easy way to inspect the contents of a ROM cartridge, specially if it has banked ROM. The "Add hex editor" is confusing because it has too many options. If a inexperienced user inserts a "file.rom" they will not understand if they should select "main RAM", "file.rom", "file.rom romblocks" or even "slotted memory". 

An easy way to address this is to create short cuts to the hex editor directly from the symbol table context menu. Even an experienced user would prefer to open a hex editor from a menu item than to navigate through layers of menus and move through pages and pages of the hex editor just to find a symbol in the symbol table. I would like to see this done for disasm widget too, but that can wait a little.

>     * Second. _How_ will this work from the _user point of view_. It's OK if this is still vague at this point. E.g. it could be stuff like "an extra column in the symbol manager with slot and segment information. And how the _user_ provide this extra information (is it somehow obtained from symbol file, does to user manually edit it, ...). This extra info is used to "..." (e.g. to set an extra condition when you create a breakpoint via the symbol manager).

The tools the developer uses to create programs (specially tools in active development) can provide most of the information about segments. They may not know the slot and subslot because this is hardware specific, but this dialog could be extended to also allow the developer to insert the corresponding symbol file and OpenMSX would already know what slot should be used:

image

>     * And only as the last step, explain your ideas for how this could actually be implemented in the code.
> 
> 
> I have the impression that you're starting with this last step, without having a clear idea yet _what_ it is you're trying to achieve. (My reason for this impression is: you first add detailed slot stuff, including edit functionality, but then replace that with segment info, and no indication at all of how this info will be used).

My first approach is to replicate what was in the debugger. But as you said, if it is not necessary to display slot, subslot and segment for each and every symbol in the symbol table, maybe it should be associated with the file from which the symbol was loaded, so that's what I had in mind.

m9710797:
> ... You wanted the segment column gone, but ...

To be clear. I don't _want_ anything. I was mainly trying to guess what you were trying to achieve, and then based on my limited (and maybe partly wrong) understanding, _suggest_ improvements.

So I'm happy that in your next post you're clarifying your goals.

pvmm:
OpenMSX now search for a debuggable in memory with the same name (without the extension) of the symbol file and open that ROM file in the hex editor if it was found.
image
* [x] ~write segment name in a different colour if segment != 0;~
* [x] context menu for symbol table items;
* [x] show segment column;
* [x] ~open hex editor pointing at the symbol address;~
* [ ] ~allow user to rename symbol by clicking on it;~
* [ ] ~save omds file;~
* [ ] ~load omds file;~
* [ ] ~ask user to save omds file if symbol was renamed;~

pvmm:
> &lt;wouterv&gt; i read a bit of the code .. seems like the "open in hex editor" function is meant for symbols from roms.   I think we should also make it work for other symbols.   e.g. symbols for the standard MSX BIOS calls or system variables

It already works for both. If you load a symbol file called `MSX BIOS with BASIC ROM.noi` with the BIOS calls at the correct addresses, the context menu will open the hex editor at that location. The same goes for system variables if you put them in a file called `memory.noi`

m9710797:
> It already works for both. If you load a symbol file called `MSX BIOS with BASIC ROM.noi` with the BIOS calls at the correct addresses, the context menu will open the hex editor at that location. The same goes for system variables if you put them in a file called `memory.noi`

Hmm, but that's very limiting. I doubt many people today already have symbol files with such names. And now that I think of it: typically in your program you'll have various lines like `CHMOD equ #005f`, so BIOS-symbols and game-symbols will often be mixed in the same symbol file.

If I read the code correctly, then I think you're using the symbol-value for the file-offset in the ROM file. That's wrong. E.g. a typical ROM mapper get mapped in the MSX address space starting from address 0x4000. So for example the entry-location for the game might be at address 0x4010, but that's at file-offset 0x10.

I don't see a solution to this problem. Not without adding extra information in the symbol file: we'd separately need the file-offset and the location in the MSX address space. And I don't see how one can be calculated from the other. Not even when you add the extra knowledge about the ROM start-address and mapper-segment-size.

Maybe you have a solution for this? But I don't. The only solution I see is to extend the symbol file to contain both values. But that's not a practical solution. To be usable it should be possible for the user to recompile his project and immediately start using the new ROM + symbol-file in openMSX (without any manual tuning or edits).

So personally I don't believe the current PR is the right approach. I also doubt you'd often want to go from a symbol to the corresponding location in the hex-editor. E.g. if you're developing a ROM and you have the source code, that source code is a much more valuable resource to lookup such stuff. And the few times that you really want to see it in a hex editor, it's not too much effort to manually calculate the file offset.

I think a much more frequent scenario is that you want to go from a symbol to the corresponding location in the disassembly view. And I think that's easily implemented by taking the (16 bits) symbol-value. (And I understand that to see the correct data, also the correct slot and segment needs to be selected).

Another scenario might be that you want to place a breakpoint on the symbol value. As a first approximation you could also just use the (16-bit) symbol value(*1). But I understand you might want to refine this with the correct slot-information (that info is currently missing in the PR). And the correct segment number.

At some point in this PR you assume that the upper-bits of the symbol-value are the segment number, but at other points you assume that the full symbol-value is the file-offset. Both cannot be true at the same time. Many symbol file formats only support 16-bit values. Some allow larger values, but I don't know what the meaning of those higher bits is. Can you point me to some documentation? In any case I think we should also support well the (very common?) case that segment information is missing (so don't treat it the same as-if it were segment=0).

-- 

(*1) Maybe we should simplify this PR (a lot), at least for an initial version. Maybe a simple first step could be:
In the symbol manager, in the symbol table, a right-click context menu with two options:
* Go to disassembly view
* Set breakpoint
In this first version both options will only look at the 16-bit symbol value (no slot or segment info). This also avoids the problem that symbol filenames need to follow a very strict naming convention.

pvmm:
> At some point in this PR you assume that the upper-bits of the symbol-value are the segment number, but at other points you assume that the full symbol-value is the file-offset.

No, I don't. As you can see in SymbolManager.hh, Symbol value is still uint16_t. So the symbol is still limited to 64kB address even if I read a 24-bit number. So I never use it as a file offset:
``` 
struct Symbol
{
        Symbol(std::string n, uint16_t v, uint16_t s)
                : name(std::move(n)), value(v), segment(s) {} // clang-15 workaround

        std::string name;
        uint16_t value;
        uint16_t segment;

        auto operator<=>(const Symbol&) const = default;
};
```
I just added segment, that is also uint16_t. There is some segment-to-offset to calculate, but I haven't written it yet because I am still figuring out the UI details.

> Both cannot be true at the same time. Many symbol file formats only support 16-bit values. Some allow larger values, but I don't know what the meaning of those higher bits is.

I cannot answer for all the symbol file formats, but in the case of NoICE files (my priority), this is the meaning of symbol-value that is bigger than 16-bit values (page 67 of SDCC manual):

> When linking your objects you need to tell the linker where to put your segments. To do this you use the following command line option to SDCC: -Wl-b BANK1=0x18000 (See 3.3.5). This sets the virtual start address of this segment. It sets the banknumber to 0x01 and maps the bank to 0x8000 and up. The linker will not check for overflows, again this is your responsibility.

In this case, segment (0x01) and bank size allow us to point the symbol to the right file offset (file offset = SEGMENT# * BANK_SIZE + ADDRESS % BANK_SIZE), while the address part (0x8000) is also used to decode the assembly properly with jumps and calls pointing to the right addresses.

> In any case I think we should also support well the (very common?) case that segment information is missing (so don't treat it the same as-if it were segment=0).

In the case of NOI files, missing segment information will not allow the compiler (and hex2bin) to produce a valid ROM file. So if a NOI file that creates a banked ROM file has segments somewhere, symbols that lack segment information mean segment 0. The omds file format can be extended to add the missing information (base address and bank size of a collection of symbols).

pvmm:
> Maybe you have a solution for this? But I don't. The only solution I see is to extend the symbol file to contain both values. But that's not a practical solution. To be usable it should be possible for the user to recompile his project and immediately start using the new ROM + symbol-file in openMSX (without any manual tuning or edits).

This is possible if we save the reference to the .NOI file, detect and reload it when it changes. The previous information about slot#-subslot# and base address is reused because it is saved as session data outside the symbol file:
image

pvmm:
> I also doubt you'd often want to go from a symbol to the corresponding location in the hex-editor. E.g. if you're developing a ROM and you have the source code, that source code is a much more valuable resource to lookup such stuff. And the few times that you really want to see it in a hex editor, it's not too much effort to manually calculate the file offset.

I disagree. Otherwise using the hex editor for a different slot/subslot has no reason for existing too, but it's already there in the "Add hex editor" submenu. And it's a short cut: there is no fast way of accessing different slot/subslot/address without selecting a ROM from a list of ROM files and scrolling through pages and pages of memory dump.

pvmm:
I've simplified this PR a bit so we can get this done already.
* [x] displays slot/subslot & segment info in columns (disabled by default);
* [x] non-zero segment is coloured yellow;
* [x] allow user to change slot/subslot for each symbol file;
* [x] save and restore slot/subslot info in/from imgui.ini;
* [x] create breakpoint with condition `[pc_in_slot <slot> <subslot> <segment>]` from symbol manager;

image

src/unittest/SymbolManager_test.cc
```diff
@@ -107,7 +107,7 @@ TEST_CASE("SymbolManager: detectType")
 TEST_CASE("SymbolManager: loadLines")
 {
 	auto dummyParser = [](std::span<std::string_view> tokens) -> std::optional<Symbol> {
-		if (tokens.size() == 1) return Symbol{std::string(tokens[0]), 123};
+		if (tokens.size() == 1) return Symbol{std::string(tokens[0]), 123, 0, 0, 0};
 		return {};
 	};
```
m9710797, line 145: Could you also extend this unittest with a few cases that test whether the segment is parsed correctly?

We briefly discussed an heuristic to set the segment to invalid when all segments in a file are 0. If/when you implement that, then please also extend the unittest.

src/debugger/SymbolManager.hh
```diff
@@ -19,11 +19,14 @@ class CommandController;
...
 {
-	Symbol(std::string n, uint16_t v)
-		: name(std::move(n)), value(v) {} // clang-15 workaround
+	Symbol(std::string n, uint16_t v, std::optional<uint16_t> s1, std::optional<uint16_t> s2, std::optional<uint16_t> s3)
+		: name(std::move(n)), value(v), slot(s1), subslot(s2), segment(s3) {} // clang-15 workaround
 
 	std::string name;
 	uint16_t value;
+	std::optional<uint16_t> slot;
+	std::optional<uint16_t> subslot;
```
m9710797: I understand why `segment` is 16 bit.  But `slot` and `subslot` shouldn't need 16 bit.

Actually, I think a lot of this PR becomes simpler if you just combine `slot` and `subslot` into a single `slot` variable (and 8 bit should be sufficient for this combination). E.g. the lower 2 bits could represent the primary slot, the next 2 bits the secondary slot.

In most cases you're just passing along slot+subslot, that becomes simpler when both are combined. Only in a few cases (e.g. displaying and setting a breakpoint) you care about primary/secondary slot separately, those few cases then need a bit more code.

Currently it's possible to have a primary slot, but no secondary slot. That won't be possible anymore if you combine both, but I also don't think it's needed. In many other places in openMSX we also only store this combination, and then by convention if a slot is not expanded, we store subslot=0. Only in the display routine we check whether the current MSX machine has the slot expanded, and if not then we don't show the subslot.

src/debugger/SymbolManager.hh
```diff
@@ -44,7 +47,10 @@ struct SymbolFile
 	};
 	[[nodiscard]] static zstring_view toString(Type type);
 	[[nodiscard]] static std::optional<Type> parseType(std::string_view str);
+	[[nodiscard]] auto& getSymbols() { return symbols; }
 
+	std::optional<uint16_t> slot;
+	std::optional<uint16_t> subslot;
```
m9710797: Same here, and in many other places (e.g. many places where we pass parameters). I won't repeat this comment for those other places.

src/debugger/SymbolManager.hh
```diff
@@ -72,14 +78,15 @@ public:
...
 	enum class LoadEmpty { ALLOWED, NOT_ALLOWED };
-	bool reloadFile(const std::string& filename, LoadEmpty loadEmpty, SymbolFile::Type type);
+	bool reloadFile(const std::string& filename, LoadEmpty loadEmpty, SymbolFile::Type type, std::optional<uint16_t> slot = {}, std::optional<uint16_t> subslot = {});
 
 	void removeFile(std::string_view filename);
 	void removeAllFiles();
 
-	[[nodiscard]] const auto& getFiles() const { return files; }
+	[[nodiscard]] auto& getFiles() { return files; }
+	[[nodiscard]] std::optional<int> findFile(std::string filename);
```
m9710797: `findFile()` returns an index that should be used in the result returned from `getFiles()`. And that's why `getFiles()` now returns a non-const result, right?

I would suggest to instead return a (non-const) `SymbolFile` pointer from `findFile()`, with `nullptr` meaning "not found". That's a more direct way (no need to go via `getFiles()`).
`getFiles()` can then again return const.

Also `findFile()` should take a `const std::string&` parameter (not a `std::string` by value). Or alternatively a `std::string_view`.

src/debugger/SymbolManager.hh
```diff
@@ -72,14 +78,15 @@ public:
...
 	[[nodiscard]] std::span<Symbol const * const> lookupValue(uint16_t value);
-	[[nodiscard]] std::optional<uint16_t> parseSymbolOrValue(std::string_view s) const;
+	[[nodiscard]] std::optional<uint32_t> parseSymbolOrValue(std::string_view s) const;
```
m9710797: I'm not 100% sure, but I think all uses of `parseSymbolOrValue()` still expect to see a 16 bit value. Or did I miss something.

Best not change stuff that doesn't need to change. It might have unexpected/unintended effects.

src/debugger/SymbolManager.hh
```diff
@@ -88,8 +95,8 @@ public:
 	// These should not be called directly, except by the unittest
 	[[nodiscard]] static std::optional<unsigned> isHexDigit(char c);
 	[[nodiscard]] static std::optional<uint16_t> is4DigitHex(std::string_view s);
-	[[nodiscard]] static std::optional<uint16_t> parseValue(std::string_view str);
-	[[nodiscard]] static std::optional<Symbol> checkLabel(std::string_view label, uint16_t value);
+	[[nodiscard]] static std::optional<uint32_t> parseValue(std::string_view str);
+	[[nodiscard]] static std::optional<Symbol> checkLabel(std::string_view label, uint32_t value);
```
m9710797, line 98: For these two I agree they should change from 16 to 32 bit.

src/debugger/SymbolManager.cc
```diff
@@ -99,6 +99,8 @@ SymbolManager::SymbolManager(CommandController& commandController_)
 	SymbolFile result;
 	result.filename = filename;
 	result.type = type;
+	result.slot = std::nullopt;
+	result.subslot = std::nullopt;
```
m9710797: No need to explicitly set `slot`/`subslot` (or later the combined `slot` field) to `std::nullptr`. It is already initialized to that value. (Similar to how `result.symbols` is already initialized to an empty vector.)

Same remark for lines 205, 239 and 302.

src/debugger/SymbolManager.cc
```diff
@@ -351,21 +359,35 @@ SymbolManager::SymbolManager(CommandController& commandController_)
...
+				return loadLinkMap(filename_, buffer_);
+			case NOICE:
+				return loadNoICE(filename_, buffer_);
+			case VASM:
+				return loadVASM(filename_, buffer_);
+			default: UNREACHABLE;
+		}
+	};
+
+	auto symbolFile = loadFile(filename, buffer, type);
```
m9710797: This is indeed a good use case for a lambda function. But it can be simplified to:
```
auto symbolFile = [&]{
    switch (type) {
        case ...
            return ...;
    }
}();
```
Notice:
* No parameters for the lambda (everything is passed via the capture list).
* We don't give the lambda a name (no `loadFile`).
* We immediately invoke this unnamed lambda (the `()` at the end).

src/imgui/ImGuiSymbols.hh
```diff
@@ -17,23 +17,29 @@ struct SymbolRef {
...
-			: filename(std::move(f)), error(std::move(e)), type(t) {} // clang-15 workaround
+		FileInfo(std::string f, std::string e, SymbolFile::Type t, std::optional<int> s, std::optional<int> ss, int i)
+			: filename(std::move(f)), error(std::move(e)), type(t), slot(s), subslot(ss), index(i) {} // clang-15 workaround
 
 		std::string filename;
 		std::string error;
 		SymbolFile::Type type;
+		std::optional<int> slot;
+		std::optional<int> subslot;
+		int index; // slot-subslot index
```
m9710797: (I'd suggest to also here combine `slot` and `subslot` into a single field.)

I don't think you need the `index` field. It should always contain the same information as the (combined) `slot` field. In other words: it's redundant information.

I understand that in your current implementation you need this because of `ImGui::Combo()`. I'll come back to that later (we can use something better than `ImGui::Combo()`.

src/imgui/ImGuiSymbols.hh
```diff
@@ -48,7 +54,16 @@ public:
 	bool show = false;
 
 private:
-	void loadFile(const std::string& filename, SymbolManager::LoadEmpty loadEmpty, SymbolFile::Type type);
+	bool showSlot = false;
+	bool showSeg = false;
```
m9710797: Remove these two boolean. They (effectively) aren't used.

src/imgui/ImGuiSymbols.cc
```diff
@@ -45,10 +50,20 @@ void ImGuiSymbols::save(ImGuiTextBuffer& buf)
 	for (const auto& file : symbolManager.getFiles()) {
 		buf.appendf("symbolfile=%s\n", file.filename.c_str());
 		buf.appendf("symbolfiletype=%s\n", SymbolFile::toString(file.type).c_str());
+		if (file.slot && file.subslot) {
+			buf.appendf("slotsubslot=%d-%d\n", *file.slot, *file.subslot);
+		} else if (file.slot) {
+			buf.appendf("slotsubslot=%d\n", *file.slot);
+		}
```
m9710797, line 54: When `slot` and `subslot` are combined, this code becomes a lot simpler.
This is not meant to be a human readable file format. So it's not a problem that e.g. "3-2" is instead saved as "11" (2 * 4 + 3).

src/imgui/ImGuiSymbols.cc
```diff
@@ -65,12 +80,29 @@ void ImGuiSymbols::loadLine(std::string_view name, zstring_view value)
...
+			int slot;
+			int subslot;
+			char dash;
+
+			if (ss >> slot) {
+				fileError.back().slot = slot;
+				if (ss >> dash && dash == '-' && ss >> subslot) {
+					fileError.back().subslot = subslot;
+				}
+			}
```
m9710797, line 89: And then especially this parsing code becomes a lot simpler.

src/imgui/ImGuiSymbols.cc
```diff
@@ -100,12 +132,12 @@ void ImGuiSymbols::loadFile(const std::string& filename, SymbolManager::LoadEmpt
...
 			it->type = type;
 		} else {
-			fileError.emplace_back(filename, e.getMessage(), type); // set error
+			fileError.emplace_back(filename, e.getMessage(), type, it->slot, it->subslot, it->index); // set error
 		}
 	}
 }
 
-static void checkSort(const SymbolManager& manager, std::vector<SymbolRef>& symbols)
+static void checkSort(SymbolManager& manager, std::vector<SymbolRef>& symbols)
```
m9710797: As discussed on IRC: this function should be adapted for the extra columns (it currently crashes).

Also make `SymbolManager` const again. That's possible after you make `getFiles()` const again. (And otherwise it was better to have both a const and a non-const `getFiles()`).

src/imgui/ImGuiSymbols.cc
```diff
@@ -129,8 +161,19 @@ static void checkSort(const SymbolManager& manager, std::vector<SymbolRef>& symb
 		UNREACHABLE;
 	}
 }
+
+void ImGuiSymbols::drawContext(MSXMotherBoard* motherBoard, const SymbolRef& sym)
+{
+	if (ImGui::MenuItem("Set breakpoint", nullptr, nullptr, bool(sym.slot(symbolManager)))) {
```
m9710797: When `slot()` returns `std::nullopt` we should still allow to set a breakpoint. But then a breakpoint without an extra `pc_in_slot` condition.

src/imgui/ImGuiSymbols.cc
```diff
@@ -129,8 +161,19 @@ static void checkSort(const SymbolManager& manager, std::vector<SymbolRef>& symb
 		UNREACHABLE;
 	}
 }
+
+void ImGuiSymbols::drawContext(MSXMotherBoard* motherBoard, const SymbolRef& sym)
+{
+	if (ImGui::MenuItem("Set breakpoint", nullptr, nullptr, bool(sym.slot(symbolManager)))) { 
+		std::string cond = strCat("[pc_in_slot ", sym.slot(symbolManager) ? narrow<int>(*sym.slot(symbolManager)) : 0, ' ', sym.subslot(symbolManager) ? narrow<int>(*sym.subslot(symbolManager)) : 0, ' ', sym.segment(symbolManager) ? narrow<int>(*sym.segment(symbolManager)) : 0, ']');
```
m9710797: This is one of the few places where a combined primary+secondary `slot` field is a bit more work. Here you'll have to split that field again. Note that it is allowed to pass `0` (or any value) for the secondary slot when the primary slot is not expanded. So you do not have to test here whether or not the slot is expanded, just extract the secondary slot value and pass it.

src/imgui/ImGuiSymbols.cc
```diff
@@ -141,12 +184,15 @@ static void drawTable(ImGuiManager& manager, const SymbolManager& symbolManager,
 	            ImGuiTableFlags_Resizable |
 	            ImGuiTableFlags_Reorderable |
 	            ImGuiTableFlags_Sortable |
+	            ImGuiTableFlags_Hideable |
```
m9710797, line 199: Very good. But now _all_ columns are hideable. Below we should probably pass `ImGuiTableColumnFlags_NoHide` to the `name` and `value` column. Maybe it's OK if the `file` column is now also hideable.

src/imgui/ImGuiSymbols.cc
```diff
@@ -158,11 +204,32 @@ static void drawTable(ImGuiManager& manager, const SymbolManager& symbolManager,
 
 			if (ImGui::TableNextColumn()) { // name
 				im::ScopedFont sf(manager.fontMono);
-				ImGui::TextUnformatted(sym.name(symbolManager));
+				im::StyleColor(sym.segment(symbolManager) && *sym.segment(symbolManager) > 0, ImGuiCol_Text, getColor(imColor::YELLOW), [&]{
```
m9710797: You're calling `sym.segment(symbolManager)` twice. That's not super expensive, but also not needed. And very easy to avoid: do it once and store the result in a local variable.

Though in this specific case there's a useful shortcut: you can directly compare a `std::optional<uint16_t>` with `> 0` (a `std::nullopt`-value compares smaller than any other `uint16_t` value).

But I also wonder why it's useful to show non-zero segments in a different color. What's special about segment zero. Especially after you try harder to detect symbol files with no segment information (and put those to `std::nullopt`).

src/imgui/ImGuiSymbols.cc
```diff
@@ -158,11 +204,32 @@ static void drawTable(ImGuiManager& manager, const SymbolManager& symbolManager,
...
+					ImGui::TextUnformatted(sym.name(symbolManager));
+				});
+				auto symNameMenu = tmpStrCat("symbol-manager##", sym.name(symbolManager)).data();
+				if (ImGui::IsItemClicked(ImGuiMouseButton_Right)) {
+					ImGui::OpenPopup(symNameMenu);
+				}
+				im::Popup(symNameMenu, [&]{ drawContext(motherBoard, sym); });
```
m9710797, line 225: We discussed on IRC to try to rework this so that right-click in the whole row work, not only on the text item itself.

src/imgui/ImGuiSymbols.cc
```diff
@@ -158,11 +204,32 @@ static void drawTable(ImGuiManager& manager, const SymbolManager& symbolManager,
...
 			}
 			if (ImGui::TableNextColumn()) { // value
 				im::ScopedFont sf(manager.fontMono);
-				ImGui::StrCat(hex_string<4>(sym.value(symbolManager)));
+				ImGui::TextUnformatted(tmpStrCat(hex_string<4>(sym.value(symbolManager))).c_str());
```
m9710797: Why did you change this from `ImGui::StrCat(...)` to `ImGui::TextUnformatted(tmpStrCat(...).c_str())`? A left over from another experiment?

src/imgui/ImGuiSymbols.cc
```diff
@@ -158,11 +204,32 @@ static void drawTable(ImGuiManager& manager, const SymbolManager& symbolManager,
...
+			}
+			if (ImGui::TableNextColumn()) { // slot
+				im::ScopedFont sf(manager.fontMono);
+				if (sym.slot(symbolManager) && sym.subslot(symbolManager)) {
+					ImGui::TextUnformatted(tmpStrCat(narrow<int>(*sym.slot(symbolManager)), "-", narrow<int>(*sym.subslot(symbolManager))));
+				} else if (sym.slot(symbolManager)) {
+					ImGui::TextUnformatted(tmpStrCat(*sym.slot(symbolManager)));
+				} else {
+					ImGui::TextUnformatted("-");
+				}
```
m9710797: This is the 2nd place where a combine primary+secondary slot field would be a bit more work. And here it would be useful to test whether the primary slot is actually expanded in the current MSX machine.

src/imgui/ImGuiSymbols.cc
```diff
@@ -171,10 +238,23 @@ static void drawTable(ImGuiManager& manager, const SymbolManager& symbolManager,
...
+{
+	if (!slot) return 0;
+
+	auto& cpuInterface = motherBoard->getCPUInterface();
+	int slotIndex = 1;
+	for (int ps = 0; ps < slot; ++ps) {
+		slotIndex += cpuInterface.isExpanded(ps) ? 4 : 1;
+	}
+	return slotIndex + (subslot ? *subslot : 0);
+}
```
m9710797: When we remove the `index` field from `FileInfo` this helper function isn't needed anymore. More details below.

Note that `motherBoard` can be `nullptr` (when there's no active MSX machine). In that case this function would have crashed.  (When there's no active machine you can e.g. assume that all primary slots are expanded).

src/imgui/ImGuiSymbols.cc
```diff
@@ -188,16 +268,58 @@ void ImGuiSymbols::paint(MSXMotherBoard* /*motherBoard*/)
...
+									nSlots++;
+								}
+							}
+							auto arrowSize = ImGui::GetFrameHeight();
+							auto extra = arrowSize + 2.0f * style.FramePadding.x;
+							ImGui::SetNextItemWidth(ImGui::CalcTextSize("3-3").x + extra);
+						        ImGui::AlignTextToFramePadding();
+
+							// slot
+							ImGui::Combo(tmpStrCat("Slot##", info.filename).data(), &info.index, slotInfo.c_str(), nSlots);
```
m9710797: You're using `ImGui::Combo()` and because of that you have to construct the `slotInfo` string and store `info.index`. This `ImGui::Combo()` function is a good idea if the list of items in the combo-box is fixed. But for a more dynamic list it's not recommended.

Instead you can use `ImGui::BeginCombo()`, `ImGui::EndCombo()` (or better our `im::Combo()` wrapper around it) combined with `ImGui::Selectable()`. Search the code for some examples.

The basic idea is: with `im::Combo()` you start a combo box context. Inside that context you have your loop of primary and optionally secondary slot. For each existing slot you make a call to `ImGui::Selectable()` (depending in `info.slot` you pass the correct value for the `selected` parameter to `Selectable()`. `Selectable()` also returns a `bool` result. If `true` you update the `info.slot` field to the current slot (i mean the slot in your for loop), and also only then you need to update the slot field in the symbols.

src/imgui/ImGuiSymbols.cc
```diff
@@ -188,16 +268,58 @@ void ImGuiSymbols::paint(MSXMotherBoard* /*motherBoard*/)
...
+								auto& file = symbolManager.getFiles()[*fileIdx];
+								// detect slot/subslot changes
+								if (file.slot != slotSubslots[info.index].first || file.subslot != slotSubslots[info.index].second) {
+									file.slot = slotSubslots[info.index].first;
+									file.subslot = slotSubslots[info.index].second;
+									for (auto& symbol: file.getSymbols()) {
+										symbol.slot = slotSubslots[info.index].first;
+										symbol.subslot = slotSubslots[info.index].second;
+									}
+								}
```
m9710797, line 290: I mean this stuff. You only need to do this when the slot combo box was actually changed. And that's what the `bool` result from `ImGui::Selectable()` tells you.

m9710797 (changes requested):
Thanks again for working on this. I think this PR is getting much closer to being ready. However I do have several remarks:
* I think a lot can be simplified.
* I didn't repeat (all) of what we discussed on IRC (that also requires some code changes).

I'll do another review once you've addressed my comments.

src/debugger/SymbolManager.hh
pvmm: The compiler doesn't complain, but on SymbolManager.cc:452 parseSymbolValue ends in:
```
        // also not found, then try to parse as a numerical value
        return parseValue(str);
}               
```
which returns an optional 32-bit value.

src/imgui/ImGuiSymbols.cc
pvmm: I don't follow how you got 11. To be generic, I would encode it like this:

slot 0 covers 0,1,2,3
slot 1 covers 4,5,6,7
slot 2 covers 8,9,10,11
slot 3 covers 12,13,14,15

So I would write "3-2" as "14" ((3)*4 + (2)).

m9710797: You're right.  I mistakenly swapped prim/sec slots.

m9710797:
Thanks! I had a _very_ quick look, and I see a lot of improvements! But I don't have enough time right now for a full review.

Hopefully I can make time for that this evening. Then I can also actually try/run this code.

src/imgui/ImGuiSymbols.cc
pvmm: Code is getting a bit hard to maintain. Maybe `Reactor::getMotherBoard()` should implement the [null object pattern](https://en.wikipedia.org/wiki/Null_object_pattern) and replace `motherBoard` with a dummy object instead of returning a `nullptr`.

src/debugger/SymbolManager.cc
```diff
@@ -168,15 +198,26 @@ SymbolManager::SymbolManager(CommandController& commandController_)
...
 		auto def   = tokens[0];
 		auto label = tokens[1];
 		auto value = tokens[2];
 		if (StringOp::casecmp cmp; !cmp(def, "def")) return {};
-		return checkLabelAndValue(label, value);
+		// special case: detecting segment information above 16bits
+		if (parseValue32(value) > 0xFFFF) {
+			segmentFlag = true;
+		}
+		return checkLabelSegmentAndValue(label, value);
```
m9710797: Detail, optimization:
You call `parseValue32()` and then soon after inside `checkLabelSegmentAndValue()` `parseValue32()` gets called again. Maybe rewrite like:
```
auto symbol = checkLabelSegmentAndValue(...);
anySegments |= symbol.segment.has_value();
return symbol;
```

src/debugger/SymbolManager.cc
```diff
@@ -168,15 +198,26 @@ SymbolManager::SymbolManager(CommandController& commandController_)
...
 	};
-	return loadLines(filename, buffer, SymbolFile::Type::NOICE, parseLine);
+	auto file = loadLines(filename, buffer, SymbolFile::Type::NOICE, parseLine);
+	if (segmentFlag) {
+		for (auto& symbol: file.getSymbols()) {
+			symbol.segment = symbol.segment ? symbol.segment : 0;
```
m9710797: Detail, personal preference.
I took me a moment to interpret this expression (but i think it's correct).
Maybe this is clearer (at least it is to me):
`if (!symbol.segment) symbol.segment = 0;`
Maybe also add one or two sentence comments to document the heuristic. Something like:
`// Heuristic: if all segments in the symbol file are 0, then assume the file contains no segment information.`

src/debugger/SymbolManager.cc
```diff
@@ -351,21 +392,31 @@ SymbolManager::SymbolManager(CommandController& commandController_)
...
+				return loadLinkMap(filename, buffer);
+			case NOICE:
+				return loadNoICE(filename, buffer);
+			case VASM:
+				return loadVASM(filename, buffer);
+			default: UNREACHABLE;
+		}
+	}();
+
+	// Update slot and subslot info for every file and symbol
```
m9710797: Detail:
* `and subslot`: not wrong, but maybe confusing because now there's only a single `slot` member anymore (yes, internally that does encode also the subslot, but here that's not relevant).
* `every file`: it's only one file, no?

src/debugger/SymbolManager.cc
```diff
@@ -114,6 +114,27 @@ SymbolManager::SymbolManager(CommandController& commandController_)
...
+	if (str.starts_with('%')) { // bin
+		str.remove_prefix(1);
+		return StringOp::stringToBase<2, uint32_t>(str);
+	}
+	// this recognizes the prefixes "0x" or "0X" (for hexadecimal)
+	//                          and "0b" or "0B" (for binary)
+	// no prefix in interpreted as decimal
+	// "0" as a prefix for octal is intentionally NOT supported
+	return StringOp::stringTo<uint32_t>(str);
+}
```
m9710797, line 137: It's unfortunate that this function is now duplicated (once for 16 bit, once for 32 bit). But never mind. After this PR is merged I'll turn this into a template that can be instantiated for either case.

src/imgui/ImGuiSymbols.cc
```diff
@@ -65,12 +76,22 @@ void ImGuiSymbols::loadLine(std::string_view name, zstring_view value)
...
 				SymbolFile::parseType(value).value_or(SymbolFile::Type::AUTO_DETECT);
 		}
+	} else if (name == "slotsubslot") {
+		if (!fileError.empty()) {
+			std::istringstream ss(value.data());
+			int slot;
+
+			if (ss >> slot) {
+				fileError.back().slot = slot;
+			}
```
m9710797, line 89: This is now parsing a simple `int`. We have helper functions for that:
```
if (auto slot = StringOp::stringTo<int>(value)) {
    fileError.back().slot = slot;
}
```
(And then remove the `#include <sstream>` again.)

src/imgui/ImGuiSymbols.cc
```diff
@@ -122,15 +143,44 @@ static void checkSort(const SymbolManager& manager, std::vector<SymbolRef>& symb
 	case 1: // value
 		sortUpDown_T(symbols, sortSpecs, [&](const auto& sym) { return sym.value(manager); });
 		break;
-	case 2: // file
+	case 2: // file or slot
+		sortUpDown_String(symbols, sortSpecs, [&](const auto& sym) { return sym.file(manager); });
+		break;
```
m9710797, line 158: This is weird (and wrong?):
shouldn't this sort on slot instead of file?

src/imgui/ImGuiSymbols.cc
```diff
@@ -122,15 +143,44 @@ static void checkSort(const SymbolManager& manager, std::vector<SymbolRef>& symb
...
+
+void ImGuiSymbols::drawContext(MSXMotherBoard* motherBoard, const SymbolRef& sym)
+{
+	if (ImGui::MenuItem("Set breakpoint", nullptr, nullptr, motherBoard != nullptr)) {
+		std::string cond;
+		if (sym.slot(symbolManager)) {
+			strAppend(cond, "[pc_in_slot ", narrow<int>(*sym.slot(symbolManager) & 3), ' ', narrow<int>((*sym.slot(symbolManager) >> 2) & 3));
+			if (sym.segment(symbolManager)) {
+				strAppend(cond, ' ', narrow<int>(*sym.segment(symbolManager)));
+			}
```
m9710797, line 178: * You repeat the expressions `sym.slot(symbolManager)` and `sym.segment(symbolManager)` multiple times. That's not wrong, and not super expensive. But best to avoid that anyway.
* I also don't immediately see why you need that `narrow<int>(...)`, because the expression inside is already an `int`, no?

 Maybe write it like this:
```
if (auto slot = sym.slot(symbolManager)) {
    strAppend(cond, "[pc_in_slot ", *slot & 3, ' ', (*slot >> 2) & 3);
    if (auto segment = sym.segment(symbolManager)) {
        strAppend(cond, ' ', *segment);
    }
}
```

src/imgui/ImGuiSymbols.cc
```diff
@@ -141,14 +191,17 @@ static void drawTable(ImGuiManager& manager, const SymbolManager& symbolManager,
...
 		ImGui::TableSetupScrollFreeze(0, 1); // Make top row always visible
-		ImGui::TableSetupColumn("name");
-		ImGui::TableSetupColumn("value", ImGuiTableColumnFlags_DefaultSort | ImGuiTableColumnFlags_WidthFixed);
+		ImGui::TableSetupColumn("name", ImGuiTableColumnFlags_NoHide);
+		ImGui::TableSetupColumn("value", ImGuiTableColumnFlags_NoHide | ImGuiTableColumnFlags_DefaultSort | ImGuiTableColumnFlags_WidthFixed);
+		ImGui::TableSetupColumn("slot", ImGuiTableColumnFlags_DefaultHide);
+		ImGui::TableSetupColumn("segment", ImGuiTableColumnFlags_DefaultHide);
 		if (!FILTER_FILE) {
-			ImGui::TableSetupColumn("file");
+			ImGui::TableSetupColumn("file", ImGuiTableColumnFlags_NoHide);
```
m9710797: I'm not sure, but maybe it is useful to be able to hide the `file` column?
Hiding the `name` and `value` columns should not be possible I think.

src/imgui/ImGuiSymbols.cc
```diff
@@ -158,23 +211,48 @@ static void drawTable(ImGuiManager& manager, const SymbolManager& symbolManager,
...
 			}
+			if (ImGui::TableNextColumn()) { // slot
+				im::ScopedFont sf(manager.fontMono);
+				if (motherBoard != nullptr && sym.slot(symbolManager)) {
+					auto ps = *sym.slot(symbolManager) & 3;
+					if (isSlotExpanded(motherBoard, ps)) {
+						auto ss = (*sym.slot(symbolManager) >> 2) & 3;
+						ImGui::TextUnformatted(tmpStrCat(ps, "-", ss));
+					} else {
+						ImGui::TextUnformatted(tmpStrCat(ps));
```
m9710797: Detail:
The combination of `ImGui::TextUnformatted()` and `tmpStrCat()` is very common. And that's exactly what our `ImGui::StrCat()` helper function does.

Very very much a detail:
For single characters inside `StrCat()` prefer `'-'` instead of `"-"`. (Generates a tiny bit more efficient code.)

src/imgui/ImGuiSymbols.cc
```diff
@@ -189,15 +267,60 @@ void ImGuiSymbols::paint(MSXMotherBoard* /*motherBoard*/)
...
 							ImGui::TextUnformatted(info.error);
 						}
 						im::StyleColor(ImGuiCol_Text, getColor(imColor::TEXT), [&]{
+							auto arrowSize = ImGui::GetFrameHeight();
+							auto extra = arrowSize + 2.0f * style.FramePadding.x;
+							ImGui::SetNextItemWidth(ImGui::CalcTextSize("3-3").x + extra);
+						        ImGui::AlignTextToFramePadding();
+
+							// build slot-subslot preview string
+							std::string preview = file->slot ? (isSlotExpanded(motherBoard, *file->slot & 3) ? tmpStrCat(*file->slot & 3, '-', (*file->slot >> 2) & 3).c_str() : tmpStrCat(*file->slot & 3).c_str()) : tmpStrCat("-").c_str();
```
m9710797: This is a very long line, not easy to see what's going on. It also contains some weird stuff:
* You repeat the same subexpressions multiple times. Local variables are useful ;-)
* `tmpStrCat("-").c_str()` is just `"-"`, no?
* You want the result as `std::string`, then you should use `strCat()` not `tmpStrCat()` (this saves a conversion from `TemporaryString` to `const char*` (via `.c_str()`) to `std::string`.
```
auto preview = [&]{
    if (file->slot) {
        int ps = *file->slot & 3;
        int ss = (*file->slot >> 2) & 3;
        return isSlotExpanded(motherBoard, ps) ? strCat(ps, '-', ss)
                                               : strCat(ps)
    } else {
        return std::string("-");
    }
}();
```
Sure that's a bit longer than 1 line, but IMHO much clearer.

src/imgui/ImGuiSymbols.cc
```diff
@@ -189,15 +267,60 @@ void ImGuiSymbols::paint(MSXMotherBoard* /*motherBoard*/)
...
+
+				                        im::Combo(tmpStrCat("Slot##", info.filename).data(), preview.c_str(), [&]{
+								int psSs = 0;
+
+								// initial state
+								if (ImGui::Selectable("-", !file->slot)) {
+									file->slot = {};
+									for (auto& symbol: file->getSymbols()) {
+										symbol.slot = file->slot;
+									}
```
m9710797: This loop is repeated 3 times (here on line 291, and also on lines 302 and 311).
Maybe factor it out in a small lambda function, and call this helper 3 times;
```
auto setSlot = [&](std::optional<uint8_t>& newSlot) {
    file->slot = newSlot;
    for (auto& symbol : file->getSymbols()) {
        symbol.slot = newSlot;
    }
};
```

src/imgui/ImGuiSymbols.cc
```diff
@@ -189,15 +267,60 @@ void ImGuiSymbols::paint(MSXMotherBoard* /*motherBoard*/)
...
+								}
+
+								for (uint8_t ps = 0; ps < 4; ++ps) {
+									if (isSlotExpanded(motherBoard, ps)) {
+										for (uint8_t ss = 0; ss < 4; ++ss) {
+											auto slotInfo = tmpStrCat(ps, "-", ss);
+											if (ImGui::Selectable(slotInfo.c_str(), file->slot && psSs == file->slot)) {
```
m9710797: Detail: it is allowed to compare a `std::optional` containing `std::nullopt`.
So `file->slot && psSS == file->slot` can be simplified to just `psSS == file->slot`.

src/imgui/ImGuiSymbols.cc
```diff
@@ -189,15 +267,60 @@ void ImGuiSymbols::paint(MSXMotherBoard* /*motherBoard*/)
...
+												file->slot = psSs;
+												for (auto& symbol: file->getSymbols()) {
+													symbol.slot = file->slot;
+												}
+											}
+											psSs += 1 << 2; // add to subslot bits
```
m9710797: I think this is wrong. After the 4 sub slots have been processed, you want `psSS` to be increased by `1` not by `16`.
I think it's simpler to not try to calculate it incrementally. Just write `int psSS = ps + (ss << 2);` inside the for loop over `ss`. And in the slot-not-expanded case below, just replace `psSS` with `ps`.

src/unittest/SymbolManager_test.cc
```diff
@@ -107,7 +115,7 @@ TEST_CASE("SymbolManager: detectType")
 TEST_CASE("SymbolManager: loadLines")
 {
 	auto dummyParser = [](std::span<std::string_view> tokens) -> std::optional<Symbol> {
-		if (tokens.size() == 1) return Symbol{std::string(tokens[0]), 123};
+		if (tokens.size() == 1) return Symbol{std::string(tokens[0]), 123, 0, 0};
 		return {};
 	};
```
m9710797, line 145: Thanks for expanding the unit-test.
Could you also add a test for the heuristic:
* if all segments in a file are 0, then assume the file has no information on segments.

In other words:
* Add a test where all segments are 0. This should result in no segments (segment is `std::nullopt`).
* Add a test with a mix of 0 and non-0 segments. Now the 0-segments should be 0 (not `std::nullopt`).

m9710797 (commented):
Several small comments. A few (maybe) bugs. But most are about small details. This clearly indicates the PR is getting close to being ready. Thanks a lot for your persistence!

src/imgui/ImGuiSymbols.cc
```diff
@@ -122,15 +139,44 @@ static void checkSort(const SymbolManager& manager, std::vector<SymbolRef>& symb
 	case 1: // value
 		sortUpDown_T(symbols, sortSpecs, [&](const auto& sym) { return sym.value(manager); });
 		break;
-	case 2: // file
+	case 2: // slot
+		sortUpDown_String(symbols, sortSpecs, [&](const auto& sym) { return sym.file(manager); });
```
m9710797: You fixed the comment, but the code itself still sorts on file instead of slot.

src/imgui/ImGuiSymbols.cc
```diff
@@ -189,15 +263,69 @@ void ImGuiSymbols::paint(MSXMotherBoard* /*motherBoard*/)
...
+							auto preview = [&]{
+								if (file->slot) {
+									int ps = *file->slot & 3;
+									int ss = (*file->slot >> 2) & 3;
+									return isSlotExpanded(motherBoard, ps) ? strCat(ps, '-', ss)
+									                                       : strCat(ps);
+								} else {
+									return std::string("-");
+								}
+							}();
```
m9710797: I didn't notice this in my review yesterday (sorry).
This `preview` helper function is almost the same as the logic that draws the slot column (around line 224).
Maybe both should use the same helper?
Note: currently both are similar but not exactly the same (different behavior when `motherBoard == nullptr`), but that difference is probably not intended (a bug?).
```
static std::string formatSlot(std::optional<uint8_t> slot, MSXMotherBoard* motherBoard)
{
    if (!slot) return "-";
    int ps = (*slot >> 0) & 3;
    int ss = (*slot >> 2) & 3;
    return isSlotExpanded(motherBoard, ps) ? strCat(ps, '-', ss)
                                           : strCat(ps);
}
```
And then calls as:
* `ImGui::TextUnformatted(formatSlot(sym.slot(symbolManager), motherBoard));` for the slot column (this immediately avoids the repeated calls to `sym.slot(...)`).
* `auto preview = formatSlot(file->slot, motherBoard);` for the combo-box preview.

src/imgui/ImGuiSymbols.cc
```diff
@@ -189,15 +262,56 @@ void ImGuiSymbols::paint(MSXMotherBoard* /*motherBoard*/)
...
+								for (uint8_t ps = 0; ps < 4; ++ps) {
+									if (isSlotExpanded(motherBoard, ps)) {
+										for (uint8_t ss = 0; ss < 4; ++ss) {
+											auto slotInfo = tmpStrCat(ps, "-", ss);
+											if (ImGui::Selectable(slotInfo.c_str(), psSs == file->slot)) {
+												setSlot(psSs);
+											}
+											psSs += 1 << 2; // add to subslot bits
+										}
+										psSs++;
```
m9710797: I still believe this is wrong. The 2x2 bits that represent the ps and ss are now correct, but after the loop over ss and after this statement `psSS` has increased by 17. So bits 4-7 are no longer zero. And that will cause problems in the comparison `psSS == file->slot` because that compares all 8 bits.

You could fix that by adding an extra `& 15` in the comparison. But I think it's simpler to:
* remove lines 292, 300, 302, 307
* add this after line 296: int psSs = (ss << 2) + ps;
* in line 304 change `psSS` to `ps`

m9710797 (commented):
Very close to ready now. Only one bug(?) remaining.

pvmm:
What about this weird unit test error? The error message doesn't make much sense.

m9710797:
> What about this weird unit test error? The error message doesn't make much sense.

Ah sorry, I didn't notice that yet.

Add these 3 lines right after the implementation of `parseValue()` in `SymbolManager.cc`. I tested it: that fixes it for me.
```
// explicitly instantiate for uint16_t and uint32_t (needed for unittest)
template std::optional<uint16_t> SymbolManager::parseValue<uint16_t>(std::string_view);
template std::optional<uint32_t> SymbolManager::parseValue<uint32_t>(std::string_view);
```

The reason for the error is that the translation-unit where the unittest gets compiled doesn't see the template definition of `getValue()` (it only sees the declaration). So it cannot create new instances of that template, and thus it can only call already exitsing instantiations of the template.

You could say that inside `SymbolManager.cc` the template does get instantiated for both `uint16_t` and `uint32_t`, so the unittest doesn't need *new* instantiations. In a way that's true, however ... Templates are implicitly `inline` so the compiler is not required to keep a non-inlined version around. The instantiation for `uint16_t` is called twice, the one for `uint32_t` is only called once (I mean inside `SymbolManager.cc`). If a function is only called once, there's no reason not to inline that function (i mean it always leads to faster and smaller code). When the function is called multiple times there's a tradeoff (it might be faster, but likely it also makes the total code size larger). So it seems that the compiler (and then specifically this version of the compiler) decided to inline the `uint32_t` version, but not the `uint16_t` version. And that's why you get linking errors on `uint32_t` but not on `uint16_t`.

These 3 extra lines explicitly instantiate the template for `uint16_t` and `uint32_t`. This forces the compiler to create a non-inlined version. And that then resolves the linker error.


EDIT:
An alternative fix could be to move the definition of parseValue() from the `.cc` to the `.hh` file. Then the unittest does see the definition, and then it can create _new_ instantiations.

pvmm:
Seems like Clang doesn't trigger the unit test error.

m9710797:
> Seems like Clang doesn't trigger the unit test error.

Yes, I tried to explain that in my previous message: it's because of arbitrary inline decisions. Different compilers, or different versions of the same compiler may make other decisions.

pvmm:
I guess it's ready. Can we merge this?

PR #1646, opened 2024-06-04
[macOS] Raise main window to front on start-up

grauw:
If I launch openMSX with additional windows opened (like the debugger & console), the openMSX window does not receive focus. Only the additional windows show in front, and openMSX window itself goes behind the window that opened it (such as the terminal or VSCode). Also key input is not received by the MSX until I click on the main window to focus it. It works fine as normal when there are no additional windows opened.

These are actually two separate issues:

1. When there are multiple windows open on startup, then by default the main msx window should start with focus, not some other window.
2. When starting openMSX it can happen that the main msx window opens behind some other (non-openmsx) window.

### DIFF
--- a/src/imgui/ImGuiLayer.cc
+++ b/src/imgui/ImGuiLayer.cc
@@ -32,9 +32,8 @@ void ImGuiLayer::paint(OutputSurface& /*surface*/)
 		ImGuiDockNodeFlags_NoDockingOverCentralNode |
 		ImGuiDockNodeFlags_PassthruCentralNode);
 
-	if (first) {
+	if (ImGui::GetFrameCount() == 1) {
 		// on startup, focus main openMSX window instead of the GUI window
-		first = false;
 		ImGui::SetWindowFocus(nullptr);
 	}
 
--- a/src/imgui/ImGuiLayer.cc
+++ b/src/imgui/ImGuiLayer.cc
@@ -51,6 +51,14 @@ void ImGuiLayer::paint(OutputSurface& /*surface*/)
 		ImGui::UpdatePlatformWindows();
 		ImGui::RenderPlatformWindowsDefault();
 		SDL_GL_MakeCurrent(backup_current_window, backup_current_context);
+
+#ifdef __APPLE__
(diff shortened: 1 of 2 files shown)


grauw:
Ok, my test was not good enough it seems. My focus stealing issue only occurs if the debugger memory view is open. I don’t know how I was then able to reproduce it with just the console open. Maybe it was intermittent? But I can’t repro that at all now anymore.

Wouterv rightfully remarked that the “fix” to focus was, in fact, equivalent to the original code, so I’ve removed that commit.

grauw:
In the third patch I resolve the remaining issue where if I boot up openMSX with anything open, it does not have focus as demonstrated by the gray OS window title bar. If I have the debugger memory view open then that’s clearly the one taking it, however also when I have the console open more often than not I have the gray title bar (it seems kinda random).

I’m not super satisfied with this patch though;

1. Firstly, this magic value. Why not frame 2.
2. Secondly, the delay (the user may be quick and be doing something already).
3. Thirdly, I’m not sure if this focus stealing part of the issue is macOS-only.

grauw:
I notice now that if I have the console open in addition to the main window, then cmd-tab away to a different window, then cmd-tab back to openMSX, the console always gets the focus, always. Perhaps ImGui thinks the main window isn’t focusable or something, and that’s what lies underneath these focus issues…

Sorry for the messy PR, it’s a bit of a quagmire :).

grauw:
Ok, I’m taking off the last patch, trying to fix the focus issues is much too complicated at this point for me to address, and goes into the depths of ImGui. Even opening the Help menu in 2x scale mode removes focus from the main window (title bar turns gray), because it falls outside the main window and thus creates a new one. That fix also is also not a good one at all. Perhaps better to just wait for the ImGui focus refactor.

I’m quite ok with the second patch though. The fact that the main window doesn’t get pulled to the foreground seems to be a macOS ideosyncracy when an app opens multiple windows on start-up, and this seems a reasonable fix until at least until ImGui may possibly do it by itself sometime in the future?

@m9710797 Are these two commits ok with you?

m9710797:
Merged. Thanks!

PR #1661, opened 2024-06-04
Fix error passing nonexistent texture to ImGui::Image().

grauw:
Caused the following warning to be logged on start-up on M1 Macs:

> UNSUPPORTED (log once): POSSIBLE ISSUE: unit 0 GLD_TEXTURE_INDEX_2D is unloadable and bound to sampler type (Float) - using zero texture because texture unloadable

### DIFF
--- a/src/imgui/ImGuiOsdIcons.cc
+++ b/src/imgui/ImGuiOsdIcons.cc
@@ -197,11 +197,13 @@ void ImGuiOsdIcons::paint(MSXMotherBoard* /*motherBoard*/)
 			}();
 
 			const auto& ic = state ? icon.on : icon.off;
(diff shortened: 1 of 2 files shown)


m9710797:
Thanks! I'm sure this PR fixes the issue for you, but I'm concerned that it fixes the symptom and not the root cause.
How come that texture was 0 for you? Was it because you explicitly disabled some OSD icons?
(I don't have time right now to investigate myself, I'll do that this evening).

grauw:
It is the pause & throttle OSD icons, they don't have an image for the off state. This is intentional, they should disappear when off. However the code would still try to use the texture unconditionally even if one was not specified.

m9710797:
Merged. Thanks!

PR #1662, opened 2024-06-08
Add support for Circuit Designer RD dongle

sndpl:
This PR adds support for a new dongle that is needed to start the Circuit Designer RD software of The Falcon (Rene Derkx). If the dongle is not present it will add a reset command to the BLOAD hook (that will reset the MSX if you try to load something).

Thanks to Tsjakoe and Lars The 18Th for finding the copy protection and info on how it works.

See: https://www.generation-msx.nl/software/the-falcon/circuit-designer-rd/release/8309/

### DIFF
--- a/doc/manual/user.html
+++ b/doc/manual/user.html
@@ -983,7 +983,7 @@ Tetris 2 Special Edition from R.A.M., an Italian MSX group needs the dongle in p
     Circuit Designer dongle from The Falcon, needs the dongle in port B too start the program:</p>
 
 <div class="commandline">
-    <a class="external" href="commands.html#plugunplug">plug</a> joyportb circuitdesignerrd-protection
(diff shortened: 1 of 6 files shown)


m9710797:
Thanks!

I only have one question about the name. You use both:
* circuitdesignerrd-protection
* Circuit Designer RD dongle

Why mix the terms "protection" and "dongle"? Why put a hyphen between "rd" and "protection", but not between "circuit" and "designer"?

In other words: is it OK if I rename "circuitdesignerrd-protection" to "circuit-designer-rd-dongle"?

sndpl:
@m9710797 well, since i'm not a C++ dev, I made a copy of the Tetris dongle and that one is called 'tetris2-protection', so I changed the tetris2 into circuitdesignerrd :) But I like your proposal better, so I changed the name in the code and the manual.

m9710797:
Merged, with some minor tweaks/corrections (e.g. also add the new files in the msvc build files).
Thanks!

PR #1663, opened 2024-06-08
AmdFlash: Miscellaneous improvements

grauw:
A couple of miscellaneous AmdFlash improvements that I think are good for review and push.

### DIFF
--- a/src/memory/AmdFlash.cc
+++ b/src/memory/AmdFlash.cc
@@ -47,21 +47,28 @@ AmdFlash::AmdFlash(const std::string& name, const ValidatedChip& validatedChip,
 	: motherBoard(config.getMotherBoard())
 	, chip(validatedChip.chip)
 {
-	auto numSectors = chip.geometry.sectorCount;
-	assert(writeProtectSectors.size() <= numSectors);
+	assert(writeProtectSectors.size() <= chip.geometry.sectorCount);
+
+	sectors.reserve(chip.geometry.sectorCount);
+	for (size_t address = 0; const Region& region : chip.geometry.regions) {
+		for (size_t regionSector = 0; regionSector < region.count; ++regionSector, address += region.size) {
+			sectors.push_back({
+				.address = address,
+				.size = region.size,
+				.writeProtect = sectors.size() < writeProtectSectors.size() && writeProtectSectors[sectors.size()]
+			});
+		}
+	}
 
 	size_t writableSize = 0;
 	size_t readOnlySize = 0;
-	writeAddress.resize(numSectors);
-	for (size_t sector = 0; const AmdFlash::Region& region : chip.geometry.regions) {
-		for (size_t regionSector = 0; regionSector < region.count; regionSector++, sector++) {
-			if (sector < writeProtectSectors.size() && writeProtectSectors[sector]) {
-				writeAddress[sector] = -1;
-				readOnlySize += region.size;
-			} else {
-				writeAddress[sector] = narrow<ptrdiff_t>(writableSize);
-				writableSize += region.size;
-			}
+	for (Sector& sector : sectors) {
+		if (sector.writeProtect) {
+			sector.writeAddress = -1;
+			readOnlySize += sector.size;
+		} else {
+			sector.writeAddress = narrow<ptrdiff_t>(writableSize);
+			writableSize += sector.size;
 		}
 	}
 	assert((writableSize + readOnlySize) == size());
@@ -124,43 +131,39 @@ AmdFlash::AmdFlash(const std::string& name, const ValidatedChip& validatedChip,
 		}
 	}
 
-	readAddress.resize(numSectors);
 	auto romSize = rom ? rom->size() : 0;
 	size_t offset = 0;
-	for (size_t sector = 0; const AmdFlash::Region& region : chip.geometry.regions) {
-		for (size_t regionSector = 0; regionSector < region.count; regionSector++, sector++) {
-			auto sectorSize = region.size;
-			if (isSectorWritable(sector)) {
-				readAddress[sector] = &(*ram)[writeAddress[sector]];
-				if (!loaded) {
-					auto* ramPtr = const_cast<uint8_t*>(
-						&(*ram)[writeAddress[sector]]);
-					if (offset >= romSize) {
-						// completely past end of rom
-						ranges::fill(std::span{ramPtr, sectorSize}, 0xFF);
-					} else if (offset + sectorSize >= romSize) {
-						// partial overlap
-						auto last = romSize - offset;
-						auto missing = sectorSize - last;
-						const uint8_t* romPtr = &(*rom)[offset];
-						ranges::copy(std::span{romPtr, last}, ramPtr);
-						ranges::fill(std::span{&ramPtr[last], missing}, 0xFF);
-					} else {
-						// completely before end of rom
-						const uint8_t* romPtr = &(*rom)[offset];
-						ranges::copy(std::span{romPtr, sectorSize}, ramPtr);
-					}
-				}
-			} else {
-				assert(rom); // must have rom constructor parameter
-				if ((offset + sectorSize) <= romSize) {
-					readAddress[sector] = &(*rom)[offset];
+	for (Sector& sector : sectors) {
+		if (isWritable(sector)) {
+			sector.readAddress = &(*ram)[sector.writeAddress];
+			if (!loaded) {
+				auto* ramPtr = const_cast<uint8_t*>(
+					&(*ram)[sector.writeAddress]);
+				if (offset >= romSize) {
+					// completely past end of rom
+					ranges::fill(std::span{ramPtr, sector.size}, 0xFF);
+				} else if (offset + sector.size >= romSize) {
+					// partial overlap
+					auto last = romSize - offset;
+					auto missing = sector.size - last;
+					const uint8_t* romPtr = &(*rom)[offset];
+					ranges::copy(std::span{romPtr, last}, ramPtr);
+					ranges::fill(std::span{&ramPtr[last], missing}, 0xFF);
 				} else {
-					readAddress[sector] = nullptr;
+					// completely before end of rom
+					const uint8_t* romPtr = &(*rom)[offset];
+					ranges::copy(std::span{romPtr, sector.size}, ramPtr);
 				}
 			}
-			offset += sectorSize;
+		} else {
+			assert(rom); // must have rom constructor parameter
+			if ((offset + sector.size) <= romSize) {
+				sector.readAddress = &(*rom)[offset];
+			} else {
+				sector.readAddress = nullptr;
+			}
 		}
+		offset += sector.size;
 	}
 	assert(offset == size());
 
@@ -169,18 +172,19 @@ AmdFlash::AmdFlash(const std::string& name, const ValidatedChip& validatedChip,
 
 AmdFlash::~AmdFlash() = default;
 
-AmdFlash::GetSectorInfoResult AmdFlash::getSectorInfo(size_t address) const
+size_t AmdFlash::getSectorIndex(size_t address) const
 {
 	address &= size() - 1;
-	auto it = chip.geometry.regions.begin();
-	size_t sector = 0;
-	while (address >= it->count * it->size) {
-		address -= it->count * it->size;
-		sector += it->count;
-		++it;
-		assert(it != chip.geometry.regions.end());
+	size_t sectorIndex = 0;
+	for (const Region& region : chip.geometry.regions) {
+		if (address < region.count * region.size) {
+			return sectorIndex + address / region.size;
+		} else {
+			address -= region.count * region.size;
+			sectorIndex += region.count;
+		}
 	}
-	return {sector + address / it->size, it->size, address % it->size};
+	UNREACHABLE;
 }
 
 void AmdFlash::reset()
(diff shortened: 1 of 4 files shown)


src/memory/AmdFlash.hh
```diff
@@ -12,6 +12,7 @@
 #include <cstdint>
 #include <memory>
 #include <span>
+#include <vector>
```
m9710797, line 16: Detail: #include "MemBuffer.hh" is then no longer needed.

src/memory/AmdFlash.hh
```diff
@@ -227,12 +228,25 @@ public:
...
+	struct Sector {
+		const size_t address = 0;
+		const power_of_two<size_t> size = 1;
+		const bool writeProtect = false;
+		ptrdiff_t writeAddress = -1;
+		const uint8_t* readAddress = nullptr;
+
+		std::weak_ordering operator<=>(const Sector& sector) const { return address <=> sector.address; }
+		bool operator==(const Sector& sector) const = default;
+	};
```
m9710797, line 245: clang-tidy gives warnings about the 3 `const` members in this struct:
```
member 'address' of type 'const size_t' (aka 'const unsigned long') is const qualifiedclang-tidy(cppcoreguidelines-avoid-const-or-ref-data-members)
```
Last year I would also have made these members `const`, now I tend to agree with the guildines, although I still think it's unfortunate. See here for more details:
* https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/avoid-const-or-ref-data-members.html
* https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-constref

Summary: having `const` members makes the struct non-copy-assignable, and non-movable (it's still is copy-constructable). And when you put such objects in a vector, it's a useful optimization if you can move (instead of copy). (Though in this specific case move and copy are basically the same, but who knows what other members will be added in the future).

m9710797, line 245: I like how you (correctly) thought about the return type for `operator<=>`.

src/memory/AmdFlash.hh
```diff
@@ -227,12 +228,25 @@ public:
...
+
 private:
 	AmdFlash(const std::string& name, const ValidatedChip& chip,
 	         const Rom* rom, std::span<const bool> writeProtectSectors,
 	         const DeviceConfig& config);
-	struct GetSectorInfoResult { size_t sector, sectorSize, offset; };
-	[[nodiscard]] GetSectorInfoResult getSectorInfo(size_t address) const;
+
+	[[nodiscard]] size_t getSectorIndex(size_t address) const;
+	[[nodiscard]] Sector& getSector(size_t address) { return sectors[getSectorIndex(address)]; };
```
m9710797, line 253: The non-const version of `getSector()` isn't needed ATM (I checked). And it also shouldn't be needed in the future (I think).

src/memory/AmdFlash.cc
```diff
@@ -47,21 +47,28 @@ AmdFlash::AmdFlash(const std::string& name, const ValidatedChip& validatedChip,
...
+	sectors.reserve(chip.geometry.sectorCount);
+	for (size_t address = 0; const Region& region : chip.geometry.regions) {
+		for (size_t regionSector = 0; regionSector < region.count; ++regionSector, address += region.size) {
+			sectors.push_back({
+				.address = address,
+				.size = region.size,
+				.writeProtect = sectors.size() < writeProtectSectors.size() && writeProtectSectors[sectors.size()]
+			});
+		}
+	}
```
m9710797, line 54: While reviewing this code, my immediate concern was that this only partially initializes the `Sector` objects. But when I continued reading I did see that also `writeAddress` and `readAddress` is correctly initialized.

Would it be cleaner to merge this loop and the loop below (the one that initializes `writeAddress`)? I mean it is technically possible, and not difficult, but I wonder if it makes the code cleaner.

And then the next question is: is it possible, and cleaner, to also include initialization of `readAddress` in the same loop? Maybe not.

m9710797, line 54: I like how this is now the only remaining location where we have this non-trivial nested loop.

src/memory/AmdFlash.cc
```diff
@@ -436,9 +440,9 @@ uint16_t AmdFlash::peekCFI(size_t address) const
 	}
 }
 
-bool AmdFlash::isSectorWritable(size_t sector) const
+bool AmdFlash::isWritable(const Sector& sector) const
 {
-	return vppWpPinLow && (sector == one_of(0u, 1u)) ? false : (writeAddress[sector] != -1) ;
+	return vppWpPinLow && (sector == one_of(sectors[0], sectors[1])) ? false : !sector.writeProtect;
```
m9710797: This may go wrong when there is only a single sector (thus `sectors[1]` is out of bounds). But I guess that's not possible. And I also see that this code is still re-written in later patches.

m9710797 (commented):
AmdFlash: Add sectors vector.

src/memory/AmdFlash.cc
```diff
@@ -134,7 +134,7 @@ AmdFlash::AmdFlash(const std::string& name, const ValidatedChip& validatedChip,
 	auto romSize = rom ? rom->size() : 0;
 	size_t offset = 0;
 	for (Sector& sector : sectors) {
-		if (isWritable(sector)) {
+		if (sector.writeAddress != -1) {
```
m9710797: Maybe it's useful to add a comment that it's not correct to use `isWritable()`.

src/memory/AmdFlash.cc
```diff
@@ -442,7 +442,14 @@ uint16_t AmdFlash::peekCFI(size_t address) const
 
 bool AmdFlash::isWritable(const Sector& sector) const
 {
-	return vppWpPinLow && (sector == one_of(sectors[0], sectors[1])) ? false : !sector.writeProtect;
+	if (vppWpPinLow) {
+		const auto range = chip.misc.writeProtectPinRange;
+		if ((range > 0 && sector.address <  sectors[range + size_t(0)].address) ||
```
m9710797: Is it ever possible that `abs(writeProtectPinRange) == sectorCount`? It should never be strictly greater, but can it be equal?
If that's possible then there's an out-of-bounds access (for `range > 0`, `range < 0` happens to work).

m9710797 (commented):
AmdFlash: Write protect pin improvements.

src/memory/AmdFlash.cc
```diff
@@ -676,39 +676,37 @@ bool AmdFlash::checkCommandQuadrupleByteProgram()
...
 	static constexpr std::array<uint8_t, 2> cmdSeq = {0xaa, 0x55};
-	if (chip.program.bufferCommand && partialMatch(cmdSeq)) {
-		if (cmd.size() <= 2) return true;
-		if (cmd.size() >= 3 && cmd[2].value == 0x25) {
-			if (cmd.size() <= 3) return true;
+	if (chip.program.bufferCommand && partialMatch(cmdSeq) && (cmd.size() <= 2 || cmd[2].value == 0x25)) {
+		if (cmd.size() <= 3) {
+			return true;
+		} else if (cmd.size() <= 4) {
+			if (cmd[3].value < chip.program.pageSize && getSector(cmd[3].addr) == getSector(cmd[2].addr)) {
```
m9710797, line 663: This patch itself is fine. Though while reading it, i was still thinking about `Sector::operator==()` (introduced in the first patch in this PR).

That operator compares the full struct, and that's correct. But wouldn't it also be correct to only compare the `address` field? My reasoning: no two sectors within the same AmdFlash have the same `address`, so that field by itself is already sufficient to uniquely identify the sector.

And that would also make `operator==` and `operator<=>` 'consistent'. Now that I think about it, why do you need `operator<=>`? It isn't used (I checked). Remove it?

m9710797 (commented):
AmdFlash: Improve checkCommandBufferProgram().

src/memory/AmdFlash.cc
```diff
@@ -455,6 +455,7 @@ bool AmdFlash::isWritable(const Sector& sector) const
 uint8_t AmdFlash::read(size_t address)
 {
 	const uint8_t value = peek(address);
+	address %= size();
```
m9710797, line 434: This is weird. I would have expected `addess %= size()` as the first statement in this method.
Even more weird: `address` isn't used anymore after this line, so this statement has no effect.

I'm fine with either dropping this line (`peek()` already does the same), or moving it to the start.

src/memory/AmdFlash.hh
grauw: It will be needed, because I’m going to add modifyable members, such as:

* `erasePending` To track which sectors are marked for erase. It is possible to issue multiple sector erase commands in one go (under a time-out). Currently working on that as well as implementing proper timing in general.
* `modified` To track which sectors have been modified from the original. On the backburner a bit, but will come back to that soon.

src/memory/AmdFlash.cc
grauw: I don’t think it necessarily makes it cleaner. Now the code sections have a single responsibility. The write/read address logic is just a bit too complex to all blend together into a single loop.

Additionally I have a patch queued up which will actually get rid of `writeAddress` and `readAddress`, instead using a single TrackedRAM to store (and persist) the Flash memory. That will simplify things, but it’s not ready for submission yet. Keeping it as it is now will make that diff a bit cleaner.

src/memory/AmdFlash.cc
grauw: To my knowledge there is no hardware where the write protect pin protects the entire memory, but I’ll strengthen the checks and logic to make sure it doesn’t go out of bounds regardless.

src/memory/AmdFlash.cc
grauw: Yeah, indeed never occurs in practice, and consider it a temporary “hack” just to keep it working, that’s addressed in the next patch.

m9710797 (commented):
AmdFlash: Modulo addresses to the size

I agree with this change, because indeed the hardware behaves like this.

--

Though in general I don't like "defensive programming". I'll try to explain:

IMHO better terminology is "a function with a wide contract" (~= defensive programming) versus "a function with a narrow contract".

Functions with a wide contract have no preconditions. It's valid to call these with any input. The function never crashes, it always computes _some_ result.

Functions with a narrow contract do have preconditions. It is possible to violate the precondition (call it with invalid values). But that's a bug in the program (more specifically a bug in the caller). And then the function may do whatever. Preferably it will crash (hopefully with a meaningful error message, e.g. an assert), but it may also return a _bogus_ result.

At first sight it may seem that functions with a wide contract are safer, and lead to more correct programs. But I think the opposite is true. Let's give an example:
```
// wide contract
double sqrt(double x) {
    if (x < 0.0) return 0.0;
    return compute_sqrt(x);
}

// narrow contract
double sqrt(double x) {
    assert(x >= 0.0);
    return compute_sqrt(x);
}
```
(Depending on the context) trying to calculate the square-root of a negative value is a bug. The program does not become more correct if we give `sqrt()` a wide contract. On the contrary this may only hide the bug. If instead the program crashes it's much more likely the bug will be discovered/fixed.

Of course `sqrt()` is maybe a too simple example. But I think the idea of a wide vs narrow contract is generally applicable. And having a narrow contract (=preconditions) helps with reasoning about bugs (*1) and locating them (violating preconditions means a bug in the caller).

(*1) E.g. a static-analysis tool may flag code that will result in an assert. But it cannot do that if instead `sqrt()` had a wide contract, maybe returning zero is the correct behavior (i mean it cannot know that it's not)?

--

But again in this context your patch is fine. The real hardware really ignores the upper address bits (those address lines are simply not connected). So it's indeed as-if the hardware has a wide contract.

m9710797 (commented):
AmdFlash: Remove empty SRAM check

OK, no comments.

src/memory/AmdFlash.cc
grauw: Hmm, I was under the assumption that `bool operator==(const Sector& sector) const = default;` used the `operator<=>` implementation, but that’s not the case. I’ll fix it.

I can make use of the `operator<=>` in `isWritable()`.

src/memory/AmdFlash.cc
grauw: In `peek()` it also masks, so if I do it here as well then it masks twice. I figured I’d place it a line below in case at some point in the future something is added to `read()` after the `peek()` call which also uses the address, as a reminder of sorts. At the moment indeed it’s not used so never actually compiled in.

But it’s a bit strange I agree, also what if I want to do something with address before the call to `peek()`. The cost of the `&` operation is negligible, so I’ll just move it to the front. I would like to keep it present in all the public entry points.

> At first sight it may seem that functions with a wide contract are safer, and lead to more correct programs. But I think the opposite is true.

I definitely agree with you. And ideally it’d be constrained through the type system, the robustness of `assert()` is somewhat lacking in my opinion since it’s compiled out.

Here I used the phrase “defensive programming” to mean that it avoids a class of potential issues if the narrow contract is not sufficiently enforced elsewhere in the code. But it’s not the primary motivation of the change.

src/memory/AmdFlash.cc
m9710797: IIRC in the initial proposal for operator<=> it worked like that. operator<=> would be used to derive an operator== (i think even without an explicit `operator==() =default` request). But that was removed from the proposal before it was approved for the c++20 standard.

The reason is performance. Often a dedicated operator== can be (much) faster than using operator<=> to derive operator==. And the fear was that if writing operator== was no longer needed, too many people would omit it (and then get a working but slower version). Many other programming languages (e.g. Rust) also work like this: you need to both implement equality and comparison.

A typical example is a string class. When given two strings with different size, operator== can immediately return false. But operator<=> must loop over the characters until it finds a mismatch.

src/memory/AmdFlash.hh
grauw: Hmm yes unfortunate. I’ll change it.

PR #1707, opened 2024-07-14
code analyzer widget

pvmm:
Inspired by [Spectrum Analyser](https://colourclash.co.uk/spectrum-analyser/), I wrote this mockup of a similar widget for OpenMSX.
image

Use case: the user wants to better understand a program or game in ROM by letting the emulator execute the code and make annotations about what it found.

The `Activate analyzer` checkbox calls a Tcl script while the program is running on the virtual machine. The script annotates portions of the memory as `CODE` or `DATA` according to how it is used. The user can later rename the automatic symbols created by the inspection. They are named automatically as something like `function@4010` for instance. The `Code parameters` section (for a lack of a better name for now), allows the user to define comments or how immediate values are displayed to the user (decimal, hexadecimal, octal or binary) on an instruction-by-instruction basis.

### DIFF
--- a/share/scripts/codeanalyzer.tcl
+++ b/share/scripts/codeanalyzer.tcl
@@ -1,0 +1,1091 @@
+namespace eval codeanalyzer {
+
+# TODO:
+# * detect when interrupt routine is called;
+# * finish all z80 instructions;
+#   ** INI,IND,INIR,INDR,OUTI,OUTIR,OUTD,OUTDR
+# * full BIOS support;
+# * detect ROM size (16k, 32k or MEGAROM);
+# * detect segment type (ROM or memory mapper) and usage when executing code or reading/writing data;
+# * detect code copying to RAM and include it in the analysis;
+# * detect instructon size using disassembler;
+# * write assembly output to file;
+# * annotate unscanned code/data when writing output file;
+# * allow user to set markers on the code;
+# * try to detect functions;
+#   * detect call/ret combinations;
+# * try to detect all types of RAM-to-VRAM memory copying:
+#   * pattern generator table (PGT)
+#   * name table;
+#   * colour table;
+#   * sprite generator table;
+#   * sprite attribute table;
+# * try to detect keyboard input;
+# * try to detect joystick port input;
+# * try to detect sound generation (PSG);
+
+variable mem_type ""
+
+variable t ;# memory type array
+variable l ;# label array
+variable c ;# comment array
+variable x ;# extended address
+variable b ;# bios label array
+variable pc
+variable tmp_pc {}
+variable slot
+variable segment
+variable ss ""
+variable is_mapper 0
+variable r_wp {}
+variable w_wp {}
+variable start_point {}
+variable entry_point {}
+variable end_point 0xBFFF ;# end of page 2
+variable comment "" ;# stored comment message
+variable label_size 14
+variable cmt_pos    19
+variable error {} ;# error propagation
+
+variable hkeyi_wp ;# HKEYI watchpoint
+variable hkeyi {} ;# HKEYI routine address
+variable intpc 0  ;# interrupted PC address
+
+# bookkeeping
+variable old_pc {}
+variable last_mem_read {}
+
+# info
+variable DATA_recs 0 ;# data records
+variable CODE_recs 0 ;# code records
+variable BOTH_recs 0 ;# both DATA and CODE records
+
+# constants
+proc NAUGHT	{} { return {}}
+proc CODE	{} { return 1 }
+proc DATA	{} { return 2 }
+proc BOTH	{} { return 3 }
+proc HKEY_HOOK  {} { return 64924 } ;# last position of HKEY (0xfd9c)
+
+set_help_proc codeanalyzer [namespace code codeanalyzer_help]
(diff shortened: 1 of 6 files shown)


m9710797:
Hi pvmm,

Good idea. Thanks for working on this. Though I'd like to give some suggestions with as goal:
* On the openMSX side:
** Make this feature more general (make it solve more problems than just this specific one).
** Make the implementation simpler? (Of course that's just an educated guess at this moment).
* Outside of openMSX (the Tcl script in your proposal):
** Make the functionality more generally usable, also for non-openMSX programs.

About the code analysis (reverse engineering) functionality. You give some examples:
* Annotate chunks as code or as data.
* Generate symbols names with (manual) rename possibility.
* Format data as decimal, binary, hex, ...

That's a good start, but it's not enough:
* You want to be able to add comments. A long comment at the start of a line (e.g. before a function). Or a short comment after an instruction or data line.
* Add blank lines. E.g. to separate the different steps within a large function.
* Some data should be formatted in 8 columns (e.g. sprite pattern data), others in 4 column (e.g. sprite attribute data), or various other formats. Sometimes it's most descriptive to use a mix of decimal, binary, hex within the same data block.
* Some code or data is very repetitive, and you may want to use macros to simplify it.

In other words: you start from fully disassembled code. And then, step-by-step, transform this to a form that's closer to what the original source code could have been. For this you want to allow arbitrary formatting with arbitrary edits in the (reverse engineered) code. But at any intermediate step you still want to be able to step through it in the (openMSX) debugger.

In modern tool-chains that's a solved problem. E.g. when I compile my c++ code with gcc, I can then use gdb to either:
* Step through it at the CPU-instruction-level. (This part we already have in openMSX).
* Step through my _original_ c++ code, thus including all comments and the exact formatting. Including data sections in exactly the same way as written/formatted in my source code.
* Maybe one line in the source code corresponds to multiple CPU instructions (e.g. a macro). You can step a single CPU-instruction, or step over the current source code line.

For this to work the debugger needs extra information, this is called "debug information". What and how exactly this debug info is represented is a complex topic. But in it's simplest form it's a mechanism to map every possible (valid) program counter (in our case PC plus possibly a segment-number) to a source filename and line number.

With (only) this mapping and the source code, we have sufficient information to do source-level debugging. At any point (after any break, step-in, step-out, step-over) we can map the current PC to a source file and the exact line within that file. And then the debugger can show this (original or reverse engineered) source code (with whatever formatting that code was using, including complex stuff like macros).

Above I talk about "modern tool-chains", but actually source-level debugging already exists for a _very_ long time. I wouldn't be surprised if also some of the existing MSX tools (assemblers or C compilers) can produce some form of debug information. Someone (you?) should do a survey of what kind of debug info already exists. Unless really needed, we shouldn't invent our own thing. E.g. a listing file can be a primitive form of debug information, but maybe something better exists as well.

So I propose to replace this "code analyzer" widget with a "source level debugging" feature.
* This solves multiple requests with the same feature.
* This keeps the openMSX code simpler: we only need to implement a "viewer" for debug information.

All edits to the (reverse engineered) source code and corresponding debug information are intentionally left outside of openMSX.
* Because this can become very complex very quickly.
* In your original proposal for this PR you invoked a Tcl script for this. So you basically did the same (move the complex part outside of openMSX).
* There's no advantage with integrating this functionality inside openMSX. For non-trivial edits an external editor will always be more comfortable than the builtin ImGui editors.

As already mentioned: the automatic analysis part can be pretty complex. Will you develop some version of this? It might be good to have extra help/volunteers for this (not me, sorry). And it's probably easier to attract such volunteers if this is a standalone project (not integrated in openMSX). So something that produces useful results also outside of openMSX.

---

My comments above were mostly from an openMSX point of view. In this part I'll give some comment mostly from a reverse engineering tool point of view.
These are very early thoughts, I literally only thought about this for a few minutes. Probably much less than the time you spend on it (so apologies if I missed something important).

Here's my starting point:
* You start from a .ROM file (but could also be a .COM or .BIN file).
* You can (automatically) disassemble this .ROM file. But it's important that this disassembled file is in a format that can be re-assembled again. And that it re-produces 100% the exact same .ROM file.
* I'm also assuming that the assembler can produce some form of debug information. So you can map (all/most) parts of the .ROM file to a line in the source code.

With this starting point and these assumptions:
* You have one main source file. You can do arbitrary edits in it. And important, you can verify that the edits are valid: after assembling it should still produce the exact same output.
* And after any (valid) edit you still have debug info (the assembler can re-generate it). So you always keep a correct mapping between binary and source code.

That's already sufficient for manual edits. But automatic analysis and transformations are also possible:
* A tool can look at either the binary code or the source code (both are guaranteed in sync, and you always have an up-to-date mapping between the two). So the analysis can be done on whichever format is most suitable.
* On that page about "Spectrum Analyser" I read something about an emulator-trace being used to annotate sections as code. This could also work for openMSX: openMSX can _export_ a file with a trace of addresses classified as opcode-fetches (most likely code), as reads, or as writes (most likely data). Then the analysis tool can combine this trace-file with the above data and make some transformations. And then we can re-assemble to verify that the transformation was correct and to update the debug info.
* My point is that we don't need to integrate this functionality in openMSX. A much smaller openMSX-feature: "export a trace file" is sufficient.

Issue #1702, opened 2024-07-01
Debugger: symbols that are slot/segment sensitive


MBilderbeek:
@turbor wrote:

> Symbols depending on current memory layout/slot selection/active rompage in cartridge. I'm currently working on a bigger project, and several routines have the same address but in different rompages, having the correct symbol name would be nice. At the moment if multiple symbols have the same value you're never sure if you see the correct name. Of course, if there are multiple rompages that have routines at the same address, you are never sure if you display the correct label especially if you have some interslot call routines (Remember the MSX BIOS :RST #30,DB destslot,DW destaddress). But otherwise when stepping through your code that you are debugging using the symbolnames with regards to the current slotlayout should be "close enough". 

I think @pvmm is working on something like this via https://github.com/openMSX/openMSX/pull/1632 - but let's keep track of this functionality in this issue.

pvmm:
Currently OpenMSX ignores segment information from the symbol manager when it sets labels in the disasm widget. So PR #1632 didn't fix this, because it only sets break points,  but I will take a look.

pvmm:
@turbor, I think #1709 solves parte of the problem, but not all of it.

m9710797:
pvmm, can you talk about what part(s) are still missing. And maybe if you have any idea's for how to tackle those?

pvmm:
The mnemonic column in the disassembler will always display symbols instead of addresses (`jp MYLABEL`) even if MYLABEL is in a different segment and we know this may not be wrong. Maybe symbol attributes like "local" or "global" can help us determine whether the replacement makes sense. These attributes can be saved as session variables, so they are not lost when a symbol file is rewritten/reloaded.

m9710797:
I 100% agree that the current situation is not always correct (that we show the wrong label in some cases).
Unfortunately I don't see a good and practical solution. Extra annotations will of course help, but the problem I see with that is that none(?) of the current symbol file formats carry that information, and then the user will have to manually annotate this information. And sure that info can be preserved over different sessions. But when you're still developing your program (and the debugger is most useful), the labels will change and then you have to redo these manual annotations.

The following proposal is only a _partial_ solution. But is it automatic (I mean doesn't require any manual work), and it may already solve +80% of the cases.

[...]

PR #1709, opened 2024-07-17
Display labels in disassembler widget address column from the right slot/segment  

pvmm:
Should fix issue #1702 

Thanks to @plainspooky for his test code that I use as a tool.

### DIFF
--- a/src/imgui/ImGuiDebugger.cc
+++ b/src/imgui/ImGuiDebugger.cc
@@ -636,13 +636,39 @@ void ImGuiDebugger::drawDisassembly(CPURegs& regs, MSXCPUInterface& cpuInterface
 								}
 							});
 
+							enum Priority {
+								MISSING_BOTH = 0, // from lowest to highest
+								MISSING_ONE,
+								SLOT_AND_SEGMENT
+							};
+							Priority currentPriority = Priority::MISSING_BOTH;
+							std::vector<std::string_view> candidates;
+							auto add = [&](const Symbol* sym, Priority newPriority) {
+								if (newPriority < currentPriority) return; // we already have a better candidate
+								if (newPriority > currentPriority) candidates.clear(); // drop previous candidates, we found a better one
+								currentPriority = newPriority;
+								candidates.push_back(sym->name); // cycle symbols in the same priority level
+							};
+
+							auto slot = getCurrentSlot(cpuInterface, debugger, addr16);
+							auto psss = (slot.ss.value_or(0) << 2) + slot.ps;
 							auto addrLabels = symbolManager.lookupValue(addr16);
-							std::string_view displayAddr = addrLabels.empty()
-								? std::string_view(addrStr)
-								: std::string_view(addrLabels[cycleLabelsCounter % addrLabels.size()]->name);
+							for (const Symbol* symbol: addrLabels) {
+								// skip symbols with any mismatch
+								if (symbol->slot && *symbol->slot != psss) continue;
+								if (symbol->segment && *symbol->segment != slot.seg) continue;
+								// the info that's present does match
+								if (symbol->slot && symbol->segment == slot.seg) {
+									add(symbol, Priority::SLOT_AND_SEGMENT);
+								} else if (!symbol->slot && !symbol->segment) {
+									add(symbol, Priority::MISSING_BOTH);
+								} else {
+									add(symbol, Priority::MISSING_ONE);
+								}
+							}
 							ImGui::SetCursorPos(pos);
 							im::Font(manager.fontMono, [&]{
-								ImGui::TextUnformatted(displayAddr);
+								ImGui::TextUnformatted(candidates.empty() ? addrStr : candidates[cycleLabelsCounter % candidates.size()]);
 							});
 							if (!addrLabels.empty()) {
 								simpleToolTip([&]{


m9710797:
Hi. Thanks once again for picking up another issue!

I understand this PR is still a draft. So I won't do a detailed code review yet. However I did have a quick peek, and I'd like to adjust the direction you're heading in.

I think we should make a distinction between labels in the "address" column and the "mnemonic" column in the disassembly window:
* For the "address" column the situation is clear: we do know what slot (and segment) this address refers to, so we can only replace the numeric representation with a symbol name if the slot matches with the slot of the label.
* But for the "mnemonic" column the situation is less clear.

Suppose I have code like:
`   0x8123    call #005F`         (located in page 2 (0x8000-0xbfff) which has rom selected)
The instruction itself and the destination of the `call` are in different memory pages. The page that contains the instruction has ROM selected, but the destination of the call is in some other page, most likely that's the CHMOD BIOS routine. So even though the CHMOD label is associated with slot 0, and the current slot for this instruction is different, I _DO_ want the numeric value to be replaced with the symbol name.

Another example is code like:
```
   LD IX,#005F
   LD IY,(#FCC0)
   CALL #001C
```
It doesn't matter in what slot this code is located. In these 3 instructions I always want these numbers to be replaced with names.

Actually I can't immediately think of an example where I don't want the number to be replaced with a symbol. Sure there will be cases where this replacement is not correct. But for all the cases I can think of, I can also think of other cases where in the _same_ type of instruction I _do_ want the replacement. Correct me if I'm wrong. And I think it's better to do too many replacements than too few (you can always still get the numeric value via the popup, or see it in the "opcode" column).

A large part of your PR deals with changing the behavior of the `dasm()` function. But in light of the above discussion, I think, those changes should all be reverted. (I think) we should only change how the replacements are done in the "address" column (and that only(?) requires changes in `ImGuiDebugger.cc`).

Of course it's very well possible that I'm wrong. And that we do need to make changes to the `dasm()` function. In that case I'd still like to make a remark:
* In the future we might want to make it possible to disassemble other debuggables than only the Z80 memory. So we should try to make the `dasm()` function less dependent on the specific environment. Your PR actually makes it more dependent, e.g. it builds-in knowledge about pages and visible devices (for other debuggables these concepts don't make sense).
* The idea of the `appendAddr` parameter in `dasm()` was to make this abstraction possible. All the extra logic you added in `dasm()` (=looking up the slot/segment number) could also be done in the `appendAddr` callback function (*1). And I think that's better: that would keep the `dasm()` function more general, increasing the chance that it can be (re)used for both the current disassembly view (and the `debug disasm` Tcl command) and the future "disassemble debuggable" feature.

(*1) It's possible we need to pass extra information to the `appendAddr` callback to make this possible. That's fine.

pvmm:
Thanks, @m9710797 . I forgot that `lookupValue` returns a `std::span`, so I didn't have to write the segment in the `lookupValueCache` key and send `slot` and `segment` info to `dasm()`. The solution is much simpler than I first visualised but it only solves the issue for the address column for now.

m9710797:
Hi pvmm,

I think the PR is "technically correct", but it doesn't fully implement the behavior "that we want". So let's first discuss what behavior do we want. The next few paragraphs are just my opinion. Feel free to disagree and make your own proposal.

This proposal is specific for values in the "address" column, it does not apply to values in the "mnemonic" column. The difference is that for the "address" column we know for sure what the corresponding slot/segment is. But for a value in the "mnemonic" column we (often) don't know this. (And also cannot always know this?)

Obviously we only consider symbols with a matching (16 bit) value. So the list of symbols returned from `symbolManager.lookupValue(addr16)`. From this list we need to pick the "best" candidate. I'd like to discuss _how_ to pick this "best" candidate:

Here's my proposal:
1) If there is a symbol that has both slot and segment information, and both match with the current slot/segment, then this is the best candidate.
2) If there is a symbol that has slot information but _no_ segment information, and the slot matches, then this is the 2nd best candidate (so we only pick this if there aren't any matches on the previous rule).
3) Symbols that don't have any slot information are also candidates (but only if neither of the above rules found any better candidates).
4) Symbols that do have slot (and segment) information, but that slot (and/or segment) doesn't match, can never be candidates. These should never be shown in the "address" column.

Note that rules 1-3 can find multiple candidates. I think we should keep all these candidates (and pick one based on the `cycleLabelsCounter` variable).

Background info: clicking on (any) label increments "cycleLabelsCounter". That allows to see all possibilities. The assumption is that multiple candidates are rare, so it's sufficient to have only one (global) "cycleLabelsCounter" variable (not needed to have such a counter per address).

I am assuming that symbols _with_ segment information but _without_ slot information don't make sense. We treat such symbols as if they don't have any segment info. Is this assumption correct?

--

If I compare this proposal to your PR:
* In the PR you require that both the slot and the segment match. But having only a slot without segment information is also an important use case.
* When there is no exact match on slot (+segment), the PR returns the full list (like we did before). But we should exclude the symbols with an explicit slot/segment _mis_match.
* The PR stops on the first exact slot(+segment) match. But if there are multiple candidates we should allow to cycle through them all.

--

You probably know this, but just in case: selecting the best (list of) candidates can be done in a single pass with an algorithm like this (pseudo code):

```
    Priority current_priority = lowest_priority;
    vector<Symbol*> best_candidates; // none found so far

    auto add_candidate = [&](Symbol* sym, Priority new_priority) {
        if (new_priority < current_priority) return; // we already have a better candidate
        if (new_priority > current_priority) best_candidates.clear(); // drop previous candidates, we found a better one
        current_priority = new_priority;
        best_candidates.push_back(sym);
    };

    for (const auto& candiate : symbolManager.lookupValue(addr16)) {
        if (sym has slot and segment info and both match) {
            add_candidate(sym, highest_priority);
        } else if (sym has slot (but no segment) info and it matches) {
            add_candidate(sym, medium_priority);
        } else if (sym has no slot info) {
            add_candidate(sym, lowest_priority);
        } else {
            // symbols that do have slot info, but that doesn't match
            // are NOT added
        }
    }

    use  best_candidates[cycleLabelsCounter % best_candidates.size()]->name
```


Detail: I think this line:
`    auto psss = slot.ss ? (*slot.ss << 2) + slot.ps: slot.ps;`
can be simplified to:
`    auto psss = (slot.ss.value_or(0) << 2) + slot.ps;`

src/imgui/ImGuiDebugger.cc
```diff
@@ -629,13 +629,39 @@ void ImGuiDebugger::drawDisassembly(CPURegs& regs, MSXCPUInterface& cpuInterface
 								}
 							});
 
+							enum Priority {
+								LOWEST_PRIORITY = 0,
+								MEDIUM_PRIORITY,
+								HIGHEST_PRIORITY
+							};
```
m9710797, line 643: detail:
Maybe shorten the names to "HIGH", "MEDIUM", "LOW". And then use them like "Priority::HIGH"?

Or if you find inspiration you could also try more descriptive names?  Maybe "SLOT_AND_SEGMENT", "ONLY_SLOT", "NO_SLOT"??   (And add a comment like: ordered from lower to higher priority?

src/imgui/ImGuiDebugger.cc
```diff
@@ -629,13 +629,39 @@ void ImGuiDebugger::drawDisassembly(CPURegs& regs, MSXCPUInterface& cpuInterface
...
-								: std::string_view(addrLabels[cycleLabelsCounter % addrLabels.size()]->name);
+							for (const Symbol* symbol: addrLabels) {
+								if (symbol->slot) {
+									if (*symbol->slot != psss) continue;
+									if (symbol->segment && symbol->segment != slot.seg) continue;
+									if (symbol->segment == slot.seg) {
+										add_candidate(symbol, HIGHEST_PRIORITY);
+									} else {
+										add_candidate(symbol, MEDIUM_PRIORITY);
+									}
```
m9710797: I think this is correct, but (I think) it's more complex than needed (therefor it confused me a bit). Would the following be simpler? Or am I missing something?
```
if (symbol->slot) {
    if (*symbol->slot != psss) continue; // slot mismatch -> don't show
    if (symbol->segment) {
        if (*symbol->segment != segment) continue; // segment mismatch -> don't show
        add(symbol, Priority::SLOT_AND_SEGMENT); // both slot and segment match
    } else {
        add(symbol, Priority::SLOT); // slot matches, no segment info
    }
} else {
    add(symbol, Priority::NO_SLOT); // no slot info (segment doesn't matter)
}
```

src/imgui/ImGuiDebugger.cc
```diff
@@ -629,13 +629,39 @@ void ImGuiDebugger::drawDisassembly(CPURegs& regs, MSXCPUInterface& cpuInterface
...
+								} else if (!symbol->segment) {
+									add_candidate(symbol, LOWEST_PRIORITY);
+								}
+							}
 							ImGui::SetCursorPos(pos);
 							im::Font(manager.fontMono, [&]{
-								ImGui::TextUnformatted(displayAddr);
+								ImGui::TextUnformatted(bestCandidates.empty() ? addrStr : bestCandidates[cycleLabelsCounter % bestCandidates.size()]);
```
m9710797: detail:
I know I used the name `bestCandidates`, but now that I see it in real code it is quite long.
Would just `candidates` also be fine? (I mean equally clear in this context). I'll let you decide.

src/imgui/ImGuiDebugger.cc
```diff
@@ -629,13 +629,39 @@ void ImGuiDebugger::drawDisassembly(CPURegs& regs, MSXCPUInterface& cpuInterface
...
+							for (const Symbol* symbol: addrLabels) {
+								if (symbol->slot) {
+									if (*symbol->slot != psss) continue;
+									if (symbol->segment && symbol->segment != slot.seg) continue;
+									if (symbol->segment == slot.seg) {
+										add_candidate(symbol, HIGHEST_PRIORITY);
+									} else {
+										add_candidate(symbol, MEDIUM_PRIORITY);
+									}
+								} else if (!symbol->segment) {
```
m9710797: I think this is a small bug:
if a symbol has no slot info, then it doesn't matter if it has segment info or not. We should always include it, no?. The current patch would never include symbols without slot but with segment info.

src/imgui/ImGuiDebugger.cc
```diff
@@ -629,13 +629,39 @@ void ImGuiDebugger::drawDisassembly(CPURegs& regs, MSXCPUInterface& cpuInterface
...
 							});
 
+							enum Priority {
+								LOWEST_PRIORITY = 0,
+								MEDIUM_PRIORITY,
+								HIGHEST_PRIORITY
+							};
+							Priority currentPriority = LOWEST_PRIORITY;
+							std::vector<std::string_view> bestCandidates;
+							auto add_candidate = [&](const Symbol* sym, enum Priority newPriority) {
```
m9710797: Very much a detail:
In C it's required to repeat `enum` before `Priority`, in C++ it's not (and usually it's left out).

m9710797 (commented):
One small bug (I think).
And a few small suggestions to simplify.

But this is looking very good. This (almost) ready to merge.
Thanks a lot!

src/imgui/ImGuiDebugger.cc
pvmm: I was following what you said earlier:

> I am assuming that symbols with segment information but without slot information don't make sense. We treat such symbols as if they don't have any segment info. Is this assumption correct?

And also this:

>  if a symbol has no slot info, then it doesn't matter if it has segment info or not. We should always include it, no?. The current patch would never include symbols without slot but with segment info.

It does matter, otherwise any symbol from any slot will show, even if it is from a wrong segment. This is wrong behaviour because BLA (segment 4) should not be visible in segment 2:
image

src/imgui/ImGuiDebugger.cc
```diff
@@ -629,13 +629,39 @@ void ImGuiDebugger::drawDisassembly(CPURegs& regs, MSXCPUInterface& cpuInterface
...
+									if (*symbol->slot != psss) continue; // slot mismatch -> don't show
+									if (symbol->segment) {
+										if (*symbol->segment != slot.seg) continue; // segment mismatch -> don't show
+										add(symbol, Priority::SLOT_AND_SEGMENT); // both match
+									} else {
+										add(symbol, Priority::MISSING_ONE); // only slot matches
+									}
+								} else if (symbol->segment == slot.seg) {
+									add(symbol, Priority::MISSING_ONE); // only segment matches
+								}
```
m9710797, line 667: I don't see any `add()` with `Priority::NO_MATCH`, that can't be right. Can I also ask how you tested this PR? Maybe in your next tests also check if symbols without slot/segment info still show up.

If we come in the else-branch of `if (symbol->slot) {` we don't yet know anything about the segment (not whether this info is present and not whether it matches). So maybe this should become something like this?
```
if (symbol->slot) {
    ...
} else if (symbol->segment) { // no slot info, but there is segment info
    if (*symbol->segment != slot.seg) continue; // segment mismatch -> don't show
    add( ... MISSING_ONE);
} else {
   add(... BOTH_MISSING);
}
```

BTW: maybe `MISSING_BOTH` is a better name than `NO_MATCH`?   (because if there is a mismatch (~no match) we won't include the symbol at all).

BTW2: In my previous comments/reviews I was wrong: I assumed segment isn't important when slot is missing (I even explicitly stated that as one of my assumptions). If you see such mistakes then please speak up. You don't have to apply or agree with all my requests for changes.

m9710797: Update:
The if-then-else logic above is again becoming quite complex. Maybe this alternative is simpler?
```
// skip symbols with any mismatch
if (symbol->slot && *symbol->slot != psss) continue;
if (symbol->segment && *symbol->segment != slot.seg) continue;

// the info that's present does match
if (symbol->slot && symbol->segment) {
    add(... SLOT_AND_SEGMENT);
} else if (!symbol->slot && !symbol->segment) {
    add(... MISSING_BOTH);
} else {
    add(... MISSING_ONE);
}
```
This is maybe also closer to what you started with in one of your previous revisions of this PR? So again sorry if I mislead you with my (wrong) comments.

pvmm: > I don't see any `add()` with `Priority::NO_MATCH`, that can't be right. Can I also ask how you tested this PR? Maybe in your next tests also check if symbols without slot/segment info still show up.

I tested it thoroughly this time and I think it's OK now.

m9710797:
Thanks!

PR #1712, opened 2024-07-21
fix crash when symbol file is not found

pvmm:
The null object (`nullfile`) solution makes the code simpler to understand.

### DIFF
--- a/src/imgui/ImGuiSymbols.cc
+++ b/src/imgui/ImGuiSymbols.cc
@@ -117,7 +117,7 @@ void ImGuiSymbols::loadFile(const std::string& filename, SymbolManager::LoadEmpt
 			it->error = e.getMessage(); // overwrite previous error
 			it->type = type;
 		} else {
-			fileError.emplace_back(filename, e.getMessage(), type, it->slot); // set error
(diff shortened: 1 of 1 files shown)


m9710797:
Thanks a lot for debugging this! I took your solution and modified it a bit:
* It was weird that "wrong" files still had an editable slot. And changing the slot in one such file modified it in all wrong files. (Typically when you use the "null object pattern" the null-object should be immutable).
* I experimented with graying-out the slot-combo-box (that would solve the edit-problem). But I found it looked better if I completely left it out.
* Also showing an empty table, or a grayed-out table, for a wrong file looked weird. I decided to also fully remove it.
* So basically wrong files only have a "reload" and "remove" button anymore.
* And with these changes there was no advantage anymore to having a `nullFile`.
* I did an assert that we either have a valid (non-nullptr) `file`, or that we process this as an error (and not both). I inspected the code, and I believe that assert should hold.

m9710797:
BTW: if you want to look at my commit, make sure to ignore white-space changes. The indentation level of a large block of code has changed. But ignoring that, the commit is fairly small.

Issue #302, opened 2015-06-22
[Bug] Console keys passed to emulated MSX [sf#1]


openMSX-import:
**Reported by manuelbi on 2002-05-01 09:58 UTC**
When the console is on (F10), the keys are still sent
to the MSX. Maybe this is intended, but I think it
should not be the case (it could be optional though).

openMSX-import:
**Updated by shevek on 2002-05-08 18:18 UTC**
- **status**: open --> closed

openMSX-import:
**Commented by manuelbi on 2002-06-24 08:06 UTC**
Logged In: YES 
user_id=78178

With the latest CVS builds, the problem returned. I think it
was since the introduction of Console Background.

openMSX-import:
**Updated by manuelbi on 2002-06-24 08:06 UTC**
- **status**: closed --> open

openMSX-import:
**Commented by m9710797 on 2002-07-31 14:30 UTC**
Logged In: YES 
user_id=356949

Fixed (again). 
EventListeners now have a priority. Higher priority listeners can block 
events for lower priority listeners. This works indepently from the 
registration order of the two listeners.

openMSX-import:
**Updated by m9710797 on 2002-07-31 14:30 UTC**
- **status**: open --> closed

PR #1717, opened 2024-08-06
dasm extension: disassemble any Tcl binary string

pvmm:
To make this kind of thing possible:
image

Use case: my source code analyser/generator script could leverage OpenMSX's capacity to disassemble z80 code from any source. And of course if the name `disasm_bin` is not the best choice, we can change it to something else.


src/cpu/CPUCore.cc
```diff
@@ -518,6 +518,29 @@ static constexpr void toHex(byte x, std::span<char, 3> buf)
...
+	}
+	unsigned len = dasm(bin, address, outBuf, dasmOutput);
+	dasmOutput.resize(19, ' ');
+	result.addListElement(dasmOutput);
+	std::array<char, 3> tmp; tmp[2] = 0;
+	for (auto i : xrange(len)) {
+		toHex(outBuf[i], tmp);
+		result.addListElement(tmp.data());
+	}
+}
```
m9710797, line 539: This function returns a Tcl list. The first element of this list is the actual disassembly-string (padded with spaces). The next elements in this list are the bytes that make up the instruction (formatted as 2 digit hex values). Indirectly, from the length of the list, you can derive the length of the instruction (this list-length minus 1).

I understand you copied this API from the existing `disasm` command. But maybe it's not the best in this context:
* The input to this command is a binary blob, so you do already have the instruction bytes. Then there's not much value in returning the same bytes in the result list.
* Except for the length of the instruction. But we can also return a list where the 2nd element in the list is an integer (1-4) for the length of the instruction.

The last parameter `addr` is optional, if not specified it defaults to the `PC`. That makes sense for the existing `disasm` command. But I think for the `disasm_bin` command you're likely not disassembling memory where the PC points to (otherwise it was simpler to use the existing command). Therefor I propose to make the `addr` parameter required.

src/cpu/Dasm.cc
```diff
@@ -111,15 +112,37 @@ unsigned dasm(const MSXCPUInterface& interface, uint16_t pc, std::span<uint8_t,
...
+unsigned dasm(const MSXCPUInterface& interface, uint16_t pc, std::span<uint8_t, 4> buf,
+              std::string& dest, EmuTime::param time,
+              function_ref<void(std::string&, uint16_t)> appendAddr)
+{
+	uint8_t bin[4];
+
+	for (int i = 0; i < 4; ++i) {
+		bin[i] = interface.peekMem(pc + narrow_cast<uint16_t>(i), time);
+	}
+	return dasm(std::span<uint8_t, 4>{bin}, pc, buf, dest, appendAddr);
```
m9710797: This version has the _old_ signature and now delegates to the _new_ generalized `dasm()` function. Very good. Much better than duplicating this code.

Though I don't like that we now always call peekMem() 4 times (in some cases this can be expensive). The old version called peekMem() the minimum number of times (the length of the instruction).

Similar for the `instructionLength()` function below. The old version usually only did 1 or sometimes 2 calls to `peekMem()`, but the new version always does 4 calls.

Maybe we can make a new helper function `fetchInstruction()` that is a combination of getting the instruction length and reading (only those) bytes from memory. Maybe something like this?
```
std::span<uint8_t> fetchInstruction(const MSXCPUInterface&, uint16_t addr, std::span<uint8_t, 4> buffer) {
    // read 1st byte into buffer
    // calculate the instruction length, possibly this also reads the 2nd byte into buffer
    // if not already done, also read remaining bytes into buffer
    return buffer.subspan(0, length);  // return the given buffer, but trimmed to the correct length
}
```

src/cpu/Dasm.hh
```diff
@@ -16,6 +16,19 @@ class MSXCPUInterface;
...
+  * @param buf The bytes that form this opcode (max 4). The buffer is only
+  *            filled with the min required bytes (see return value)
+  * @param dest String representation of the disassembled opcode
+  * @param time TODO
+  * @return Length of the disassembled opcode in bytes
+  */
+unsigned dasm(std::span<const uint8_t> bin, uint16_t pc, std::span<uint8_t, 4> buf,
+              std::string& dest,
+              function_ref<void(std::string&, uint16_t)> appendAddr = &appendAddrAsHex);
+
```
m9710797, line 27: This function has a weird API:
* We pass `bin` as input, this is the to-be-disassembled binary data.
* As return value we get the length of this instruction (let's call this `length`).
* The buffer passes as the `buf` parameter is filled with `length` number of bytes, copied from `bin`. But this doesn't give us anything new: we already have the bytes in `bin` and we have the length as the return value. It's extra work for nothing.

In the old `dasm()` function the `buf` parameter did make some sense. The function itself did call `peekMem()`, so returning those bytes then avoids having to call this (possibly expensive) function again (e.g. to show the hex opcode).

Maybe we should split this `dasm()` function into two parts?
* One part could be the `fetchInstruction()` function I talked about above.
* And the other part takes this result and turns it into a string (without doing and fetches or buffer copies itself).

And in case of `disasm_bin` the 1st part only calculates the length. And then we pass a correctly sized subspan to the 2nd part (so only the 2nd part is common between `disasm` and `disasm_bin`.

src/cpu/MSXCPU.cc
```diff
@@ -354,6 +354,14 @@ void MSXCPU::update(const Setting& setting) noexcept
...
 // Command
 
+void MSXCPU::disasmBinCommand(
+	Interpreter& interp, std::span<const TclObject> tokens,
+	TclObject& result) const
+{
+	z80Active ? z80 ->disasmBinCommand(interp, tokens, result)
+	          : r800->disasmBinCommand(interp, tokens, result);
+}
+
```
m9710797, line 364: I fully understand you copied the existing pattern. So this is a remark against the code I've written.

I don't like that `disasmCommand()` and now also `disasmBinCommand()` is implemented in CPUCore: to disassemble something you don't need access to the active CPU state. There are two exceptions:
* The `PC` (for the `disasm` command, I already proposed to not use `PC` for `disasm_bin`). But we can instead query the PC from the active CPU instead of implementing disassembling in the CPU emulation code.
* The active time, because that's needed in peekMem(). This only applies to `disasm`, not `disasm_bin`. But again we can query this time and keep the actual disassembly code outside the CPUCore class.

I'm adding this as a review comment, but I don't expect you to 'fix' this very old design decision. You can of course, but I think it should be done in a separate commit (within this PR or as a separate PR). Or I can do it afterwards.

src/debugger/Debugger.cc
```diff
@@ -241,6 +241,7 @@ void Debugger::Cmd::execute(
 		"step",              [&]{ debugger().motherBoard.getCPUInterface().doStep(); },
 		"cont",              [&]{ debugger().motherBoard.getCPUInterface().doContinue(); },
 		"disasm",            [&]{ debugger().cpu->disasmCommand(getInterpreter(), tokens, result); },
+		"disasm_bin",        [&]{ debugger().cpu->disasmBinCommand(getInterpreter(), tokens, result); },
```
m9710797: I was thinking about the name for this new sub-command. `disasm_bin` is not a bad name at all. Other names that came to my mind were `disasm_blob` or `disasm_binary`.

src/cpu/CPUCore.cc
```diff
@@ -518,6 +518,29 @@ static constexpr void toHex(byte x, std::span<char, 3> buf)
 	buf[1] = toHex(x & 15);
 }
 
+template<typename T> void CPUCore<T>::disasmBinCommand(
+	Interpreter& interp, std::span<const TclObject> tokens, TclObject& result) const
+{
+	if (tokens.size() < 3) {
+		throw CommandException("wrong # args: should be \"debug disasm_bin value ?addr?\"");
```
m9710797: Detail: we have a helper function for this (this ensures constently formatted error message in the whole code). Search for examples using `checkNumArgs()`.

src/cpu/Dasm.cc
```diff
@@ -111,15 +112,37 @@ unsigned dasm(const MSXCPUInterface& interface, uint16_t pc, std::span<uint8_t,
 	return i;
 }
 
+unsigned dasm(const MSXCPUInterface& interface, uint16_t pc, std::span<uint8_t, 4> buf,
+              std::string& dest, EmuTime::param time,
+              function_ref<void(std::string&, uint16_t)> appendAddr)
+{
+	uint8_t bin[4];
```
m9710797: Detail: we prefer `std::array<uint8_t, 4>` over C-style arrays. Doesn't matter here, and it is a bit more verbose. But then it's consistent in our whole code base. (And from time-to-time I run clang-tidy which flags uses of C-style arrays).

src/cpu/Dasm.cc
```diff
@@ -23,30 +23,31 @@ void appendAddrAsHex(std::string& output, uint16_t addr)
...
 }
 
-unsigned dasm(const MSXCPUInterface& interface, uint16_t pc, std::span<uint8_t, 4> buf,
-              std::string& dest, EmuTime::param time,
-              function_ref<void(std::string&, uint16_t)> appendAddr)
+unsigned dasm(std::span<const uint8_t> bin, uint16_t pc, std::span<uint8_t, 4> buf,
+              std::string& dest, function_ref<void(std::string&, uint16_t)> appendAddr)
 {
+	auto length = instructionLength(bin);
+	assert(bin.size() > 0 && length <= bin.size());
```
m9710797: detail: `bin.size() > 0` is usually written like `!bin.empty()`.
In this case that doesn't make any difference. But in other cases using `empty()` can be more efficient than `size()`. So it's a good habit.

m9710797 (commented):
Hi,

I like the idea, but I'd also like to know more about the use case you have in mind.
* For example is this mainly to disassemble single isolated instructions?
* Or is it meant to be used in a loop to disassemble a larger block.

In case of the latter it might be more efficient to use `debug read_block` only once and then pass parts of this larger block to the new `debug disasm_bin` command. And then my question is:
* Is it easy in Tcl to get a sub-range from a larger binary blob? (I googled, and it seems that `string range` should be able to do this, but we should test this).
* Or alternatively we can change the API of `debug disasm_bin` to accept a larger binary blob, and an offset within that blob from where to start disassembling.

I have a slight preference for the first of these two options, IF testing confirms that `string range` indeed works fine.

I do have some other remarks about the API, but that's easier to discuss next to the code, so see below.

And some more general remarks (also see below). Some of these apply to your PR, but others apply to my code (some written +20 years ago).

pvmm:
**Keeping a record of the conversation from Discord:**
**&lt;pvmm&gt;** I mirrored the API in this case because dasm is used for single instructions and the looping version is implemented in **_disasm.tcl**. With that said, I still need the single instruction translation first and foremost because there are special cases I am trying to consider. If there are overlapping machine language instructions, I use the single instruction dasm to break them up. And if memory is marked as `DATA` I will not use dasm, but write `db xxh` directly until I find memory marked as `CODE` again. So the looping disasm is not very useful after all.
**&lt;pvmm&gt;** And `string range` works fine.

src/cpu/CPUCore.cc
pvmm: There is no easy way to use it now, since I need to create a class that inherits from `Command` to use `checkNumArgs()` and since `disasmBin` should not be connected to `CPUCore` anyway, it needs to live inside its own command (something like a `DasmCommand` class).

src/debugger/Debugger.cc
```diff
@@ -801,6 +802,7 @@ string Debugger::Cmd::help(std::span<const TclObject> tokens) const
 		"    break             break CPU at current position\n"
 		"    breaked           query CPU breaked status\n"
 		"    disasm            disassemble instructions\n"
+		"    disasm_blob       disassemble a instruction in Tcl binary string\n"
```
pvmm, line 805: Renamed it to `disasm_blob`.

src/cpu/CPUCore.cc
```diff
@@ -518,6 +519,27 @@ static constexpr void toHex(byte x, std::span<char, 3> buf)
...
+		throw CommandException("Wrong binary string size");
+	}
+	unsigned len = dasm(bin, word(tokens[3].getInt(interp)), dasmOutput);
+	dasmOutput.resize(19, ' ');
+	result.addListElement(dasmOutput);
+	std::array<char, 3> tmp; tmp[2] = 0;
+	for (auto i : xrange(len)) {
+		toHex(bin[i], tmp);
+		result.addListElement(tmp.data());
+	}
```
m9710797: I'm sorry if this is still work in progress, and this is still on your TODO list. But since you marked this as resolved ...

I still think we shouldn't have this loop over bin[0..len]. We're literally just returning (part of) the input parameter again. This doesn't tell anything useful. Instead we should just return `len`. (I mean a list containing both the string and len).

For the existing `dasm` command returning these bytes does provide new information.

src/cpu/CPUCore.cc
```diff
@@ -518,6 +519,27 @@ static constexpr void toHex(byte x, std::span<char, 3> buf)
...
 
+template<typename T> void CPUCore<T>::disasmBlobCommand(
+	Interpreter& interp, std::span<const TclObject> tokens, TclObject& result) const
+{
+	if (tokens.size() < 4) {
+		interp.wrongNumArgs(2, tokens, "value addr");
+	}
+	std::string dasmOutput;
+	auto bin = tokens[2].getBinary();
+	if (bin.empty() || instructionLength(bin) > bin.size()) {
```
m9710797: I think there's a bug in this check:
* Suppose you pass a binary blob of length only 1 byte.
* That byte is the start of a `CB` instruction ... meaning the `instructionLength()` function also needs to read the 2nd byte to be able to determine the length,
* In that case `instructionLength()` will return 0. And this check will not throw an exception.
* And then later the `dasm()` function will try to read more data than is available in `bin`.

m9710797: The implementation of `instructionLength()` already checks for `bin.empty()`. No need to repeat that here (nor in all other places where you check this right before calling `instructionLength()`.

src/cpu/Dasm.cc
```diff
@@ -111,8 +111,38 @@ unsigned dasm(const MSXCPUInterface& interface, uint16_t pc, std::span<uint8_t,
 	return i;
```
m9710797, line 118: Maybe here add an assert:
```assert(i == length)```

Or maybe better:
* Is it correct that all callers of this `dasm()` function now already now the instruction length? (either via calling `instructionLength()` or `fetchInstruction()`)
* In that case they can pass an appropriately sized `bin` parameter to this function.
* And then there no point anymore in returning the length from this function (so the return type can become `void`),
* Calling `instructionLength()` (at line 39) in this function is then no longer useful,
* And we can end this function with `assert(i == bin.size());`.

File: src/cpu/Dasm.cc [hunk 13]
```diff
@@ -111,8 +111,38 @@ unsigned dasm(const MSXCPUInterface& interface, uint16_t pc, std::span<uint8_t,
...
+	}
+
+	// if DD, ED or FD instruction, read 0 to 3 more bytes into buffer (4 bytes total)
+	buffer[1] = interface.peekMem(addr + 1, time);
+	auto len2 = instr_len_tab[64 * len1 + buffer[1]];
+	for (unsigned i = 2; i < len2; ++i) {
+		buffer[i] = interface.peekMem(addr + narrow_cast<uint16_t>(i), time);
+	}
+	return buffer.subspan(0, len2);  // return the given buffer, but trimmed to the correct length
+}
```
m9710797, line 136: The tail of this function is duplicated code. Maybe structure it like this:
```
// read 1st byte
uint16_t idx = 0;
buffer[idx++] = interface.peekMem(addr, time);
auto len = instr_len_tab[buffer[0]];
if (len >= 4) {
    buffer[idx++] = interface.peekMem(addr + 1, time);
    len = instr_tab_len[64 * len + buffer[1]];
}
while (idx < len) {
    buffer[idx++] = interface.peekMem(addr + idx, time);   // maybe needs narrow_cast to silence warning?
}
return buffer.subspan(0, len);
```

src/cpu/Dasm.cc
```diff
@@ -111,8 +111,38 @@ unsigned dasm(const MSXCPUInterface& interface, uint16_t pc, std::span<uint8_t,
...
+
+unsigned dasm(const MSXCPUInterface& interface, uint16_t pc, std::span<uint8_t, 4> buf,
+              std::string& dest, EmuTime::param time,
+              function_ref<void(std::string&, uint16_t)> appendAddr)
+{
+	return dasm(fetchInstruction(interface, pc, buf, time), pc, dest, appendAddr);
+}
```
m9710797, line 146: If you rework `dasm()` to return `void`. This can become:
```
auto opcodes = fetchInstruction(...)
dasm(opcodes, ...);
return opcodes.size();
```

Or alternatively you can change the return type from `unsigned` to `std::span<uint8_t>`. So similar to how the return is handled in `fetchInstruction()`.

m9710797 (commented):
Already a little bit more detailed review. But I understand this is still work in progress.

src/cpu/Dasm.cc
```diff
@@ -23,96 +23,122 @@ void appendAddrAsHex(std::string& output, uint16_t addr)
...
-              function_ref<void(std::string&, uint16_t)> appendAddr)
+unsigned instructionLength(std::span<const uint8_t> bin)
+{
+	if (bin.empty()) return 0;
+	auto t = instr_len_tab[bin[0]];
+	if (t < 4) return t;
+
+	if (bin.size() < 2) return 0;
+	return instr_len_tab[64 * t + bin[1]];
+}
```
m9710797, line 42: Would it be cleaner if this function returns a `std::optional<unsigned>` instead of using value `0` to indicate an error?
* That better separates the valid values from the error case ("could not calculate instruction length")
* Callers are less likely to forget to check for this error.

Or alternatively you could indicate errors with `unsigned(-1)`. That way `bin.size() < len` is sufficient to check both errors "could not calculate instruction length" and "blob size too small for instruction".

My personal preference is `std::optional`.

src/cpu/CPUCore.cc
```diff
@@ -518,14 +519,32 @@ static constexpr void toHex(byte x, std::span<char, 3> buf)
...
+	Interpreter& interp, std::span<const TclObject> tokens, TclObject& result) const
+{
+	if (tokens.size() < 4) {
+		interp.wrongNumArgs(2, tokens, "value addr");
+	}
+	std::string dasmOutput;
+	auto bin = tokens[2].getBinary();
+	unsigned len = instructionLength(bin);
+	if (len == 0 || len > bin.size()) {
+		throw CommandException("Wrong binary string size");
```
m9710797: Detail: "wrong size" is technically correct. But maybe a more helpful error message is "too small binary size" or "binary does not contain a complete instruction", or ...  (giving good error messages is hard).

src/cpu/CPUCore.cc
```diff
@@ -518,14 +519,32 @@ static constexpr void toHex(byte x, std::span<char, 3> buf)
 	buf[1] = toHex(x & 15);
 }
 
+template<typename T> void CPUCore<T>::disasmBlobCommand(
+	Interpreter& interp, std::span<const TclObject> tokens, TclObject& result) const
+{
+	if (tokens.size() < 4) {
+		interp.wrongNumArgs(2, tokens, "value addr");
+	}
+	std::string dasmOutput;
```
m9710797: Very very much a detail:
In general prefer to declare local variables as close as possible to where they are first needed.
In this case you construct an empty string `dasmOutput`. Then you check for (more) possible errors. So if that error triggers constructing the string was for nothing.
This is easily avoided by moving this line right before the call to `dasm()`.

So very much a detail: it can be a small performance cost (in corner cases). But more important to also improves the readability of the code.

src/cpu/Dasm.cc
```diff
@@ -23,96 +23,122 @@ void appendAddrAsHex(std::string& output, uint16_t addr)
...
-			              ' ');
-			return 2;
-		case ' ': {
-			dest.resize(7, ' ');
-			break;
-		}
-		default:
-			dest += s[j];
-			break;
+	i = [&]() -> unsigned {
```
m9710797: Maybe I'm overlooking something, but why did you turn this into a lambda?
This new lambda captures `i`, modifies it (it modifies the same `i` variable outside the lambda), returns it, and then assigns it to the same `i` variable (so this assignment does nothing). Unless I missed something I propose to revert this change.

Maybe you did this for symmetry with the code above that initializes `s` and `i`? But that lambda has a different purpose. Without that lambda the code should have been:
```
const char* s; // uninitialized
unsigned i; // uninitialized
switch (...) {
    case ...:
        s = ...;
        i = ...;
}
```
So this is a way to avoid having uninitialized variables (and have the compiler help enforce this). But this is a tradeoff, the lambda does make the code more complex.

m9710797 (commented):
PR is close to ready. Comments are more and more about details.

src/cpu/Dasm.cc
pvmm: It's because the final `assert` test at the end of the `dasm()` should always compare `i` with `bin.size()`, but the for-loop has some `return`s in the switch-case which skip the assertion test. And the attribution to `i` is there because the `@`, `#` and `!` switch-cases replace `i` with a new value (by directly returning to the caller). So I figured a less modified version of the code that still allowed us to run the final assert is putting the for-loop inside a lambda.

m9710797: Aha, I indeed missed that.

I don't like that, I mean my own code. Maybe we can make the switch more regular? So not have a mix between `break` and `return` in the different cases, but only use `break`. The `@`, `#` and `!` are always the last character in the format string, so it should be possible.

Of course we have to think about what to do exactly with the `return` statement. One possibility is to replace `return <n>` with `i = <n>; break;`
* For `!` (only used for `ED`) we can also leave out `i = 2;`, it already has this value, at the time `!` is processed.
* For `@` (used for `DD/FD` instructions unrelated to the `IX/IY` register) the `i = 1;` assignment is needed.
* For `#` the current implementation is weird. We print `db #DD,#CB,#nn` (so 3 bytes), but then only report an instruction length of 2 (`i = 2;`). I think that's a bug, we should either only print 2 bytes (my preference) or skip over 3 instead of 2 bytes.

This is a weird structure and a bug in the existing code. Feel free to tackle it in this PR, but it's also fine if you leave it for later.

Background info: All 3 of `!`, `@` and `#` are only used to disassemble _invalid_ instructions. Our implementation in `CPUCore` emulates these in the same way as how a real Z80 would handle them. But that's _not_ the same as how this `dasm()` function prints them. (Practical problem: the real Z80 has no upper limit on the instruction length (at least not for invalid instructions, e.g. DD DD DD DD DD ... DD 00). Taking that into account in the `dasm()` function introduces extra complexity for no practical gain.)

m9710797: Now that I think about it some more. In case of invalid instructions that `assert(i == bin.size())` will not always be true.

For example `DD CB nn 00` will (currently) be disassembled as `db #DD,#CB,#nn` and will skip over 2 bytes (inconsistent). BUT to be able to determine that this is an invalid instruction we DO need to look at 4 bytes in `bin` (we need to see this `00` value).

And something similar happens for `@`, we only report a length of 1 byte (and print e.g. `db #DD`), but we needed to look at 2 bytes in `bin` to determine that.

So maybe we should skip the assert in case of invalid instructions (and then the `return` in the switch was not such a bad idea ... but an extra comment explaining this would be nice). We should also fix the content of the `instr_len_tab[]` array so that it does fetch enough bytes so that `dasm()` won't crash on invalid instructions (= won't read past the end of `bin`).

pvmm:
This is experimental code that puts dasm through the `unittest` routines and tries to find some solutions to solve the instructionLength/dasm/fetchInstruction size mismatch. Unit testing code is slow because it is comprehensive, but I will change it to test all valid z80 instructions and just a few invalid ones.

src/cpu/Dasm.cc
```diff
@@ -23,82 +22,90 @@ void appendAddrAsHex(std::string& output, uint16_t addr)
...
+	0x96, 0x9e, 0xa6, 0xae, 0xb6, 0xbe, 0xc6, 0xce, 0xd6, 0xde, 0xe6, 0xee, 0xf6, 0xfe
+};
+
+std::optional<unsigned> instructionLength(std::span<const uint8_t> bin)
+{
+	if (bin.empty()) return {};
+	auto t = instr_len_tab[bin[0]];
+	if (t < 4) return t;
+
+	if (bin.size() < 2) return 1;
```
m9710797: I don't agree with this change. Let's take an example: `ED 44 (neg)`, this is a valid instruction.
But if we only pass `ED` then this function now returns `1`. Instead it should report an error (via `std::nullopt`), meaning "insufficient data to calculate the instruction length".

src/cpu/Dasm.cc
```diff
@@ -23,82 +22,90 @@ void appendAddrAsHex(std::string& output, uint16_t addr)
...
+	t = instr_len_tab[64 * t + bin[1]];
+	if (t != bin.size()) return 1;
```
m9710797: This is weird: we pass the `bin` parameter because we don't know what instruction-length it contains. But then this `if` checks that the size we passed must already be equal to the instruction length we just calculated???

src/cpu/Dasm.cc
```diff
@@ -23,82 +22,90 @@ void appendAddrAsHex(std::string& output, uint16_t addr)
...
+	if (bin.size() == 4 && !std::binary_search(ddcb__.begin(), ddcb__.end(), bin[3])) return 1;
```
m9710797: I think you added this to handle invalid instructions, right?
The intention being: only emit a 1-byte `db #xx` line in the disassembly, and try to _recover_ disassembly at the next byte. At least that's how I (partially) implemented this in the old code.

The last few days (because of your PR) I've been thinking about this, and I've changed my mind. I no longer think this _fast_recovery_ is a good idea. I'll try to explain.

Let's take the sequence `ED 01 02 03` as an example. The first two bytes `ED 01` form an invalid instruction.
* With fast recovery we would disassemble this as `db #ED ; ld bc,#0302`.
* Or we could disassemble it as `db #ED,#01 ; ld (bc),a ; inc bc`. So print the _full_ invalid sequence as a "db-string". And just continue from there.

Let's look at how a real Z80 behaves. It fetches the 1st `ED` byte. From this value it _knows_ it has to fetch a 2nd byte `01`. But this sequence `ED 01` does not correspond to a valid instruction. From experiments we know a real Z80 executes this sequence as a `nop` instruction. And then the Z80 will continue with the _next_ byte `02`. It will not try to recover by re-fetching the `01` byte.

Both alternatives for the disassembly are valid:
* The 1st is maybe something you manually write in your assembly source code.
* The 2nd is closer to how the real Z80 sees the byte-stream (*1).

IMHO in the context of the openMSX debugger we should pick the latter. We're stepping through the code, so we should interpret the bytes similar to how the Z80 sees them.

(*1) And yes, in case of repeated `DD` or `FD` prefix bytes (all invalid instructions), the current implementation of the disassembly is still not in sync with how the real Z80 interprets this sequence. But that's a rare corner case, and that's done to keep the implementation simple (otherwise there's no upper limit to the instruction length, now we can assume an instruction takes max 4 bytes).

src/cpu/Dasm.cc
```diff
@@ -23,82 +22,90 @@ void appendAddrAsHex(std::string& output, uint16_t addr)
...
+	return t;
+}
```
m9710797, line 42: I think it's best if you revert this function to the previous version.

src/cpu/Dasm.cc
```diff
@@ -23,82 +22,90 @@ void appendAddrAsHex(std::string& output, uint16_t addr)
...
+		case '@': // used when bin contains invalid z80 instruction
+			dest = strCat("db     #", hex_string<2>(bin[0]),
 			              "         ");
+			// skip assertion at the end of function
+			return 1;
+		case '#': // used when bin contains invalid z80 instruction
+			dest = strCat("db     #", hex_string<2>(bin[0]),
+			              "         ");
+			// skip assertion at the end of function
 			return 1;
```
m9710797: For these 3 cases `!`, `@` and `#`, I now think we should disassemble them as:
`db hex_string<2>(bin[0..i])`
and don't change `i` or `return` in the case.

In other words: print the byte sequence that forms an invalid instruction and continue. Don't try to recover. That's also how the real Z80 handles this.

src/cpu/Dasm.cc
```diff
@@ -108,18 +115,50 @@ unsigned dasm(const MSXCPUInterface& interface, uint16_t pc, std::span<uint8_t,
...
+	buffer[idx++] = interface.peekMem(addr, time);
+	auto len = instr_len_tab[buffer[0]];
+	if (len >= 4) {
+		buffer[idx++] = interface.peekMem(addr + 1, time);
+		len = instr_len_tab[64 * len + buffer[1]];
+	}
+	while (idx < len) {
+		buffer[idx++] = interface.peekMem(addr + idx, time);
+	}
+	if (len == 4 && !std::binary_search(ddcb__.begin(), ddcb__.end(), buffer[3])) len = 1;
```
m9710797: Remove this line again. We don't want this "recover" behavior. But also it wouldn't work:
* Suppose we do set `len = 1`.
* Then in the next line we only return a `span` of length = 1.
* We pass this to `dasm()`.
* But the implementation of `dasm()` _will_ read `bin[1]`, and that's undefined behavior. E.g. when compiled with `-D_GLIBCXX_DEBUG` it will trigger an assert.

src/cpu/Dasm.cc
```diff
@@ -108,18 +115,50 @@ unsigned dasm(const MSXCPUInterface& interface, uint16_t pc, std::span<uint8_t,
...
 	auto op0 = interface.peekMem(pc, time);
 	auto t = instr_len_tab[op0];
 	if (t < 4) return t;
 
 	auto op1 = interface.peekMem(pc + 1, time);
-	return instr_len_tab[64 * t + op1];
+	t = instr_len_tab[64 * t + op1];
+
+	auto op3 = interface.peekMem(pc + 3, time);
+	if (std::find(ddcb__.begin(), ddcb__.end(), op3) == ddcb__.end()) return 1;
```
m9710797: Same here.

src/unittest/Dasm.cc
```diff
@@ -0,0 +1,55 @@
+#include "Dasm.hh"
+#include "EmuTime.hh"
+#include "MSXCPUInterface.hh"
```
m9710797: Detail: `EmuTime` and `MSXCPUInterface` aren't needed.

src/unittest/Dasm.cc
```diff
@@ -0,0 +1,55 @@
...
+TEST_CASE("instructionLength, dasm")
+{
+	std::array<uint8_t, 4> opcode = {0, 0, 0, 0};
+	std::string dest;
+
+	bool end = false;
+	while (!end) {
+		auto len = instructionLength(opcode).value_or(0);
+		assert(1 <= len);
+		assert(len <= 4);
```
m9710797: In the context of `catch2` it might be better to replace `assert` with `REQUIRE`.

src/unittest/Dasm.cc
```diff
@@ -0,0 +1,55 @@
...
+			std::stringstream ss;
+			ss << std::hex;
+
+			for (unsigned j = 0; j < opcode.size(); ++j) {
+				ss << std::setw(2) << std::setfill('0') << int(opcode[j]) << ' ';
+			}
+			ss << '(' << dest << ")\n";
+			INFO( "opcode: " << ss.str());
+
+			auto instructionLength = len;
```
m9710797: This line confused me for a moment. I guess it's a left-over from a previous version.
Just use `len` in the next line? Or on line 18 rename `len` to `instructionLength`?

src/unittest/Dasm.cc
```diff
@@ -0,0 +1,55 @@
...
+					end = true;
+					break;
+				}
+				--idx;
+			} else {
+				++opcode[idx];
+				break;
+			}
+		}
+	}
```
m9710797, line 60: In my version of the test I also checked that we had exactly `2904196` iterations of this loop. I admit I did not calculate what this value should be. I only did some quick manual inspection of the output and then trusted the number that I got from running the test. But at least this will detect future (accidental?) changes.

I don't know if it's a good idea to add this check.

m9710797, line 60: In my test I also checked that each call to `dasm()` (with different input in `opcode`) returned a different string. I understand that check won't work in your case (when doing "recovery"). But we want to remove this recovery strategy.

I don't know if it's a good idea to check this. But at least it gave me more confidence that the logic in "goto next instruction" is correct.

Ah, and to make this pass I had to change how `ED 63 nn nn` gets disassembled. Currently it's disassembled to `ld (nn),hl` but that collides with `22 nn nn`. To fix that I changed the disassembly of this longer encoding to `ld (nn),hl*`, but feel free to pick a better string. I don't expect to see this string often (or at all) in normal code.

m9710797 (commented):
This started as a simple addition of `disasm_blob`, but it has grown a lot bigger. Also fixing bugs and changing behavior in the existing code. Thanks for your persistence.

In this latest revision you tried to enhance the "recovery" strategy on invalid instructions. The last days (thanks to you) I've given this some more thought, and now I think we shouldn't try to recover. 1) because that's not what the real Z80 does 2) because that's too complex (e.g. the current PR contains several bugs related to that). More details in the review comments below.

Thanks to your effort, I think, I now have a reasonably good idea of what this PR should look like. How do you want to continue?
* Continue normally, you make changes and I review?
* Or shall I make changes and you review?

src/cpu/Dasm.cc
pvmm: For the case when we reach line 39 with t = 3, but bin.size() = 2. Here, I have rewritten it.

src/cpu/Dasm.cc
pvmm: I made them return 0 in case of invalid Z80 instruction, otherwise the comparison with `instructionLength(std::span<const uint8_t> bin)` will fail in unittest.

src/cpu/Dasm.cc
pvmm: I left the binary search there because it is needed by `disasmBlobCommand()` but reverted everything else.

src/cpu/Dasm.cc
m9710797: These should return the length of the invalid instruction, not 0. Or actually didn't we already change the return type of dasm() to void?

From the disassembly point of view, there are no invalid instructions. For example the sequence `DD 00` is an instruction with length 2 (not 0) that produces the string `db #DD,#00`. And similar for all other _invalid_ instructions. And this is also how the real Z80 sees this sequence: it's a 2-byte instruction with the same effect as `nop`.

So these `return` statements in `case '#' '!' '@'` should be removed. And the statement above should be adjusted to print all `i` bytes.

The unittest should not do anything special for _invalid_ instructions. Again, from the disassembly point of view there are no invalid instructions.

m9710797:
Ok, I think I start to understand the confusion. In your commit message you wrote
```
Make disassembly more simple and consistent by processing a single byte if
the buffer doesn't contain a valid Z80 instruction (that's the only rule now)
```
* That's a valid rule. And it may be a good rule if you're trying to reverse engineer some block of binary data (e.g. figure out what is data and what is code).
* I have another rule in mind: try to disassemble instructions as close as possible to how the real Z80 sees them.

I think in the context of the openMSX debugger the 2nd rule is better. But of course that's just my opinion. My reasoning is the following:
* The debugger is meant to find bugs in your code. It should show as much as possible how the program behaves or mis-behaves.
* Suppose because of a bug (e.g. you've accidentally overwritten the stack) your program jumps to an invalid location (e.g. to a data-block). That block contains the sequence `DD 01 02 03`.
* Rule 1 would disassemble this as: `db #DD ; ld bc,#0302`
* Rule 2 would disassemble this ad `db #DD,#01 ; ld (bc),a ; inc bc`
* The latter is also how the Z80 will execute this sequence. It will not load the value 0x302 in BC, instead it will write to the memory location pointed to by BC.
* IMHO it will be easier to understand a misbehaving program (and thus find your bug) when we show the 2nd rule.

Before this PR, the debugger implemented a mix of these two rules. But while reviewing your code, I noticed this, and now I think we should change this (I already talked about this in my review comments, maybe you missed that?). And a nice side effect is that rule 2 also simplifies the code.

m9710797:
@pvmm I created this PR: https://github.com/openMSX/openMSX/pull/1721
It's mostly your PR with the changes I proposed. Can you review? If you agree, can we commit that one?
I thought it was easier to just show you what I mean rather than describe it in comments.

In this version the unittest passes now.
I'm not an expert in 'catch2', so I don't know if my changes in the unittest are idiomatic (but I think they are correct).
While debugging the unittest I found that the uses of `CHECKED_ELSE` and `INFO` didn't work as expected (they weren't helpful to locate the problem). So I rewrote those with regular tests and prints. (I found some doc that says the INFO must come before the CHECK (not after like in your version), but I'm not sure).

pvmm:
> That's a valid rule. And it may be a good rule if you're trying to reverse engineer some block of binary data (e.g. figure out what is data and what is code).

OK, but since yesterday, this text is out of date and this was no longer the rule. But I agree that we can close this and merge #1721 instead.

pvmm:
> That's a valid rule. And it may be a good rule if you're trying to reverse engineer some block of binary data (e.g. figure out what is data and what is code).

OK, but since yesterday, this text is out of date and this was no longer the rule. But I agree that we can close this and merge https://github.com/openMSX/openMSX/pull/1721 instead.

PR #1721, opened 2024-08-11
dasm extension: disassemble any Tcl binary string

m9710797:
This is an alternative for pvmm's PR:
   https://github.com/openMSX/openMSX/pull/1717

This branch is heavily based on pvmm's work. I only made some small adjustments.


pvmm:
My `binary_search` code from #1717 was included to address problems like this:
image
It should return that this is not valid z80 code. Can we do something about it?

m9710797:
The returned string starts with `db<space>` if (and only if) it's an invalid instruction.
Is that sufficient?

m9710797:
Thanks to pvmm!
I only reviewed and made some final tweaks.

PR #1732, opened 2024-08-25
disasmBlob should use the symbol manager

pvmm:
Allow user to give `disasm_blob` their own symbol manager, like this:
```
debug disasm_blob "\30\02" 0x100 {260 label1}
{jr   label1    } 2
```

### DIFF
--- a/src/debugger/Debugger.cc
+++ b/src/debugger/Debugger.cc
@@ -375,14 +375,27 @@ void Debugger::Cmd::disasm(std::span<const TclObject> tokens, TclObject& result,
 
 void Debugger::Cmd::disasmBlob(std::span<const TclObject> tokens, TclObject& result) const
 {
-	checkNumArgs(tokens, 4, Prefix{2}, "value addr");
+	checkNumArgs(tokens, Between{4, 5}, Prefix{2}, "value addr ?function?");
 	std::span<const uint8_t> bin = tokens[2].getBinary();
 	auto len = instructionLength(bin);
 	if (!len || *len > bin.size()) {
 		throw CommandException("Blob does not contain a complete Z80 instruction");
 	}
 	std::string dasmOutput;
-	dasm(bin.subspan(0, *len), word(tokens[3].getInt(getInterpreter())), dasmOutput);
+	unsigned addr = tokens[3].getInt(getInterpreter());
+	dasm(bin.subspan(0, *len), word(addr), dasmOutput,
+		[&](std::string& output, uint16_t a) {
+			zstring_view result;
+			if (tokens.size() > 4) {
+				auto command = makeTclList(tokens[4], a);
+				result = command.executeCommand(getInterpreter()).getString();
+			}
+			if (result.size() > 0) {
+				strAppend(output, result);
+			} else {
+				appendAddrAsHex(output, a);
+			}
+		});
 	dasmOutput.resize(19, ' ');
 	result.addListElement(dasmOutput);
 	result.addListElement(*len);
@@ -1008,11 +1021,14 @@ string Debugger::Cmd::help(std::span<const TclObject> tokens) const
 		"  Note that openMSX comes with a 'disasm' Tcl script that is much "
 		"more convenient to use than this subcommand.";
 	auto disasmBlobHelp =
-		"debug disasm_blob <value> <addr>\n"
+		"debug disasm_blob <value> <addr> [<function>]\n"
 		"  This is a more generic version of the disasm subcommand, but it "
 		"works on a Tcl binary string (see Tcl manual) to disassemble a "
 		"single instruction. The given address is used when relative "
-		"address to jump to is necessary.\n";
+		"address to jump to is necessary. The optional fuction will be "
+		"called with an address as parameter and may return a symbol name "
+		"that replaces that address if a symbol that matches the address "
+		"is found.\n";
 	auto symbolsHelp =
 		"debug symbols <subcommand> [<arguments>]\n"
 		"  Possible subcommands are:\n"


src/debugger/Debugger.cc
```diff
@@ -375,14 +375,24 @@ void Debugger::Cmd::disasm(std::span<const TclObject> tokens, TclObject& result,
...
 	if (!len || *len > bin.size()) {
 		throw CommandException("Blob does not contain a complete Z80 instruction");
 	}
 	std::string dasmOutput;
-	dasm(bin.subspan(0, *len), word(tokens[3].getInt(getInterpreter())), dasmOutput);
+	unsigned addr = tokens[3].getInt(getInterpreter());
+	dasm(bin.subspan(0, *len), word(addr), dasmOutput,
+		[&](std::string& output, uint16_t a) {
+			TclObject key(a);
+			auto label = tokens[4].getDictValue(getInterpreter(), key).getString();
```
m9710797: Unconditionally using `tokens[4]` when the user only passes 4 (not 5) arguments may crash.

src/debugger/Debugger.cc
```diff
@@ -375,14 +375,24 @@ void Debugger::Cmd::disasm(std::span<const TclObject> tokens, TclObject& result,
...
+			if (!label.empty()) {
+				strAppend(output, label);
+			} else {
+				appendAddrAsHex(output, a);
+			}
+		});
 	dasmOutput.resize(19, ' ');
```
m9710797, line 399: Now you require to pass the full dictionary, in other words you _eagerly_ need to account for all possible labels, What if we allow to _lazily_ translate values to names. So instead of passing an (optional) dictionary to the `disasm_blob` function we pass the name of a proc that can translate a value to a string-representation. For example:
```
proc formatLabel {value} {
    if {exists $::sym($value)} {
        return $::sym($value)
    } else {
        return [format "#%04x" $value]
    }
}
```
This example uses the global `$::sym` array, this contains all symbols known by the openMSX symbol manager. But of course you can also create a proc that does a lookup in your own dict.

m9710797 (changes requested):
I don't like the current approach, see review comments.

src/debugger/Debugger.cc
```diff
@@ -375,14 +375,24 @@ void Debugger::Cmd::disasm(std::span<const TclObject> tokens, TclObject& result,
...
 	if (!len || *len > bin.size()) {
 		throw CommandException("Blob does not contain a complete Z80 instruction");
 	}
 	std::string dasmOutput;
-	dasm(bin.subspan(0, *len), word(tokens[3].getInt(getInterpreter())), dasmOutput);
+	unsigned addr = tokens[3].getInt(getInterpreter());
+	dasm(bin.subspan(0, *len), word(addr), dasmOutput,
+		[&](std::string& output, uint16_t a) {
+			if (tokens.size() > 4) {
+				TclObject command(tmpStrCat(tokens[4].getString(), ' ', a));
```
m9710797: Better write this as:
```auto command = makeTclList(tokens[4], a);```
Caveat: I did not test this.

Short reasoning: the old ways is slower and also wrong in corner cases.

Longer reasoning:
Tcl is a weakly typed language. That is: given a variable you cannot tell whether it's a string, an integer, a list, ... Instead, depending on how it's used, a variable can be _interpreted_ as list, integer, ... E,g, when evaluating an expression like `$a + 1` the variable `a` will be interpreted as an integer (I'm making some simplifications here). Such interpretation may fail of course (e.g. "abc" can not be interpreted as an integer). Similarly in a command like `eval $a`, the variable `a` will (try to) be interpreted as a list (with the first element in the list the name of the proc/command, and the other elements in the list the arguments). In a way that's what happening here: `command.executeCommand(...)` will try to interpret the value of `command` as a list.

The string-representation of a simple Tcl list is just the elements separated by spaces. For example `a b c` can be interpreted as a list containing the 3 elements `a`, `b` and `c`. However elements (I mean their string representation) can themself contain spaces. Therefor there exist special encoding rules for lists. For example `{a b} c` represents a list of 2 elements, `a b`, and  `c`. `a b` can then be interpreted as a list of `a` and `b` or it could be the string "a space b" (there's no way to know, it really depends on how it's used). In the original code you tried to create a list by manually constructing the string representation, so appending a space followed by (the string representation of) the number `a`. That works in simple cases, but not when `tokens[4]` itself contains spaces.

From a Tcl user point of view it may seem that everything in Tcl is a string (which can be interpreted in different ways), but internally Tcl _does_ use a smarter representation. E.g. if some proc computes an integer, the TcllObject containing that result really holds an integer. That integer is not yet converted to a string representation. Only when we need the value of that variable in a string-context we do create the string-representation (and this conversion is then cached inside the TclObject). Or when we use some variable in an integer-context, we first check whether the object already contains a valid integer interpretation, if not only then do we try to parse the string representation as an integer. So the vast majority of conversions between string and integer can be avoided,

And the same for list representations. A TclObject can internally also hold a list of other TclObjects. So most of the time we don't need to compute the string-representation of that list (that's only done when _really_ needed, and often that's either a debug print or a programming error). Note that calculating (a correct) string-representation isn't trivial. Similarly when creating a new Tcl list, we don't first need to generate a string-representation and then parse that (parsing a string as a list is even more tricky than generating the string).

And detail: the original code did `tokens[4].getString()` this forces the string-representation of `tokens[4]` to be generated, while the new code just uses `tokens[4]`. In this specific case (when passing the name of a proc) most likely the string-representation is already available, but it's anyway a good idea to not force such conversions. And note that we don't create a string representation of `a` which then later needs to be parsed again into an integer. The old code did convert `a` to string.

File: src/debugger/Debugger.cc [hunk 4]
```diff
@@ -1008,11 +1018,14 @@ string Debugger::Cmd::help(std::span<const TclObject> tokens) const
 		"  Note that openMSX comes with a 'disasm' Tcl script that is much "
 		"more convenient to use than this subcommand.";
 	auto disasmBlobHelp =
-		"debug disasm_blob <value> <addr>\n"
+		"debug disasm_blob <value> <addr> <function>\n"
```
m9710797: `<function>` is optional. I think (please check) in other places we then write it as `[<function>]`.

m9710797 (changes requested):

src/debugger/Debugger.cc
pvmm: I tested your suggestion and it works.

src/debugger/Debugger.cc
```diff
@@ -1008,11 +1018,14 @@ string Debugger::Cmd::help(std::span<const TclObject> tokens) const
...
-		"debug disasm_blob <value> <addr>\n"
+		"debug disasm_blob <value> <addr> [<function>]\n"
 		"  This is a more generic version of the disasm subcommand, but it "
 		"works on a Tcl binary string (see Tcl manual) to disassemble a "
 		"single instruction. The given address is used when relative "
-		"address to jump to is necessary.\n";
+		"address to jump to is necessary. The optional fuction will be "
+		"called with an address as parameter and may return a symbol name "
+		"that replaces that address if a symbol that matches the address "
+		"is found.\n";
```
m9710797, line 1031: Detail: this help text says "... and _may_ return a symbol name ...". But in the current implementation the function _must_ return a string. And that string is always used, also if it is e.g. an empty string.
An alternative could be to fallback to `appendAddrAsHex()` if the function returns an empty string.
Either alternative is fine for me, pick which you like best: tweak the help text or change the implementation.

Can you also test what happens when the function returns an error instead of a string (e.g. as a test intentionally write an error in your function). What should happen is that the `debug disasm_blob` command fails with that error. E.g. it should not crash openMSX. I think the current implementation might already be fine, but we should make sure.

Can you also document `disasm_blob` in `doc/manual/commands.html`. Search for `debug disasm`, a similar single line documentation should be sufficient (no need for any details, the help text you added in openMSX itself already has the details).

m9710797 (changes requested):
Almost ready to merge.

src/debugger/Debugger.cc
pvmm: > Can you also test what happens when the function returns an error instead of a string (e.g. as a test intentionally write an error in your function). What should happen is that the `debug disasm_blob` command fails with that error. E.g. it should not crash openMSX. I think the current implementation might already be fine, but we should make sure.

It seems to return a TclException.
image
Tried with invalid data and it returns "invalid command name"

m9710797:
Merged and made some small tweaks.

PR #1744, opened 2024-09-09
MLTS2: remove debug output

zeilemaker54:
MLTS2: remove debug output, not needed anymore (bank switching works ok) and it floods the error output


MBilderbeek:
Thanks! And congratulations on your first PR on openMSX :)

PR #1749, opened 2024-09-18
Highlight known symbols in hex memory view

zappatic:
This patch highlights the known symbols in the memory view. This helped me keeping track of which memory locations I had already taken a look at and which were currently unknown. 
It also shows the symbol name of the current address at the bottom of the window. All behind a toggle in the popup menu.

memory-symbols

The color of the highlight can obviously be changed if necessary, I just picked something with enough contrast.

### DIFF
--- a/src/imgui/DebuggableEditor.cc
+++ b/src/imgui/DebuggableEditor.cc
@@ -34,6 +34,7 @@ using namespace std::literals;
 
 static constexpr int MidColsCount = 8; // extra spacing between every mid-cols.
 static constexpr auto HighlightColor = IM_COL32(255, 255, 255, 50); // background color of highlighted bytes.
+static constexpr auto HighlightSymbolColor = IM_COL32(148, 95, 35, 255); // background color of known symbol bytes.
 
 DebuggableEditor::DebuggableEditor(ImGuiManager& manager_, std::string debuggableName_, size_t index)
 	: ImGuiPart(manager_)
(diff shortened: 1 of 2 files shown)


m9710797:
Hi,

Thanks a lot for this PR!

This weekend we plan to make the official openMSX-20 release. The days before such a release we tend to be super conservative in adding/extending new features (we only do bug fixes). But soon after I think we can merge this.

I already had a quick view at the code changes. Looks good.

Here are some general remarks/ideas I had while reviewing the code. But no need to make any changes, I think we can (later) merge your PR as is:
* This patch only highlights a single byte per symbol. But it could also be that a symbol refers to a 16-bit value. Or it could be that a symbol refers to code (not data), and in that case it's not clear how many bytes to highlight. Though I don't think there's anything we can do about that now: the current symbols (and in particular the symbol files) don't contain sufficient information to address this.
* The current implementation adds an extra status line with the name of the symbol. I agree that's useful information. But I can also imagine that some users don't like the extra space that it takes (e.g. depending on the use case, when most of the time the cursor won't point to a symbol). I don't have a good solution in mind yet, but maybe we can find something to (optionally) have highlighting without that extra status line? And maybe the symbol name can (also/only) be shown when hovering the mouse cursor over the highlighted region? Anyway that's something for a later follow-up PR (if we want to change this at all).

So thanks again. I'll probably merge this PR soon after the release, hopefully early next week.

Wouter

PR #1751, opened 2024-09-18
Added support for loading symbol files generated by WLA-DX

mdagois:
This PR adds support for loading `no$gmb` symbol files generated by WLA-DX.
The `no$gmb` symbol format is used by other assembly tools, so the `loadNoGmb` should be reusable.
However, the auto-detect feature will only work for files generated by WLA-DX right now.

Note that WLA-DX can also generate symbol files in its own format, but this custom format is not supported by this PR.

### DIFF
--- a/src/debugger/SymbolManager.cc
+++ b/src/debugger/SymbolManager.cc
@@ -23,13 +23,14 @@ zstring_view SymbolFile::toString(Type type)
 {
 	switch (type) {
 		using enum Type;
-		case AUTO_DETECT: return "auto-detect";
-		case ASMSX:       return "asMSX";
-		case GENERIC:     return "generic";
-		case HTC:         return "htc";
-		case LINKMAP:     return "linkmap";
-		case NOICE:       return "NoICE";
-		case VASM:        return "vasm";
+		case AUTO_DETECT:   return "auto-detect";
+		case ASMSX:         return "asMSX";
+		case GENERIC:       return "generic";
+		case HTC:           return "htc";
+		case LINKMAP:       return "linkmap";
+		case NOICE:         return "NoICE";
+		case VASM:          return "vasm";
+		case WLALINK_NOGMB: return "wlalink";
 		default: UNREACHABLE;
 	}
 }
@@ -44,6 +45,7 @@ std::optional<SymbolFile::Type> SymbolFile::parseType(std::string_view str)
 	if (str == "linkmap")     return LINKMAP;
 	if (str == "NoICE")       return NOICE;
 	if (str == "vasm")        return VASM;
+	if (str == "wlalink")     return WLALINK_NOGMB;
 	return {};
 }
 
@@ -76,6 +78,8 @@ SymbolManager::SymbolManager(CommandController& commandController_)
 			return GENERIC;
 		} else if (StringOp::containsCaseInsensitive(line, "Sections:")) {
 			return VASM;
+		} else if (StringOp::containsCaseInsensitive(line, "; this file was created with wlalink")) {
+			return WLALINK_NOGMB;
 		} else {
 			// this is a blunt conclusion but I don't know a way
 			// to detect this file type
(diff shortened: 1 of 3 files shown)


m9710797:
Hi,

Thanks a lot for this PR!

This weekend we plan to make the official openMSX-20 release. The days before such a release we tend to be super conservative in adding/extending new features (we only do bug fixes). But soon after I think we can merge this.

I had a quick look at the code changes. They all look fine. My only request is that you also extend the unittest in `src/unittest/SymbolManager_test.cc`. Or I could do that.

To build the unittest we use the build-flavour located in `build/flavour-unittest.mk`. On linux you build this with `make OPENMSX_FLAVOUR=unittest` but I have to admit: I don't know how to build these on windows. Does somebody else know?

Thanks again. I'll probably merge this PR soon after the release, hopefully early next week.

Wouter

mdagois:
> This weekend we plan to make the official openMSX-20 release. The days before such a release we tend to be super conservative in adding/extending new features (we only do bug fixes). But soon after I think we can merge this.

Perfectly understandable, and great news!
Version 20 is a huge improvement over 19. :)

> I had a quick look at the code changes. They all look fine. My only request is that you also extend the unittest in `src/unittest/SymbolManager_test.cc`. Or I could do that.

I updated the test file. Please have a look.

> To build the unittest we use the build-flavour located in `build/flavour-unittest.mk`. On linux you build this with `make OPENMSX_FLAVOUR=unittest` but I have to admit: I don't know how to build these on windows.

I am trying to build on WSL. Compilation is very slow, but at least it seems to work for now.

> Thanks again. I'll probably merge this PR soon after the release, hopefully early next week.

Sure, no rush at all.

mdagois:
```
(base) mdagois@HomePC:~/openMSX$ ./derived/x86_64-linux-unittest/bin/openmsx
===============================================================================
All tests passed (14543989 assertions in 159 test cases)
```
All clear. :)

m9710797:
Merged. Thanks again.

I have a follow-up question. Currently only lines starting with `00:` are accepted. Any idea what that value (zero) represents? Could it be the segment number?

Earlier @pvmm extended our symbols with segment information. I wonder if that could be compatible with the no$gbm symbol file format.

mdagois:
> Merged. Thanks again.

Thanks!

> I have a follow-up question. Currently only lines starting with `00:` are accepted. Any idea what that value (zero) represents? Could it be the segment number?

It is the bank number.
Right now, WLA-DX always outputs `00:`, unless it targets a specific SNES emulator.

> Earlier @pvmm extended our symbols with segment information. I wonder if that could be compatible with the no$gbm symbol file format.

What does `segment` mean in the context of openMSX?
I am new to MSX development, I understand the slots and pages system, but that's about it.

m9710797:
Background info (I'm making some slight simplifications):
At the hardware level, the MSX slot (and subslot) stuff is handled at the level of the main board. A MSX cartridge (e.g. containing a game ROM) is "located" in exactly one slot, meaning that the hardware read and write signals for that cartridge only get activated when that specific slot is selected. So from this point of view, an cartridge only has a 16-bit (64kB) address space. To overcome this limitation the hardware on the cartridge itself can implement additional switching logic. E.g. writes to a specific region (e.g. to a region where ROM is located) may be interpreted as switching to a different bank. And then for example reads from one address in this 16-bit address space get remapped to a different location in a (larger than 64kB) ROM chip. I assume this is all very similar to how bank-switching works on other systems. The terms "bank" and "segment" are more or less synonyms in this context.

I have no idea if the bank-number that can be produced by WLA-DX can be meaningful in the context of MSX. We could ignore it for now. At least until we find a good use case for it.

mdagois:
> Background info (I'm making some slight simplifications)...

Thanks for the explanation!
Yes, that does sound a lot like bank switching on the systems I know.

> I have no idea if the bank-number that can be produced by WLA-DX can be meaningful in the context of MSX. We could ignore it for now. At least until we find a good use case for it.

Once I get used to MSX development, I will check if I can add a meaningful bank number to WLA-DX.
Right now, I am setting up one single 32KB bank (from $4000 to $BFFF) when developing for MSX, which happens to work well, as page 1 is set to the cartridge slot at boot up, and I can switch page 2 to the cartridge slot at initialization time.

PR #1758, opened 2024-09-26
Add context menu with 'Show in Disassembly' entry to addresses in breakpoints window

zappatic:
I didn't add a context menu for watchpoints, which should jump to the memory viewer, because I'm not sure which memory viewer to address (there can be multiple, ...), and how to best solve that issue.


m9710797:
Hi. Thanks a lot!
I took your PR, simplified the implementation a bit, and merged it into master.

Issue #747, opened 2015-06-24
[Bug] hq/hqlite scaling of 512-pixel-wide modes to 3x or 4x not working [sf#353]


openMSX-import:
**Reported by manuelbi on 2009-07-27 21:56 UTC**
Scaling a screen 7 image to 3x or 4x scalefactor with hq and hqlite isn't working (it's not implemented according to wouter). This is unexpected, as it works fine in SDL, but there you don't have the benefit of hardware acceleration...

MBilderbeek:
One of our users who was using SDL with hq on 3x scaling also noticed this and reported it as a regression. Look at the difference:
SDL 3x hq:
![image](https://github.com/user-attachments/assets/539d818f-8252-4444-b103-3b677f5b41a7)
and on SDLGL-PP:
![image](https://github.com/user-attachments/assets/c4d3a156-d257-4af6-add5-b3bb7fd4b117)

Can we do something to fix this regression?

PR #1761, opened 2024-09-29
Make opengl scalers work with both 320 and 640 width screens

balyuf:
This patch wants to solve https://github.com/openMSX/openMSX/issues/747 ...
It works for me in WSL, not tried on Windows (where I would like to see it fixed mainly).
I don't have much experience with C++ and even less with scalers, so the code may be less than optimal, and perhaps not even entirely correct (functionally and/or in terms of memory handling).

Please review.
I have tried to fix all scalers that needed fixing, not only the HQ and HQ Lite scalers.
I have not fixed any existing bugs, other than making sure that the scalers also run in 640 width modes.
Hence, the TV scaler still has problems with its first and last line, the RGB scaler only works fine in 3x mode and the ScaleNX scaler still looks a bit weird in 4x mode.

### DIFF
--- a/src/video/scalers/GLHQLiteScaler.cc
+++ b/src/video/scalers/GLHQLiteScaler.cc
@@ -28,7 +28,7 @@ GLHQLiteScaler::GLHQLiteScaler(GLScaler& fallback_)
 
 	// GL_LUMINANCE_ALPHA is no longer supported in newer openGL versions
 	auto format = (OPENGL_VERSION >= OPENGL_3_3) ? GL_RG : GL_LUMINANCE_ALPHA;
-	edgeTexture.bind();
+	edgeTexture320.bind();
 	glTexImage2D(GL_TEXTURE_2D,    // target
 	             0,                // level
 	             format,           // internal format
@@ -38,11 +38,22 @@ GLHQLiteScaler::GLHQLiteScaler(GLScaler& fallback_)
 	             format,           // format
 	             GL_UNSIGNED_BYTE, // type
 	             nullptr);         // data
+	edgeTexture640.bind();
+	glTexImage2D(GL_TEXTURE_2D,    // target
+	             0,                // level
+	             format,           // internal format
+	             640,              // width
+	             240,              // height
+	             0,                // border
+	             format,           // format
+	             GL_UNSIGNED_BYTE, // type
+	             nullptr);         // data
 #if OPENGL_VERSION >= OPENGL_3_3
 	GLint swizzleMask1[] = {GL_RED, GL_RED, GL_RED, GL_GREEN};
 	glTexParameteriv(GL_TEXTURE_2D, GL_TEXTURE_SWIZZLE_RGBA, swizzleMask1);
 #endif
-	edgeBuffer.setImage(320, 240);
+	edgeBuffer320.setImage(320, 240);
+	edgeBuffer640.setImage(640, 240);
 
 	const auto& context = systemFileContext();
 	glPixelStorei(GL_UNPACK_ALIGNMENT, 1);
@@ -75,16 +86,16 @@ void GLHQLiteScaler::scaleImage(
 	unsigned dstStartY, unsigned dstEndY, unsigned dstWidth,
 	unsigned logSrcHeight)
 {
-	unsigned factorX = dstWidth / srcWidth; // 1 - 4
+	unsigned factorX = ((srcWidth / 320) * dstWidth) / srcWidth; // 1 - 4
 	unsigned factorY = (dstEndY - dstStartY) / (srcEndY - srcStartY);
 
-	if ((srcWidth == 320) && (factorX > 1) && (factorX == factorY)) {
+	if ((srcWidth % 320 == 0) && (factorX > 1) && (factorX == factorY)) {
 		setup(superImpose != nullptr);
 		src.setInterpolation(true);
 		glActiveTexture(GL_TEXTURE3);
 		offsetTexture[factorX - 2].bind();
 		glActiveTexture(GL_TEXTURE2);
-		edgeTexture.bind();
+		if (srcWidth == 320) edgeTexture320.bind(); else edgeTexture640.bind();
 		glActiveTexture(GL_TEXTURE0);
 		execute(src, superImpose,
 		        srcStartY, srcEndY, srcWidth,
@@ -104,9 +115,12 @@ void GLHQLiteScaler::uploadBlock(
 	unsigned srcStartY, unsigned srcEndY, unsigned lineWidth,
 	FrameSource& paintFrame)
 {
-	if ((lineWidth != 320) || (srcEndY > 240)) return;
+	if ((lineWidth % 320 != 0) || (srcEndY > 240)) return;
 
-	std::array<Endian::L32, 320 / 2> tmpBuf2; // 2 x uint16_t
+	std::array<Endian::L32, 320 / 2> tmp320; // 2 x uint16_t
+	std::array<Endian::L32, 640 / 2> tmp640; // 2 x uint16_t
+	std::span<Endian::L32> tmpBuf2;
+	if (lineWidth == 320) tmpBuf2 = tmp320; else tmpBuf2 = tmp640;
 	#ifndef NDEBUG
 	// Avoid UMR. In optimized mode we don't care.
 	ranges::fill(tmpBuf2, 0);
@@ -118,6 +132,7 @@ void GLHQLiteScaler::uploadBlock(
 	auto next = paintFrame.getLine(narrow<int>(srcStartY) + 0, buf2);
 	calcEdgesGL(curr, next, tmpBuf2, EdgeHQLite());
 
+	gl::PixelBuffer<uint16_t> &edgeBuffer = (lineWidth == 320) ? edgeBuffer320 : edgeBuffer640;
 	edgeBuffer.bind();
 	if (auto* mapped = edgeBuffer.mapWrite()) {
 		for (auto y : xrange(srcStartY, srcEndY)) {
@@ -125,12 +140,12 @@ void GLHQLiteScaler::uploadBlock(
 			std::swap(buf1, buf2);
 			next = paintFrame.getLine(narrow<int>(y + 1), buf2);
 			calcEdgesGL(curr, next, tmpBuf2, EdgeHQLite());
-			memcpy(mapped + 320 * size_t(y), tmpBuf2.data(), 320 * sizeof(uint16_t));
+			memcpy(mapped + lineWidth * size_t(y), tmpBuf2.data(), lineWidth * sizeof(uint16_t));
 		}
 		edgeBuffer.unmap();
 
 		auto format = (OPENGL_VERSION >= OPENGL_3_3) ? GL_RG : GL_LUMINANCE_ALPHA;
-		edgeTexture.bind();
+		if (lineWidth == 320) edgeTexture320.bind(); else edgeTexture640.bind();
 		glTexSubImage2D(GL_TEXTURE_2D,                      // target
 		                0,                                  // level
 		                0,                                  // offset x
--- a/src/video/scalers/GLHQLiteScaler.hh
+++ b/src/video/scalers/GLHQLiteScaler.hh
@@ -25,9 +25,9 @@ public:
 
 private:
 	GLScaler& fallback;
-	gl::Texture edgeTexture;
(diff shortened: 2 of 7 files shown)


balyuf:
The Code Analysis complains that there is too much duplicate code for my added lines, but I guess that this is simply a consequence of the pre-existing duplication, indicating a refactoring is needed, so I think we can ignore this check failure here?

What is more worrisome is that in the automatic builds ScaleNX scaler does not seem to kick in for both Windows flavors (mingw and vc) while for the Linux build it works (as it is working in my wsl build).  The other scalers seem to be consistent.  Not sure why the Windows builds would display different behavior from the Linux builds?

balyuf:
Remark that even in the 20.0 stock installation ScaleNX does not kick in for 256 (320) screens where it does in Linux builds.  So whatever the problem is, it was no introduced by me, it seems ...

balyuf:
So, for me, my goal to fix HQ (Lite) scaling on 512 (640) screens is reached, while I have not fixed pre-existing bugs, nor introduced new ones.  Pending code review of course ... :-)

m9710797:
Hi balyuf,

Thanks a lot for this PR, I'll try to make time for this one of the following days.

I haven't looked at your patch yet. But here's some background info anyway...
Today we only have scaling factors 2x, 3x and 4x. "Eventually" I'd like to make it possible to have arbitrary scaling factors. So also larger than 4x (because on 4k screen resolution, 4x may still be too small). But also non-integer factors (though I'm not sure yet whether non-integer scaling will look nice in all cases). And then I hope that will allow to (re-)enable docking into the main openMSX window. The main window can then be arbitrary sized and the MSX screen itself will take up the remaining space of that window (currently it always fills the full window).

Hopefully this PR can be a step in that direction.

Wouter

balyuf:
Thanks Wouter for the background, interesting plans for the future also.  My fix indeed only touches the src side, not the dst side, so should be orthogonal to your plans.  BTW, I had a look at the ScaleNX scaler and it seems that it should do the same as the default scaler.  So perhaps the Windows builds have it right and the Linux builds do something weird ...  Anyway ... you will see that my main contribution is to split the buffers in two, one for 320 and one for 640 and replacing the hardcoded 320's by size functions.  I am not sure if I have done that in the most efficient (or elegant) way?  Also I don't fully understand why the dst size scales with the scale factor (e.g. 1280 or 960 etc) but not with the src size (e.g. switching from 256 to 512 width does not change the dst size).  So this makes for a bit of a hack in the calculation of the X factor, as you will see.  Also in the RGB scaler there is one place where the srcWidth was set to 320 in case it was 1, and I have set it now to 640 as I don't know if I can make it dependent on some size of an object?  So, I hope that this change is not trashing some memory etc ...  For the rest I think the changes are fairly straightforward ...

m9710797:
Quick update: I started looking into this, but it's taking more work than expected. I don't think the approach in your PR is correct (but I'm also still evaluating).

balyuf:
Wouter, in what sense do you suspect it is not correct?  In terms of memory management?  Or in terms of doing the correct interpolation at all times?

m9710797:
Hi balyuf,

As I said, I was still evaluating. I'm not 100% sure about any of the following points. But since you ask, I'll try to explain anyway:

Two minor things:
* About memory management. At first sight it seems correct, though maybe sub-optimal. E.g. instead of allocating one texture for 320 pixels wide and another for 640, it should be possible to only have the largest one, and only use half of it in case of 320 (and then figure out how to draw only halve of it). But that's a detail.
* At some point you changed `srcWidth == 320` into `(srcWidth % 320) == 0`. That indeed makes it trigger for both 320 and 640. But also for 1280, and that can happen for the (unofficial) V9990 B7 mode (1024 pixels wide + borders). That would now crash. But that's a small oversight, easy to fix.

But my main concern was the following, and I'm least sure about this point:
* The HQ{Lite}Scale algorithms scale X and Y with the same amount. Let's take screen 7 as an example. Internally openMSX renders a screen 7 image (512x212 + borders) to a 640x240 buffer (I'm slightly simplifying). From that buffer the scaler should produce the output screen. Let's take 4x scaling as an example, in that case the output screen has a resolution of 1280x960 pixels. In principle the HQ4x algorithm blows up every input pixel to 4x4 output pixels, thus in principle it would produce an output of 2560x960 pixels. But that's twice too wide.
* So what I think is happening in your current approach is the following. The GLShader program runs for every output pixel, so in principle only for halve the number of pixels of what HQ4x would/should have produced. I think that's as-if you take the full theoretical HQ4x output and then throw away every other horizontal pixel.
* Maybe that's an acceptable result? Though it's not the same as what the old software implementation (I mean non-openGL implementation) of our HQ scaler did for screen7. That one would calculate the full output, and then instead of throwing away halve of the pixels, average pairwise pixels. (Or at least in principle, internally it was implemented more efficiently).
* At some point in the code we check `factorX == factorY` to ensure we scale both X and Y by the same factor. And that check is still present after your change. So I'm not sure how you make that work. I think you changed the meaning of `factorX` somehow, but it's not clear to me what the new meaning is.

So now that I've written down my points (now that I thought about it some more). My concern is not so much about the "correctness", but more about the "quality" of the result. (All without having tested it yet) I think it may look worse than our old software implementation (that did work on screen 7).

--

I also haven't studied your changes in GLRGBScaler and GLScaleNxScaler yet. Though your change in GLRGBScaler looks strange, or at least unexpected.

--

I plan to continue working on this, but I had busy week, and also a busy weekend coming up.

Wouter

balyuf:
Wouter,

Yes, this scaling factor mystery is what puzzles me too.  As I said, I don't understand why when switching from a 256 screen to a 512 screen the display geometry is not doubling?  It adapts itself to the scaling factor, but not to the screen width.  That to me is the strange thing we need to understand.  In order to compensate for that I had to introduce a bit of magic indeed for factorX, which happens here:
unsigned factorX = ((srcWidth / 320) * dstWidth) / srcWidth; // 1 - 4
As you can see I introduce a correction factor for the dstWidth (the rendered display width) which will be a doubling in case the msx screen (srcWidth) is 512 (640).

But, yes, it is a bit of a miracle indeed, now that I think of it, that it is displayed correctly.  I guess the opengl system itself will do another interpolation?  Which renders acceptable quality, though not entirely in line with HQ?  BTW, what happens when I enable fullscreen?  It seems that whatever scaling is in place, is stretched again and you can see that the anti-aliasing is different from windowed mode.  This, I feel, must involve the same issue, not?

Regards,
Filip.

balyuf:
PS: on my laptop screen, 4x HQ scaling gives a window which is bigger than my poor laptop screen (which BTW caused problems initially with 20.0 as I had my settings on 4x but SDL, so effectively only 3x, and now suddenly the window was bigger than the screen which resulted in the menu being invisible).  So fullscreen mode actually shrinks the scaling iso of stretching it, in my case.  Anyway, you can see the difference in anti-aliasing ...

balyuf:
PPS: the change in GLRGBScaler is itself in a existing hack to translate the one pixel to a full line.  However we don't know at this point what a full line width is.  Therefore 320 was taken, which was correct in the past, since only that line width was allowed.  Now we have (at least) two line width possibilities, so I adjusted it to the maximum (or so I thought) of 640, in the hope that it will do the job and not overwrite any memory in case the width is only 320 ...

m9710797:
Hi,

I've pushed 1afaa52df47c34e41de78c87a4caf661d36ddd21. That basically solves this issue in the same way like the approach taken in this PR. The differences are:
* Also support 1280 pixel wide modes (gfx9000 B7) (next to only 640 and 320).
* Share the same buffers/textures instead allocating one for each size.

Merging this PR took a lot longer than I hoped. The reason is that I first tried to implement the same algorithm as the old SDL version (main difference is: averaging pixel pairs instead of dropping one of the pixels in a pair, see comments above for details). Ideally the solution also steps in the direction of allowing non-integer scale-factors. But that makes it even more complex (e.g. one of the complications is that within a single frame you can e.g. have a screen-split between a 256 and 512 pixel wide mode).

Anyway, I lost some time because of this. But in the end I decided to already go with this intermediate solution. It's a clear improvement compared to doing nothing. And it already looks good enough (for now?). 

@balyuf thanks again for your work.
I'm closing this PR, but I'd still appreciate your help with testing (I only had time for some very basic tests).

m9710797:
Reopened because the implementation of this idea in openMSX gives a different result than this PR.
It's not yet clear which of the two is correct. But for sure this is unexpected: the result should have been the same.

I'll investigate, but I'm not sure when I'll have time.

m9710797:
I can already confirm the implementation in this PR is correct, the current implementation in openMSX is not.
I did not have time yet to investigate what is wrong.

m9710797:
Finally, I had enough time to debug/fix this. In the end it was a simple fix, but finding the problem took a lot of time.
@balyuf Can you help testing again. For me the output looks the same now. But I'd also like you to verify it.

balyuf:
Ok, indeed when it works, it works fine (i.e. the same as my proposal), however now that I start to do other tests, I see it only works for scale 4x.  For scale 3x and 2x it seems to fall back to the default scaler in case of 512 (640) modes.  Can you reproduce that?

m9710797:
@balyuf Thanks again for testing/reporting. It should be fixed now. Please try again.

balyuf:
Yes, everything is OK now, thanks!  There is however still the strange mystery that with the Linux binary in WSL ScaleNx works, but in Windows-VC it does not kick in (default scaler seems to be active), both in 256 and 512 modes.  This was also the case with my proposal ... I don't know how that's even possible ...  Anyway, I don't care too much about that scaler ...  So feel free to drop investigation ...

balyuf:
PS: we may even care less for the RGBTriplet scaler, but there is still this hack in src/video/scalers/GLRGBScaler.cc where in one case (1 pixel) the line width is set to 320 to force a full line, which probably should be reconsidered now that lines can be multiples of 320?  Anyway I tested it and I don't see anything wrong (apart from the fact it only works fine in one specific scale factor of course).

m9710797:
> ... but there is still this hack in src/video/scalers/GLRGBScaler.cc where in one case (1 pixel) the line width is set to 320 to force a full line, which probably should be reconsidered now that lines can be multiples of 320?

Any value different than 1 is fine (i mean one any of the supported line widths). This is needed to force rendering vertical border lines as display-area lines (needed because this is the only scaler where border lines look distinctly different from single-color display area lines). But whether we render it as a 256 or 512 pixel wide mode doesn't really matter.

balyuf:
> Yes, everything is OK now, thanks! There is however still the strange mystery that with the Linux binary in WSL ScaleNx works, but in Windows-VC it does not kick in (default scaler seems to be active), both in 256 and 512 modes. This was also the case with my proposal ... I don't know how that's even possible ... Anyway, I don't care too much about that scaler ... So feel free to drop investigation ...

I removed my branch as your approach covers everything I wanted to achieve, just in a better way, but this also closed this PR.  So, I am not sure if you want to proceed with this ScaleNX difference between Windows and Linux, but this can probably happen outside of this PR?

balyuf:
When running the Windows binaries from command line (both VC and mingw) and switching to ScaleNx, there is output indicating what might be wrong with it:
Warning(s) compiling shader "scale2x.vert":
ERROR: 0:14: Unary operation not folded into constant

Warning(s) compiling shader "scale2x.vert":
ERROR: 0:14: Unary operation not folded into constant

I think this refers to the following lines (not present in the other vert files):
float pi = 4.0 * atan(1.0);
float pi2 = 2.0 * pi;

Any idea what is going on and how to fix this?  Something that in Linux works fine but not in Windows?

balyuf:
Indeed, when I define pi and pi2 like this, it works fine:
float pi = 3.1415926535897932384626433832795;
float pi2 = 6.283185307179586476925286766559;

balyuf:
I tried some combinations and it is this one which causes troubles:
float pi = 4.0 * atan(1.0);

I think what happens is that atan(1.0) does not evaluate to a constant at compile time in the Windows compiler, while it does in the Linux compiler (or it is not required to evaluate to a constant)?

balyuf:
Also it seems ScaleNX does not kick in (default scaler) when set to 2x in 512 modes (ok in 320 modes), this both in Windows and in Linux.

m9710797:
Thanks a lot for investigating this!
So did I understand correctly that the following works?
```
float pi = 3.14159265358;
float pi2 = 2.0 * pi;
```

balyuf:
Yes, that combination should work.  At least it will work for 4x and 3x, not sure what the other issue is with 2x ...

PR #1767, opened 2024-09-30
Update FreeType to 2.13.3

MBilderbeek:
Another update...

### DIFF
--- a/build/3rdparty/3rdparty.props
+++ b/build/3rdparty/3rdparty.props
@@ -12,7 +12,7 @@
     <OpenMSXIntDir>$(BuildDir)\build</OpenMSXIntDir>
     <OpenMSXOutDir>$(BuildDir)\install</OpenMSXOutDir>
     <OpenMSXConfigDir>$(BuildDir)\config</OpenMSXConfigDir>
-    <LibNameFreeType>freetype-2.11.1</LibNameFreeType>
+    <LibNameFreeType>freetype-2.13.3</LibNameFreeType>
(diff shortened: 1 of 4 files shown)


mthuurne:
```
            Creating library ..\..\derived\x64-VC-Release\install\openmsx.lib and object ..\..\derived\x64-VC-Release\install\openmsx.exp
     2>freetype.lib(ftinit.obj) : error LNK2001: unresolved external symbol ft_sdf_renderer_class [D:\a\openMSX\openMSX\build\msvc\openmsx.vcxproj]
     2>freetype.lib(ftinit.obj) : error LNK2001: unresolved external symbol ft_svg_renderer_class [D:\a\openMSX\openMSX\build\msvc\openmsx.vcxproj]
     2>freetype.lib(ftinit.obj) : error LNK2001: unresolved external symbol ft_bitmap_sdf_renderer_class [D:\a\openMSX\openMSX\build\msvc\openmsx.vcxproj]
     2>..\..\derived\x64-VC-Release\install\openmsx.exe : fatal error LNK1120: 3 unresolved externals [D:\a\openMSX\openMSX\build\msvc\openmsx.vcxproj]
 ```
You mentioned on IRC that you disabled SVG support, but it looks like the SVG implementation was indeed disabled and therefore `ft_svg_renderer_class` does not exist, but the FreeType init code in `ftinit.obj` is still trying to reference it. So I would recommend looking into the source code that produces `ftinit.obj` to see how optional features like SVG support are handled there.

PR #1770, opened 2024-10-05
Panasonic FS-JH1 Joy Handle emulation

theNestruo:
Disclaimer: I am not familiar with C++ development; most of the changes have been done copying from other source codes (such as JoyMega, Paddle, BooleanInput, ...)

An attempt to support Panasonic FS-JH1 Joy Handle emulation. My steps so far:
- Cloned _MSXJoystick_ as _JoyHandle_,
- added it in _PluggableFactory_, _ImGuiConnector_ and _ImGuiSettings_,
- added handling of _BooleanJoystickAxis_ + _JoystickAxisMotionEvent_ in _signalMSXEvent()_ to update _analogValue_, and
- used the _analogValue_ in _read()_.

Code is compiling, Joy Handle can be plugged in its inputs can be mapped, ~~but it seems that the _analogValue_ is never updated~~.
_[Edit: fixed in a260b848eb48734950f1977d52ebaa9dd01a40d2]_

Other things in the _to do_ list:
- _analogValue_ in _JoyHandleState_.
- _analogValue_ serialization.
- Proper input mapping window (right now it is just a clone of MSX Joystick).

### DIFF
--- a/build/msvc/openmsx.vcxproj
+++ b/build/msvc/openmsx.vcxproj
@@ -510,6 +510,7 @@
     <ClCompile Include="$(OpenMSXSrcDir)\input\JoystickPort.cc" />
     <ClCompile Include="$(OpenMSXSrcDir)\input\JoyTap.cc" />
     <ClCompile Include="$(OpenMSXSrcDir)\input\JoyMega.cc" />
+    <ClCompile Include="$(OpenMSXSrcDir)\input\JoyHandle.cc" />
     <ClCompile Include="$(OpenMSXSrcDir)\input\Keyboard.cc" />
     <ClCompile Include="$(OpenMSXSrcDir)\input\KeyboardSettings.cc" />
     <ClCompile Include="$(OpenMSXSrcDir)\input\MSXJoystick.cc" />
--- a/build/msvc/openmsx.vcxproj.filters
+++ b/build/msvc/openmsx.vcxproj.filters
@@ -1462,6 +1462,7 @@
     <ClCompile Include="$(OpenMSXSrcDir)\fdc\RawTrack.cc" />
     <ClCompile Include="$(OpenMSXSrcDir)\fdc\DMKDiskImage.cc" />
     <ClCompile Include="$(OpenMSXSrcDir)\input\JoyMega.cc" />
+    <ClCompile Include="$(OpenMSXSrcDir)\input\JoyHandle.cc" />
     <ClCompile Include="$(OpenMSXSrcDir)\memory\RomDooly.cc" />
     <ClCompile Include="$(OpenMSXSrcDir)\sound\ResampledSoundDevice.cc" />
     <ClCompile Include="$(OpenMSXSrcDir)\video\scalers\GLHQLiteScaler.cc" />
--- a/src/PluggableFactory.cc
+++ b/src/PluggableFactory.cc
@@ -4,6 +4,7 @@
 #include "MSXMotherBoard.hh"
 #include "Reactor.hh"
 #include "JoyMega.hh"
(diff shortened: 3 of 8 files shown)


m9710797:
Hi,

Thanks for working on this.

I understand it's still in draft, so maybe a detailed code review is not yet appropriate?
If you have any questions, or you need help with something, feel free to ask.

I had a very quick look at the code already. My only remark is about the `checkTime()` function. And then specifically about the `lastTime` state. Imagine the MSX only reads the joystick ports every 1000ms. Then in the current implementation we toggle `cycle` every 1000ms instead of every 500ms (like the real hardware does). Or in other words: reading every second on real hardware should result in reading the same `cycle` every time, though in this implementation we see `cycle` toggling.

A possible solution is to calculate `cycle` like this:
```
Clock<2> clock(EmuTime::zero()); // ticks at 2Hz
uint8_t cycle = clock.getTicksTill(time) & 1;
```
So we use the `Clock` helper class, instantiated with `2`, meaning it ticks at 2Hz (and the first tick happens at time zero). Then we ask: how many ticks did this clock produce between the start (zero) and now? And with `& 1` we transform that to `0` or `1` (even or odd cycle).

As a bonus this allows to remove the `lastTime` and `cycle` member variables from the `JoyHandle` class. (`cycle` becomes a local variable inside the `read()` method.) And having less state inside a class also allows to simplify stuff like serialization.

theNestruo:
Clock change makes sense. I adapted my own checkTime() from JoyMega, but obviously it does not work the way I thought it did.

About detailed code review: it is ok, but keep in mind that I am not familiar with C++ development and, likely, I won't be able to fix everything by myself :/

m9710797:
> Clock change makes sense. I adapted my own checkTime() from JoyMega, but obviously it does not work the way I thought it did.

The main difference is that in JoyMega the time interval starts from the last write, so the MSX controls this timing.
As far as I understand (but I admit: I don't know _exactly_ how the FS-JH1 works) in this case the cycle toggles at 2Hz regardless of whether the MSX read or writes the joystick ports.

 
> About detailed code review: it is ok, but keep in mind that I am not familiar with C++ development and, likely, I won't be able to fix everything by myself :/

No problem at all. I'm happy you're willing to work on this.
At any time feel free to ask for help, or ask for clarifications, ...

m9710797:
Just checking ...
What's the status of this PR?
Is there anything you need help with?

theNestruo:
Right now, the wheel of the FS-JH1 can be emulated mapping an analog stick.

I have been unable to validate the behaviour using a real FS-JH1, as the unit I had access was malfunctioning and we were unable to open it to fix it.

Neither the joystick/wheel switch nor the autofire sliders are emulated (1). The joystick/wheel switch can be easily emulated (it just disables left/right input from either the joystick or the wheel), but I don't know how to show a popup (something like "FS-JH1 is now on joystick mode").

Besides the actual emulation, the settings (at ImGuiSettings.cc) have two main issues:
- Right now, the graphic is copied from msxjoystick (2). I don't know how to draw the FS-JH1 with code.
- The code of that class was full of conditionals to alternate between msxjoystick and joymega. Now, there are even more conditionals to alternate between msxjoystick, joymega, and joyhandle. I guess it would be better to have some kind of interface/inheritance there to remove all those conditionals (3)... but I don't have the knowledge to do that in C++.

From my point of view, using MoSCoW prioritization:
- The graphic in the settings window (2): must have
- The missing switch/sliders (1): should have
- The code improvement (3): could have

theNestruo:
Abandoned in favor of https://github.com/openMSX/openMSX/pull/2145

PR #1771, opened 2024-10-05
Run auto-run on rewind instead of insert tape.

### LINKED ISSUE #1724: Resetting MSX shouldn't rewind tape
...probably I implemented this myself many years ago, but it seems wrong to rewind the tape when the MSX is reset. I guess it was done to trigger the autorun again. But on a real MSX, this of course doesn't happen either. And if you want to load the next game on a multi-game tape and reset in between, there's no way to continue from where you left off... so, I regret this implementation.

Instead:
- perhaps it would make sense to run the auto run after rewinding
- we should automatically rewind when inserting a tape (but that should be a no-op) and thus trigger the auto run
- the tape should not need to be reinserted automatically on a BOOT event. 

It's a little tricky to make this correct, i.e. to keep a valid state of the cassetteplayer in all cases of recording, rewinding, playing and all the cassetteplayer commands that influence it... but I guess we should look how far we can get.

MBilderbeek:
This will make sure that at boot the tape is not reinserted. That was a bad idea, because it will make it impossible to reset and continue loading a next item on the tape. A real MSX also doesn't rewind the tape when resetting it.

Fixes #1724.

### DIFF
--- a/src/cassette/CassettePlayer.cc
+++ b/src/cassette/CassettePlayer.cc
@@ -38,7 +38,6 @@
 #include "Reactor.hh"
 #include "GlobalSettings.hh"
 #include "CommandException.hh"
-#include "EventDistributor.hh"
 #include "FileOperations.hh"
 #include "WavWriter.hh"
 #include "TclObject.hh"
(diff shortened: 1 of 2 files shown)


MBilderbeek:
@m9710797 can you review this patch, please?

MBilderbeek:
Note that inserting a tape still triggers auto-run, as is explained in #1724 - we rely on the fact that rewind is always executed and we couple the auto-run to rewind.

m9710797:
Merged. Thanks!

PR #1773, opened 2024-10-06
Allow speed settings smaller than 1 (float value).

### LINKED ISSUE #1742: Cannot slow down speed more than 1%
I thought it would be nice to demo the real time debugger, so I tried to set the speed really low, but then discovered it can't be lower than 1%, because the speed setting is an integer setting.

Suggest to make it a float setting, and when doing so, we'll have to check whether the old Catapult isn't broken due to the change.

MBilderbeek:
This does break updates from openMSX to the Catapult slider. But that's it, the display value is still correct, and control still works towards openMSX.

Fixes #1742.

### DIFF
--- a/src/SpeedManager.cc
+++ b/src/SpeedManager.cc
@@ -7,11 +7,11 @@ namespace openmsx {
 SpeedManager::SpeedManager(CommandController& commandController)
 	: speedSetting(commandController, "speed",
 	       "controls the emulation speed: higher is faster, 100 is normal",
-	       100, 1, 10000, Setting::Save::NO)
+	       100.0, 0.01, 10000.0, Setting::Save::NO)
 	, fastforwardSpeedSetting(commandController, "fastforwardspeed",
 	       "controls the emulation speed in fastforward mode: "
-		   "higher is faster, 100 is normal",
-	       2000, 1, 10000)
+		   "higher is faster, 100 is normal speed",
+	       2000.0, 0.01, 10000.0)
 	, fastforwardSetting(commandController, "fastforward",
 	       "select emulation speed:\n"
 	       " on -> fastforward speed ('fastforwardspeed' setting)\n"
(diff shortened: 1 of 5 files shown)


MBilderbeek:
There is still a bit of discussion about the usability of the speed slider. So, please don't merge until we have finished the discussion.

src/SpeedManager.cc
```diff
@@ -7,7 +7,7 @@ namespace openmsx {
 SpeedManager::SpeedManager(CommandController& commandController)
 	: speedSetting(commandController, "speed",
 	       "controls the emulation speed: higher is faster, 100 is normal",
-	       100, 1, 10000, Setting::Save::NO)
+	       100.0, 0.01, 10000.0, Setting::Save::NO)
```
m9710797, line 10: For consistency, also change the constants (e.g. add `.0`) in `fastforwardSpeedSetting`.

src/SpeedManager.cc
```diff
@@ -33,7 +33,7 @@ SpeedManager::~SpeedManager()
 void SpeedManager::updateSpeed()
 {
 	speed = (fastforwardSetting.getBoolean() ? fastforwardSpeedSetting : speedSetting)
-	        .getInt() * (1.0 / 100.0);
+	        .getFloat() / 100.0;
```
m9710797: Please revert this. There's a technical reason why we wrote it like this. See https://godbolt.org/z/6hfee6GPn.

We want a multiplication here. But the compiler is not allowed to transform this division into multiplication by itself, because this may have different rounding behavior (because the extra 1.0 / 100.0 also gets rounded).

In the past we compiled with `-ffast-math`. That does give permission to the compiler for these transformation. But it gave problems in some Dear ImGui calculation.

m9710797 (commented):
If you don't want your PR to be merged yet, you can mark it as a "draft".

Do you also still plan to investigate using a logarithmic slider?

src/SpeedManager.cc
MBilderbeek: I don't know why I'm not seeing it here, but I did so in my last force-push....

src/SpeedManager.cc
MBilderbeek: Done in the latest version.

src/SpeedManager.cc
MBilderbeek: OK, apparently I only updated the commitmessage and I did not commit the actual changes... fixed now. Sorry about the fuss.

PR #1775, opened 2024-10-07
Add motorcontrol to tape deck window.

### LINKED ISSUE #1734: Allow to disable cassette player remote control in GUI
It's a rare use case, but at least for the CD- Sequential use case this is needed, so it would be nice if this could be controlled on the GUI as well, preferably on the Tape Deck window.

MBilderbeek:
Fixes #1734.

### DIFF
--- a/src/cassette/CassettePort.hh
+++ b/src/cassette/CassettePort.hh
@@ -58,6 +58,11 @@ public:
 	*/
 	virtual void setLaserdiscPlayer(LaserdiscPlayer *laserdisc) = 0;
 #endif
+
+	/**
+	* Get the cassette player (if available)
+	*/
+	virtual CassettePlayer* getCassettePlayer() = 0;
 };
 
 class CassettePort final : public CassettePortInterface, public Connector
@@ -72,6 +77,7 @@ public:
 	void setLaserdiscPlayer(LaserdiscPlayer* laserdisc) override;
 #endif
 	[[nodiscard]] bool lastOut() const override;
+	CassettePlayer* getCassettePlayer() override { return cassettePlayer; }
 
 	// Connector
 	[[nodiscard]] std::string_view getDescription() const override;
@@ -105,6 +111,7 @@ public:
 	void setLaserdiscPlayer(LaserdiscPlayer *laserdisc) override;
 #endif
 	[[nodiscard]] bool lastOut() const override;
+	CassettePlayer* getCassettePlayer() override { return nullptr; }
 };
 
 } // namespace openmsx
--- a/src/imgui/ImGuiMedia.cc
+++ b/src/imgui/ImGuiMedia.cc
@@ -6,7 +6,9 @@
 #include "ImGuiUtils.hh"
 
 #include "CartridgeSlotManager.hh"
+#include "CassettePlayer.hh"
 #include "CassettePlayerCLI.hh"
+#include "CassettePort.hh"
 #include "DiskImageCLI.hh"
 #include "DiskImageUtils.hh"
 #include "DiskManipulator.hh"
@@ -587,11 +589,11 @@ void ImGuiMedia::showMenu(MSXMotherBoard* motherBoard)
 
 		// cassetteplayer
 		elementInGroup();
-		if (auto cmdResult = manager.execute(TclObject("cassetteplayer"))) {
+		if (auto* player = motherBoard->getCassettePort().getCassettePlayer()) {
 			ImGui::MenuItem("Tape Deck", nullptr, &cassetteMediaInfo.show);
 			simpleToolTip([&]() -> std::string {
-				auto tip = cmdResult->getListIndexUnchecked(1).getString();
-				return !tip.empty() ? std::string(tip) : "Empty";
+				auto current = player->getImageName().getResolved();
+				return current.empty() ? "Empty" : current;
 			});
 		} else {
 			ImGui::TextDisabled("No cassette port present");
@@ -714,8 +716,8 @@ void ImGuiMedia::paint(MSXMotherBoard* motherBoard)
 	}
 
 	if (cassetteMediaInfo.show) {
-		if (auto cmdResult = manager.execute(TclObject("cassetteplayer"))) {
-			cassetteMenu(*cmdResult);
+		if (auto* player = motherBoard->getCassettePort().getCassettePlayer()) {
+			cassetteMenu(*player);
 		}
 	}
 }
@@ -1386,14 +1388,14 @@ static void RenderRecord(gl::vec2 center, ImDrawList* drawList)
 }
 
 
-void ImGuiMedia::cassetteMenu(const TclObject& cmdResult)
+void ImGuiMedia::cassetteMenu(CassettePlayer& cassettePlayer)
 {
 	ImGui::SetNextWindowSize(gl::vec2{29, 20} * ImGui::GetFontSize(), ImGuiCond_FirstUseEver);
 	auto& info = cassetteMediaInfo;
 	auto& group = info.group;
 	im::Window("Tape Deck", &info.show, [&]{
 		ImGui::TextUnformatted("Current tape"sv);
-		auto current = cmdResult.getListIndexUnchecked(1).getString();
+		auto current = cassettePlayer.getImageName().getResolved();
 		im::Indent([&]{
 			if (current.empty()) {
 				ImGui::TextUnformatted("No tape inserted"sv);
@@ -1412,9 +1414,9 @@ void ImGuiMedia::cassetteMenu(const TclObject& cmdResult)
 
 		ImGui::TextUnformatted("Controls"sv);
 		im::Indent([&]{
-			auto status = cmdResult.getListIndexUnchecked(2).getString();
+			auto status = cassettePlayer.getState();
 			auto size = ImGui::GetFrameHeightWithSpacing();
-			if (ButtonWithCustomRendering("##Play", {2.0f * size, size}, status == "play", RenderPlay)) {
+			if (ButtonWithCustomRendering("##Play", {2.0f * size, size}, status == CassettePlayer::State::PLAY, RenderPlay)) {
 				manager.executeDelayed(makeTclList("cassetteplayer", "play"));
 			}
 			ImGui::SameLine();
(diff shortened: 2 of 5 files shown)


src/imgui/ImGuiMedia.cc
```diff
@@ -1501,6 +1501,15 @@ void ImGuiMedia::cassetteMenu(const TclObject& cmdResult)
 			if (auto* mute = dynamic_cast<BooleanSetting*>(controller.findSetting("cassetteplayer_ch1_mute"))) {
 				Checkbox(hotKey, "Mute tape audio", *mute, [](const Setting&) { return std::string{}; });
 			}
+			if (auto cmdResult2 = manager.execute(TclObject("cassetteplayer motorcontrol"))) {
```
m9710797: This creates a TclObject from a string. That string gets parsed (e.g. split into tokens) into a command. That command gets interpreted. That delegates to our c++ code (CassettePlayer::TapeCommand::execute()). There we do string comparisons to figure out which subcommand to execute. That subcommand creates another string. Via the Tcl interpreter that string gets returned to this code. And then finally here we parse that string. All this 60 times per second. All this just to query a boolean. 

I think instead we should add a public getter in CassettePlayer to query this boolean.

But I've written exactly the same pattern for "cassetteplayer getpos" and "cassetteplayer getlength". So I assume you copied my "bad example". We should also fix that. And probably the same anti-pattern is still present in various other places in our code.

So for the GUI we shouldn't do queries via Tcl command. At least not when those queries are repeated every frame. When they only happen occasionally, e.g. only when interacting with a widget, or maybe in the value of a tooltip (which are only shown when hovered), it's more acceptable. On the other hand, when making changes we _must_ do this via Tcl commands. The reason is that currently our replay mechanism is counting on those commands. (E.g. in this case changing the motorcontrol is something that should end up in the replay).

src/utils/zstring_view.hh
```diff
@@ -54,6 +54,10 @@ public:
 		return view().find(c, pos);
 	}
 
+	[[nodiscard]] constexpr auto find(std::string_view sv, size_type pos = 0) const {
+		return view().find(sv, pos);
+	}
```
m9710797: Good idea. I like this.

You could even go a step further and implement [contains()](https://en.cppreference.com/w/cpp/string/basic_string/contains) (because that's actually what you need above). The `contains()` method is new in c++23, so we cannot use it yet in our code. But because `zstring_view` is our own class (with an API that intentionally mimics the API of `std::string`) there's no reason we cannot yet add it.

That being said. If we would rewrite the above code to not use Tcl for queries, then this helper won't be used anymore. (But I don't mind adding it anyway).

m9710797 (commented):
I like the feature. But I think it can/should be implemented better.
But that's actually critique against my own code, because I assume you only copied my bad example from above.

src/imgui/ImGuiMedia.cc
MBilderbeek: I force-pushed a version where I add access to the plugged in cassette device, and added the motor control status query on the CassetteDevice API. Can you check whether that's the right direction, i.e. what you meant?

src/imgui/ImGuiMedia.cc
```diff
@@ -1501,6 +1503,14 @@ void ImGuiMedia::cassetteMenu(const TclObject& cmdResult)
 			if (auto* mute = dynamic_cast<BooleanSetting*>(controller.findSetting("cassetteplayer_ch1_mute"))) {
 				Checkbox(hotKey, "Mute tape audio", *mute, [](const Setting&) { return std::string{}; });
 			}
+			auto& player = reactor.getMotherBoard()->getCassettePort().getPluggedCasDev();
+			bool enabled = player.isMotorControlEnabled();
+			bool changed = ImGui::Checkbox("Motor control enabled", &enabled);
+			if (changed) {
+				manager.execute(TclObject(strCat("cassetteplayer motorcontrol ", enabled ? "on" : "off")));
```
m9710797: Detail. Here strCat() constructs a string, but one of the first things that the Tcl interpreter does with this string is to split it into a list of tokens. It's more efficient, and more important also easier, to directly create a Tcl list. Something like:
```
manager.execute(makeTclList("cassetteplayer", "motorcontrol", enabled ? "on" : "off"));
```

m9710797:
I didn't do a detailed review (I could do that this evening). Just a quick reflection on the direction you're taking here.
* You look at the device that's actually plugged in into the cassette-port.
* Today that can only be `CassettePlayer` or nothing (`DummyCassetteDevice`).
* Then you operate on this abstract device.
* Because of that you must make `isMotorControlEnabled()` a virtual method.

That's a valid solution. But it does introduce an inconsistency. The rest of the "Tape Deck" window directly operates on the `CassettePlayer`, whether or not it's plugged in (because that's how the `cassetteplayer` Tcl command works). So some (today most) of the window doesn't care about the plugged-in status. But the new motorcontrol does care. I'm not saying which is better, I'm only saying it's different.

In practice this probably doesn't matter: today we have only one device that fits into the cassetteport. And because of the auto-plug script, when there is a cassetteport we plug the cassetteplayer by default.

So an alternative implementation could be to not query the device that's plugged into the cassetteport, but instead directly query the cassetteplayer. (This could be a new query method on `CassettePort`). And then `isMotorControlEnabled()` could be implemented only in `CassettePlayer` (not in `DummyCassetteDevice`) and then it can be non-virtual.

I think this will also simplify the implementation. But it is a different behavior, and I'm not sure which behavior is better. (But I also think that in practice it doesn't matter).

MBilderbeek:
Yes, I did consider making the device directly available as well, but it resulted in some quirks:
1. there was already a way to query the (abstract) CassetteDevice, as a private method. It felt a bit double to add another method to get the actual cassette player.
2. there is no real access to the actual CassettePlayer, other than some weird raw pointer that is taken from the `unique_ptr` of the cassette player that is directly moved into the pluggingcontroller via `registerPluggable`

I'm not sure if all existing stuff works if we unplug the CassettePlayer, but I guess it does.

So, I guess we have to make a decision on how to continue. Regarding 'getpos' and `getlength`... I guess it's not too weird to have that also on the abstract interface and use the same path as I took here. But then again, what you propose can also be done, but then please give some guidance on how to make the actual `CassettePlayer` object properly available (if we choose that direction).

src/imgui/ImGuiMedia.cc
MBilderbeek: OK, addressed this in a new push.

m9710797:
The simplest solution is to add:
```CassettePlayer& getCassettePlayer() { return *cassettePlayer; }```
as a public method in CassettePort.

And yes it's a bit weird to have both an abstract and a concrete getter. The former is needed to allow for the possbility of unplugging the cassette player. The latter is needed for this new code. So the question is where to place that new getter.

From a pragmatic point of view CassettePort is the only place that keeps a concrete pointer to the CassettePlayer, all other places instead use a pointer to the base class. So therefor it's logical to place the getter there. If we ever (i don't think we will) introduce other CassetteDevices then we can revisit this decision. 

Another posibility is to search in PluggingController for the pluggable with name "cassetteplayer", and then cast this `Pluggable` to the type `CassettePlayer`. But IMHO that's more ugly.  (And also we do this every frame, so searching by name is again not the best idea from a performance point of view).

**Comment by MBilderbeek** [comment]
Yes, that was actually the first thing I did. It also means a little more, because the method must also be in `CassettePortInterface`. So it's not too different from the alternative implementation.

So what do you prefer?

**Comment by m9710797** [comment]
You're right. I didn't do a detailed enough review yet, I missed this.

What about this:
* in `CassettePortInterface`:   `virtual CassettePlayer* getCassettePlayer()` 
* in `CassettePort` this returns a valid non-nullptr.
* in `DummyCassettePort` this returns a nullptr.

Then in `ImGuiMedia.cc`:717 we can call this method to see if there is a cassette port in this machine. Instead of trying to execute the `cassetteplayer` Tcl command and then check if this gave an error or not. We can then immediately pass (a reference to) the `CassettePlayer` to `cassetteMenu()`. Inside that function we can then perform all queries directly on that object. So something like:
```
if (auto* player = motherBoard->getCassettePort().getCassettePlayer()) {
    cassetteMenu(*player);
}
```

And similar in `ImGuiMedia.cc`:590. In both cases we need to replace parsing the result of the `cassetteplayer` Tcl command with more direct queries.

You don't need to do that all in this PR. But it's allowed of course.

**Comment by MBilderbeek** [comment]
OK, so you prefer the direct access to the CassettePlayer object via that weird raw pointer in CassettePort. I'll update the patch later.

**Comment by MBilderbeek** [comment]
OK, so I implemented my original idea (which matches your idea, apparently) and simplified the code a bit. Not everywhere yet, but I can do that in a next iteration. Also, I think I'd better just revert the change in `zstring_view.hh`. Perhaps you are interested to put that in, separately from this PR?

File: src/utils/zstring_view.hh (hunk 2 above)
Comment (MBilderbeek) [reply src/utils/zstring_view.hh]: I propose to revert it for this PR. Please add it separately if you like it. In any way you like :)

File: src/cassette/CassettePort.hh [hunk 4]
```diff
@@ -58,6 +58,11 @@ public:
 	*/
 	virtual void setLaserdiscPlayer(LaserdiscPlayer *laserdisc) = 0;
 #endif
+
+	/**
+	* Get the cassette player connected to the port.
+	*/
```
m9710797, line 64: I agree with the implementation, but not with the comment.
This method returns the cassette player (if it exists). Whether or not it is plugged into the port.

src/utils/zstring_view.hh
m9710797: Yes, maybe it's indeed better to drop this.
I don't plan to add it until we need it.

m9710797 (commented):
Thanks. It's looking good. I may pull this PR this evening (when I've time to do a more detailed review).

Could you still adapt the code around line 590? I think that's the last remaining cassetteplayer query done via a Tcl command.

MBilderbeek:
Thanks! Please don't pull it until I processed your comments. I probably have time late this evening.

m9710797:
> Thanks! Please don't pull it until I processed your comments. 

Then mark the PR as "draft", then you cannot pull it. (I've now done that).

m9710797:
@MBilderbeek Can I also ask that you re-tune the initial window size on line 1393?
Because now, with empty `imgui.ini`, when you first open the window, the content already doesn't fit.

MBilderbeek:
> @MBilderbeek Can I also ask that you re-tune the initial window size on line 1393? Because now, with empty `imgui.ini`, when you first open the window, the content already doesn't fit.

image
This is with empty imgui.ini. What doesn't fit?

MBilderbeek:
> > Thanks! Please don't pull it until I processed your comments.
> 
> Then mark the PR as "draft", then you cannot pull it. (I've now done that).

It wasn't draft until you added comments :)

src/cassette/CassettePort.hh
MBilderbeek: Fixed.

m9710797:
Merged after re-tuning the window size.

MBilderbeek:
> > @MBilderbeek Can I also ask that you re-tune the initial window size on line 1393? Because now, with empty `imgui.ini`, when you first open the window, the content already doesn't fit.
> 
> image This is with empty imgui.ini. What doesn't fit?

Compare:
image

So, it ate a blank line. I didn't see that at first.... so I guess it is a good idea after all to add some extra line of space :)

Issue #1780, opened 2024-10-10
Toolbar buttons have inconsistent sizes and shapes


gilbertfrancois:
A minor (cosmetic) UI issue. The toolbar buttons have all different shapes and sizes, depending on the size of the icon. It would be nice(r) if all buttons have the same shape and size, when they only show an icon and no text label.

openMSX release version 20.0, running on macOS 14.6

<img width="334" alt="Screenshot 2024-10-10 at 23 00 09" src="https://github.com/user-attachments/assets/061220f4-4322-4010-b62a-18c86e7ba153">

m9710797:
"Fixed" in commit 459f221f56328ba0bbd97bd988edef6d825dffd5. Maybe not yet ideal, but good enough?
@gilbertfrancois good enough to close this ticket? Feel free to re-open (but then I need help to properly fix it).

gilbertfrancois:
You gave me the idea to make an adjustment, based on your commit 459f221f56328ba0bbd97bd988edef6d825dffd5. I just made another PR with commit cdc4b8daef1e5d5f145264f0b510ec9a991d4f23. It boils down to change the button type from image button to a regular button and print the glyph as a text label. This is easier to align and justify in the center of the button without scaling the glyph itself.

PR #1782, opened 2024-10-12
Issue#1780: Toolbar buttons have inconsistent sizes and shapes

gilbertfrancois:
The suggestion in this commit is to change the ImageButton to a Button and convert the glyph symbol to a null terminated character. This makes it easier to align and center the symbol in the button without undesired scaling. As an extra, the buttons have now a bit of margin and the symbols don't look crammed in the button. It takes now a 28x28 pixel size or the largest glyph size, if the default is too small.

See issue #1780, commit 459f221f56328ba0bbd97bd988edef6d825dffd5

I'm not sure if this is a better way, or if you want to keep the buttons as ImageButton type.

The buttons look now like in the screenshot. 
Screenshot 2024-10-12 at 23 27 15

### DIFF
--- a/src/imgui/ImGuiDebugger.cc
+++ b/src/imgui/ImGuiDebugger.cc
@@ -11,6 +11,7 @@
 #include "ImGuiUtils.hh"
 #include "ImGuiVdpRegs.hh"
 #include "ImGuiWatchExpr.hh"
+#include "imgui_internal.h"
 
 #include "CPURegs.hh"
 #include "Dasm.hh"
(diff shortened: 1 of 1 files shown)


m9710797:
Thanks! Looks promising.
I'm busy now, but I'll try to merge this soon. (There are some details I'd like to check first).

m9710797:
Thanks for your PR. It's a good idea, but unfortunately it's not quite correct.
You can make it go wrong when you select a different font size:
```main menu bar > Settings > GUI > Select font```
Then select an extreme font size for the "normal" font. This will make part of the icons invisible.

I played around with your patch, but I couldn't easily fix it.
I think the root of the problem is the following:
* Internally Dear ImGui calls `CalcTextSize()`, that result is combined with  the "ImGuiStyleVar_ButtonTextAlign" setting to center the text position  (BTW your PR sets this setting to the already default value).
* But `CalcTextSize()` assumes that all glyphs have the same height. That's true for regular text. But for these icons that assumption breaks.

I couldn't find an easy way to "compensate" for this.

Maybe an alternative is to draw a (text) button with an invisible text, and then render the icon (as an image) over it. That way we have full control over how the image is rendered. In `ImGuiMedia.cc` around line 1342 we already do something similar.

gilbertfrancois:
Or we change back to ImageButton and instead of using glyphs with inconsistent sizes that scale ugly, we use png icons and make the button size fixed, e.g. 28x28 px.

m9710797:
I've pushed b7e26f5d022. That one uses our helper function `ButtonWithCustomRendering()`. See commit message for more details.

Using png files was a valid alternative (e.g. we do that for the LED icons). I choose not to do that here for "logistical" reasons: then we have to ship those extra files, load them as textures, etc. All technically doable, but it seemed more hassle than the current solution.

@gilbertfrancois thanks for the discussion and inspiration.

gilbertfrancois:
I checked the commit. It looks good. Thanks.

Issue #1689, opened 2024-07-01
Make it easier to find your replay from the list: add data from the replay file


MBilderbeek:
A luxurious feature for a replay selector: show some data from the replay file, like:

- length of the replay in time format
- current machine name
- currently running software
- current machine time
- rerecord count :)

MBilderbeek:
That PR adds showing the save time of the replay. This makes it a lot easier to find it back. I wonder whether the other information adds anything useful, so let's keep this low priority until we think it does.

m9710797:
One possibility is to add extra columns that are hidden by default. And then also lazily retrieve the information (because it requires parsing the .omr files).
Though personally I don't think it's useful (enough).

MBilderbeek:
Agreed. So, let's work on other things for now :)

Issue #1688, opened 2024-07-01
Make it easier to find your replay from the list: add creation date/time


MBilderbeek:
When selecting/searching a reply, also make this date/time of the files visible, so it's easier to find back your old reply.

MBilderbeek:
That PR fixes this issue.

PR #1790, opened 2024-10-19
Make savestates and replays easier to find back by showing more information

MBilderbeek:
See #1688 and #1689.

### DIFF
--- a/src/imgui/ImGuiReverseBar.cc
+++ b/src/imgui/ImGuiReverseBar.cc
@@ -62,10 +62,32 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
 		}
 		ImGui::Separator();
 
+		auto formatFileTime = [](std::filesystem::file_time_type fileTime) {
+			// As we still want to support gcc-11 (and not require 13 yet), we have to use this method
+			// as workaround instead of using more modern std::chrono and std::format stuff.
+
+			// Convert file_time_type to system_clock::time_point (the standard clock since epoch)
+			auto sctp = std::chrono::time_point_cast<std::chrono::system_clock::duration>(
+					fileTime - std::filesystem::file_time_type::clock::now() +
+					std::chrono::system_clock::now());
+
+			// Convert system_clock::time_point to time_t (time since Unix epoch in seconds)
+			std::time_t cftime = std::chrono::system_clock::to_time_t(sctp);
+
+			// Convert time_t to local time (broken-down time in the local time zone)
+			std::tm* local_time = std::localtime(&cftime);
+
+			// Get the local time in human-readable format
+			std::stringstream ss;
+			ss << std::put_time(local_time, "%F %T");
+			return ss.str();
+		};
+
 		loadStateOpen = im::Menu("Load state ...", [&]{
 			if (!loadStateOpen) {
 				// on each re-open of this menu, we recreate the list of states
 				stateNames.clear();
+				stateNamesChanged = true;
 				for (auto context = userDataFileContext(STATE_DIR);
 				     const auto& path : context.getPaths()) {
 					foreach_file(path, [&](const std::string& fullName, std::string_view name) {
@@ -76,7 +98,6 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
 						}
 					});
 				}
-				ranges::sort(stateNames, std::greater<>{}, &StateNames::ftime);
 			}
 			if (stateNames.empty()) {
 				ImGui::TextUnformatted("No save states found"sv);
@@ -84,38 +105,77 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
 				im::Table("table", 2, ImGuiTableFlags_BordersInnerV, [&]{
 					if (ImGui::TableNextColumn()) {
 						ImGui::TextUnformatted("Select save state"sv);
-						im::ListBox("##list", ImVec2(ImGui::GetFontSize() * 20.0f, 240.0f), [&]{
-							for (const auto& [name_, _] : stateNames) {
-								const auto& name = name_; // clang workaround
-								if (ImGui::Selectable(name.c_str())) {
-									manager.executeDelayed(makeTclList("loadstate", name));
-								}
-								if (ImGui::IsItemHovered(ImGuiHoveredFlags_DelayShort) &&
-								    (previewImage.name != name)) {
-									// record name, but (so far) without image
-									// this prevents that on a missing image, we don't continue retrying
-									previewImage.name = std::string(name);
-									previewImage.texture = gl::Texture(gl::Null{});
+						int flags = ImGuiTableFlags_RowBg |
+							ImGuiTableFlags_BordersV |
+							ImGuiTableFlags_BordersOuter |
+							ImGuiTableFlags_Resizable |
+							ImGuiTableFlags_Sortable |
+							ImGuiTableFlags_Hideable |
+							ImGuiTableFlags_Reorderable |
+							ImGuiTableFlags_ContextMenuInBody |
+							ImGuiTableFlags_ScrollY |
+							ImGuiTableFlags_SizingStretchProp;
+						im::Table("##select-savestate", 2, flags, ImVec2(ImGui::GetFontSize() * 40.0f, 240.0f), [&]{
+							ImGui::TableSetupScrollFreeze(0, 1); // Make top row always visible
+							ImGui::TableSetupColumn("Name");
+							ImGui::TableSetupColumn("Date/time", ImGuiTableColumnFlags_DefaultSort | ImGuiTableColumnFlags_PreferSortDescending | ImGuiTableColumnFlags_WidthFixed);
+							ImGui::TableHeadersRow();
+							// check sort order
+							auto* sortSpecs = ImGui::TableGetSortSpecs();
+							if (sortSpecs->SpecsDirty || stateNamesChanged) {
+								sortSpecs->SpecsDirty = false;
+								stateNamesChanged = false;
+								assert(sortSpecs->SpecsCount == 1);
+								assert(sortSpecs->Specs);
+								assert(sortSpecs->Specs->SortOrder == 0);
 
-									std::string filename = FileOperations::join(
-										FileOperations::getUserOpenMSXDir(),
-										STATE_DIR, tmpStrCat(name, ".png"));
-									if (FileOperations::exists(filename)) {
-										try {
-											gl::ivec2 dummy;
-											previewImage.texture = loadTexture(filename, dummy);
-										} catch (...) {
-											// ignore
+								switch (sortSpecs->Specs->ColumnIndex) {
+								case 0: // name
+									sortUpDown_String(stateNames, sortSpecs, &StateNames::name);
+									break;
+								case 1: // time
+									sortUpDown_T(stateNames, sortSpecs, &StateNames::ftime);
+									break;
+								default:
+									UNREACHABLE;
+								}
+							}
+							for (const auto& [name_, ftime] : stateNames) {
+								const auto& name = name_; // clang workaround
+								if (ImGui::TableNextColumn()) {
+									if (ImGui::Selectable(name.c_str())) {
+										manager.executeDelayed(makeTclList("loadstate", name));
+									}
+									if (ImGui::IsItemHovered(ImGuiHoveredFlags_DelayShort) &&
+									    (previewImage.name != name)) {
+										// record name, but (so far) without image
+										// this prevents that on a missing image, we don't continue retrying
+										previewImage.name = std::string(name);
+										previewImage.texture = gl::Texture(gl::Null{});
+
+										std::string filename = FileOperations::join(
+											FileOperations::getUserOpenMSXDir(),
+											STATE_DIR, tmpStrCat(name, ".png"));
+										if (FileOperations::exists(filename)) {
+											try {
+												gl::ivec2 dummy;
+												previewImage.texture = loadTexture(filename, dummy);
+											} catch (...) {
+												// ignore
+											}
 										}
 									}
+									im::PopupContextItem([&]{
+										if (ImGui::MenuItem("delete")) {
+											confirmCmd = makeTclList("delete_savestate", name);
+											confirmText = strCat("Delete savestate '", name, "'?");
+											openConfirmPopup = true;
+										}
+									});
 								}
-								im::PopupContextItem([&]{
-									if (ImGui::MenuItem("delete")) {
-										confirmCmd = makeTclList("delete_savestate", name);
-										confirmText = strCat("Delete savestate '", name, "'?");
-										openConfirmPopup = true;
-									}
-								});
+								if (ImGui::TableNextColumn()) {
+									ImGui::TextUnformatted(formatFileTime(ftime));
+								}
 							}
 						});
 					}
@@ -190,7 +250,7 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
 				ImGui::TextUnformatted("No replays found"sv);
 			} else {
 				ImGui::TextUnformatted("Select replay"sv);
-				int flags = ImGuiTableFlags_RowBg |
+int flags = ImGuiTableFlags_RowBg |
 					ImGuiTableFlags_BordersV |
 					ImGuiTableFlags_BordersOuter |
 					ImGuiTableFlags_Resizable |
@@ -227,7 +287,7 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
 					}
 					for (const auto& [fullName_, displayName_, ftime] : replayNames) {
 						const auto& fullName = fullName_; // clang workaround
-						if (ImGui::TableNextColumn()) {
+if (ImGui::TableNextColumn()) {
 							const auto& displayName = displayName_; // clang workaround
 							if (ImGui::Selectable(displayName.c_str(), false, ImGuiSelectableFlags_SpanAllColumns | ImGuiSelectableFlags_AllowOverlap)) {
 								manager.executeDelayed(makeTclList("reverse", "loadreplay", fullName));
@@ -241,24 +301,7 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
 							});
 						}
 						if (ImGui::TableNextColumn()) {
-							// As we still want to support gcc-11 (and not require 13 yet), we have to use this method
-							// as workaround instead of using more modern std::chrono and std::format stuff.
-
-							// Convert file_time_type to system_clock::time_point (the standard clock since epoch)
-							auto sctp = std::chrono::time_point_cast<std::chrono::system_clock::duration>(
-									ftime - std::filesystem::file_time_type::clock::now() +
-									std::chrono::system_clock::now());
-
-							// Convert system_clock::time_point to time_t (time since Unix epoch in seconds)
-							std::time_t cftime = std::chrono::system_clock::to_time_t(sctp);
-
-							// Convert time_t to local time (broken-down time in the local time zone)
-							std::tm* local_time = std::localtime(&cftime);
-
-							// Get the local time in human-readable format
-							std::stringstream ss;
-							ss << std::put_time(local_time, "%F %T");
-							ImGui::TextUnformatted(ss.str().c_str());
+							ImGui::TextUnformatted(formatFileTime(ftime));
 						}
 
 				}});
--- a/src/imgui/ImGuiReverseBar.cc
+++ b/src/imgui/ImGuiReverseBar.cc
@@ -20,6 +20,10 @@ using namespace std::literals;
 
 namespace openmsx {
 
+
+static constexpr std::string_view STATE_EXTENSION = ".oms";
+static constexpr std::string_view STATE_DIR = "savestates";
+
 void ImGuiReverseBar::save(ImGuiTextBuffer& buf)
 {
 	savePersistent(buf, *this, persistentElements);
@@ -58,13 +62,23 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
 		}
 		ImGui::Separator();
 
-		auto existingStates = manager.execute(TclObject("list_savestates"));
-		im::Menu("Load state ...", existingStates && !existingStates->empty(), [&]{
+		im::Menu("Load state ...", [&]{
 			im::Table("table", 2, ImGuiTableFlags_BordersInnerV, [&]{
 				if (ImGui::TableNextColumn()) {
 					ImGui::TextUnformatted("Select save state"sv);
 					im::ListBox("##list", ImVec2(ImGui::GetFontSize() * 20.0f, 240.0f), [&]{
-						for (const auto& name : *existingStates) {
+						std::vector<std::string> names;
+						for (auto context = userDataFileContext(STATE_DIR);
+						     const auto& path : context.getPaths()) {
+							foreach_file(path, [&](const std::string&, std::string_view name) {
+								if (name.ends_with(STATE_EXTENSION)) {
+									name.remove_suffix(STATE_EXTENSION.size());
+									names.emplace_back(name);
+								}
+							});
+						}
+						ranges::sort(names, StringOp::caseless{});
+						for (const auto& name : names) {
 							if (ImGui::Selectable(name.c_str())) {
 								manager.executeDelayed(makeTclList("loadstate", name));
 							}
@@ -77,7 +91,7 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
 
 								std::string filename = FileOperations::join(
 									FileOperations::getUserOpenMSXDir(),
-									"savestates", tmpStrCat(name, ".png"));
+									STATE_DIR, tmpStrCat(name, ".png"));
 								if (FileOperations::exists(filename)) {
 									try {
 										gl::ivec2 dummy;
@@ -111,7 +125,7 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
 		saveStateOpen = im::Menu("Save state ...", [&]{
 			auto exists = [&]{
 				auto filename = FileOperations::parseCommandFileArgument(
-					saveStateName, "savestates", "", ".oms");
+					saveStateName, STATE_DIR, "", STATE_EXTENSION);
 				return FileOperations::exists(filename);
 			};
 			if (!saveStateOpen) {
@@ -120,7 +134,7 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
 					saveStateName = result->getString();
 					if (exists()) {
 						saveStateName = stem(FileOperations::getNextNumberedFileName(
-							"savestates", result->getString(), ".oms", true));
+							STATE_DIR, result->getString(), STATE_EXTENSION, true));
 					}
 				}
 			}
@@ -139,7 +153,7 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
 			}
 		});
 		if (ImGui::MenuItem("Open savestates folder...")) {
-			SDL_OpenURL(strCat("file://", FileOperations::getUserOpenMSXDir(), "/savestates").c_str());
+			SDL_OpenURL(strCat("file://", FileOperations::getUserOpenMSXDir(), '/', STATE_DIR).c_str());
 		}
 
 		ImGui::Separator();
--- a/src/imgui/ImGuiReverseBar.cc
+++ b/src/imgui/ImGuiReverseBar.cc
@@ -161,7 +161,6 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
 			} else {
 				im::Table("table", 2, ImGuiTableFlags_BordersInnerV, [&]{
 					if (ImGui::TableNextColumn()) {
-						ImGui::TextUnformatted("Select save state"sv);
 						im::Table("##select-savestate", 2, selectionTableFlags, ImVec2(ImGui::GetFontSize() * 25.0f, 240.0f), [&]{
 							setAndSortColumns(stateNamesChanged, stateNames);
 							for (const auto& [_, name_, ftime] : stateNames) {
@@ -206,12 +205,16 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
 						});
 					}
 					if (ImGui::TableNextColumn()) {
-						ImGui::TextUnformatted("Preview"sv);
-						ImVec2 size(320, 240);
+						gl::vec2 size(320, 240);
 						if (previewImage.texture.get()) {
 							ImGui::Image(previewImage.texture.getImGui(), size);
 						} else {
+							gl::vec2 pos = ImGui::GetCursorPos();
 							ImGui::Dummy(size);
+							auto text = "No preview available..."sv;
+							gl::vec2 textSize = ImGui::CalcTextSize(text);
+							ImGui::SetCursorPos(pos + 0.5f*(size - textSize));
+							ImGui::TextUnformatted(text);
 						}
 					}
 				});
@@ -275,7 +278,6 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
 			if (replayNames.empty()) {
 				ImGui::TextUnformatted("No replays found"sv);
 			} else {
-				ImGui::TextUnformatted("Select replay"sv);
 				im::Table("##select-replay", 2, selectionTableFlags, ImVec2(ImGui::GetFontSize() * 25.0f, 240.0f), [&]{
 					setAndSortColumns(replayNamesChanged, replayNames);
 					for (const auto& [fullName_, displayName_, ftime] : replayNames) {
--- a/src/imgui/ImGuiReverseBar.cc
+++ b/src/imgui/ImGuiReverseBar.cc
@@ -77,53 +77,57 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
 				}
 				ranges::sort(stateNames, StringOp::caseless{});
 			}
-			im::Table("table", 2, ImGuiTableFlags_BordersInnerV, [&]{
-				if (ImGui::TableNextColumn()) {
-					ImGui::TextUnformatted("Select save state"sv);
-					im::ListBox("##list", ImVec2(ImGui::GetFontSize() * 20.0f, 240.0f), [&]{
-						for (const auto& name : stateNames) {
-							if (ImGui::Selectable(name.c_str())) {
-								manager.executeDelayed(makeTclList("loadstate", name));
-							}
-							if (ImGui::IsItemHovered(ImGuiHoveredFlags_DelayShort) &&
-							    (previewImage.name != name)) {
-								// record name, but (so far) without image
-								// this prevents that on a missing image, we don't continue retrying
-								previewImage.name = std::string(name);
-								previewImage.texture = gl::Texture(gl::Null{});
+			if (stateNames.empty()) {
+				ImGui::TextUnformatted("No save states found"sv);
+			} else {
+				im::Table("table", 2, ImGuiTableFlags_BordersInnerV, [&]{
+					if (ImGui::TableNextColumn()) {
+						ImGui::TextUnformatted("Select save state"sv);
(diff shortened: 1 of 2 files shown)


src/imgui/ImGuiReverseBar.cc
```diff
@@ -58,13 +62,23 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
 		}
 		ImGui::Separator();
 
-		auto existingStates = manager.execute(TclObject("list_savestates"));
-		im::Menu("Load state ...", existingStates && !existingStates->empty(), [&]{
+		im::Menu("Load state ...", [&]{
```
m9710797: Before when there were no savestates, the menu-item was greyed-out. Now it'll show an empty list.
Probably that's acceptable (it's certainly easier to implement). But I wonder if this change was a conscious decision or if it was accidental.

Or maybe, if there are no savestates yet, we could inside this sub-menu replace the ListBox with a greyed-out text (via `ImGui::TextDisabled()`) "no savestates found".

src/imgui/ImGuiReverseBar.cc
```diff
@@ -58,13 +62,23 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
...
 					ImGui::TextUnformatted("Select save state"sv);
 					im::ListBox("##list", ImVec2(ImGui::GetFontSize() * 20.0f, 240.0f), [&]{
-						for (const auto& name : *existingStates) {
+						std::vector<std::string> names;
+						for (auto context = userDataFileContext(STATE_DIR);
+						     const auto& path : context.getPaths()) {
+							foreach_file(path, [&](const std::string&, std::string_view name) {
+								if (name.ends_with(STATE_EXTENSION)) {
+									name.remove_suffix(STATE_EXTENSION.size());
+									names.emplace_back(std::string(name));
```
m9710797: Detail: replace this last line with `names.emplace_back(name);`   (caveat: not tested, but I think it'll work).

Background info:
In a `vector<T>` there are two main ways to insert elements at the back:
* `push_back(const T&)`: this makes a copy of the given element.
* `emplace_back(Args&&... args): this takes (a variable amount) of constructor parameters, and constructs the new element in-place in the back of the vector. This can save a copy.

So when you call `names.emplace_back(std::string(name))'`, you first create a (temporary) `string` object, and pass that to `emplace_back`. That temporary is used to construct another `string` (via the copy constructor).

If you instead call `names.emplace_back(name)`, you pass a `string_view` object to `emplace_back`. And that object is used to construct a `string`.

src/imgui/ImGuiReverseBar.cc
```diff
@@ -139,7 +153,7 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
 			}
 		});
 		if (ImGui::MenuItem("Open savestates folder...")) {
-			SDL_OpenURL(strCat("file://", FileOperations::getUserOpenMSXDir(), "/savestates").c_str());
+			SDL_OpenURL(strCat("file://", FileOperations::getUserOpenMSXDir(), "/", STATE_DIR).c_str());
```
m9710797: Very much a detail: replace "/" with '/'  (when appending sub-strings that have length=1).

src/imgui/ImGuiReverseBar.cc
```diff
@@ -62,23 +62,26 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
 		}
 		ImGui::Separator();
 
-		im::Menu("Load state ...", [&]{
+		loadStateOpen = im::Menu("Load state ...", [&]{
+			if (!loadStateOpen) {
```
m9710797: Is this correct?  (I haven't tested myself.)

I think `im::Menu()` returns `true` when the menu is _currently_ open. You want to know when it changed from closed to open. So I think your current code will never execute.

Maybe this could be a solution:
```
bool isOpen = im::Menu(..., [&]{
    if (!wasOpen) {
        wasOpen = isOpen;
        ...
    });
```
Another way to see that there's something fishy going on is:
* In this function you unconditionally overwrite `loadStateOpen` before using it.
* And this function is the only user of `loadStateOpen`.
* So in the current structure it wasn't needed to "remember" it (= make it a member variable).
* But we do need to compare state from the previous frame with this frame, so we do need some "memory".

src/imgui/ImGuiReverseBar.cc
```diff
@@ -62,23 +62,26 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
...
+				// on each re-open of this menu, we recreate the list of states
+				stateNames.clear();
+				for (auto context = userDataFileContext(STATE_DIR);
+				     const auto& path : context.getPaths()) {
+					foreach_file(path, [&](const std::string&, std::string_view name) {
+						if (name.ends_with(STATE_EXTENSION)) {
+							name.remove_suffix(STATE_EXTENSION.size());
+							stateNames.emplace_back(std::string(name));
```
m9710797: Same comment as above about `emplace_back()`.
... Ah, but that's because this commit moves that code.

src/imgui/ImGuiReverseBar.cc
```diff
@@ -161,27 +164,24 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
...
-			im::ListBox("##select-replay", [&]{
-				struct Names {
-					Names(std::string f, std::string d) // workaround, needed for clang, not gcc or msvc
-						: fullName(std::move(f)), displayName(std::move(d)) {} // fixed in clang-16
-					std::string fullName;
-					std::string displayName;
-				};
-				std::vector<Names> names;
+		loadReplayOpen = im::Menu("Load replay ...", reverseEnabled, [&]{
+			if (!loadReplayOpen) {
```
m9710797: Same comment about `isOpen` vs `wasOpen`.

m9710797 (commented):
Going in the right direction. Some suggestions for technical improvements. And I think one bug (same bug in 2 locations).

src/imgui/ImGuiReverseBar.cc
MBilderbeek: Yes, I realize that this was changed here. But I actually couldn't think of a way to find out that there are 0 savestates without leaving the Tcl query or file operations to run every frame. But now that I think of it, I guess I could just get the data one menu level higher, when that opens. I'll try that and adjust the code to make this a little nicer.

src/imgui/ImGuiReverseBar.cc
MBilderbeek: I realize that, but I didn't try that, because the original code seemed to do the same thing: https://github.com/openMSX/openMSX/blob/bb18df8c864469563db24f95ea55a0baeee0db8f/src/imgui/ImGuiReverseBar.cc#L165 
Is this then a different situation?

src/imgui/ImGuiReverseBar.cc
MBilderbeek: Um, well, it did seem to work and in fact, I thought I was using the same construction as already present in the code, see https://github.com/openMSX/openMSX/blob/bb18df8c864469563db24f95ea55a0baeee0db8f/src/imgui/ImGuiReverseBar.cc#L192C9-L192C23 in combination with https://github.com/openMSX/openMSX/blob/bb18df8c864469563db24f95ea55a0baeee0db8f/src/imgui/ImGuiReverseBar.cc#L186

src/imgui/ImGuiReverseBar.cc
m9710797: I have a slight preference for showing "no savestates found" inside the sub-menu. That's more descriptive (e.g. if "Load state ..." is grayed out you may still wonder why it's not available, do I need to enable something first?).
And it also keeps the code simpler (more localized).

src/imgui/ImGuiReverseBar.cc
m9710797: Yes, it's a different situation: this calls the constructor of `Names`. That one needs two `string` parameters. (And `string_view` does not _implicitly_ convert to `string`, only _explicitly_).

Anyway, this was only a detail. Initialization in c++ can be tricky (if you care about optimal code).

src/imgui/ImGuiReverseBar.cc
m9710797: Aha. I understand now. I was wrong.

The assignment to 'loadStateOpen' only happens _after_ the lambda has been executed. So inside that lambda you still see the "old" value of this variable.

src/imgui/ImGuiReverseBar.cc
MBilderbeek: OK, I implemented it in the 3rd commit. Please resolve if you're satisfied :)

src/imgui/ImGuiReverseBar.cc
MBilderbeek: Should be resolved now, please resolve the conversation if you're satisfied.

src/imgui/ImGuiReverseBar.cc
MBilderbeek: Should be resolved now, please resolve the conversation if you're satisfied.

src/imgui/ImGuiReverseBar.cc
MBilderbeek: So, please resolve the conversation if you're happy how things are (after all).

src/imgui/ImGuiReverseBar.cc
MBilderbeek: OK, please resolve the conversation if you're satisfied.

src/imgui/ImGuiReverseBar.cc
MBilderbeek: Please resolve the conversation if you're satisfied.

src/imgui/ImGuiReverseBar.cc
m9710797: In the 1st commit you indeed resolve this, but then in your 2nd commit you accidentally revert it.

MBilderbeek: Oops, that happened during a rebase that had conflicts and somehow I got it partially reverted :( Should be fixed now.

src/imgui/ImGuiReverseBar.cc
```diff
@@ -190,22 +190,52 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
...
-						const auto& displayName = displayName_; // clang workaround
-						if (ImGui::Selectable(displayName.c_str())) {
-							manager.executeDelayed(makeTclList("reverse", "loadreplay", fullName));
-						}
-						im::PopupContextItem([&]{
+						if (ImGui::TableNextColumn()) {
+							const auto& displayName = displayName_; // clang workaround
+							if (ImGui::Selectable(displayName.c_str(), false, ImGuiSelectableFlags_SpanAllColumns | ImGuiSelectableFlags_AllowOverlap)) {
+								manager.executeDelayed(makeTclList("reverse", "loadreplay", fullName));
+							}						im::PopupContextItem([&]{
```
m9710797: weird layout

src/imgui/ImGuiReverseBar.cc
```diff
@@ -190,22 +190,52 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
...
 								confirmText = strCat("Delete replay '", displayName, "'?");
 								openConfirmPopup = true;
 							}
 						});
-					}
-				});
+
+						}
+						if (ImGui::TableNextColumn()) {
+							const auto& ftime = ftime_; // clang workaround
```
m9710797: Is this workaround needed? (Did it fail without this on clang?)
The "problem" occurs when an element of a structured-binding (the `const auto& [a, b, c]` stuff) is later used as a capture in a lambda function. I don't believe that is the case here.

Is it too soon to bump the required clang version to clang-16? Then we no longer need these workarounds.

src/imgui/ImGuiReverseBar.cc
```diff
@@ -190,22 +190,52 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
...
+				int flags = ImGuiTableFlags_RowBg |
+					ImGuiTableFlags_BordersV |
+					ImGuiTableFlags_BordersOuter |
+					ImGuiTableFlags_Resizable |
+					ImGuiTableFlags_Sortable |
+					ImGuiTableFlags_Hideable |
+					ImGuiTableFlags_Reorderable |
+					ImGuiTableFlags_ContextMenuInBody |
+					ImGuiTableFlags_SizingStretchProp;
+				im::Table("##select-replay", 2, flags, {}, [&]{
```
m9710797: I understand this is still WIP.
Next steps are these?
* Add a table-header.
* Implement the sort functionality.

src/imgui/ImGuiReverseBar.cc
```diff
@@ -190,22 +190,52 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
...
+							// Convert file_time_type to system_clock::time_point (the standard clock since epoch)
+							auto sctp = std::chrono::time_point_cast<std::chrono::system_clock::duration>(
+									ftime - std::filesystem::file_time_type::clock::now() +
+									std::chrono::system_clock::now());
+
+							// Convert system_clock::time_point to time_t (time since Unix epoch in seconds)
+							std::time_t cftime = std::chrono::system_clock::to_time_t(sctp);
+
+							// Convert time_t to local time (broken-down time in the local time zone)
+							std::tm* local_time = std::localtime(&cftime);
```
m9710797: Maybe add some comment here that we're using a workaround, because (IIRC) `clock_cast`, `zoned_time`, and `std::format` are only supported starting from gcc-13.

src/imgui/ImGuiReverseBar.cc
```diff
@@ -190,22 +190,52 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
...
+
+							// Get the local time in human-readable format
+							std::stringstream ss;
+							ss << std::put_time(local_time, "%F %T");
+							ImGui::TextUnformatted(ss.str().c_str());
```
m9710797: I haven't tried your patch yet, so I haven't seen how it actually looks. But I think it might be a good idea to show only one of the date/time (depending on whether date==today). Many other programs (e.g. file managers) do this as well. Reason is to keep the table more compact (it's a sub menu after all, and those shouldn't be too big).

m9710797 (commented):
I understand this is WIP. But here are some comments anyway.

src/imgui/ImGuiReverseBar.cc
MBilderbeek: Oops, will fix that.

src/imgui/ImGuiReverseBar.cc
MBilderbeek: I have no idea what the status is on macOS now. So I can't tell. @grauw which clang version is safe to use on macOS, can you shed some light on this?

src/imgui/ImGuiReverseBar.cc
MBilderbeek: Please see the commit message for some ideas on what's next :)

src/imgui/ImGuiReverseBar.cc
MBilderbeek: Will do.

src/imgui/ImGuiReverseBar.cc
MBilderbeek: The biggest problem is that the whole list is way too long (as I put in the commit message). Well, two issues:
1. the first time it seems to appear *behind* the openMSX window: 
image
2. and the 2nd time on the place you expect. But yeah, way too long:
image

The first problem seems to be some more generic Z-level issue, more GitHub issues have mentioned it. No idea what to do there.
The fact that it is so long, what do you recommend to do about it? The size of the original listbox would be fine.

As for the date/time, I think it looks fine, even without refining this. What do you think?

src/imgui/ImGuiReverseBar.cc
MBilderbeek: OK, I couldn't resist and already added the headers and sorting functionality. But the code looks a bit clunky. Hints most welcome to improve it.

src/imgui/ImGuiReverseBar.cc
m9710797: For the ListBox we passed the size: `ImVec2(ImGui::GetFontSize() * 20.0f, 240.0f)`.
We could do the same for the table. See `src/imgui/ImGuiCpp.hh` lines 454-474. There are various overloads of `im::Table()`, the size is the 4th (optional) parameter.

src/imgui/ImGuiReverseBar.cc
```diff
@@ -184,28 +184,89 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
...
+								ranges::sort(replayNames, std::less<>{}, &ReplayNames::displayName);
 							}
-						});
+							break;
+						case 1: // time
+							if (sortSpecs->Specs->SortDirection == ImGuiSortDirection_Descending) {
+								ranges::sort(replayNames, std::greater<>{}, &ReplayNames::ftime);
+							} else {
+								ranges::sort(replayNames, std::less<>{}, &ReplayNames::ftime);
+							}
```
m9710797: Using our helper functions `sortUpDown_String()` and `sortUpDown_T()` may make it a bit less clunky.

src/imgui/ImGuiReverseBar.cc
MBilderbeek: OK, I tried that, but it did not have any effect at all. What am I doing wrong? I simply passed this exact same size as the 4th parameter, but I still get the super long table... Yes, I'm sure I compiled it etc.

MBilderbeek: Ah, the width does have effect. But the height does nothing, it seems.

src/imgui/ImGuiReverseBar.cc
MBilderbeek: Thanks, that works well. I noticed the sorting was only working the first time, so I fixed that. Each time we recreate the list, it must also be sorted again, despite the dirty-flag saying it doesn't need to.

MBilderbeek: I now noticed what I did to fix this isn't working. But I don't understand at all why.... can you help?

src/imgui/ImGuiReverseBar.cc
MBilderbeek: OK, setting ScrollY flag helps. It behaves a bit weird when you get to the end of the list though... Try to scroll in the last item... I can't really do it somehow.

src/imgui/ImGuiReverseBar.cc
grauw: The Clang version on macOS is Apple Clang version 15.0.0 (clang-1500.3.9.4) as shipped in Xcode 15.4.

Note that Apple Clang version numbers do not map to LLVM Clang, this is not the same as LLVM Clang 15. So if you’re asking which LLVM Clang version is safe to use, I can’t answer that question. If you don’t have access to macOS, just let the CI build it and see if it complains. That’s how I deal with MSVC and GCC build issues as well.

src/imgui/ImGuiReverseBar.cc
m9710797: Could it be you're testing with a non-assert build? Because when I try your PR, I immediately trigger an assert (with a semi-helpful message explaining what's wrong).
You need to move the line containing `TableSetupScrollFreeze()` before the lines with `TableSetupColumn()` (see also various examples in our code, or in `imgui_demo.cc`).

src/imgui/ImGuiReverseBar.cc
m9710797: I did some debugging. In the end I found that:
* I add a debug-print right before the call to `im::Table()` (around line 200), and one inside the lambda we pass to that function.
* Most of the time both prints trigger. _Except_ on the first frame that draws the sub-menu.
* But only on that first frame you have `loadReplayOpen == false`.

Speculation from my part: could it be that internally on the first frame ImGui only calculates the available size for the table, and then only on the next frames draws the content of the table??

So I'm afraid you'll have to add a workaround. Maybe add a `bool alreadySorted` that you reset to `false` when you re-populate `replayNames`?

m9710797: I think the current approach of re-sorting is fine. But just as a background remark (I needed this to convince myself it's fine):

In general when sorting columns in a table, you want to be able to first sort on column X, followed by sorting on column Y. More specifically, items that are equal in column Y should still remain sorted on column X. (That's why the `SortUpDown_xxx()` utility functions use _stable_-sort).
In principle when you re-populate the data, you have to keep this history in mind. Though in this specific case, because there won't be duplicates in the columns, it doesn't matter.

At least there won't be any duplicate names. In principle there could be duplicate times, but that should happen often, or at all.

src/imgui/ImGuiReverseBar.cc
MBilderbeek: Thanks, solved and from now on using devel builds :)

src/imgui/ImGuiReverseBar.cc
MBilderbeek: OK, I didn't think about it in this detail, but I did think it was good enough. Now it's also very clear why.

src/imgui/ImGuiReverseBar.cc
MBilderbeek: Thanks. We should probably do some follow up commits to see whether the workaround is still needed on the latest macOS images.

src/imgui/ImGuiReverseBar.cc
```diff
@@ -241,24 +301,7 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
...
-							std::time_t cftime = std::chrono::system_clock::to_time_t(sctp);
-
-							// Convert time_t to local time (broken-down time in the local time zone)
-							std::tm* local_time = std::localtime(&cftime);
-
-							// Get the local time in human-readable format
-							std::stringstream ss;
-							ss << std::put_time(local_time, "%F %T");
-							ImGui::TextUnformatted(ss.str().c_str());
+							ImGui::TextUnformatted(fileTimeToString(ftime).c_str());
```
m9710797: Detail: this comment already applied to the version before this patch, but I only notice it now.
It's more efficient (and shorter) to leave out the `.c_str()` part.

There are various overloads of `ImGui::TextUnformatted()`.
* The version with a single `const char*` parameter internally calls `strlen()` to calculate the end of the string.
* There's also a Dear ImGui version where you pass 2 `const char*`, the beginning and the end of the string, this avoids the `strlen()` calculation.
* We added two overloads that take a `std::string` or `std::string_view` parameter. These delegate to the version with 2 pointers. Thus also avoiding the `strlen()` calculation.

This comment applies here, but also above.

src/imgui/ImGuiReverseBar.cc
```diff
@@ -62,10 +62,32 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
...
+			std::time_t cftime = std::chrono::system_clock::to_time_t(sctp);
+
+			// Convert time_t to local time (broken-down time in the local time zone)
+			std::tm* local_time = std::localtime(&cftime);
+
+			// Get the local time in human-readable format
+			std::stringstream ss;
+			ss << std::put_time(local_time, "%F %T");
+			return ss.str();
+		};
```
m9710797, line 90: Good idea to factor this out.

This is also a good place to implement the idea of "show exactly one of date or time, depending on whether date==today". It'll help keep the sub menu smaller.  (For sure `std::chrono` offers a way to extra the date part from a time-point, including from `system_clock::now()`. But you'll have to google a bit.)

Nitpick: I don't know if this is a universal convention. But I prefer the name `xxxToString` when there's a (semi-)unique string-representation for `xxx`. And the name `formatXxx` when this function chooses one of many possible formattings.

src/imgui/ImGuiReverseBar.cc
```diff
@@ -76,46 +98,84 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
 						}
 					});
 				}
-				ranges::sort(stateNames, std::greater<>{}, &StateNames::ftime);
 			}
 			if (stateNames.empty()) {
 				ImGui::TextUnformatted("No save states found"sv);
 			} else {
 				im::Table("table", 2, ImGuiTableFlags_BordersInnerV, [&]{
```
m9710797, line 166: Another idea was to add a 3rd column with small previews of the screenshots. And only show the large (original sized) screenshot when hovering over the small one. This will also save a lot of space in the sub menu.

m9710797 (commented):
Good work.

About reduce duplication:
The first idea that came to mind was unify the types `StateNames` and `ReplayNames` (it doesn't matter that not all fields are used all the time). Then you can start factoring out common code. E.g. a function to scan a directory, or a function to (re-)sort, ...

I tried, but I cannot reproduce your Z-order issue. The sub-menu indeed doesn't fit in the main window. But for me it's displayed as expected. Also if I place the openMSX window fully to the right, the sub menu is instead displayed on the left of the parent menu (also partly outside the main window).

But for sure we should further investigate this Z-issue issue. Let's discuss later on IRC.

However I also think the current sub-menu is _excessively_ wide. Below I give some ideas for how to reduce this size.

src/imgui/ImGuiReverseBar.cc
MBilderbeek: Yeah, but I'm not sure I like it. IMHO it doesn't add much to the current solution. Can you really see something in such a small preview anyway?

src/imgui/ImGuiReverseBar.cc
MBilderbeek: I'm not sure it's a good idea to do this actually. If you have saved many states/replays on one day, but that was yesterday, you can't even see anymore the times at which you saved them...

src/imgui/ImGuiReverseBar.cc
MBilderbeek: OK, will change it in the last patch (if you don't mind I will not put it in the patch before).

src/imgui/ImGuiReverseBar.cc
m9710797: I didn't repeat this part, but the idea was that if you hover over the date, it still shows the full date + time.

src/imgui/ImGuiReverseBar.cc
m9710797: What it adds is that it saves _a lot_ of horizontal space. E.g. currently the menu is not usable in full-screen, scale_factor=2, it's wider than 640 pixels, it will never fit on screen. (We have other windows that are as wide or wider, but AFAIK only debug windows, none that you'd want to use in full-screen mode).

You don't loose much: currently you only see one screenshot at a time, and you must first click to see it. In this proposal you see a small preview of many savestates, and, if you want, also one big one (by hovering instead of clicking).

I didn't clearly say this, but the small preview I had in mind could be something line 32x20 pixels (that should already be reasonably recognizable) So taller than one normal text line. But that's a tradeoff, it means fewer rows are visible at the same time. (Very similar like in a file-browser with previews of your image (e.g. jpg) files).

src/imgui/ImGuiReverseBar.cc
```diff
@@ -94,7 +132,7 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
 						if (name.ends_with(STATE_EXTENSION)) {
 							name.remove_suffix(STATE_EXTENSION.size());
 							std::filesystem::file_time_type ftime = std::filesystem::last_write_time(fullName);
-							stateNames.emplace_back(std::string(name), ftime);
+							stateNames.emplace_back("", std::string(name), ftime); // we don't care about full name
```
m9710797: There's no harm in filling in `fullName` anyway. That would allow to factor out a function to scan all files in a directory (you pass `STATE_DIR` and `STATE_EXTENSION` as parameters).

And having the fullname available is actually useful:
* Then searching for the corresponding `.png` file becomes simpler (just replace `.oms` with `.png`).
* I haven't checked, but maybe the `loadstate` command accepts full names (that save a lookup during execution of that command). By itself maybe not a useful tweak, but it again makes the savestate and replay code more similar.

src/imgui/ImGuiReverseBar.hh
```diff
@@ -31,24 +31,18 @@ private:
...
 		std::string fullName;
 		std::string displayName;
 		std::filesystem::file_time_type ftime;
 	};
-	std::vector<ReplayNames> replayNames;
+	std::vector<StateOrReplayInfo> stateNames;
+	bool stateNamesChanged = false;
+	bool saveReplayOpen = false;
+	bool loadReplayOpen = false;
+	std::vector<StateOrReplayInfo> replayNames;
```
m9710797: Just an idea, it could be a bad idea (only way to find out is to try it):

You could group "names", "open" and "changed" into another struct. Something like:
```
struct Info {
    struct Entry {
        std::string fullName;
        std::string displayName;
        file_time_type ftime;
    };
    std::vector<Entry> entries;
    bool subMenuOpen = false;
    bool needSort = false;
};
Info replayInfo;
Info saveStateInfo;
```
I think that would allow to factor out even more code.

src/imgui/ImGuiReverseBar.cc
MBilderbeek: When going to full-screen, the native resolution of the screen is used. So the scale factor doesn't matter. The menu is usable fine in full-screen on my machine. But if you have a really low resolution screen it probably won't fit indeed.
We can also make the column with the name smaller and horizontally scrollable. Would that be helpful?

Also, currently you don't have to click to see the screenshot. It is already show in the preview window if you only hover on the name in the list/table. If you click, you already load the savestate.

src/imgui/ImGuiReverseBar.cc
MBilderbeek: Alright, so I tried to make something like this:
```
		auto formatFileTime = [](std::filesystem::file_time_type fileTime) {
			// As we still want to support gcc-11 (and not require 13 yet), we have to use this method
			// as workaround instead of using more modern std::chrono and std::format stuff.

			// Convert file_time_type to system_clock::time_point (the standard clock since epoch)
			auto sctp = std::chrono::time_point_cast<std::chrono::system_clock::duration>(
					fileTime - std::filesystem::file_time_type::clock::now() +
					std::chrono::system_clock::now());

			// Convert system_clock::time_point to time_t (time since Unix epoch in seconds)
			std::time_t cftime = std::chrono::system_clock::to_time_t(sctp);

			// Convert time_t to local time (broken-down time in the local time zone)
			std::tm* local_time = std::localtime(&cftime);

			std::time_t t_now = std::time(0);   // get time now
			std::tm* now = std::localtime(&t_now);

			const std::string format = ((now->tm_mday == local_time->tm_mday) &&
						    (now->tm_mon  == local_time->tm_mon ) &&
						    (now->tm_year == local_time->tm_year)) ? "%T" : "%F";

			// Get the local time in human-readable format
			std::stringstream ss1, ss2;
			ss1 << std::put_time(local_time, format.c_str());
			ss2 << std::put_time(local_time, "%F %T");
			return std::pair{ss1.str(), ss2.str()};
		};
```
But for some reason totally unclear to me, I only get the current date/time as results.... Even though I'm putting `local_time`'s value in the `stringstream`s.  How can querying "now" affect the results of `local_time`?

m9710797: `localtime()` is a C-function (even if it's made available in the c++ `std` namespace). It returns a _pointer_ to some `tm` struct. Calling `localtime()` a second time may return the same pointer (and thus overwrite the data in the same struct).

Possible solutions are:
* Replace `localtime()` with `localtime_r()`, in the latter you pass a pointer to your own `struct tm` that will be filled in. Though it's unclear how "portable" this function is (it works on Linux, but I'm not sure about Windows).
* Make a copy of the returned struct: `std::tm tm1 = *std::localtime(...);`

src/imgui/ImGuiReverseBar.cc
m9710797: An alternative is to not have a 3rd preview column, but instead show the full-sized screenshot as a popup, when hovering over the name.
Though this has the disadvantage that the popup will block (most of) the rest of the table. Maybe this can be mitigated by using the `ImGuiHoveredFlags_NoSharedDelay` flag, see the documentation in `src/3rdparty/imgui/imgui.h`, lines 1379-1386.

src/imgui/ImGuiReverseBar.cc
m9710797: In this prototype returning both formats is fine. In the final version I think it might be better to split this into two functions.
Reasoning: most of the time you only need one of these two: on most frames the tooltip isn't visible, And then it's wasteful to eagerly calculate both.
And yes, in case you do need both then you redo part of the calculation, but that's a minority of the cases.

src/imgui/ImGuiReverseBar.cc
MBilderbeek: We discussed on IRC that we can leave it as it is now, until we find an application where the 640 pixel width is important.

src/imgui/ImGuiReverseBar.cc
MBilderbeek: Done, thanks for all the suggestions and hints.

MBilderbeek:
@m9710797 I was thinking: we put in the 'currenttime' in the replay, so that you could continue where you left off when saving the replay. So, I tried to implement jumping to that time, but I ran into two issues:
1. I have no idea how to make the interpreter execute this:  `reverse goto [dict get [reverse status] current]` so with the `[]` stuff. I tried nested lists and all tokens as separate items. I got Syntax error
2. But worse; this doesn't even work on the console, because the current time becomes 0 immediately. So the saved 'current time' is lost :(

m9710797:
> @m9710797 I was thinking: we put in the 'currenttime' in the replay, so that you could continue where you left off when saving the replay ...

Maybe we better discuss this in a separate item (it's off-topic for this PR). But short answer:
```
help reverse
reverse loadreplay -goto savetime myreplay.omr 
```
I didn't remember we had this feature. I also had to look it up.

src/imgui/ImGuiReverseBar.cc
```diff
@@ -206,12 +205,11 @@ void ImGuiReverseBar::showMenu(MSXMotherBoard* motherBoard)
 						});
 					}
 					if (ImGui::TableNextColumn()) {
-						ImGui::TextUnformatted("Preview"sv);
 						ImVec2 size(320, 240);
 						if (previewImage.texture.get()) {
 							ImGui::Image(previewImage.texture.getImGui(), size);
 						} else {
-							ImGui::Dummy(size);
+							ImGui::TextUnformatted("No preview available..."sv);
```
m9710797: The point of this `ImGui::Dummy()` call is to reserve space (as big as the image would be). The effect is that when you move the mouse up/down, so hover over different rows, some with and some without preview, the size of the submenu remains the same. And that might be important when it's close to the window or even screen border, to avoid that the submenu changes position.

So maybe we should keep this? One way to do that is with something like this:
```
auto pos = ImGui::GetCursorPos();
ImGui::Dummy(size);
ImGui::SetCursorPos(pos);
ImGui::TextUnformatted("...");
```
In other words: get and restore the imgui cursor position to draw "overlapping" elements (but one of the elements only takes space but is itself invisible).

Optionally, in the call to `SetCursorPos()` you could also add something to the y-coordinate to make the text appear vertically centered.

MBilderbeek: Totally agreed.... See update where I address it and center it nicely in the box. Perhaps the centering could be put in ImGuiUtils or something. Not sure how to do that though, as you need to capture the cursor pos before drawing the thing you want to center in.

src/imgui/ImGuiReverseBar.cc
grauw: If you do a PR or test in a fork (or a branch when you add it to on: push: branches: in build_snapshots.yml) the CI will build without having to push it to master.

MBilderbeek: @grauw we actually already have that. See below, this PR is built automatically. No need to change the yml file.

grauw: You need to change the yml file if you are working on a private branch without opening a PR.

MBilderbeek: Ah, okay. Well, why not just open a PR then?

grauw: Just outlining multiple ways to get the CI to trigger other than opening a PR, to test Apple Clang, for future reference :). Since PRs cause all kinds of notifications to people, and additional ones for any iterations on it, depending on what you’re trying to do one may not always want that before the patch is ready.

Issue #1756, opened 2024-09-25
Menu bar doesn't (reliably) appear on full screen in Windows, in contrast to Linux


MBilderbeek:
When using full screen on Linux I get this behavior: go to full screen (e.g. Alt-Enter) and wait until the main menu bar disappears (or click in the main window outside the menu bar). Then see how you can let the menu appear again: on my Linux box it just appears again if I move the (hidden) host PC mouse cursor to the top of the screen. So I just move upwards with my mouse. 

But on Windows, this isn't working. According to @f-fix there is no way to move the (invisible) cursor, you can't get the menu bar to reappear. Note that he doesn't have separate mouse buttons on his laptop, and didn't have an external mouse.
He also said that after trying the touch screen of his laptop, he was /sometimes/ able to make the menu appear by touching near the top. But it didn't work reliably.

Another windows user said:

> By connecting an external mouse I discovered what the problem is. I not only have to move to the top edge in fullscreen mode, but also click to activate the menu bar - did I miss that instruction, or is it not supposed to be like that? And with the mousepad I now see that it is not enough to first slide up and then click, I have to click while moving. So I can't do that by tapping but by clicking the upper left mouse button at the same time as moving. I would never have thought of that without an external mouse ... By the way, even with that external mouse I have to click while I keep moving, stopping and then clicking doesn't work. A bit kinky though 😉

This behavior was more or less confirmed by @f-fix as well. It didn't work reliably for him though.

@m9710797 is this different between Windows and Linux on our side, or should we ask ImGUI folks to help?

m9710797:
This might be fixed in revision b3f3d00d3becaa743a648154e0cc19f69ca7ea0f. Can you check?

MBilderbeek:
Filip wrote (the "other user"):
> If you mean this one nothing has changed for me regarding the fullscreen menubar: https://archive.org/download/openMSX_Windows_Installer/openmsx-20.0-13-gb3f3d00d3-windows-vc-x64-bin-msi.zip

balyuf:
Some further behavior I noticed while using the Linux version of openMSX in WSL.  There, sometimes there is an in between behavior where the menubar does not fade when moving the mouse down (even though the fade option is on) and only fades when clicking below it.  I don't know reliably how to get into this mode.

[...]

Issue #892, opened 2015-06-24
[Bug] Mouse emulation lacks arithmetic saturation [sf#498]


openMSX-import:
**Reported by sd-snatcher on 2013-08-31 17:03 UTC**
This bug is related to the #477.

Interesting enough, just like what happened on Trackball.cc, the mouse emulation also fails arithmetic saturation tests. This means that the same overflow bug happens with the mouse, albeit they're much rarer because the offsets are 8-bit.

But it seems that there was already an attempt to implement the arithmetic saturation on Mouse.cc, but the code is bugged and end up imposing limits to the actual X/Y coordinates instead of the offsets.

The fixed code for the lines 240-243 of Mouse.cc seems to be:

int newX = curxrel - mev.getX();
int newY = curyrel - mev.getY();
int deltaX = max(127, min(-128, newX - curxrel));
int deltaY = max(127, min(-128, newY - curyrel));

openMSX-import:
**Updated by m9710797 on 2013-09-01 19:00 UTC**
- **status**: open --> closed-fixed

openMSX-import:
**Commented by m9710797 on 2013-09-01 19:00 UTC**
You're right that the original clipping code in the mouse emulation didn't work (it actually clipped on -256..254 instead of -128..127). But I just tested a real MSX mouse (Philips SBC3810) and it doesn't do any clipping at all (so by accident the buggy emulation code was already (closer to) the real behavior).

In revision ebeb2989 I've completely removed the mouse-offset clipping code.

openMSX-import:
**Commented by sd-snatcher on 2013-09-02 02:16 UTC**

The Philips mouse may not have clipping, but I have a MK-Mouse II here and it clearly has.

The clipping seems to be implementation dependent. Some mice have, others don't. This happens because the MSX mice are implemented by software on a small microcontroller, so each mouse has a unique algorithm.

If you want to, I may record a video of my MK-Mouse II, similar to the one Werner recorded for the trackball.

openMSX-import:
**Commented by m9710797 on 2013-09-02 08:10 UTC**
You can make a video, but I don't think it's really needed (the trackball video _was_ very useful). This is how I tested:

  10 A=PAD(12)
  20 X=PAD(13)
  30 Y=PAD(14)
  40 PRINT X,Y
  50 FORI=0to250;NEXT
  60 GOTO 10

[...]

PR #1791, opened 2024-10-21
Use different SDL grab method

balyuf:
Do not actually grab cursor when in fullscreen mode as this both messes up the mouse handling and makes the menubar hard to activate in Windows.

This is a fix for https://github.com/openMSX/openMSX/issues/1756 ... remark that setting the grabInput option is still messing up things in Windows, but at least automatic fullscreen grab is no longer done (was that ever useful, as we also hide the cursor in parallel?) ...

The SDL incompatibility here lies in fact with this function:
```
void InputEventGenerator::setGrabInput(bool grab) const
{
        SDL_SetRelativeMouseMode(grab ? SDL_TRUE : SDL_FALSE);

        // TODO is this still the correct place in SDL2
        // TODO get the SDL_window
        //SDL_Window* window = ...;
        //SDL_SetWindowGrab(window, grab ? SDL_TRUE : SDL_FALSE);
}

```
The SDL_SetRelativeMouseMode(SDL_TRUE) messes up the mouse updates (reversed, wildly jumping or frozen after initial jump) and also makes detection of hovering over the menubar fail.  I have no idea why this is only messed-up in Windows (and Linux VMs on Windows) and not in (native) Linux.  Anyway, I think for now we have a good middle way.  This needs to be tested on native Linux as I do not have that option.

### DIFF
--- a/src/events/InputEventGenerator.cc
+++ b/src/events/InputEventGenerator.cc
@@ -312,6 +312,33 @@ void InputEventGenerator::handle(const SDL_Event& evt)
 		break;
 	case SDL_MOUSEMOTION:
 		event = MouseMotionEvent(evt);
+		{
+			auto *window = SDL_GL_GetCurrentWindow();
+			if (SDL_GetWindowGrab(window)) {
+				int w, h;
+				SDL_GetWindowSize(window, &w, &h);
+				int x, y;
+				SDL_GetMouseState(&x, &y);
+				int xn = x, yn = y;
+				// There seems to be a bug in Windows in which the mouse can be locked on the right edge
+				// only when moving fast to the right (not so fast actually) when grabbing is enabled
+				// which stops generating mouse events
+				// When moving to the left, events resume, and then moving even slower to the right fixes it
+				// This only occurs when grabbing is explicitly enabled in windowed mode,
+				// not in fullscreen mode (though not sure what happens with multiple monitors)
+				// To reduce the impact of this bug, long range warping (e.g. to the middle of the window)
+				// was attempted but that caused race conditions with fading in of gui elements
+				// So, in the end it was decided that to go for the least kind of trouble
+				// The value of 10 below is a heurstic value which seems to balance all factors
+				// such as font size and the overall size of gui elements
+				// and the speed of crossing virtual barriers
+				if (x <= 10)     xn = 10;
+				if (x >= w - 11) xn = w - 11;
+				if (y <= 10)     yn = 10;
+				if (y >= h - 11) yn = h - 11;
+				if (xn != x || yn != y) SDL_WarpMouseInWindow(window, xn, yn);
+			}
+		}
 		break;
 
 	case SDL_JOYBUTTONUP:
@@ -443,12 +470,7 @@ bool InputEventGenerator::signalEvent(const Event& event)
 
 void InputEventGenerator::setGrabInput(bool grab) const
 {
-	SDL_SetRelativeMouseMode(grab ? SDL_TRUE : SDL_FALSE);
-
-	// TODO is this still the correct place in SDL2
-	// TODO get the SDL_window
-	//SDL_Window* window = ...;
-	//SDL_SetWindowGrab(window, grab ? SDL_TRUE : SDL_FALSE);
+	SDL_SetWindowGrab(SDL_GL_GetCurrentWindow(), grab ? SDL_TRUE : SDL_FALSE);
 }
 
 


MBilderbeek:
With this patch, fullscreen and mouse still work fine on Linux, at least on my system.

balyuf:
Now that the Windows VC build is available I tried it out and it indeed solves the menubar problem, however when moving down with the mouse the OS mouse pointer appears again.  Not so bad, but in case two monitors are connected, moving to e.g. the left allows the mouse to escape openMSX and not moving the MSX mouse anymore.  So this explains why we need grabbing next to making the cursor invisible.  I now feel it is not the way to go, and hence we need to better understand why the grabbing is so dysfunctional in Windows in combination with the menubar.  I will investigate further when I have some time, hopefully tonight ...  Edit: even the OS mouse pointer appearing at the bottom is problematic as the MSX mouse and OS mouse are typically out of sync, so one cannot always reach the lower parts of the MSX screen.  So, it is really important that the mouse moves relative to the MSX screen and the OS mouse is invisibly rebounding from the borders of the screen, which is accomplished by grabbing ...

balyuf:
Update: it seems Linux SDL is the one which implements SDL_SetRelativeMouseMode in a buggy way.  The expected behavior is that the absolute mouse remains centered in the window.  So the hoverMenuBar test should never trigger.  As is indeed the case in Windows.  However in Linux there seems to be a bug which allows absolute mouse movements even in relative mode.  That's why the hoverMenuBar triggers.  Well, that's my theory as I cannot test it.  I verified in Windows that indeed the absolute mouse coordinates do not (or barely) change.  However, it seems that before the addition of hoverMenubar the other two tests were sort of sufficient for Linux. @m9710797 can you throw some light on this issue?

m9710797:
@balyuf Thanks a lot for investigating this! But I'm not sure how I can help. At this point you seem to know more than me about the (grab)mouse behavior on linux/widows (and how it's supposed to work according to the SDL documentation).

Maybe we should take a step back. How do we want the mouse to behave (in full-screen mode). I mean first from a user point of view (not at an implementation level). Then we should ask how to achieve that in the implementation. And maybe (unfortunately) we'll have to take bugs in the SDL implementation into account (but we should double-check that, maybe by re-reading the SDL documentation, the SDL implementation, or asking the SDL developers).

balyuf:
This different approach to grabbing the mouse looks promising ...  I would like to try out the Windows VC version from the workflow and if someone can try out the native Linux version ...

MBilderbeek:
I checked out that branch and built it and I can still use the menu bar in full screen on Linux. Just as before :)

MBilderbeek:
Any results on Windows so far?

balyuf:
Well, when moving the mouse too fast, the borders are not reachable.  It's better than the previous attempt, but not good enough.  I have some ideas to improve the situation this time, but today was too busy, tomorrow should give more time to look into this ...

balyuf:
I've been debugging the mouse movements today, but nothing conclusive yet, hopefully I can continue tomorrow in the evening ...

balyuf:
Ok, I went deep into the mouse handling and I discovered that the remaining problem is the famous mouse vs trackball detection.  There is this second partial scanning of the mouse which only goes through the XHIGH reading, skipping the XLOW, YHIGH and YLOW reading.  This second scanning is apparently to detect if there is a trackball rather than a mouse connected.  The method seems to involve checking the deviation in the second read from the first read.  If there is again a large difference it decides that it is a trackball (I think) and the delta is clipped to (-8, +7), in effect reducing some +-127 to that range.  (see also ticket https://github.com/openMSX/openMSX/issues/892)  I am in the process of creating a fix for that, which already works in BASIC with a PAD reading little program (it does not lose anything no matter how fast I move the mouse) but unexpectedly it does not work yet with Shanghai, so will need to investigate a bit more, but I think I am on a good track (there maybe multiple similar problems involved) ...

m9710797:
Sorry, I'm confused now.

I thought this issue was about grabbing the host mouse cursor. But now in your last post you talk about the difference between MSX mouse / trackball??

What am I missing?

balyuf:
Oh, ok, I may have omitted some steps here, sorry ...

So, the alternate grabbing method does not enable relative mouse mode.  This is needed to give the absolute mouse a chance to reach the menubar coordinates so that it can show itself.  As discussed previously the original method which enabled relative mouse mode seems buggy on Linux where it apparently does allow absolute mouse movements, in effect making menubar activation possible.  Now, the issue with the mouse grabbing which does not enable relative mouse movements, is that when it reaches a border it will not continue to generate mouse movements.  In Windows, the original method warps the mouse to the center of the window to solve this issue.

So, actually this need not be a problem if all mouse movements are correctly received by MSX software (i.e. not losing tracking).  Because in that case, once you have the OS cursor in a good place relative to the MSX cursor, normally you will never hit virtual barriers.  Since I did hit those virtual barriers, I figured something else must be wrong (not the fault of the MSX software).  And indeed I discovered that we lose events when they are too large due to this trackball detection.  This is a general problem with openMSX mouse handling (the trackball handling takes great care to avoid the opposite problem through smearing out the delta's, but no such care was taken in the mouse handling, for which I am now looking for an easy fix).  Of course, when no grabbing is in effect, this problem is hardly noticed and easy to correct.  Not so with grabbing: when you lose deltas, you can get stuck with virtual barriers ...

balyuf:
Update: I found that the mouse also behaves perfectly in Video Graphics, and that the trouble in Shanghai seems to be due to a bug (feature?) in its program.  I noticed that I also hit a virtual barrier in Shanghai in relative mode grabbing, but since we continue to receive mouse events at window borders (in MS Windows through mouse warping), next (slow) mouse movement easily crosses the barrier.  So, the bug is exacerbated in absolute mode grabbing which causes impenetrable virtual barriers.  *sigh* I am not sure if it can be solved satisfactorily in this way.  I am starting to think that we need to leave the accessibility of the menubar in Windows an unsolved issue ... but maybe I'll come up with another tentative solution ...

balyuf:
I think I have an acceptable solution for Windows at least.  @MBilderbeek can you try again on Linux to check if everything remains workable?

MBilderbeek:
It seems to work exactly the same as before, so as I would expect. I tried using the mouse in full screen mode. The menu bar can be easily activated by moving the (invisible) cursor to the top.

balyuf:
Ok, and in a program which uses the mouse, the mouse still moves smoothly as before?

MBilderbeek:
Yes, I tried with Shanghai and the Philips NMS 8220 built in drawing program.  Works as normal.

balyuf:
Great, then I guess @m9710797 can have a look at this request to see if it can be pulled as is?

MBilderbeek:
You might want to clean up the commits to make them a set of logical 'first time right' commits, using interactive rebase. At least it would make the commits on themselves be easier to understand. Don't forget that they'll end up in the commit history "for ever".

balyuf:
Oh, really?  With an MR in Gitlab, only one summary commit is created ... I was not aware of that difference ...

m9710797:
Thanks. I'll try to review one of the following days.

Could you also document a bit what you tried (maybe what didn't work), and why you've chosen for the current solution. That info might be important for the future (e.g. so that when we migrate to SDL3 we won't re-introduce the problem).

When merging a PR there are 2 possibilities (I assume both in gitlab and in github). Squash the PR in a single commit, or merge the full history. Typically we choose the latter.

balyuf:
Ok, should I document it as code comments or here in the PR?

m9710797:
Either is fine. In the commit message works as well.

Personally (in general) what I try to do:
* In the code I (mostly) only comment about the the current code itself.
* In the commit message that makes the change I talk about why that change was made. E.g. why the old code was wrong or changed. And maybe why it was changed like that, and not to some other alternative.

But the form itself isn't that important. You did a lot of investigation into this topic. And it would be nice to not fully loose/forget this knowledge in the future. So any form is better than nothing.

balyuf:
Ok, I squashed the commits on the branch, put some comments in the code, and some context in the PR.  After all not that much to say, just took a long time (on and off) to discover what was going on.

m9710797 (commented):
The change in `InputEventGenerator.cc` is fine. But I'm not sure about the change in `Mouse.{cc,hh}`. Or at least I don't believe that's the right location for this code.

I believe the intention is to move the host mouse cursor away from the host window border, right? That has nothing to do with MSX-mouse emulation, so it does not belong in the `Mouse` class.
* For example when the MSX mouse isn't plugged into the emulated joystick port the code in `Mouse` doesn't trigger. 
* It's possible some other emulated MSX device is plugged into the joystick port that does react the host mouse movements. E.g. search for `MouseMotionEvent` in the code. These device react to that: `ArkanoidPad`, `Mouse`, `Paddle`, `TouchPad`, `Trackball`. I think all of these could benefit from your fixes, not just `Mouse`.

So maybe your fixes should be moved to the place where we handle the host mouse event? Maybe somewhere around `InputEventGenerator.cc` line 314. This is where we translate the SDL mouse-motion event into our internal openMSX event (that then gets handled by the various MSX devices).

balyuf:
I guess you mean the mouse warping code only?  The xLow tracking is very specific to MSX mouse handling.

I tried to put it in that place, but it was somehow choking openMSX (I don't understand why).  When the mouse is not plugged, the warping is not necessary, as the mouse is then only useful for ImGui anyway.  But your point about multiple MSX devices benefiting from the warping, I find a good point, yes.

balyuf:
Ok, I moved the code again (as it is now) and it works fine, was probably an intermediate version of code which was causing interference or something.  Anyway, good that it works now.

balyuf:
BTW, I had to put the code in a block, because the compiler was complaining about the window variable and jumping across labels ...

balyuf:
PS: I cannot view the SonarCloud Code Analysis ... it demands logging in, though I briefly seem to get the analysis ... so what does it complain about?

src/events/InputEventGenerator.cc
```diff
@@ -312,6 +312,20 @@ void InputEventGenerator::handle(const SDL_Event& evt)
 		break;
 	case SDL_MOUSEMOTION:
 		event = MouseMotionEvent(evt);
+		{
+			auto *window = SDL_GL_GetCurrentWindow();
+			if (SDL_GetWindowGrab(window)) {
+				int x, y, xn, yn, w, h;
+				SDL_GetWindowSize(window, &w, &h);
+				SDL_GetMouseState(&x, &y);
```
m9710797, line 321: Nitpick: we try to avoid uninitialized variables. But here we cannot avoid it because of the `SDL_xxx()` functions. Nevertheless it would be cleaner to keep the time between declaration and initialization as small as possible. Concretely in this case that would mean:
* Declare only `w` and `h` before `SDL_GetWindowSize()`.
* Declare `x` and `y` right before `SDL_GetMouseState()`.
* No need to forward declare `xn` and `yn`, better write this as `int xn = x;` and `int yn = y;`.

src/events/InputEventGenerator.cc
```diff
@@ -312,6 +312,20 @@ void InputEventGenerator::handle(const SDL_Event& evt)
...
+				xn = x; yn = y;
+				if (x <= 10)     xn = 10;
+				if (x >= w - 2)  xn = w/2;
+				if (y <= 10)     yn = 10;
+				if (y >= h - 2)  yn = h/2;
+				if (xn != x || yn != y) SDL_WarpMouseInWindow(window, xn, yn);
```
m9710797, line 339: I don't understand this part. Or at least I don't understand the asymmetry:
* You set `xn` to `max(x, 10)`. Ok, I assume you want to keep the cursor 10 pixels away from the left border.
* But then I'd also expect the symmetrical check for 10 pixels away from the right border.
* Instead you check for `x >= w - 2`. Why `2` instead of `10`?
* And even more strange: why then assign `w / 2` instead of `w - 2`? (Warp to the middle instead of 2 pixels from the side).

Maybe this could be simplified to:
```
int xn = std::clamp(x, 10, w - 10);
int yn = std::clamp(y, 10, h - 10);
```
Or more fancy using vector-operations (not tested):
```
gl::ivec2 windowSize;
SDL_GetWindowSize(window, &windowSize.x, &windowSize.y);
gl::ivec2 pos;
SDL_GetMouseState(&pos.x, &pos.y);
gl::ivec2 margin{10, 10};
gl::ivec2 warp = clamp(pos, margin, windowSize - margin);
if (warp != pos) {
    SDL_WarpMouseInWindow(window, warp.x, warp.y);
}
```

src/input/Mouse.cc
```diff
@@ -239,10 +241,14 @@ void Mouse::write(uint8_t value, EmuTime::param time)
...
-				curXRel -= xRel;
-				curYRel -= yRel;
+				// Only introduce the next delta when we had a skipped xLow reading
+				// This to avoid trackball detection in the alternate cycle
+				if (xLow) {
+					xRel = std::clamp(curXRel, -127, 127);
+					yRel = std::clamp(curYRel, -127, 127);
+					curXRel -= xRel;
+					curYRel -= yRel;
+				} else xRel = yRel = 0;
```
m9710797: Sorry I don't understand this change. Maybe it is correct, but then it needs more comments.

This is my current understanding:
* In the normal case (when there's no timeout) the variable `phase` cycles through these 4 values `XHIGH`, `XLOW`, `YHIGH` and `YLOW` (in this order).
* When the `XHIGH` state is entered, we latch the variables `xRel` and `yRel`, these are the values that will be returned on reading (in the `read()` method).
* But when there is no timeout, the variable `xLow` can only be `true` in the phase `XHIGH` or `XLOW`. And in that case in this code block (when changing from `YLOW` to `XHIGH` we will always execute the `else` branch, and thus always set `xRel = yRel = 0`.
* This means that if a MSX program constantly reads the mouse (I mean more than once per frame, so not waiting for the timeout) it will effectively see no mouse movement. It's not impossible, but I strongly doubt that's the case with real MSX hardware.

Also the real MSX mouse hardware does not know when the MSX software actually reads from it (it constantly puts a high or low signal on the joystick port input pins (input-direction = from mouse to MSX), but it has no idea when the AY8910 chip actually 'uses' those signals). 

That means the real MSX mouse hardware cannot change behavior on whether or not the data was read. Maybe an alternative could be to do something (on line 215) when the timeout triggers. Real hardware can measure the time between *changes* in the output pins (output-direction = from MSX to mouse) (it cannot detect all writes, only writes that actually make a change).

So to summarize. probably this PR does make a good fix. But for sure it does it in a different way than how a real MSX mouse behaves. And I do believe this difference may break other MSX programs (e.g. programs that read in a tight loop).

@sdsnatcher73 Can you comment on this? Do you happen to know (in detail) how the MSX mouse protocol works, and how it interacts with trackball detection?

src/events/InputEventGenerator.cc
balyuf: Fair enough, will improve the code in that way ...

src/events/InputEventGenerator.cc
balyuf: Yes, this code has grown out of experiment.  Ideally, I'd like to always test edges (x,y <= 1 && x,y >= w,h - 2) and warp to the middle of the window.  However when doing that with y, there is often interference with the menubar appearing: sometimes it starts appearing and yet the mouse warps away, annoyingly needing another attempt.  For x that would not be necessary, but there I wanted to create at least a bit of symmetry.  The truth is however that I had first done the other way around (x,y <= 10 && x,y >= w,h - 11) but for some reason now horizontally the mouse could hang on the last pixel (and even become visible there).  I tried to increase the margin, but to no avail, and being tired of trying to finger this yet another strange anomaly, I decided to go for the middle of the window (and have symmetry).  So, in short: y needs small top edge, x needs large right edge.

src/input/Mouse.cc
balyuf: Ok, here is what I have seen for mouse scanning in every case (I assume through BIOS):
write xHigh, read xHigh
write xLow, read xLow
write yHigh, read yHigh
write yLow, read yLow
write xHigh, read xHigh
write xLow
write yHigh
write yLow
[next cycle]

So my change will make sure that we only set the next delta (potentially large, which will never happen in HW, only with a virtual mouse) upon entering the xHigh state at the beginning of the cycle, and not in the middle.  In the middle the check happens for trackball.  So if the middle xHigh read is high, clipping to within trackball range happens in the next cycle.  Is that more clear now?

I never observed timeout BTW.  I guess it could happen, and I confess not having thought about what the effect will be, so I could think about that.  But overall, I do think this the right way to do this.  Remember that we also clamp between -127 and +127, not because the HW does it, but because our virtual mouse situation demands it.  This would be a similar fix ...

balyuf: PS: and so we should think about the case that the MSX program bypasses the BIOS and attempts its own mouse scanning routine?  So if the program does not do this second half-cycle, indeed it would only receive 0 deltas.  I wonder if we could detect this case?

src/events/InputEventGenerator.cc
m9710797: I'm sorry but I'm still confused.
* What I meant is that left vs right (or top vs bottom) symmetry is broken in the code. But you say there's a good reason for that? It would be very useful to document such things in the code. Because such knowledge is currently impossible to retrieve.
* And not only are the values different (`10` for left margin, `2` for right margin), but also the behavior itself: left is clamped to `10`, right is warped to the middle. That's also very unexpected behavior and (if intentional) should be documented.
* The current code is symmetrical for X vs Y. But from what I understood from your explanation the real behavior for X and Y is NOT symmetrical. If so, then maybe we shouldn't artificially make it symmetrical in the code?

May I ask to treat the 4 cases (left, right, top, bottom) separately. Start with a common simple implementation for all 4, but then individually introduce exceptions. And for each exception clearly document why it's needed.

Or of course if there is a common implementation that works for all 4 (e.g. unconditionally always warp to the middle?) that would be even better.

balyuf: Ok, let me try to explain better ...

Naively, the first implementation would be to detect that the mouse is at the penultimate pixel at any edge and when the mouse goes forward, warp it just one pixel backwards.  This works, but makes for a very slow compensation when continuing the movement in the forward direction.

So, a bigger margin is needed.  It turns out 10 pixels is a good heuristic.  Now the movement to compensate for virtual barriers can be processed fast enough.  Now to translate this into numbers, be careful with width and height.  The first pixel is of course x=0 or y=0, but the last pixel is x=w-1 or y=h-1.  So this already introduces a bit of asymmetry.  So for the small margin this becomes x=1, y=1 and x=w-2 and y=h-2.  For the big margin this becomes x=10, y=10 and x=w-11, y=h-11.

This would be good.  Warping it further towards the middle of the window will probably improve the situation, yet only marginally, and would introduce the issue with the interrupted attempt to raise the menubar (as described previously).

Now enter this strange anomaly where at right edge, moving towards the right, the Windows mouse would suddenly become visible and block at the last pixel (because of the grabbing).  In that state, moving the mouse up or down does no longer cause mouse events.  The mouse can go back to the left, becomes invisible again and generate mouse events.  But this is a very weird mess-up again.  I thought I had miscalculated the edge (with the -1 or -2) but increasing the edge to w-11 or even w-22 or so did not solve it.  Brute forcing it to w/2 did solve it.  As I said I did not feel like figuring this one out again, or try to find the right interpolation value.  Asymmetry would be the result anyway, so w/2 is as good as any then, and has some esthetic value at least.

It is possible that the same anomaly does not happen at the bottom edge, because of the activation of ImGui which happens there (because of the leds and the timeline).  So the only question left is: do we also go for y=h/2 or do we keep it at y=h-11?  Or else we need to understand why this anomaly ... and then keep everything at 10 pixels margin ...

balyuf: PS: I wonder now if ImGui does not occupy the right edge, and hence this behavior?  I am not sure because when at the bottom ImGui is activated (e.g. over the leds or when the timeline raises) also the menubar raises, and I don't remember that this happened in the case of the right edge anomaly ...  I'd need to check again ...

src/input/Mouse.cc
m9710797: Ok thanks, this part is becoming clear to me.
First of all: I think this is a separate (orthogonal) issue than the change in `InputEventGenerator.cc`, right?
* 1st part is about the (windows) host mouse grab handling.
* 2nd part is about msx mouse vs trackball detection confusion

It may be the case that the 2nd part now triggers more frequently(*), but it was in fact an already existing problem.

(*) Or actually does part 1 influence part 2? I wouldn't expect that. Warping the host mouse cursor should not influence what the MSX mouse sees. Or do I miss something?

In any case I believe it's better to split this over two commits (maybe you had done this initially, but were then asked to cleanup your commits and merge them, sorry if that's the case).

--

I did some research on the MSX mouse protocol, and how the MSX BIOS (or other MSX software) handles the mouse. It's hard to find good documentation. This [MRC wiki article](https://www.msx.org/wiki/Mouse/Trackball) is the best I could find.
* At the hardware level, the MSX mouse is close to what we emulate in openMSX. It has indeed 4 phases, with a reset to the 1st phase after some timeout.
* MSX software that directly accesses the hardware (I mean reading the mouse without using the BIOS) indeed closely follows this protocol. So just cycling through these 4 phases, thus reading 4 nibbles.
* What's new info (at least to me) is that the MSX BIOS reads **2 times** 4 phases. The expectation is that when you read the mouse a 2nd time (very close in time after the 1st read) then the real mouse will not have had the time to move very far(*2), and then you'll read an offset equal to zero. (An MSX trackball will instead reset the counter to -8 (no movement is represented by -8 instead 0), and that allows to distinguish between a mouse or a trackball).
* In openMSX this assumption breaks because we try to "smear" very large offsets over multiple successive cycles (with cycle I mean a group of 4 phases).

(*2) this "very far" part isn't true for negative offsets of course, also the smallest negative offset already has a non-zero upper nibble.

You rightfully claim that the real mouse does not have the smearing/clamping behavior, and that's indeed the root cause of breaking the detection heuristic. We did introduce this extra behavior to solve [this issue](https://github.com/openMSX/openMSX/issues/892). So I would like to keep this.

In this patch you try to detect the 2nd cycle, and then don't apply the smearing. I think that's a good solution, I only like to change the trigger for when to apply this. (Because as we discussed above, the current code may break MSX software that doesn't use the MSX BIOS to read the mouse).

The first idea that comes to my mind is:
* Instead of having 4 phases have 8 phases. So split `XHIGH` into `XHIGH1` and `XHIGH2` etc.
* These 4 extra phases behave identical to their original counterpart. So we don't change the MSX mouse protocol.
* Except for the *extra* smearing behavior that we added (that is not part of the actual mouse protocol), that will only be applied in `YLOW2`, not in `YLOW1`. (YLOW**2** is not a typo, it must be applied in the phase that comes before `XHIGH1`).
* The timeout will also reset to phase `YLOW2`, that way MSX software that doesn't use the BIOS (and assuming there's sufficient time between two read cycles), will always have the extra smearing heuristic.

src/events/InputEventGenerator.cc
m9710797: For your experiments you could try disabling the leds and reverse bar. Of course the final solution should also work when they are enabled.

Thanks for the extra explanation. I think it's valuable to document such weird quirks/exceptions in the code. One possibility is to add a comment referring to this discussion (at least I think you can get a stable URL for these individual replies?).

src/input/Mouse.cc
balyuf: Ok, so yes, this approach with 8 phases works fine.  I have to clarify that I never saw timeouts.  What I should have said, is that I never saw timeouts which had to change phase, i.e. they always happened when already in YLOW.  Setting it to YLOW2 will again normally not change the phase for BIOS driven mouse scans, but when this is bypassed, it can now typically timeout from YLOW1 to YLOW2, which should **theoretically** work fine (is there a known piece of MSX software that wants to read the mouse itself?).

I was able to link to this discussion in a comment in the code.  I have also made the variable initializations for the other point as you requested (I hope), but was not yet successful in solving the mouse warping issue (though it looks more and more as yet another SDL bug unfortunately).  I will see later what I can do.  For now I wanted to force push the code changes already, so that you can have a look at them.

You are right that these are in fact two different issues.  The relative mode issue and the mouse smearing out issue.  And indeed, the latter was always present but showed up clearly when I first try to use the alternate way of enabling relative mode (without automatic warping of the mouse).  It gave me virtual barriers.  However, now that I warp the mouse myself, the other issue is again less pressing, but of course we want to solve it, now that we know of it.  So, I will split both in two commits later.  For now let's live with one commit ...

MBilderbeek: You could try SymbOS. I'm certain it does not use the BIOS to read out the mouse.l

balyuf: Yes, SymbOS works fine.  But since its mouse is slower (seen that with other MSX programs) a 10 pixel border is not enough.  Thinking about warping the mouse to the middle from all directions.  Need to test further the interaction of that method with the menubar ...

balyuf:
Ok, @m9710797 I came to a conclusion on this one.  I have kept one asymmetry, explained in the code for the the upper border, in orde to avoid a race condition between ImGui menu rising and the mouse warping.  I hope everything is clear now for merging in ...

src/input/Mouse.cc
```diff
@@ -128,14 +132,22 @@ uint8_t Mouse::read(EmuTime::param /*time*/)
...
+		case PHASE_YLOW1:
 			return  (yRel       & 0x0F) | status;
+		case PHASE_XHIGH2:
+		case PHASE_XLOW2:
+		case PHASE_YHIGH2:
+		case PHASE_YLOW2:
+			// Keep the delta zero in the alternate cycle of the scan round
+			// to avoid trackball detection for large (absolute) values
+			// Timeout resets the state to YLOW2 so that each scan round syncs to XHIGH1
+			return status;
```
m9710797: I agree with this 1st commit. Though I propose to simplify the implementation a bit.
I propose to implement this function as:
```
switch (phase) {
    case PHASE_XHIGH1: case PHASE_XHIGH2:
        ...
    case PHASE_XLOW1: case PHASE_XLOW2:
        ...
    case PHASE_YHIGH1: case PHASE_YHIGH2:
        ...
    case PHASE_YLOW1: case PHASE_YLOW2:
        ...
}
```
So do exactly the same for the '1' and '2' phases. Instead we'll implement the different behavior in write(). That keeps the trick (=clamp with workaround for trackball detection) more localized.

src/input/Mouse.cc
```diff
@@ -212,23 +224,39 @@ void Mouse::write(uint8_t value, EmuTime::param time)
...
+			// still gets continuous delta updates in each scan round
+			phase = PHASE_YLOW2;
 		}
 		lastTime = time;
 
 		switch (phase) {
-		case PHASE_XHIGH:
-			if ((value & STROBE) == 0) phase = PHASE_XLOW;
+		case PHASE_XHIGH1:
+			if ((value & STROBE) == 0) phase = PHASE_XLOW1;
```
m9710797: Cases `PHASE_XHIGH1`, `PHASE_XHIGH2`, `PHASE_YHIGH1` and `PHASE_YHIGH2` basically all do the same thing. At least if we rewrite it like:
```
case PHASE_XHIGH1: case PHASE_XHIGH2:
case PHASE_YHIGH1: case PHASE_YHIGH2:
    if ((value & STROBE) == 0) ++phase;
    break;
```
And similar for `PHASE_XLOW1` and `PHASE_XLOW2`. (In the original code it wasn't worth merging the phases, but now that we've doubled the number of phases it makes more sense).

But `PHASE_YLOW1` and `PHASE_YLOW2` are really different (that was the point of introducing the 2nd set of phases):
```
case PHASE_YLOW1:
    if ((value & STROBE) != 0) {
        phase = PHASE_XHIGH2;
        xRel = yRel = 0;
    }
    break;
case PHASE_YLOW2:
    if ((value & STROBE) != 0) {
        phase = PHASE_XHIGH1;
        ... same as in your patch ...
    }
    break;
```
Notice how `case PHASE_YLOW1` prepares the result that will be returned in `read()` for the 2nd set of phases. Also notice how to difference in behavior between `PHASE_YLOW1` and `PHASE_YLOW2` is now very close together in the code.

m9710797 (commented):
I have a suggestion. Functionally it's exactly the same as your code. But it's a bit more compact and the tricky part (where we intentionally deviate from how a real MSX mouse behaves) is localized in the single `write()` function (rather than spread over both `read()` and `write()`).

m9710797 (commented):
No questions about the change in `setGrabInput()`. That seems like a good improvement.

But the change in `handle()` in `case SDL_MOUSEMOTION` still seems like "magic" to me. I mean no doubt you've spend a lot of time on it to find something that works well for you.

My understanding is that when the mouse cursor moves within 10 pixels from the border (left, right, top, bottom) you warp it back to the middle (separately for horizontal and vertical). But for some reason for the top border you clamp instead of warp. The reason you give for that is `avoid race condition between ImGui rising and mouse warping`, but I don't understand that comment :(
* What do you mean with `ImGui rising`?
* Is this in any way related to the main menu bar at the top? But if so, what then if the menu bar is undocked? (Click the triangle in the top left corner).
* Or what with other GUI elements? For example the reverse-bar in the bottom

balyuf:
Ok, I am reworking the first commit as per your suggestions ...

For the second commit, indeed I should have written race condition with the menubar rising.  But remember, that I said that normally all of them could be the same with 10 pixels.  I don't see it as clamping.  It is still warping, but within a very small range (max 10 pixels).  However there is this bug where that does not work somehow on the right edge.  That's why I introduced the warping over a long range (to the middle) because it avoids (most of the time!) the bug at the right edge.  So, I really don't want to make it more complicated than it is.  In the case the menubar is undocked or whatever, it does not matter, it still works fine.  So, how do you want me to introduce the asymmetry as simple as possible?  I figured this was a simple way.  I could also do the opposite: have them all work with 10 pixels, and only warp the right edge to the middle.  But somehow that felt less nice than what I do now ... Is this more clear now?

m9710797:
This is useful extra info. It would be good to include it somewhere in the PR (i mean either as comments in the code (preferably), or in the commit message). Also fine if it's a short summary with a reference to a longer discussion here.

But now that I know that these 10 pixels are somehow related to the size(?) of the top menu bar, it triggers a follow up question: Does this value (10) remain valid when you change the font size? E.g. try selecting a huge or a tiny font size (main menu > Settings > GUI > Select font ...).

balyuf:
Ok, force pushed once more.  Both commits have changed.

For the second one, 10 is a heuristic value.  I checked it with the smallest font possible (9 where the default is 13) and the largest font possible.  It still works.  The trick is simply to keep the mouse within the rising menubar so that it keeps rising.

m9710797:
Ok, thanks for testing.
`an heuristic ... to keep cursor within rising menu bar` <-- that's useful info to include.
BTW with "rising" you mean "fading in", right?

Did you already have time to test with an undocked menu bar?
Did you already test other GUI elements (like the reverse bar)?

balyuf:
Yes, I mean "fading in" indeed.  I tried to detach the menubar and move around the reverse bar to the top.  Everything remains compatible with the stock openMSX, and no adverse effects on the fix I am trying to introduce.  But I fail to see how it even could?  I am just restricting the warping of the mouse on the top side.  Whether there is some gui element or not, I fail to see what difference it could make?

m9710797:
What I meant was: at the top you needed to make an exception because there was a GUI element there (the main menu bar). So I was wondering why you don't need similar exceptions for the other GUI elements, like the reverse bar which is near the bottom (by default, but it can moved anywhere, near any border).

balyuf:
Ah, I see, yes, it is a potential issue ...

In fact I now realize the issue is more general than I thought.  The issue is that in fullscreen mode we have to move the mouse blindly so we tend to overreact to make sure it reaches an edge.  In the case of the reversebar, it overshoots, briefly letting it start to fade in, only to fade out again, as the mouse is warped to the middle.  Shoot, if it were not for the that bug with the right edge, I could do all edges with the short range warping.  Oh well, it is what it is, but I will pass for tonight then.  Not sure when I will have time (and enough courage) to look at it (also with the festivities coming up).  Should I completely split the two commits over two PRs so that you could already pull in the fix for the trackball confusion (first commit here)?

m9710797:
Yes, if you split off the first commit, then we can already merge that part.

balyuf:
Ok, parked the second commit here, moved the first commit to https://github.com/openMSX/openMSX/pull/1816

balyuf:
Ok, I think I have come to a final conclusion on this one too.  There is now a long explanation in the comments.  So, @m9710797 if you agree with this one as it is, can we also pull it in?

m9710797:
I've merged this PR (and then applied some very minor tweaks).
@balyuf thanks a lot for your work, testing and persistence.

balyuf:
For the record: I suddenly realized that on Windows openMSX was being scaled already by the automatic scaling mechanism (typically used on laptops).  When I configured the application to allow it to do the scaling, suddenly the locked mouse on the right edge phenomenon disappeared!  So this probably is a rounding error of sorts ...  And without Windows automatic scaling, the MSX display is more sharp too ...

m9710797:
Do you mean the HiDPI scaling stuff? (I don't really know how that works at a technical level).
But do you mean openMSX is using the wrong default? Would it be better if 'somehow' we could configure openMSX to (by default) don't do this scaling?

balyuf:
Yes, I mean the HiDPI scaling.  I had to go to the openMSX.exe and ask for the properties.  Then in the compatibility tab I had to override the HiDPI setting to "Application".  I think it would be better to include this "somehow" in the self-installer package, yes.  When I compile openMSX for Linux under WSL and run it there, no HiDPI scaling is applied either.  Of course this makes the menu font also smaller, so people may want to enlarge the font size.  I've put it to 16 iso the default 13.

PR #1800, opened 2024-11-13
Keep recent items of machines and be able to select from them.

### LINKED ISSUE #1697: Show "recent items" also when selecting machines
This way you can quickly switch between your favourite (most often used) machine types.

MBilderbeek:
TODO:
- assess duplication inside `ImGuiMachine` and with `ImGuiMedia`. Maybe it's not too bad?

Closes #1697

### DIFF
--- a/src/imgui/ImGuiMachine.cc
+++ b/src/imgui/ImGuiMachine.cc
@@ -156,32 +156,39 @@ void ImGuiMachine::paintSelectMachine(const MSXMotherBoard* motherBoard)
 			ImGui::Separator();
 		}
 
+		auto showMachine = [&](MachineInfo& info, bool doubleClickToSelect) {
+			bool ok = getTestResult(info).empty();
+			im::StyleColor(!ok, ImGuiCol_Text, getColor(imColor::ERROR), [&]{
+				bool selected = info.configName == newMachineConfig;
+				if (ImGui::Selectable(info.displayName.c_str(), selected,
+						doubleClickToSelect ? ImGuiSelectableFlags_AllowDoubleClick: ImGuiSelectableFlags_None)) {
+					newMachineConfig = info.configName;
+					if (ok && (doubleClickToSelect ? ImGui::IsMouseDoubleClicked(ImGuiMouseButton_Left) : true)) {
+						showSelectMachine = false; // close window
+						manager.executeDelayed(makeTclList("machine", newMachineConfig));
+						addRecentItem(recentMachines, newMachineConfig);
+					}
+				}
+				if (selected) {
+					if (ImGui::IsWindowAppearing()) ImGui::SetScrollHereY();
+				}
+				if (ImGui::IsItemHovered(ImGuiHoveredFlags_DelayNormal | ImGuiHoveredFlags_NoSharedDelay | ImGuiHoveredFlags_Stationary)) {
+					im::ItemTooltip([&]{
+						printConfigInfo(info);
+					});
+				}
+			});
+		};
+
 		im::TreeNode("Recently used", recentMachines.empty() ? ImGuiTreeNodeFlags_None : ImGuiTreeNodeFlags_DefaultOpen, [&]{
 			if (recentMachines.empty()) {
 				ImGui::TextUnformatted("(none)"sv);
 			} else {
-				MachineInfo* info = nullptr;
 				im::Combo("##recent", "Switch to recently used machine...", [&]{
 					for (const auto& item : recentMachines) {
-						info = findMachineInfo(item);
-						if (!info) continue;
-						bool ok = getTestResult(*info).empty();
-						im::StyleColor(!ok, ImGuiCol_Text, getColor(imColor::ERROR), [&]{
-							bool selected = info->configName == newMachineConfig;
-							if (ImGui::Selectable(info->displayName.c_str(), selected)) {
-								newMachineConfig = info->configName;
-								if (ok) {
-									showSelectMachine = false; // close window
-									manager.executeDelayed(makeTclList("machine", newMachineConfig));
-									addRecentItem(recentMachines, newMachineConfig);
-								}
-							}
-							if (ImGui::IsItemHovered(ImGuiHoveredFlags_DelayNormal | ImGuiHoveredFlags_NoSharedDelay | ImGuiHoveredFlags_Stationary)) {
-								im::ItemTooltip([&]{
-									printConfigInfo(*info);
-								});
-							}
-						});
+						if (auto* info = findMachineInfo(item)) {
+							showMachine(*info, false);
+						}
 					}
 				});
 				simpleToolTip("Replace the current with the selected machine.");
@@ -222,26 +229,7 @@ void ImGuiMachine::paintSelectMachine(const MSXMotherBoard* motherBoard)
 			im::ListClipper(filteredMachines.size(), selectedIdx, [&](int i) {
 				auto idx = filteredMachines[i];
 				auto& info = allMachines[idx];
-				bool ok = getTestResult(info).empty();
-				im::StyleColor(!ok, ImGuiCol_Text, getColor(imColor::ERROR), [&]{
-					bool selected = info.configName == newMachineConfig;
-					if (ImGui::Selectable(info.displayName.c_str(), selected, ImGuiSelectableFlags_AllowDoubleClick)) {
-						newMachineConfig = info.configName;
-						if (ok && ImGui::IsMouseDoubleClicked(ImGuiMouseButton_Left)) {
-							showSelectMachine = false; // close window
-							manager.executeDelayed(makeTclList("machine", newMachineConfig));
-							addRecentItem(recentMachines, newMachineConfig);
-						}
-					}
-					if (selected) {
-						if (ImGui::IsWindowAppearing()) ImGui::SetScrollHereY();
-					}
-					if (ImGui::IsItemHovered(ImGuiHoveredFlags_DelayNormal | ImGuiHoveredFlags_NoSharedDelay | ImGuiHoveredFlags_Stationary)) {
-						im::ItemTooltip([&]{
-							printConfigInfo(info);
-						});
-					}
-				});
+				showMachine(info, true);
 			});
 		});
 
--- a/src/imgui/ImGuiMachine.cc
+++ b/src/imgui/ImGuiMachine.cc
@@ -31,6 +31,20 @@ using namespace std::literals;
 
 namespace openmsx {
 
+void ImGuiMachine::save(ImGuiTextBuffer& buf)
+{
+	for (const auto& item : recentMachines) {
+		buf.appendf("machine.recent=%s\n", item.c_str());
+	}
+}
+
+void ImGuiMachine::loadLine(std::string_view name, zstring_view value)
+{
+	if (name == "machine.recent") {
+		recentMachines.push_back(value);
+	}
+}
+
 void ImGuiMachine::showMenu(MSXMotherBoard* motherBoard)
 {
 	im::Menu("Machine", [&]{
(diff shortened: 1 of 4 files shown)


m9710797:
About "duplicate item check isn't working". I debugged it: your code is fine, but there was a bug in `circular_buffer` itself. I've already pushed a fix (see that commit for more details).

MBilderbeek:
Thanks! Please note that the code is copied from `ImGuiMedia.cc`, so I would expect it to trigger there as well.

MBilderbeek:
@m9710797 please review. And if you have some ideas/advice on how to reduce the duplication as mentioned in the description, please let me know.

src/imgui/ImGuiMachine.cc
```diff
@@ -31,6 +31,36 @@ using namespace std::literals;
 
 namespace openmsx {
 
+void ImGuiMachine::save(ImGuiTextBuffer& buf)
+{
+	savePersistent(buf, *this, persistentElements);
```
m9710797: `persistentElements` is an empty tuple. In that case `savePersistent()` does nothing. (And the corresponding `loadOnePersistent()` is also missing). Actually this `{save,load}Persistent()` is just a helper function to replace "manual" `buf.appendf(...)` calls, so if the tuple is empty this indeed does nothing.

So I propose to just remove this line (and the corresponding line in the `.hh` file) until we need it.

src/imgui/ImGuiMachine.cc
```diff
@@ -31,6 +31,36 @@ using namespace std::literals;
...
+
+	for (const auto& item : recentMachines) {
+		if (!item.empty()) {
```
m9710797: Is this check needed? Is it possible that `item` is empty?

src/imgui/ImGuiMachine.cc
```diff
@@ -31,6 +31,36 @@ using namespace std::literals;
...
+{
+	if (auto it = ranges::find(recentItems, item); it != recentItems.end()) {
+		// was already present, move to front
+		std::rotate(recentItems.begin(), it, it + 1);
+	} else {
+		// new entry, add it, but possibly remove oldest entry
+		if (recentItems.full()) recentItems.pop_back();
+		recentItems.push_front(item);
+	}
+}
```
m9710797, line 39: This function seems like a good candidate to share. Maybe generalize it to
```
template<typename T>
void addRecentItem(circular_buffer<T>& buffer, const T& item)
{
...
}
```
And then call it from here and from ImGuiMedia.

Maybe do this is a separate commit?

src/imgui/ImGuiMachine.cc
```diff
@@ -142,6 +172,33 @@ void ImGuiMachine::paintSelectMachine(const MSXMotherBoard* motherBoard)
 			ImGui::Separator();
 		}
 
+		if (!recentMachines.empty()) {
+			ImGui::TextUnformatted("Recently used machines:"sv);
+			auto* info = findMachineInfo(recentMachines.front());
+			im::Combo("##recent", info->displayName.c_str(), [&]{
+				for (const auto& item : recentMachines) {
+					info = findMachineInfo(item);
+					bool ok = info && getTestResult(*info).empty();
```
m9710797: Here you test for `info != nullptr`, but above on line 177 you do not. I think we should either check both or check none.

This list is potentially populated with information from `imgui.ini`. It might be the case that between the time this was saved and when it's reloaded some machines have been deleted. So I think we should indeed check.

Ah, and when `info == nullptr` various things below can go wrong (line 183, 184, 185, 194).
* Line 183: `info->configName`, fallback to `item`?
* Line 184: `info->displayName`, also fallback to `item`?
* Line 184: in case of `!ok` we could pass `ImGuiSelectableFlags_Disabled` to `ImGui::Selectable()` (3rd parameter). Then lines 185-190 should never execute (and the test on `ok` inside is no longer needed). Just an idea I did not test this. We should also verify that disabled selectables can still be hovered (otherwise the user cannot see the error message)`
* Line 194: `*info`, maybe we should pass a pointer instead of a reference to `printConfigInfo()`, and then in case of a `nullptr` print a generic error message like `machine config doesn't exist`.

--

Or alternatively when loading `recentMachines` we could then check whether these machines actually exist, and if not then drop the item? This makes the situation the same as in the `Available machines` list box: then you can do `assert(info)`. (In theory it's still possible to delete a machine config file while openMSX is running. Such errors are only detected when you actually try to activate that machine).

src/imgui/ImGuiMachine.cc
```diff
@@ -142,6 +172,33 @@ void ImGuiMachine::paintSelectMachine(const MSXMotherBoard* motherBoard)
...
+								manager.executeDelayed(makeTclList("machine", newMachineConfig));
+								addRecentItem(recentMachines, newMachineConfig);
+							}
+						}
+						if (ImGui::IsItemHovered(ImGuiHoveredFlags_DelayNormal | ImGuiHoveredFlags_NoSharedDelay | ImGuiHoveredFlags_Stationary)) {
+							im::ItemTooltip([&]{
+								printConfigInfo(*info);
+							});
+						}
+					});
```
m9710797, line 178: Assuming you go for the alternative that `info` is guaranteed `!= nullptr`, then this block of code (starting after line 180, where you obtain `info`) is very similar to the block below in the `Available machines` list box. The main difference is double-click vs single-click, but that could become a parameter on the factored out code.

I propose to do this in a separate commit.

m9710797 (commented):
PR looks good.
* I have some suggestions for simplification.
* A crash on missing machine configs (config removed since last session). 
* And two suggestions for reducing code duplication (for follow-up commits).

I only briefly tested your PR yesterday. I like the functionality. But IMHO it didn't look very nice yet (i understand this is still a prototype). Though I don't really know how to improve it. Here are two ideas to try (mutually exclusive):
* Add a `ImGui::Separator()` between `Recently used machines` and `Available machines`.
* Use a `im::TreeNode("Recently used Machines`, ImGuiTreeNodeFlags_DefaultOpen)` (similar to the filter). Then it can be hidden for users who don't need it. And the combo box is also indented (maybe that looks better?).

src/imgui/ImGuiMachine.cc
MBilderbeek: Ah yeah. I wasn't sure what it was for, but I forgot to remove it when I kind of discovered I didn't need it.

src/imgui/ImGuiMachine.cc
MBilderbeek: I took it over from the loop in `ImGuiMedia.cc`, but indeed: I don't understand how it can be empty there, let alone here.... how can it be empty there?

src/imgui/ImGuiMachine.cc
MBilderbeek: Should be fixed in the latest version.

src/imgui/ImGuiMachine.cc
MBilderbeek: Should be fixed in the latest version.

src/imgui/ImGuiMachine.cc
MBilderbeek: Should be fixed in the latest version.

After our discussion, I decided upon skipping items that are not known machines (no MachineInfo present) in the list that is displayed in the GUI.

MBilderbeek:
> I only briefly tested your PR yesterday. I like the functionality. But IMHO it didn't look very nice yet (i understand this is still a prototype). Though I don't really know how to improve it. Here are two ideas to try (mutually exclusive):
> 
>     * Add a `ImGui::Separator()` between `Recently used machines` and `Available machines`.
> 
>     * Use a `im::TreeNode("Recently used Machines`, ImGuiTreeNodeFlags_DefaultOpen)` (similar to the filter). Then it can be hidden for users who don't need it. And the combo box is also indented (maybe that looks better?).

I actually didn't consider them mutually exclusive and applied both. Please take a look what you think. It's easy to change it again, and I am completely fine to do that :)

src/imgui/ImGuiMachine.cc
MBilderbeek: Done. I made a commit before the actual commit to isolate this and then use it in the actual commit.

src/imgui/ImGuiMachine.cc
MBilderbeek: OK, let's hope that I can more easily solve conflicts in case they appear when processing further feedback.

Should be fixed in latest version.

src/imgui/ImGuiMachine.cc
```diff
@@ -31,6 +31,20 @@ using namespace std::literals;
 
 namespace openmsx {
 
+void ImGuiMachine::save(ImGuiTextBuffer& buf)
+{
+	for (const auto& item : recentMachines) {
+		buf.appendf("machine.name=%s\n", item.c_str());
```
m9710797: Detail: `name` is very generic. Maybe it's more descriptive to rename this to `recent`? (and then also in `loadLine()` of course).

src/imgui/ImGuiMachine.cc
```diff
@@ -142,6 +156,46 @@ void ImGuiMachine::paintSelectMachine(const MSXMotherBoard* motherBoard)
...
+					im::ItemTooltip([&]{
+						printConfigInfo(info);
+					});
+				}
+			});
+		};
+
+		im::TreeNode("Recently used", recentMachines.empty() ? ImGuiTreeNodeFlags_None : ImGuiTreeNodeFlags_DefaultOpen, [&]{
+			if (recentMachines.empty()) {
+				ImGui::TextUnformatted("(none)"sv);
```
m9710797, line 185: In an earlier version, when `recentMachines` was empty, you wouldn't show this section at all. Now you show `(none)`. But are fine I think.

Maybe I have a slight personal preference for not showing the section. Normally this only happens once (unless you delete your `imgui.ini` file). And maybe not showing a not-yet-useful section for new users is better?

src/imgui/ImGuiMachine.cc
```diff
@@ -142,6 +156,46 @@ void ImGuiMachine::paintSelectMachine(const MSXMotherBoard* motherBoard)
...
+			} else {
+				MachineInfo* info = nullptr;
```
m9710797: Probably a leftover from some refactoring. This variable could also be declared inside the for-loop.

I know that yesterday I mentioned using a `continue` to skip over invalid entries. But now that I see that the remaining code is just a single line, you could also do the following (both are fine).
```
if (auto* info = ) {
   showMachine(...);
}
```

src/imgui/ImGuiMachine.cc
```diff
@@ -142,6 +156,46 @@ void ImGuiMachine::paintSelectMachine(const MSXMotherBoard* motherBoard)
...
+				im::Combo("##recent", "Select recently used machine...", [&]{
```
m9710797: Detail: about the text "**Select** recently used machine ...". This not just selects a machine, it replaces the current machine and start the selected one. Maybe replace *Select* with *Switch to*?

Several of the buttons in this window have tool-tips explaining what will happen if you press them. Maybe we could add a similar tooltip here on the closed combo box?  (But we'd have to verify that it's possible to have tool-tips only on the combo-box itself, not on the drop-down list).

m9710797 (commented):
Looks good. In this review I focused more on details, and small usability things.
I do like the improved look.

src/imgui/ImGuiMachine.cc
MBilderbeek: Another point of view: it can show that we have a certain feature (as a teaser), which will quickly become useful. 

And it takes only one line in the window, as I do make the treenode closed if there are no recent machines.

src/imgui/ImGuiMachine.cc
MBilderbeek: OK, changed it in the latest version.

src/imgui/ImGuiMachine.cc
MBilderbeek: OK, changed it in the latest version.

src/imgui/ImGuiMachine.cc
MBilderbeek: OK, changed it in the latest version. Maybe you have a better idea on what the text should be though, because it's quite duplicate with the 'preview' of the Combo. But indeed, the tooltip works as you expect.

m9710797: `If you click on one of these recently used machines, you immediately switch to them.`
The "If you click on..." that part is redundant. And you don't say why *immediately* switching might be bad. So I propose something like this (the *replace* word is important):
`Replace the current machine with a recently used other machine.`
Or even just this (to remove the redundancy with the *preview* text).
`Replace the current machine.`
or
`Replace the current with the selected machine.`

Apart from this detail I think the PR is ready. So feel free to merge.

MBilderbeek: OK, changed it to that last suggestion in the latest version.

PR #1801, opened 2024-11-14
Minor corrections on ROM and RAM id's for consistency

jlautenbag:
I noticed some inconsistencies in the machine xml files. Manuel invited me to create a pull request for this.

### DIFF
--- a/share/machines/Toshiba_HX-20E.xml
+++ b/share/machines/Toshiba_HX-20E.xml
@@ -81,7 +81,7 @@
         </rom>
         <mem base="0x0000" size="0x8000"/>
       </ROM>
-      <RAM id="Upper RAM">
+      <RAM id="Main RAM page 2, 3">
         <mem base="0x8000" size="0x8000"/>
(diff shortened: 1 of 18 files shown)


share/machines/Toshiba_HX-20E.xml
```diff
@@ -94,7 +94,7 @@
     <primary slot="3">
```
MBilderbeek, line 94: Isn't it more logical to name the lower RAM "Main RAM 1" and the upper part "Main RAM 2"? Or is it the same in other configurations?
Perhaps we should change it the other way around, but then more standardized, like "Main RAM Lower part", "Main RAM Upper part"?
What do you think?

jlautenbag: Well, as you guessed, other configurations are like this. I had to look twice before making the change this way. :-)

"1" and "2" are very arbitrary indeed. I like the "Lower part" and "Upper part" better.

MBilderbeek: Thanks for the updates after our discussion. I think this is addressed now.

MBilderbeek:
Thanks!

PR #1809, opened 2024-12-01
WIP: added DalSoRi R2 Moonsound clone.

### LINKED ISSUE #1805: Implement DalSoRi R2
See https://www.msx.org/forum/msx-talk/openmsx/dalsori-r2-on-openmsx for discussion.

MBilderbeek:
TODO:
- flash support
- remove duplication with MoonSound (copy pasted from it, for this prototype)

Closes #1805

### DIFF
--- a/src/sound/DalSoRiR2.cc
+++ b/src/sound/DalSoRiR2.cc
@@ -1,3 +1,5 @@
+// TODO: for now copied from MSXMoonsound. Must be cleaned up.
+
 // ATM this class does several things:
 // - It connects the YMF278b chip to specific I/O ports in the MSX machine
 // - It glues the YMF262 (FM-part) and YMF278 (Wave-part) classes together in a
@@ -7,7 +9,8 @@
 //      different class, but until there's a 2nd user of this chip, this is
 //      low priority.
 
-#include "MSXMoonSound.hh"
+#include "DalSoRiR2.hh"
+#include "MSXCPUInterface.hh"
 #include "Clock.hh"
 #include "serialize.hh"
 #include "unreachable.hh"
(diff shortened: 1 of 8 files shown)


src/sound/YMF278.cc
```diff
@@ -820,7 +821,7 @@ YMF278::YMF278(const std::string& name_, int ramSizeInKb,
 {
 	if (rom.size() != 0x200000) { // 2MB
 		throw MSXException(
-			"Wrong ROM for MoonSound (YMF278). The ROM (usually "
+			"Wrong ROM for OPL4 (YMF278b). The ROM (usually "
```
grauw: Upper-case B :).

src/sound/YMF278.cc
```diff
@@ -908,8 +909,10 @@ void YMF278::reset(EmuTime::param time)
 // for the region 0x000000-0x1FFFFF. (But this routine does not handle ROM).
 unsigned YMF278::getRamAddress(unsigned addr) const
 {
-	addr -= 0x200000; // RAM starts at 0x200000
-	if (regs[2] & 2) [[unlikely]] {
+	if (romEnabled) {
+		addr -= 0x200000; // RAM starts at 0x200000
+	}
```
m9710797: My _guess_ would be that if you enable/disable the sample-ROM, this will only (un)hide the RAM that's present in the first 2MB. And the RAM present from 2MB to 4MB remains unchanged.

Though this implementation will instead shift the lower part of the RAM to be visible in the upper part of the address space (and make the upper part of the RAM inaccessible).

src/sound/YMF278.cc
```diff
@@ -908,8 +909,10 @@ void YMF278::reset(EmuTime::param time)
...
+	if (regs[2] & 2 && ram.size() != 4096 * 1024) [[unlikely]] {
```
m9710797: I'm not sure this extra condition (size !=4MB) is correct.
But good enough for this proof-of-concept. I'll take a more detailed look later.

I'm also thinking of better optimizing the `readMem()` method (which internally calls this getRamAddress() method). Because it gets called upto 4 233 600 times per second (44100Hz x 24channels x 16-bit x 2 reads for linear interpolation).

My first idea is to build an array of 32 pointers (4MB address space in chunks of 128kB).

src/sound/DalSoRiR2.cc
```diff
@@ -0,0 +1,374 @@
...
+{
+	if ((0x3000 <= addr) && (addr < 0x4000)) {
+		// SRAM frame 0
+		return sram[addr - 0x3000];
+	} else if ((0x7000 <= addr) && (addr < 0x8000) && ((regFrame[0] & DISABLE) == 0)) {
+		// SRAM frame 1
+		return sram[(addr - 0x6000) + (regFrame[0] & REG_FRAME_SEGMENT_MASK) * 0x1000];
+	} else if ((0xB000 <= addr) && (addr < 0xC000) && ((regFrame[1] & DISABLE) == 0)) {
+		// SRAM frame 2
+		return sram[(addr - 0x9000) + (regFrame[1] & REG_FRAME_SEGMENT_MASK) * 0x1000];
```
m9710797: Line 298 should be `addr - 0x7000` (instead of `- 0x6000`), and this line should be `addr - 0xB000` (instead of `- 0x9000`). Or maybe simpler: both should be `addr & 0x0FFF`.
(I do understand where the error comes from: these were the correct offset from before you used `regFrame[]`).

src/sound/DalSoRiR2.cc
```diff
@@ -0,0 +1,374 @@
...
+	} else {
+		// read selected segment of flash
+		auto flashBank = addr >> 13;
+		auto flashSegmentAddr = addr & 0x3FFF;
+		return flash.read(flashSegmentAddr + regBank[flashBank] * 0x4000);
```
m9710797: Did you forgot to apply `REG_BANK_SEGMENT_MASK` here?
Or maybe better: we should take the actual flash ROM size into account (or does AMDFlash::read() already do that internally?)

In the PCB photos is saw that it uses a SST39SF010A chip. That's a 128kB flash ROM chip.  IIRC the documentation states that the flash size can vary between 128kB and 512kB   (That would be SST39SF010A (1Mbit = 128kB), SST39SF020A (2Mbit = 256kB) or SST39SF040A (4Mbit = 512kB).  512kB indeed requires a mask of 5 bit (32 * 16kB = 512kB). But if we only emulate a 128kB flash ROM, then this address calculation can go out-of-bounds.

I think the easiest fix is to mask on the actual flash rom size. Something like   `(address-calculation) & (flash.size() - 1)`.
EDIT: I checked: AmdFlash::read() does mask internally, ignore this last suggestion.

src/sound/DalSoRiR2.cc
```diff
@@ -0,0 +1,374 @@
...
+}
+
+void DalSoRiR2::writeMem(word addr, byte value, EmuTime::param /*time*/)
+{
+	if ((0x3000 <= addr) && (addr < 0x4000)) {
+		// SRAM frame 0
+		sram[addr - 0x3000] = value;
+	} else if ((0x7000 <= addr) && (addr < 0x8000) && ((regFrame[0] & DISABLE) == 0)) {
+		// SRAM frame 1
+		sram[addr - 0x6000] = value;
```
m9710797: Also here we need to take `regFrame[]` into account.

src/sound/DalSoRiR2.cc
```diff
@@ -0,0 +1,374 @@
...
+		// SRAM frame 2
+		sram[addr - 0x9000] = value;
+	} else if ((0x6000 <= addr) && (addr < 0x6800)) {
+		if (addr >= 0x6700) {
+			setRegCfg(value);
+		} else if (addr >= 0x6500) {
+			regFrame[(addr >> 9) & 1] = value;
+		} else if (addr < 0x6400) {
+			regBank [(addr >> 8) & 3] = value;
+		}
```
m9710797: I think this is correct, but it took me a little while to verify that it's indeed correct. E.g. it isn't obvious that [0x6400, 0x6500) has no function.
Maybe it's clearer with this structure:
```
} else if ((0x6000 <= addr) && (addr < 0x6400)) {
    regBank[..] = ..;
} else if ((0x6500 <= addr) && (addr < 0x6700)) {
    regBank[..] = ..;
} else if ((0x6700 <= addr) && (addr < 0x6800)) {
 setRegCfg(..);
} ..
```

src/sound/DalSoRiR2.cc
```diff
@@ -0,0 +1,374 @@
...
+	} else if ((regCfg & ENA_FW) != 0) {
+		auto flashBank = addr >> 13;
```
m9710797: You already found this bug yourself (`>> 13` should be `14`).
I'm adding a comment anyway, so that we don't forget about it.

src/sound/DalSoRiR2.cc
MBilderbeek: I had it at first, but I realized that the value written in regBank[x] ss always already masked to 5 bits when writing. But now I see I forgot to add that masking there. Will add it.

MBilderbeek: This cannot be done for the frame register, because it has that extra bit. So, this idea I had makes it a bit asymmetric. So I guess I'll indeed just add the mask for the regBank as well.

m9710797:
Implemented via: https://github.com/openMSX/openMSX/pull/1810

PR #1810, opened 2024-12-10
DalSoRiR2, based upon PR#1809 by Manuel

m9710797:


src/sound/DalSoRiR2.cc
```diff
@@ -123,12 +123,9 @@ void DalSoRiR2::reset(EmuTime::param time)
...
 
-	for (auto reg : xrange(4)) {
-		regBank[reg] = reg;
-	}
-	for (auto reg : xrange(2)) {
-		regFrame[reg] = reg;
-	}
+	ranges::iota(regBank, 0);
+	ranges::iota(regFrame, 0);
+
```
MBilderbeek, line 107: Does this still set the initial value correctly?

src/sound/YMF278.hh
```diff
@@ -134,12 +144,17 @@ private:
...
-	// or nullptr means this region is unmapped.
+	// or 'nullBlock' means this region is unmapped.
 	// Note: we can only _read_ via these pointers, not write.
-	std::array<const uint8_t*, 32> memPtrs = {};
+	std::array<Block128, 32> memPtrs{
+		nullBlock, nullBlock, nullBlock, nullBlock, nullBlock, nullBlock, nullBlock, nullBlock,
+		nullBlock, nullBlock, nullBlock, nullBlock, nullBlock, nullBlock, nullBlock, nullBlock,
+		nullBlock, nullBlock, nullBlock, nullBlock, nullBlock, nullBlock, nullBlock, nullBlock,
+		nullBlock, nullBlock, nullBlock, nullBlock, nullBlock, nullBlock, nullBlock, nullBlock,
+	};
```
MBilderbeek: One would expect that there is a shorter way to express this...

src/sound/DalSoRiR2.cc
```diff
@@ -150,22 +150,48 @@ void DalSoRiR2::writeIO(word port, byte value, EmuTime::param time)
...
+		return *r;
 	} else if (!dipSwitchBDIS.getBoolean()) {
-		// read selected segment of flash
-		auto flashBank = addr >> 14;
-		auto flashSegmentAddr = addr & 0x3FFF;
-		return flash.read(flashSegmentAddr + (regBank[flashBank] & REG_BANK_SEGMENT_MASK) * 0x4000);
+		auto page = addr >> 14;
+		auto bank = regBank[page] & REG_BANK_SEGMENT_MASK;
+		auto offset = addr & 0x3FFF;
+		return flash.read(0x4000 * bank + offset);
```
MBilderbeek: You changed terminology here. I tried to stick to the manual's terminology... what's best? Use that, or try to be consistent across the openMSX codebase?

src/sound/DalSoRiR2.cc
```diff
@@ -150,22 +150,48 @@ void DalSoRiR2::writeIO(word port, byte value, EmuTime::param time)
...
+}
+
+byte DalSoRiR2::peekMem(word addr, EmuTime::param /*time*/) const
+{
+	if (auto r = readCommon(addr)) {
+		return *r;
+	} else if (!dipSwitchBDIS.getBoolean()) {
+		auto page = addr >> 14;
+		auto bank = regBank[page] & REG_BANK_SEGMENT_MASK;
+		auto offset = addr & 0x3FFF;
```
MBilderbeek: Maybe this can also be factored out between peek/read....

src/sound/YMF278.hh
m9710797: The technical problem here is that Block128 has no default constructor. So we cannot first declare the array and then e.g. fill it with the correct value.

I know there are some c++ tricks using `std::index_sequence` and parameter packs. But my first idea was that those would add more complexity than just these 4 lines with 32 repeats of `nullBlock`.

However, I checked and in our `stl.hh` we _do_ already have a helper function `generate_array` that implements exactly this trick. With that helper, I think, we could simplify this code snippet to (not tested):
```
auto memPtrs = generate_array<32>([](size_t){ return nullBlock; });
```

I'll experiment and change it.

src/sound/DalSoRiR2.cc
m9710797: I don't immediately see the conflict in terminology.
* The Z80 64kB address space is divided into exactly 4 pages of 16kB.
* (I'm skipping the MSX primary/secondary slot stuff).
* In each of the 4 pages it's possible to select a specific bank. For DalSoRi2 there are 32 possible banks.
* We then combine the selected bank and the offset with that bank (or page) to calculate the full address in the flash rom.

Or if the DalSoRiR2 uses 'page' and/or 'bank' with a different meaning, then personally I think we should stick to the openMSX terminology.

MBilderbeek: In the variables I used in the original patch I directly copied the terminology from that manual: 
image
and
image

But, no deal breaker for me of course.

MBilderbeek:
Heh, I don't understand what you did now to prevent the initial ROM content to load. Can you explain it?

PR #1816, opened 2024-12-23
Only change the MSX mouse cursor when in the right phase to avoid trackball confusion

balyuf:
So, this commit is splitted off from https://github.com/openMSX/openMSX/pull/1791 so that it can be pulled-in separately ...


m9710797:
Thanks!

PR #1819, opened 2025-01-07
Use 64-bit stat structure on Windows.

### LINKED ISSUE #1755: Can't open harddisk images when they are 2GB or larger on Windows
Seen on openMSX 19.1 (Windows 64-bit version)

It works - though extremely slowly thanks to hashing time - when I make a 2000-megabyte disk image file and launch openMSX like this:
```bash
(
  dd bs=$(( 1024 * 1024 )) count=2047 if=/dev/zero of=2000MB.DSK &&
  openmsx -machine Sony_HB-F1XV -ext Musical_Memory_Mapper -ext SunriseIDE_Nextor -hda 2000MB.DSK
)
```
I can see it and partition it using `_FDISK`
image

However a 2048-megabyte disk image is not found and openMSX fails to launch:
```bash
(
  dd bs=$(( 1024 * 1024 )) count=2048 if=/dev/zero of=2048MB.DSK &&
  openmsx -machine Sony_HB-F1XV -ext Musical_Memory_Mapper -ext SunriseIDE_Nextor -hda 2048MB.DSK
)
```
No window opens, and in the console it prints this error message:
`Fatal error: Error in "SunriseIDE_Nextor" extension: 2048MB.DSK not found in this context`

I would like to be able to directly use an image of the microSD card I use with my Carnivore2+ on real MSX hardware with openMSX, but due to this size restriction I cannot, since the microSD card I am using is 16GB and contains multiple partitions larger than 2GB

MBilderbeek:
For some reason, the default is 32-bit, which gives an overflow with files larger than 2GB.

Thanks to Frits Hilderink for debugging and proposing the solution.

Closes #1755.


m9710797:
I trust this indeed fixes the stat() problem.
But now I'm wondering: is this sufficient? E.g. in `src/file/LocalFile.cc` line 93 and 105 we use the normal `fread()` and `fwrite()` calls to read/write files. Can those handle more than 2GB on windows?

MBilderbeek:
> I trust this indeed fixes the stat() problem. But now I'm wondering: is this sufficient? E.g. in `src/file/LocalFile.cc` line 93 and 105 we use the normal `fread()` and `fwrite()` calls to read/write files. Can those handle more than 2GB on windows?

I am not sure, but with this change, the large disk image (up to 4GB was tested by Frits) could be opened and read from at least.

He sent me these screenshots:
image
image

m9710797:
For sure this PR is an improvement. And it's well isolated. So we can merge it.

I was only wondering if it's sufficient. E.g. can the stuff past 2GB on such a huge hard disk image actually be read/written?
So without extra checks I don't think we can close #1755.

MBilderbeek:
> For sure this PR is an improvement. And it's well isolated. So we can merge it.
> 
> I was only wondering if it's sufficient. E.g. can the stuff past 2GB on such a huge hard disk image actually be read/written? So without extra checks I don't think we can close #1755.

@fhil can you confirm the image can be read from and written to beyond the 2GB limit?

MBilderbeek:
Update: I have been chatting with @fhil in private and we do see some weird things.
E.g. when creating a 1GB HDD file with making 2 partitions of 512MB, the 2nd one doesn't become available in Nextor for some reason, but that only happens on Windows, not in Linux. In fact, if I use a savestate of that situation made in Windows on my Linux box, I get the same, until I reset the MSX. Then I can suddenly access drive B (partition 2). 

Oddly, even on Linux, the reported size of that partition seems wrong (always 32MB, although it's FAT16).

Tested like this:
1. create a 1GB file with dd from /dev/zero
2. start openMSX with the SunriseIDE_Nextor extension
3. Then `_FDISK`
4. Add 512MB partition
5. Add another 512MB partition
6. Write partition table with W-key
7. reset
8. Then in basic do this: `?DSKF(0)` - this gives 524008; `_DSKF(1)` as well of course - but `?DSKF(2)` gives 31984; I would have expected 524008 as well there.

This is on Linux. But on Windows, you can't even do a `_DSKF(2)`, as it will tell you the drive doesn't exist.

@fhil Did I put it correctly?

MBilderbeek:
As the current patch is already an improvement for sure, I propose we already merge it whilst we try to find someone with the real Nextor based Sunrise IDE to find out about possible other issues. Agreed, @m9710797 ?

m9710797:
Agreed.

Issue #1655, opened 2024-06-04
Add status bar (that can be hidden) at the bottom of the screen with useful info


MBilderbeek:
Missing there compared to Catapult is the FPS indicator.

Other stuff that can be added:
- running software
- main properties of the current machine (RAM, VRAM)
- current screen mode
- CPU usage? (As seen in some other emulators.)

It must definitely be possible to hide this bar.

MBilderbeek:
I think this feature is done for now.

PR #1822, opened 2025-01-11
Make items visibile in status bar configurable.

MBilderbeek:
This is a follow up to issue #1655.

### DIFF
--- a/src/imgui/ImGuiManager.cc
+++ b/src/imgui/ImGuiManager.cc
@@ -728,99 +728,111 @@ void ImGuiManager::drawStatusBar(MSXMotherBoard* motherBoard)
 		im::MenuBar([&]{
 			auto frameTime = ImGui::GetIO().DeltaTime;
 
-			// limit updating to at most 10Hz
-			fpsDrawTimeOut -= frameTime;
-			if (fpsDrawTimeOut < 0.0f) {
-				fpsDrawTimeOut = 0.1f;
-				fps = reactor.getDisplay().getFps();
-			}
-			std::stringstream ssFps;
-			ssFps << std::fixed << std::setprecision(1) << fps << " fps";
-			ImGui::RightAlignText(ssFps.str(), "999.9 fps");
-			simpleToolTip("refresh rate");
-			ImGui::Separator();
-
-			auto [modeStr, extendedStr_] = [&] { // TODO: remove duplication with VDP debugger code
-				if (!motherBoard) return std::pair{"-", ""};
-				const auto* vdp = dynamic_cast<const VDP*>(motherBoard->findDevice("VDP"));
-				if (!vdp) return std::pair{"-", ""};
-
-				auto mode = vdp->getDisplayMode();
-				auto base = mode.getBase();
-				if (base == DisplayMode::TEXT1)      return std::pair{"0 (40)", "TEXT 1"};
-				if (base == DisplayMode::TEXT2)      return std::pair{"0 (80)", "TEXT 2"};
-				if (base == DisplayMode::GRAPHIC1)   return std::pair{"1", "GRAPHIC 1"};
-				if (base == DisplayMode::GRAPHIC2)   return std::pair{"2", "GRAPHIC 2"};
-				if (base == DisplayMode::GRAPHIC3)   return std::pair{"4", "GRAPHIC 3"};
-				if (base == DisplayMode::MULTICOLOR) return std::pair{"3", "MULTICOLOR"};
-				if (base == DisplayMode::GRAPHIC4)   return std::pair{"5", "GRAPHIC 4"};
-				if (base == DisplayMode::GRAPHIC5)   return std::pair{"6", "GRAPHIC 5"};
-				if (base == DisplayMode::GRAPHIC6)   return std::pair{"7", "GRAPHIC 6"};
-				if (base != DisplayMode::GRAPHIC7)   return std::pair{"?", ""};
-				return (mode.getByte() & DisplayMode::YJK)
-					? (mode.getByte() & DisplayMode::YAE) ? std::pair{"11", "GRAPHIC 7 (YJK/YAE mode)"} : std::pair{"12", "GRAPHIC 7 (YJK mode)"}
-					: std::pair{"8", "GRAPHIC 7"};
-			}();
-			auto extendedStr = extendedStr_; // pre-clang-16 workaround
-			ImGui::RightAlignText(modeStr, "0 (80)");
-			simpleToolTip([&]{
-				std::string result = "screen mode as used in MSX-BASIC";
-				if (extendedStr[0]) {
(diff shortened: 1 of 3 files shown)


src/imgui/ImGuiSettings.cc
```diff
@@ -398,7 +398,16 @@ void ImGuiSettings::showMenu(MSXMotherBoard* motherBoard)
 		im::Menu("Misc", [&]{
 			ImGui::MenuItem("Configure OSD icons...", nullptr, &manager.osdIcons->showConfigureIcons);
 			ImGui::MenuItem("Fade out menu bar", nullptr, &manager.menuFade);
-			ImGui::MenuItem("Show status bar", nullptr, &manager.statusBarVisible);
+			im::Menu("Status bar...", [&]{
```
m9710797: Detail. For items that open a sub menu we usually don't append "..." (we only do that for items that open a new window). Or at least I believe that's the common convention.

src/imgui/ImGuiSettings.cc
```diff
@@ -398,7 +398,16 @@ void ImGuiSettings::showMenu(MSXMotherBoard* motherBoard)
...
+				ImGui::MenuItem("Show bar", nullptr, &manager.statusBarVisible);
+				ImGui::Separator();
+				ImGui::MenuItem("Show FPS indicator", nullptr, &manager.statusBarItemVisibilityFps);
+				ImGui::MenuItem("Show screen mode info", nullptr, &manager.statusBarItemVisibilityScreenModeInfo);
+				ImGui::MenuItem("Show machine time", nullptr, &manager.statusBarItemVisibilityTime);
+				ImGui::MenuItem("Show actual emulation speed", nullptr, &manager.statusBarItemVisibilityActualSpeed);
+				ImGui::MenuItem("Show machine name info", nullptr, &manager.statusBarItemVisibilityMachine);
+				ImGui::MenuItem("Show running software", nullptr, &manager.statusBarItemVisibilityRunningSoftware);
```
m9710797: Some alternatives to try:
* Indent the individual visibility items (using `im::Indent`).
* Show the items grayed out if the status bar itself is not enabled (using `im::Disabled`).
* Or both. And because this is a common combination we have a utility for that: `im::DisabledIndent()`.

So maybe something like this:
```
ImGui::MenuItem("Show bar", nullptr, &manager.statusBarVisible);
im::DisabledIndent(!manager.statusBarVisible, [&]{
    ImGui::MenuItem("Show FPS indicator", nullptr, &manager.statusBarItemVisibilityFps);
    ImGui::MenuItem("Show screen mode info", nullptr, &manager.statusBarItemVisibilityScreenModeInfo);
    ...
});
```

BTW: currently these settings are located in "Settings > Misc", but wouldn't "Settings > GUI" be a better location?

--

Another alternative is via a right-click context menu in the status bar itself.
* Advantage: currently it's not so easy to change the visibility of multiple items, because for each you have to navigate into 3 (sub)menus. With a context menu, it's only 1 level deep.
* Disadvantage: the visibility of the bar itself and of the individual items is no longer together.

src/imgui/ImGuiManager.hh
```diff
@@ -190,7 +197,13 @@ private:
 		PersistentElement{"mainMenuBarUndocked", &ImGuiManager::mainMenuBarUndocked},
 		PersistentElement{"mainMenuBarFade",     &ImGuiManager::menuFade},
 		PersistentElement{"windowPos",           &ImGuiManager::windowPos},
-		PersistentElement{"statusBarVisible",    &ImGuiManager::statusBarVisible}
+		PersistentElement{"statusBarVisible",    &ImGuiManager::statusBarVisible},
+		PersistentElement{"statusbarItemvisibility.fps",                 &ImGuiManager::statusBarItemVisibilityFps},
```
m9710797: This line has inconsistent capitalization.

src/imgui/ImGuiManager.cc
```diff
@@ -728,99 +728,111 @@ void ImGuiManager::drawStatusBar(MSXMotherBoard* motherBoard)
...
 
-			if (auto result = execute(TclObject("guess_title"))) {
-				ImGui::TextUnformatted(result->getString());
-				simpleToolTip("the (probably) currently running software");
+			if (statusBarItemVisibilityRunningSoftware) {
+				if (auto result = execute(TclObject("guess_title"))) {
+					ImGui::TextUnformatted(result->getString());
+					simpleToolTip("the (probably) currently running software");
+				}
 			}
```
m9710797, line 837: In an earlier ticket you talked about some missing info in the status bar (compared to blueMSX). Do you plan to extend the status bar now with that info, possibly disabled by default.

m9710797 (commented):
Thanks. Looks good. Only some small details. And suggestions for possible alternatives. Or future extensions.

src/imgui/ImGuiSettings.cc
MBilderbeek: OK, will fix that.

src/imgui/ImGuiSettings.cc
MBilderbeek: I also thought about 'what to do if the bar isn't visible as a whole', and I like your idea with DisabledIndent. I'll push that suggestion.

About the location: I had the same idea, but I think then we should also reconsider moving "Fade out menu bar" and perhaps "Configure OSD icons" there. What do you think?

The context-menu is also a nice idea. Not sure what's best.

But, one drawback of the currently pushed solution is that you can't easily toggle multiple of these items. Each time you have to go back into the 3 submenus if you want to change more. So, maybe that pleads for the context-menu. Or, alternatively, we should perhaps use checkboxes instead of these checkable menu items. 
I'm curious what you think about this.

src/imgui/ImGuiManager.hh
MBilderbeek: Will fix.

src/imgui/ImGuiManager.cc
MBilderbeek: Compared to the info panel script that was removed (which was based on blueMSX), we are still missing:
- mapper type of the currently running software game if it's a ROM. (The thing is a bit: this only makes sense if the currently running software is indeed a ROM... maybe it's too detailed? The information can also be easily seen through the proper cartridge slot window.)
- RAM/VRAM amount of the current machine. I'm not sure it's that useful, but perhaps we can simply show the machine info details similar to the machine selection window, when hovering the machine model status bar item?

and what blueMSX has as an extra is the host CPU usage of the emulator. But I guess that is hard to calculate in a portable way and probably not so interesting nowadays. (If it was easy to do, I'd probably still add it, though ;-))

Please share your thoughts about these.

m9710797:
I've merged this with some tweaks (see commit message).
We could still tweak this further, but personally I'm already happy with the current state.

PR #1823, opened 2025-01-16
fix warning message

pvmm:
`*begin` and `*end` are `uint16_t`, so checking if they are below 0x10000 is redundant.


m9710797:
Thanks, merged.
What compiler do you use (because I didn't see this warnings).

pvmm:
> What compiler do you use (because I didn't see this warnings).

Clang 19.1.5

PR #1826, opened 2025-01-21
Add persistency to open/close status of chips in wave viewer.

MBilderbeek:
Here's a first attempt. It's a bit odd, as it only adds items with `=1` behind, but this seems to be a simpler solution than storing a list (which may get buggy if commas end up in the name).

Please have  a look at this solution.

### DIFF
--- a/src/imgui/ImGuiWaveViewer.cc
+++ b/src/imgui/ImGuiWaveViewer.cc
@@ -123,14 +123,28 @@ static void plotHistogram(std::span<const float> values, float scale_min, float 
 }
 
 
+#define SHOW_CHIP_KEY "showChip"
+
 void ImGuiWaveViewer::save(ImGuiTextBuffer& buf)
 {
 	savePersistent(buf, *this, persistentElements);
+	for (const auto& chip: openChips) {
+		buf.appendf(SHOW_CHIP_KEY"=%s\n", chip.c_str());
+	}
+}
+
+void ImGuiWaveViewer::loadStart()
+{
+	openChips.clear();
 }
 
 void ImGuiWaveViewer::loadLine(std::string_view name, zstring_view value)
 {
-	loadOnePersistent(name, value, *this, persistentElements);
+	if (loadOnePersistent(name, value, *this, persistentElements)) {
+		// already handled
+	} else if (name == SHOW_CHIP_KEY) {
+		openChips.emplace(value);
+	}
 }
 
 static void paintVUMeter(std::span<const float>& buf, float factor, bool muted)
@@ -454,7 +468,13 @@ void ImGuiWaveViewer::paint(MSXMotherBoard* motherBoard)
 		for (const auto& info: motherBoard->getMSXMixer().getDeviceInfos()) {
 			auto& device = *info.device;
 			const auto& name = device.getName();
-			if (!ImGui::CollapsingHeader(name.c_str())) continue;
+			auto it = openChips.find(name);
+			bool wasOpen = it != openChips.end();
+			bool nowOpen = ImGui::CollapsingHeader(name.c_str(), wasOpen ? ImGuiTreeNodeFlags_DefaultOpen : ImGuiTreeNodeFlags_None);
+			if (wasOpen != nowOpen) {
+			    if (nowOpen) openChips.insert(name); else openChips.erase(it);
+			}
+			if (!nowOpen) continue;
 			HelpMarker("Right-click column header to (un)hide columns.\n"
 			           "Drag to reorder or resize columns.");
 
--- a/src/imgui/ImGuiWaveViewer.hh
+++ b/src/imgui/ImGuiWaveViewer.hh
@@ -2,6 +2,7 @@
 #define IMGUI_WAVEVIEWER_HH
 
 #include "ImGuiPart.hh"
+#include <set>
 
 namespace openmsx {
 
@@ -12,6 +13,7 @@ public:
 
 	[[nodiscard]] zstring_view iniName() const override { return "wave-viewer"; }
 	void save(ImGuiTextBuffer& buf) override;
+	void loadStart() override;
 	void loadLine(std::string_view name, zstring_view value) override;
 	void paint(MSXMotherBoard* motherBoard) override;
 
@@ -22,6 +24,8 @@ private:
 	static constexpr auto persistentElements = std::tuple{
 		PersistentElement{"show", &ImGuiWaveViewer::show}
 	};
+
+	std::set<std::string> openChips;
 };
 
 } // namespace openmsx


src/imgui/ImGuiWaveViewer.cc
```diff
@@ -123,14 +123,25 @@ static void plotHistogram(std::span<const float> values, float scale_min, float
 }
 
 
+static constexpr std::string_view SHOW_CHIP_PREFIX = "showChip.";
+
 void ImGuiWaveViewer::save(ImGuiTextBuffer& buf)
 {
 	savePersistent(buf, *this, persistentElements);
+	for (const auto& chip: openChips) {
+		buf.appendf("%s%s=%d\n", std::string(SHOW_CHIP_PREFIX).c_str(), chip.c_str(), true);
```
m9710797: This converts from `string_view` to `string` (to add a zero-terminator) to `const char*`, that seems wasteful.
You could instead make the type of `SHOW_CHIP_PREFIX` directly a `const char*` (or alternatively a `zstring_view`).

And in this case you can even write it like this:
```
#define SHOW_CHIP_PREFIX "showChip."
`buf.appendf(SHOW_CHIP_PREFIX"%s=1\n", chip.c_str();)
```
To avoid calculating appending constant strings.

--

An alternative is to simplify the format itself. We could save lines like this:
```
showChip=PSG
showChip=SCC
...
```
And that would also simplify the loading code.

--

For now we don't guarantee any forwards/backwards compatibility with .ini files from different openMSX versions. So we can always pick the format that is the simplest for the current need, and change the format when the need changes.

In the future, when we rework the session management we can think harder about this. We might want to use the openMSX serialization mechanism (that we use for savestates/replays) instead of the Dear ImGui .ini files. Our mechanism has better support for versioning (and better builtin support for serializing collections).
```

File: src/imgui/ImGuiWaveViewer.cc [hunk 2]
```diff
@@ -123,14 +123,25 @@ static void plotHistogram(std::span<const float> values, float scale_min, float
...
 
 void ImGuiWaveViewer::loadLine(std::string_view name, zstring_view value)
 {
-	loadOnePersistent(name, value, *this, persistentElements);
+	if (loadOnePersistent(name, value, *this, persistentElements)) {
+		// already handled
+	} else if (name.starts_with(SHOW_CHIP_PREFIX)) {
+		if (StringOp::stringToBool(value)) {
+			openChips.insert(std::string(name.substr(SHOW_CHIP_PREFIX.size())));
+		}
```
m9710797: See above. with the above suggestion (alternative format) this could become something like:
```
if (name == "showChip") {
    openChips.insert(value);
}
```

There's one thing missing from this code: it can only insert items in the set. So when loading a new "layout" file (so not at startup, but during runtime), the `openChip` set will contain the union of the current set and the loaded values. The fix that you also need to implement the `loadStart()` method (and then clear the set in that method).

src/imgui/ImGuiWaveViewer.cc
```diff
@@ -454,7 +465,12 @@ void ImGuiWaveViewer::paint(MSXMotherBoard* motherBoard)
 		for (const auto& info: motherBoard->getMSXMixer().getDeviceInfos()) {
 			auto& device = *info.device;
 			const auto& name = device.getName();
-			if (!ImGui::CollapsingHeader(name.c_str())) continue;
+			bool open = openChips.contains(name);
+			if (!ImGui::CollapsingHeader(name.c_str(), open ? ImGuiTreeNodeFlags_DefaultOpen : ImGuiTreeNodeFlags_None)) {
+				openChips.erase(name);
+				continue;
+			}
+			openChips.insert(name);
```
m9710797: This code is correct. Nevertheless I'd like to tweak it a bit. The reason is that it runs on every frame (and for every sound chip), and it does non-trivial string-comparisons to lookup something in a binary tree (std::set is typically implemented via a red-black-tree). And then specifically, it does this _twice_: once for `contains()` and then again for either `erase()` or `insert()`.

An alternative could be something like this:
```
std::map<std::string, bool> openChips;
...
auto [it, inserted] = openChip.try_emplace(name, false); // initialize new entries to 'false'
auto& open = it->second;
if (!...) {
   open = false;
   continue
}
open = true;
```
This does only one binary-search for the lookup (and maybe the initial insert), and then changes a reference to a `bool` (instead of `erase()` or `insert()`). I know that yesterday I suggested to use a `set` instead of a `map-to-bool`, but I'm free to change my mind ;-)

m9710797 (commented):
Thanks a lot for working on this.
I believe your code is correct. But I have some suggestions for alternatives (feel free to disagree).

src/imgui/ImGuiWaveViewer.cc
MBilderbeek: OK, thanks for the good ideas. I now updated the code to address all these points, I think.

src/imgui/ImGuiWaveViewer.cc
MBilderbeek: And this as well.

src/imgui/ImGuiWaveViewer.cc
```diff
@@ -123,14 +123,28 @@ static void plotHistogram(std::span<const float> values, float scale_min, float
...
+	openChips.clear();
 }
 
 void ImGuiWaveViewer::loadLine(std::string_view name, zstring_view value)
 {
-	loadOnePersistent(name, value, *this, persistentElements);
+	if (loadOnePersistent(name, value, *this, persistentElements)) {
+		// already handled
+	} else if (name == SHOW_CHIP_KEY) {
+		openChips.insert(std::string(value));
```
m9710797: Detail. Instead of converting `zstring_view` into `string` and then placing a copy of that string into the `set` (maybe later a `map`?), you could directly construct the string in-place via: `openChips.emplace(value)`  (caveat: I have not tested it).   (or later when you use a `map` via `openChips.try_emplace(value, true)`).

src/imgui/ImGuiWaveViewer.cc
MBilderbeek: Are you sure it's worth it? Because:
1. it's a very small set
2. this only applies if the wave viewer is open
3. the actual calculations in the wave viewer seem (intuitively) a lot heavier than these checks (if not already from point 1.)

src/imgui/ImGuiWaveViewer.cc
MBilderbeek: Good point, changed it. (I still have to get used to be able to use such a fantastic thing as `emplace`.)

src/imgui/ImGuiWaveViewer.cc
m9710797: It was just a suggestion. As I said: feel free to ignore.

I was looking at the speed of "keeping the set up-to-date" in isolation. And then the speed difference will be about a factor 2 (doing almost the same operation twice instead of just once). And it's done hundreds of times per second, (about a thousand for a machine like Boosted_MSX2_EN).

And also important, it doesn't make the code structure more complex (at least IMHO):

* For a set:
```
bool open = openChips.contains(name);
if (!ImGui::CollapsingHeader(name.c_str(), open ? ImGuiTreeNodeFlags_DefaultOpen : ImGuiTreeNodeFlags_None)) {
    openChips.erase(name);
    continue;
}
openChips.insert(name);
```

* For a map:
```
auto [it, inserted] = openChips.try_emplace(name, false);
bool& open = it->second;
if (!ImGui::CollapsingHeader(name.c_str(), open ? ImGuiTreeNodeFlags_DefaultOpen : ImGuiTreeNodeFlags_None)) {
    open = false;
    continue;
}
open = true;
```
or if you prefer (simpler than the set?):
```
bool& open = openChips.try_emplace(name, false).first->second;
open = ImGui::CollapsingHeader(name.c_str(), open ? ImGuiTreeNodeFlags_DefaultOpen : ImGuiTreeNodeFlags_None);
if (!open) continue;
```

--

An alternative could be to only update the set (via `erase()` and `insert()`) when something actually changes. Then you only do the double work occasionally. But I think that will make the control flow more complex.

And if it's really performance critical we could also use a hash_set or map. Or if it's super important we could store a boolean in the SoundDevice itself. But neither of these should be needed.

MBilderbeek: Doesn't it matter that the map also contains 'closed' chips ('false' entries) and the set only contains the 'open' chips?

m9710797: That's indeed a good point. I overlooked that.

The following still saves some repeated calculation (using a `set`). But it's indeed already a bit more complex.
I'm fine with either solution.
```
auto it = openChips.find(name);
bool wasOpen = it != openChips.end();
bool nowOpen = ImGui::CollapsingHeader(name.c_str(), wasOpen ? ImGuiTreeNodeFlags_DefaultOpen : ImGuiTreeNodeFlags_None);
if (wasOpen != nowOpen) {
    if (nowOpen) openChips.insert(name); else openChips.erase(it);
}
if (!nowOpen) continue;
```

MBilderbeek:
Merging per request of @m9710797 .

PR #1837, opened 2025-02-03
Tcl script that connects to sdcdb (SDCC debugger)

pvmm:
This script allows C developers to create C-aware breakpoints on OpenMSX console. For instance:
```
sdcdb break main.c:100
bp#1: breakpoint at 0x00004816
```
creates a breakpoint at line 100 of main.c at the appropriate memory address. While the command:
```
sdcdb list main.c:100
99     draw_scene(gf);
100    uint8_t input = input_read(KEYBOARD);
101    if (input == INPUT_IGNORED) {
102        return;
103    }
106    debug("key code =", input);
107
108    switch (input) {
```
will list the contents of the file in the OpenMSX console. This last command is just for convenience, since the sdcdb does all the work.


m9710797:
Hi,

Thanks a lot for working on this, seems like a useful addition.
I understand this is still a draft (UPDATE: or at least it was a few minutes ago ;-) so I didn't do a detailed review yet.

I did have a quick look a the Tcl code, without trying to understand the full picture yet. The code looks very clean. The only minor suggestion I have so far is:
Replace stuff like this:
```
set srcpath [lindex $args 0]
set target [lindex $args 1]
```
with:
```
lassign $args srcpath target
```

Before I can do a more detailed review I'd like to ask for some more background information.
* Can you explain the role of the different programs? Or point me to some documentation?
* I'm surprised you need a 2nd simulator "sz80 simulator (µCsim)". Does that mean the program is being simulated twice (once in openmsx, once in sz80)?
* Given an address, the program counter, is it possible to query what source file and line number corresponds to that? It might be nice if `sdcdb::list` without arguments lists the current position in the source code.
* Detail: instead of "sdcdb::list" a syntax like "sdcdb list" would be more in line with the syntax of our other commands (and with the typical Tcl syntax). Ah, that may work already, but the example still uses the first syntax.

I do agree that for prototyping a Tcl script is most convenient. But maybe later we can develop this further into an integrated ImGui window?

pvmm:
> Before I can do a more detailed review I'd like to ask for some more background information.
> 
>     * Can you explain the role of the different programs? Or point me to some documentation?

SDCC has this component called SDCDB that creates a wrapper around the Z80 simulator (µCsim, which also comes with SDCC) and it provides basically the same interface of GDB. But it's not just a wrapper, since it knows and parses the symbols created by SDCC in debug mode. µCsim provides a memory structure, CPU instruction emulation and basic debugger features (next, go, step, step into, etc.). So my script basically connects through pipes to SDCDB and send commands to it while SDCDB does the same to sz80.

>     * I'm surprised you need a 2nd simulator "sz80 simulator (µCsim)". Does that mean the program is being simulated twice (once in openmsx, once in sz80)?

Yes, SDCDB needs the simulator to work, since it won't allow me to create their own breakpoints without it, even though I am not actually running any debugging commands from SDCDB yet, just creating breakpoints, parsing their position in memory and deleting them afterwards.

>     * Given an address, the program counter, is it possible to query what source file and line number corresponds to that? It might be nice if `sdcdb::list` without arguments lists the current position in the source code.

I haven't found how to do that yet since µCsim's documentation is sparse and the code is a horrible mess. There is an assembly mode (the opposite of the default C mode) in SDCDB that may show me how to do that.

>     * Detail: instead of "sdcdb::list" a syntax like "sdcdb list" would be more in line with the syntax of our other commands (and with the typical Tcl syntax). Ah, that may work already, but the example still uses the first syntax.

Yes, the docs were outdated.

> I do agree that for prototyping a Tcl script is most convenient. But maybe later we can develop this further into an integrated ImGui window?

Yes, I thought about that. But it shouldn't be exclusive to SDCDB, SDCC or µCsim.

m9710797:
I still find it a bit weird that you have to start a 2nd simulator, and basically do nothing with it (actually even actively undo stuff that it does for you).

Today for the very first time I installed and ran `sdcc` myself. I did some quick experiments. It seems the (final) debug information is produced in a .cdb file. That file is not trivial, but also not too complex to parse. It's also reasonably well documented: https://sourceforge.net/p/sdcc/wiki/CDB%20File%20Format/ so maybe it's an idea to use that file directly instead of going via `sdcdb`?

This is an orthogonal discussion to this PR. I don't mind merging your script (for now?). But I do think that in the end, if we want a fully integrated C-level debugger in openMSX, it might be a good idea to parse the .cdb file ourselves (instead of having the user startup sdcdb and ucsim).

pvmm:
I am trying just for fun something much simpler to test the idea of OpenMSX/SDCC interation: I removed from SDCDB the code that calls the µcSim binary but I left the code that connects to it by socket at port 9756 (the default port). Then I created a TCP/IP server inside OpenMSX to listen to the connection from SDCDB, effectively replacing the default Simulator with OpenMSX. Now I can translate all SDCDB commands to OpenMSX (`debug set_bp` instead of `break` for instance) to create the breakpoints and execute code. There is a very small set of commands that would need translation anyway.

pvmm:
This was replaced by #1854

PR #1841, opened 2025-02-05
Fix msvc build error

uniskie:
"struct Symbol" is used with an incorrect forward declaration. (declared as a class in forward declaration)


m9710797:
Thanks!

Issue #1465, opened 2022-11-25
Any kbd_mapping_mode has issues JAPANESE keyboard layout


uniskie:
I tried with OpenMSX 1.8.0 for Windows.
(only japanese regions)

Graphic characters and underscores are not assigned correctly.

It happens both with the default LALT and with additional assignments to RCTRL.

    RCTRL binding:
    <bind key="keyb RCTRL">keymatrixdown 6 0x04</bind>
    <bind key="keyb RCTRL,RELEASE">keymatrixup 6 0x04</bind>

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - 

--- incorrect key list ---

| keyboard      | correct  | CHARACTER | KEY      | POSITIONAL | judge |
|---------------|----------|-----------|----------|------------|-----------|
| GRAPH + ￥    | 円       | 円        | 円       | No Input   | incorrect |
| GRAPH + [     | ○       | No Input  | No Input | ○         | incorrect |
| GRAPH + :     | ♥       | No Input  | No Input | ♥         | incorrect |
| GRAPH + ]     | ●       | ○        | ○       | 円         | incorrect |
| GRAPH + ＼_ろ | ♦       | No Input  | No Input | No Input   | incorrect |
| ＼_ろ         | MTX only | \         | No Input | MTX only   | incorrect |
| SHIFT + ＼_ろ | _        | _         | No Input | No Input   | incorrect |
| CTRL + @      | CTRL+@   | No Input  | No Input | CTRL+@     | incorrect |

* Jan 21, 2025' add "CTRL+@"
* Feb 9, 2025' ,modify ＼_ろ key's rest errors & modify key event message.

* On Japanese keyboards, yen and backslash exist as separate keys.
* All kyebord mapping mode have some kind of problem.

MSX JAPANESE KEY MATRIX
https://www.msx.org/wiki/Keyboard_Matrices#Japanese_Matrices

Windows Japanese Keyboard Layout for 106
http://kbdlayout.info/kbd106/

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - 

--- all key list ---

> "GRAPH" is PC Keybord key that assigned by OpenMSX.
> In my case, "GRAPH" is "LALT" key.

| keyboard    | correct | CHARACTER | KEY      | POSITIONAL | judgement |
|-------------|---------|-----------|----------|------------|-----------|
[...]

PR #1842, opened 2025-02-09
fix POSITIONAL mode JAPANESE keymap

uniskie:
for https://github.com/openMSX/openMSX/issues/1465

1. MSX KEYMATRIX 1-4: (￥) = SDL_SCANCODE_INTERNATIONAL3
2. MSX KEYMATRIX 2-5: (ろ_) = SDL_SCANCODE_NONUSBACKSLASH 
3. Fixed the issue where key release does not occur in IME multi-byte character mode.

see.  https://wiki.libsdl.org/SDL2/SDL_Scancode

- I was able to verify only the Japanese keyboard, but NONUSBACKSLASH may also be effective on keyboards outside Japan.

- I have determined that there is no solution to the CHARACTER/KEY mode problem.
  The behavior in POSITIONAL mode was fixable, so I fixed it.


MBilderbeek:
@m9710797 can you still take a look whether you think the solution was in the right direction? (Despite the PR being withdrawn.)

PR #1843, opened 2025-02-09
fix JAPANESE HOST KEYBOARD on POTISION mode

uniskie:
for https://github.com/openMSX/openMSX/issues/1465

1. MSX KEYMATRIX 1-4: (￥) = SDL_SCANCODE_INTERNATIONAL3
2. MSX KEYMATRIX 2-5: (ろ_) = SDL_SCANCODE_NONUSBACKSLASH
3. Added special handling for the "]" key on Japanese keyboard hosts.(Corrected to US keyboard position)
5. Fixed the issue where key release does not occur in IME multi-byte character mode.

see. https://wiki.libsdl.org/SDL2/SDL_Scancode

### DIFF
--- a/src/input/Keyboard.cc
+++ b/src/input/Keyboard.cc
@@ -1135,7 +1135,13 @@ void Keyboard::processSdlKey(EmuTime::param time, SDLKey key)
 	};
 
 	if (keyboardSettings.getMappingMode() == KeyboardSettings::MappingMode::POSITIONAL) {
-		if (const auto* mapping = binary_find(scanCodeTab, key.sym.scancode, {}, &ScanCodeMsxMapping::hostScanCode)) {
+		if ((key.sym.sym == SDLK_RIGHTBRACKET) && (key.sym.scancode == SDL_SCANCODE_BACKSLASH))
+		{
+			// host: Japanese keyboard "]" -> position as -> US keyboard "GRAVE"
+			if (const auto* mapping = binary_find(scanCodeTab, SDL_SCANCODE_GRAVE, {}, &ScanCodeMsxMapping::hostScanCode)) {
+				process(mapping->msx);
+			}
+		} else if (const auto* mapping = binary_find(scanCodeTab, key.sym.scancode, {}, &ScanCodeMsxMapping::hostScanCode)) {
 			process(mapping->msx);
 		}
 	} else {


uniskie:
There are reasonable grounds to respond with SCANCODE.

- INTERNATINAL3 is often used for currency keys on non-English speaking keyboards.

- NON-US-BACKSLASH is also used as a backslash in non-English speaking countries.

- There is no key in the NON-US "]" key position on a US keyboard  (The place is taken up by ENTER). It seems that the GRAVE key is sometimes placed in the corresponding position.
  > Here's the scenario:
  > JIS "]" was determined to be at the US BACKSLASH key position according to the USB standard, but in the case of MSX it was determined to be at the US GRAVE key position.
  > 
  > (In fact, there are also US keyboards where the GRAVE key is next to ENTER.)
  > 
  > In this pull request, if it is determined that it is NON US RIGHTBRACKET, we will fill in the difference.
  >
  > _(It's not the smartest implementation, but it's the surest method with the least impact.)_

I don't understand why SDL's Keycode cannot identify these INTERNATINAL3 and NON-US-BACKSLASH key, but it is because it is intended for text input, and it is a flawed function for emulation purposes that require a physical key press state.

What is important here is that there are keys in the key matrix that are pressed down without any character input.
（NONUSBACKSLASH is exactly that.）

It is impossible to reproduce it with anything other than SCANCODE.

For the reasons mentioned above, I applied the fix to POSITIONAL mode.
At least the only way to solve the Japanese keyboard problem is to use SCANCODE.

src/input/Keyboard.hh
```diff
@@ -240,9 +240,19 @@ private:
...
+	// 
+	// // Remembers the last unicode for a key-press-keycode. To be used later
+	// // on the corresponding key-release, because those don't have unicode info.
+	// std::vector<std::pair<SDL_Keycode, uint32_t>> lastUnicodeForKeycode; // sorted on SDL_keycode
+	// ***
+
+	// Remembers the last unicode for a key-press-scancode. To be used later
 	// on the corresponding key-release, because those don't have unicode info.
-	std::vector<std::pair<SDL_Keycode, uint32_t>> lastUnicodeForKeycode; // sorted on SDL_keycode
+	std::vector<std::pair<uint32_t, uint32_t>> lastUnicodeForScancode; // sorted on SDL_Scancode
```
m9710797: Can we use `SDL_Scancode` instead of `uint32_t`?

src/input/Keyboard.cc
```diff
@@ -2068,8 +2140,18 @@ void Keyboard::serialize(Archive& ar, unsigned version)
...
-	if (ar.versionAtLeast(version, 4)) {
+	if (ar.versionAtLeast(version, 5)) {
+		ar.serialize("lastUnicodeForScancode", lastUnicodeForScancode);
+	}
+	else if (ar.versionAtLeast(version, 4)) {
+		// convert keycode to scancode
+		std::vector<std::pair<SDL_Keycode, uint32_t>> lastUnicodeForKeycode;
 		ar.serialize("lastUnicodeForKeycode", lastUnicodeForKeycode);
+		lastUnicodeForScancode.clear();
+		for (auto it : xrange(lastUnicodeForKeycode.begin(), lastUnicodeForKeycode.end())) {
```
m9710797: This for loop can be written like:
```
for (const auto& [keyCode, unicode] : lastUnicodeForKeyCode) {
    auto scanCode = SDL_GetScancodeFromKey(keyCode);
    lastUnicodeForScancode.emplace_back(scanCode, unicode);
}
```

src/input/Keyboard.cc
```diff
@@ -2068,8 +2140,18 @@ void Keyboard::serialize(Archive& ar, unsigned version)
...
+			auto ScanCode = SDL_GetScancodeFromKey(it->first);
+			lastUnicodeForScancode.push_back(std::pair<uint32_t, uint32_t>(ScanCode, it->second));
+		}
```
m9710797: Maybe it's not needed, but I'm wondering if the new `cntKeyMatrix` variable should be serialized.

src/input/Keyboard.cc
```diff
@@ -1127,6 +1155,10 @@ void Keyboard::updateKeyMatrix(EmuTime::param time, bool down, KeyMatrixPosition
 		return;
 	}
 	if (down) {
+	#if USE_KEYPRESSCNT
+		auto pos_i = pos.getRow() * pos.getColumn();
```
m9710797: Maybe use a 2D array, instead of linearizing it manually?

src/input/Keyboard.cc
```diff
@@ -1127,6 +1155,10 @@ void Keyboard::updateKeyMatrix(EmuTime::param time, bool down, KeyMatrixPosition
...
+		cntKeyMatrix[pos_i] = cntKeyMatrix[pos_i] < 255 ? cntKeyMatrix[pos_i] + 1 : cntKeyMatrix[pos_i];
```
m9710797: ```cntKeyMatrix[pos] = std::min(255, cntKeyMatrix[pos] + 1);```
and similar below
```cntKeyMatrix[pos] = std::max(0, cntKeyMatrix[pos] - 1);```

But is this really needed? Can this ever overflow in practice? Or is this extra protection for when increment/decrement are ever not properly paired?

m9710797 (commented):
Manuel, asked me to review. I had a quick look at the code, but I did NOT do any testing.

For a prototype/experiment, this code is fine (to easily test with/without). But in the final version we should remove the `USE_KEYPRESSCNT` and `KEYTEST` defines. Either we need them or we don't, but not conditionally.

I see what `cntKeyMatrix` is doing, but I don't immediately see why it's needed. Is it still an experiment? Or can you give an example of where this fixes things?

src/input/Keyboard.cc
uniskie: This is an unintentional experimental commit.
Feel free to ignore it.
As pointed out, cntKeyMatrix does not need to be serialized.

uniskie:
> Manuel, asked me to review. I had a quick look at the code, but I did NOT do any testing.
> 
> For a prototype/experiment, this code is fine (to easily test with/without). But in the final version we should remove the `USE_KEYPRESSCNT` and `KEYTEST` defines. Either we need them or we don't, but not conditionally.
> 
> I see what `cntKeyMatrix` is doing, but I don't immediately see why it's needed. Is it still an experiment? Or can you give an example of where this fixes things?

This is an unintentional experimental commit.
Feel free to ignore it.

src/input/Keyboard.hh
uniskie: (This is an unintentional experimental commit.Feel free to ignore it.)

A type mismatch occurred in the serialize macro when using SDL_Keycode.
Since it is just an experimental code, I chose uint32_t, but SDL_Keycode is preferable for the quality of the program.

src/input/Keyboard.cc
uniskie: (This is in the experimental stage, so you can ignore it.)

In fact, in cases where this protection is necessary, you should probably take precautions in advance.

src/input/Keyboard.cc
uniskie: (This is in the experimental stage, so you can ignore it.)

A formal implementation would have been a 2d array.
In reality, it is a code that ignores quality.
This is an experimental code that I don't have time for, so I avoided complicated multidimensional array templates.

src/input/Keyboard.cc
uniskie: supplement:

cntKeyMatrix is ​​only used to clarify the keyboard state on the host side, and has nothing to do with machine state.
So I don't think serialization is necessary.

In the current implementation,
Problems may occur if you switch the keyboard mapping mode while a key is pressed.

Although it is unlikely to occur, safety measures are necessary.

src/input/Keyboard.hh
uniskie: Due to the mechanism of serialize, on the enumlate side, conversion processing with the enumeration name string is required.

(The string list consumes memory, and the conversion process with strings consumes processing time.)

This is redundant, so I think primitive types are preferable.

Considering the security of the code, I would like to use SDL_Scancode as a key, but this is a bit of a problem.

Do you have any good ideas in this case?

(The keycode cannot be used because sdl may not tell us the keycode...)

@m9710797

uniskie: The way I can come up with is:

The method is to create a class of lastUnicodeForScancode, operate it via the interface, and perform the serialization process within the class.

PR #1853, opened 2025-02-09
Fatal error and message if init.tcl is not found

pvmm:
Display more useful error message instead of just crashing, since OpenMSX is still usable without the Tcl environment.

### DIFF
--- a/src/Reactor.cc
+++ b/src/Reactor.cc
@@ -543,8 +543,9 @@ void Reactor::runStartupScripts(const CommandLineParser& parser)
 	try {
 		commandController.source(
 			preferSystemFileContext().resolve("init.tcl"));
-	} catch (FileException&) {
-		// no init.tcl, ignore
+	} catch (FileException& e) {
+		throw FatalError("Couldn't execute \"<openmsx>/share/init.tcl\": ",
+		                 e.getMessage());
 	}
 
 	// execute startup scripts


m9710797:
Thanks a lot for your PR ... but I'm a bit confused.

The title or this PR and the warning message talk about `init.tcl`. But the implementation is inside `tabCompletion()` (the part that completes the first token on the line, the command name). I don't immediately see the connection between the two.

Can you give some more background info. What should be wrong in the openMSX installation, and what action should I then perform to trigger this new warning message?

You add a `try-catch` block around a rather big code block. As far as I can see, the only(?) line in there that can be related to `init.tcl` is `interpreter.getCommandNames()`, right? Wouldn't it be better (clearer) to put the `try-catch` only around that line? Or even better move the error handling inside that function, then it's also shown for all other callers of that function.

... Looking a bit deeper ... `getCommandNames()` only executes the Tcl proc `openmsx::all_command_names` which should indeed have been defined via `init.tcl`. But if something is wrong with that, then I expect several other things will go wrong as well. Shouldn't we then try to detect this problem at an earlier moment in time?  (But I'll probably understand better once you've explained how exactly to trigger this problem).

m9710797:
@pvmm ping?

pvmm:
Before this error occurred, I was unaware that there were Tcl scripts outside of `<openmsx>/share/scripts` as it is not in the console command reference and I was building a shared folder from scratch for some experiments. This PR just presents an error message rather than leaving the user with no clue as to what happened.

m9710797:
You only half answered my original question ("How to trigger the error scenario that this PR is trying to solve?"). Did I understand correctly that the scenario involves removing `init.tcl`?

That `init.tcl` fie is fundamental to a correct openMSX installation. Lots of other Tcl stuff builds upon that. So if it's missing then lots of other stuff will also go wrong.

(If my understanding is correct then) maybe we should reject this PR and instead create a different PR that gives a clearer error message when `init.tcl` could not be found (or could not be executed without errors). Maybe it should even be a hard error (refusing to start openMSX)?

pvmm:
OpenMSX now stops if init.tcl is not found.

m9710797:
Thanks! I've merged your PR with a small tweak to the error message.

I was a bit concerned about the fact that apparently we _intentionally_ ignore a missing `init.tcl`.
I did some digging through git-history ... this decision was made in 2004, in the very first commit that integrated Tcl scripting. At that point in time `init.tcl` was indeed still optional. Later it became mandatory. But it seems we never revisited this check.

PR #1854, opened 2025-02-11
Tcl script that replaces sdcdb (SDCC debugger)

pvmm:
This script allows C developers to create C-aware breakpoints on OpenMSX console. You just need to compile your code with `-debug` to create a file with symbols and their respective addresses. For instance:
```
sdcdb break main.c:155
bp#1
```
creates a breakpoint at line 155 of main.c at the appropriate memory address.
```
break bp_list
bp#1 18558 {} {sdcdb info -break}
```
And then
```
debug cont
file: main.c:155, position: 0x489C
154   :    int main() {
155   *        platform_init();
156   :
```
While the command `sdcdb next` go to the next instruction:
```
sdcdb next
file: main.c: 157, position: 0x489F
156   :
157   *        struct gamefield* gf = init_gamefield();
158   :        debug("[DEBUG] program started at: ", (uint16_t) gf);
```
And `sdcdb list` can show you more of those files.
```
sdcdb list main.c:100
99     draw_scene(gf);
100    uint8_t input = input_read(KEYBOARD);
101    if (input == INPUT_IGNORED) {
102        return;
103    }
106    debug("key code =", input);
107
108    switch (input) {
```
The `sdcdb help` shows all the commands.

This PR replaces the older #1837, because it is a completely different implementation.

### DIFF
--- a/share/scripts/sdcdb.tcl
+++ b/share/scripts/sdcdb.tcl
@@ -1,0 +1,795 @@
+# SDCDB Debugger in Tcl - Version 0.8
+#
+# Copyright 2025 Pedro de Medeiros all rights reserved
+#
+# A debugger written in Tcl that uses `SDCC --debug` parameter to create breakpoints and much more.
+#
+# The main difference between SDCC Debugger's breakpoints and the built-in OpenMSX breakpoints is the
+# C code integration:
+# [*] allow users to create breakpoints on C code:
+#     > sdcdb break main.c:55
+# [*] list C source code at line 100:
+#     > sdcdb list main.c:100
+# [*] step through the source code
+#     > sdcdb step
+# [*] and more things to come (WIP).
+#
+# All commands that the SDCDB Debugger recognizes directly. You may call them with sdcdb [COMMAND] [ARGS...]:
+#
+# open ?-recursive? <directories> [<pathToCDBFile>]
+#       - Start debugging session, reading source files and a CDB file specified by parameters.
+#         -recursive enables recursive searching in subdirectories.
+# break <file>:<line>
+#       - creates breakpoint in <file>:<line>
+# list ?<line>?
+#       - list source code at <file>:<line>[-<line>]
+# info ?-break?
+#       - Display information on source code under the program counter. The -break parameter
+#         parameter stops execution.
+# step ?<n>?
+#       - Executes next n lines of C code step by step, proceeding through subroutine calls.
+#         n defaults to 1.
+# next ?<n>?
+#       - The 'sdcdb next' is like 'sdcdb step' but will not proceed through subroutine calls.
+#         You may specify how many n times 'sdcdb next' should execute (defaults to 1).
+# whereis <fileName>
+#       - Checks if a C/assembly source file was included in a scan from `sdcdb open` and
+#         returns the path to the file.
+# map <file>:<line>
+#       - Returns memory address of <file>:<line>, accepting same parameters of 'sdcdb break',
+#         but without creating breakpoints.
+# laddr <address>
+#       - Returns the C/assembly source code associated with a memory address.
+#
+# Known limitations:
+# [*] don't create C files with same name in different folders (in projects with multiple
+#     folders), since CDB files don't keep track of directories, just file names. SDCDB
+#     will get lost and report the wrong position because of missing files during break or
+#     step.
+# [*] don't create projects with assembly and C files with the same name, since SDCC
+#     already creates an .asm file for every .c file in your project. SDCDB will get lost
+#     and report the wrong position because of missing files during break or step.
+#
+# For more information about the CDB file format: https://sourceforge.net/p/sdcc/wiki/Home/
+
+namespace eval sdcdb {
+
+variable initialized    0
+variable c_files_count                  ;# C files reference
+variable a_files_count                  ;# ASM files reference
+array set c_files       {}              ;# pool of c source files
+array set a_files       {}              ;# pool of a source files
+array set addr2file     {}              ;# array that maps to source from address
+array set g_func2addr   {}              ;# array that maps to address from source (global)
+# array set c_<filename>2addr           ;# same, but dinamically created array for each SDCC module (C file)
+# array set a_<filename>2addr           ;# same, but dinamically created array for each SDCC module (ASM file)
+# array set <filename>_func2addr        ;# array that maps function to source (static)
+variable current_file   {}              ;# debugger file
+variable current_line   {}              ;# debugger line
+variable cond           {}              ;# breakpoint check condition
+variable old_PC         {}
+variable times_left     0
+proc ASM {} { return ".asm" }
+
+set_help_proc sdcdb [namespace code sdcdb_help]
+proc sdcdb_help {args} {
+    if {[llength $args] == 1} {
+        return {The SDCDB debugger in Tcl connects OpenMSX to the CDB file created by SDCC.
+
+Recognized commands: open, break, list, step, next, info, whereis, laddr, map
+
+Type 'help sdcdb <command>' for more information about each command.
+}
+    }
+    switch -- [lindex $args 1] {
+        "open" { return {Opens a project directory and start debugging session
+
+'sdcdb open' will scan directories for source files and also search for a single CDB file in them, or you may specify the path to a CDB file directly (a single file probably named <projectName>.cdb) in addition to the source files.
+
+Syntax: sdcdb open <directories> [<.cdbFile>]
+}}
+        "break" { return {Creates a breakpoint
+
+Create a OpenMSX breakpoint, but using the C source files as reference. 'sdcdb info -break' replaces 'debug break' for extra details about C code execution.
+
+Syntax: sdcdb break <file>:<line>
+        sdcdb break <file>:<functionName>
+        sdcdb break <functionName>
+}}
+        "list" { return {Lists contents of a C source file
+
+Without parameters, 'sdcdb list' returns the C/assembly source code under the PC register. Obs.: assembly function names are not recognized.
+
+Syntax: sdcdb list
+        sdcdb list <file>:<line>
+        sdcdb list <file>:<functionName>
+        sdcdb list <functionName>
+}}
+        "laddr" { return {Lists contents of C/assembly source code in a specified memory address
+
+Syntax: sdcdb laddr mem
+}}
+        "step" { return {Steps through next n lines of C code
+
+The 'sdcdb step' command will proceed through subroutine calls. You may specify how many times 'sdcdb step' should execute (defaults to 1).
+
+Syntax: sdcdb step ?n?
+}}
+        "next" { return {Executes next n lines of C code
+
+The 'sdcdb next' command will not proceed through subroutine calls. You may specify how many times 'sdcdb next' should execute (defaults to 1).
+
+Syntax: sdcdb next ?n?
+}}
+        "info" { return {Displays information about current line of source code
+
+A '-break' parameter stops execution after displaying the information.
+
+Syntax: sdcdb info ?-break?
+}}
+        "whereis" { return {Displays if source file was found in the database
+
+Syntax: sdcdb whereis <sourceFileName>
+}}
+        "map" { return {Returns memory address from a C/assembly file line
+
+Basically a dry-run version of 'sdcdb break'. If a functionName is used, it returns the beginning and ending address of the function.
+
+Syntax: sdcdb map <file>:<line>
+        sdcdb map <file>:<functionName>
+        sdcdb map <functionName>
+}}
+    }
+}
+
+proc debug_out {args} {
+    variable debug
+    if {$debug} {
+        set chan stderr
+        set msg [string map {"\n" "\\n"} [join $args " "]]
+        puts $chan $msg
+        flush $chan
+    }
+}
+
+proc sdcdb {args} {
+    if {[catch {set result [dispatcher {*}$args]} msg]} {
+        variable debug
+        if {$debug} {
+            puts stderr $::errorInfo
+        }
+        error $msg
+    }
+    return $result
+}
+
+proc dispatcher {args} {
+    set params "[lrange $args 1 end]"
+    set cmd [lindex $args 0]
+    switch -- $cmd {
+        open    { return [sdcdb_open       {*}$params] }
+    }
+    # remaining commands need initialization
+    variable initialized
+    if {!$initialized} {
+        error "You need a CDB file to use this command. Use the 'sdcdb open' command first"
+    }
+    switch -- $cmd {
+        break   { return [sdcdb_break      {*}$params] }
+        list    { return [sdcdb_list       {*}$params] }
+        info    { return [sdcdb_info       {*}$params] }
+        step    { return [sdcdb_step       {*}$params] }
+        next    { return [sdcdb_next       {*}$params] }
+        whereis { return [sdcdb_whereis    {*}$params] }
+        laddr   { return [sdcdb_laddr      {*}$params] }
+        map     { return [sdcdb_map        {*}$params] }
+        default { error "Unknown command \"[lindex $args 0]\"." }
+    }
+}
+
+proc sdcdb_open {param args} {
+    set function normal_glob
+    variable initialized
+    if {$initialized} {
+        free
+    }
+    if {$param eq "-recursive"} {
+        set function recursive_glob
+    } else {
+        set args [linsert $args 0 $param]
+    }
+    set cdb_file {}
+    foreach path $args {
+        if {![file exists $path]} {
+            error "file '$path' not found"
+        }
+        if {[file isdirectory $path]} {
+            set new_files [$function $path *.c]
+            debug_out "adding files: [join $new_files {, }]"
+            add_files_to_database c_files $new_files
+            set new_files [$function $path *[ASM]]
+            debug_out "adding files: [join $new_files {, }]"
+            add_files_to_database a_files $new_files
+            set cdb_files [$function $path *.cdb]
+        } elseif {[file isfile $path] && [file extension $path] eq ".cdb"} {
+            set cdb_file $path
+        }
+    }
+    if {$cdb_file eq {}} {
+        if {[llength $cdb_files] > 1} {
+            error "multiple CDB files found, you may specify one as a parameter to 'scdb open'"
+        } elseif {[llength $cdb_files] == 0} {
+            error "CDB file not found"
+        }
+        set cdb_file $cdb_files
+    }
+    set initialized 1
+    read_cdb $cdb_file
+    process_data
+}
+
+proc complete {arrayname name list} {
+    variable $arrayname
+    if {![info exists ${arrayname}($name)]} {
+        debug_out "$arrayname\($name\) ignored"; return 0
+    }
+    set ${arrayname}($name) [concat {*}[set ${arrayname}($name)] {*}$list]
+    return 1
+}
+
+proc read_cdb {fname} {
+    variable c_files_count
+    variable addr2file
+    set fh [open $fname "r"]
+    # function pattern: search for "F:G$function_name$..." lines
+    set func_pat {^F:(G|F([^$]+)|L([^$]+))\$([^$]+)\$.*}
+    # ASM line number pattern: search for "L:A$filename$line:address" lines
+    set aline_pat {^L:A\$([^$]+)\$([^$]+):([^$]+)$}
+    # C line number pattern: search for "L:C$filename$line$level$block:address" lines
+    set cline_pat {^L:C\$([^$]+)\$([^$]+)\$([^$]+)\$([^$]+):(\S+)$}
+    # function begin pattern: search for "L:(G|F<name>|L<name>)$function$level$block:address" lines
+    set func_bn_pat {^L:(G|F([^$]+)|L([^$]+))\$([^$]+)\$([^$]+)\$([^$]+):(\S+)$}
+    # function end pattern: search for "L:X(G|F<name>|L<name>)$function$level$block:address" lines
+    set func_ed_pat {^L:X(G|F([^$]+)|L([^$]+))\$([^$]+)\$([^$]+)\$([^$]+):(\S+)$}
+    set c_count 0  ;# lines of C code
+    set a_count 0  ;# lines of asm code
+    set gf_count 0 ;# global functions
+    set sf_count 0 ;# static functions
+    while {[gets $fh line] != -1} {
+        set match [regexp -inline $func_pat $line]
+        if {[llength $match] == 5} {
+            lassign $match {} context {} {} funcname
+            switch -- [string index $context 0] {
+                G {
+                    variable g_func2addr
+                    set g_func2addr($funcname) {}
+                    debug_out "g_func2addr\($funcname\) created"
+                }
+                F {
+                    set arrayname [string range $context 1 [string length $context]]_func2addr
+                    variable $arrayname
+                    set ${arrayname}($funcname) {}
+                    debug_out "${arrayname}\($funcname\) created"
+                }
+                L {}
+            }
+            continue
+        }
+        set match [regexp -inline $cline_pat $line]
+        if {[llength $match] == 6} {
+            lassign $match {} filename linenum {} {} address
+            incr c_files_count($filename)
+            # Put line -> address mapping of array with dynamic name
+            set arrayname c_[file rootname $filename]2addr
+            if {$c_files_count($filename) eq 1} {
+                debug_out "Created new dynamic array $arrayname"
+            }
+            variable $arrayname
+            set ${arrayname}($linenum) [expr {"0x$address"}]
+            set addr2file([expr {"0x$address"}]) [list $filename $linenum]
+            debug_out "mapping C source ${arrayname}\($linenum\): [set ${arrayname}($linenum)]"
+            incr c_count
+            continue
+        }
+        set match [regexp -inline $aline_pat $line]
+        if {[llength $match] == 4} {
+            lassign $match {} filename linenum address
+            incr a_files_count($filename[ASM])  ;# SDCC removes the file extension for some reason
+            # Put line -> address mapping of array with dynamic name
+            set arrayname a_[file rootname $filename]2addr
+            if {$a_files_count($filename[ASM]) eq 1} {
+                debug_out "Created new dynamic array $arrayname"
+            }
+            variable $arrayname
+            set ${arrayname}($linenum) [expr {"0x$address"}]
+            # Is it possible for assembly to override the C mappings?
+            set addr2file([expr {"0x$address"}]) [list $filename[ASM] $linenum]
+            debug_out "mapping asm source ${arrayname}\($linenum\): [set ${arrayname}($linenum)]"
+            incr a_count
+            continue
+        }
+        set match [regexp -inline $func_bn_pat $line]
+        if {[llength $match] == 8} {
+            lassign $match {} context {} {} funcname {} {} address
+            # Put function begin record in array with dynamic name
+            switch -- [string index $context 0] {
+                G {
+                    variable g_func2addr
+                    if {[complete g_func2addr $funcname [list [expr {"0x$address"}]]]} {
+                        debug_out "1:mapping function g_func2addr\($funcname\): [set g_func2addr($funcname)]"
+                        incr gf_count
+                    }
+                }
+                F {
+                    set arrayname [string range $context 1 [string length $context]]_func2addr
+                    variable $arrayname
+                    if {[complete $arrayname $funcname [list [expr {"0x$address"}]]]} {
+                        debug_out "1:mapping function ${arrayname}\($funcname\): [set ${arrayname}($funcname)]"
+                        incr sf_count
+                    }
+                }
+                L {}
+            }
+            continue
+        }
+        set match [regexp -inline $func_ed_pat $line]
+        if {[llength $match] == 8} {
+            lassign $match {} context {} {} funcname {} {} address
+            # Put function end record in array with dynamic name
+            switch -- [string index $context 0] {
+                G {
+                    variable g_func2addr
+                    if {[complete g_func2addr $funcname [list [expr {"0x$address"}]]]} {
+                        debug_out "mapping function g_func2addr\($funcname\): [set g_func2addr($funcname)]"
+                    }
+                }
+                F {
+                    set arrayname [string range $context 1 [string length $context]]_func2addr
+                    variable $arrayname
+                    if {[complete $arrayname $funcname [list [expr {"0x$address"}]]]} {
+                        debug_out "mapping function ${arrayname}\($funcname\): [set ${arrayname}($funcname)]"
+                    }
+                }
+                L {
+                }
+            }
+            continue
+        }
+        debug_out "Ignored '$line'"
+    }
+    close $fh
+    variable c_files
+    variable a_files
+    puts "[array size c_files] C files added"
+    puts "[array size a_files] assembly files added"
+    puts "[array size c_files_count] C files references added"
+    puts "[array size a_files_count] assembly files references added"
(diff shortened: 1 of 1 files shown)


m9710797:
Very nice! The main external input is the `.cdb` file, right? (And the `.c` files for listing the source code). So no more communication with external processes.

I didn't read the script in detail yet (it's also still a draft PR). But at first sight it looks nice!

Quick question/suggestion:
* After doing a `cdcdb open <.cdb file>`, would it make sense to automatically do a `cdcdb add <path-of-cdb-file>`? At least in my tiny experiment, my `.cdb` and `.c` files were in the same directory.
* And maybe it's an idea to let `cdcdb add` scan the full directory tree (so recursively in sub-directories).

pvmm:
I've added the tracking of assembly files, now SDCDB won't get lost so easily. It can now  `sdcdb step` through the source code n times, just like GDB. I think it is good enough for regular use, even if `sdcdb next` is not implemented yet.

I've added your suggestions too.

pvmm:
Things to add next: the `sdcdb next` command, make it work with segments, display info on frame variables, including local variables.

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,778 @@
...
+#     step.
+# [*] don't create projects with assembly and C files with the same name, since SDCC
+#     already creates an .asm file for every .c file in your project. SDCDB will get lost
+#     and report the wrong position because of missing files during break or step.
+#
+# For more information about the CDB file format: https://sourceforge.net/p/sdcc/wiki/Home/
+
+namespace eval sdcdb {
+
+set ::env(DEBUG) 1
```
m9710797: Maybe this should not be enabled by default?

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,778 @@
...
+
+Syntax: sdcdb quit}}
+    }
+}
+
+proc output {args} {
+    set chan stdout
+    puts $chan [join $args " "]
+    flush $chan
+}
```
m9710797, line 143: Most (all) commands use this `output` proc to produce their output. An alternative is to return a string from the command.

When used interactively in the openMSX console, the outcome is the same: both alternatives print the output in the console.

But the alternative of returning a string allows the result of a `sdcdb` command to be further processed by other Tcl scripts.

Or are some of the commands taking a (very) long time? So that you already want to start outputting partial results instead of waiting for the full command to finish? Maybe that's why you have the `flush` command?

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,785 @@
...
+
+proc warn {args} {
+    if {[env DEBUG]} {
+        set chan stderr
+        set msg [string map {"\n" "\\n"} [join $args " "]]
+        puts $chan $msg
+        flush $chan
+    }
+}
```
m9710797, line 153: This is more like debug-output than warning message, right?
* Debug output is mostly (or only) useful for you (the developer of this script).
* Warning messages can be useful for the user of this script.

So maybe rename this proc? Maybe `debug` or `debug_output`?

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,785 @@
...
+
+proc env {varname {defaults {}}} {
+    if {[info exists ::env($varname)]} {
+        return $::env($varname);
+    }
+    return $defaults;
+}
```
m9710797: You used an environment variable `DEBUG`. I agree that may have been very useful for you during development.
But `DEBUG` is a very generic name, it might clash with environment variables that the user has set.

Also should it be an environment variable? Maybe it can be a regular Tcl variable in the `sdcdb` Tcl namespace? Then you can enable it via a command like `set sdcdb::debug true`.

(And then this proc can be dropped).

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,785 @@
...
+
+proc sdcdb {args} {
+    if {[env DEBUG]} {
+        if {[catch {set result [dispatcher {*}$args]} fid]} {
+            puts stderr $::errorInfo
+            error $::errorInfo
```
m9710797: Should this be `error $::errorInfo` or `error $fid`? I assume you want to re-issue the error unchanged, not modify it to include the full backtrace? (The full backtrace was already printed to `stderr`)

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,785 @@
...
+        }
+    } else {
+        set result [dispatcher {*}$args]
+    }
+    return $result
+}
```
m9710797, line 164: Good idea to add a mechanism to get more detailed debug info. But I don't like that the debug version takes a completely different code path. What about only putting the "if-debug" around the "put stderr"? Maybe something like this:
```
if {[catch {set result [dispatcher {*}$args]} msg]} {
    if {[env DEBUG]} {
        puts stderr $::errorInfo
    }
    error $msg
}
return $result
```

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,785 @@
...
+
+proc dispatcher {args} {
+    set params "[lrange $args 1 end]"
+    set cmd [lindex $args 0]
+    switch -- $cmd {
+        "open"  { return [sdcdb_open       {*}$params] }
+        add     { return [sdcdb_add        {*}$params] }
+        whereis { return [sdcdb_whereis    {*}$params] }
+        quit    { return [sdcdb_quit       {*}$params] }
```
m9710797: Detail: you have quotes around only some of the sub-command names. Either with or without is fine, but be consistent.

Same remark below.

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,785 @@
...
+    }
+    # remaining commands need initialization
+    variable initialized
+    if {!$initialized} {
+        error "SDCDB not initialized"
```
m9710797: Detail: "SDCDB not initialized" does not really help the user. And the help text doesn't talk about initialization. Maybe instead this error should talk about "no CDB file opened"? And maybe even talk about which command t use to do this?

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,785 @@
...
+        "next"  { return [sdcdb_next       {*}$params] }
+        default { error "Unknown command \"[lindex $args 0]\"." }
+    }
+}
+
+proc sdcdb_open {path} {
+    variable initialized
+    if {$initialized} {
+        error "SDCDB already initialized"
+    }
```
m9710797, line 200: I don't see a way to un-initialize. So once you've opened one `.cdb` file, you can never switch to another anymore.

Instead of giving an error, can we do an automatic `sdcdb_quit`?  And then of course `sdcdb_quit` should set `initialized` to 0 again.

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,785 @@
...
+    set initialized 1
+    set file [glob -type f -directory $path *.cdb]
+    if {[llength $file] ne 1} {
+        error "unique regular CDB file not found"
+    }
```
m9710797, line 225: I was thinking about (also) allowing to specify a `.cdb` file directly instead of a path to a directory containing the `.cdb` file (and with an error if there is more than one such file).

But maybe that's not worth the extra complexity? Will it ever happen in practice that there's more than 1 `.cdb` file?

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,785 @@
...
+    sdcdb_add $path
```
m9710797: Should this be `-recursive`?

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,785 @@
...
+        ;#warn "Ignored '$line'"
+    }
+    close $fh
+    output "[array size c_files_count] C files references added"
+    output "[array size a_files_count] assembly files references added"
+    output "$c_count C source lines found"
+    output "$a_count assembly source lines found"
+    output "$gf_count global function(s) registered"
+    output "$sf_count static function(s) registered"
+}
```
m9710797: I admit I didn't (try to) understand all this parsing code in detail. But at first sight it all looks sane.

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,790 @@
...
+        if {$old_key ne {} && [expr $key - 1] != $old_key} {
+            for {set i [expr $old_key + 1]} {$i < $key} {incr i} {
+                #warn "Setting $arrayname\($i\) to ${arrayname}\($old_key\) = [set ${arrayname}($old_key)]"
+                set ${arrayname}($i) [set ${arrayname}($old_key)]
+                incr count
+            }
+        }
+        set old_key $key
+    }
+    return $count
```
m9710797: This return value is never used. Should it be used? Or can we simplify this proc (not compute it)? Or is it sometimes useful for debugging?

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,790 @@
...
+        output "* $key: ${arrayname}($key)"
+    }
+}
+
+proc fix_blank_spaces {arrayname} {
+    variable $arrayname 
+    set count 0
+    set old_key {}
+    foreach key [lsort -integer [array names $arrayname]] {
+        if {$old_key ne {} && [expr $key - 1] != $old_key} {
```
m9710797: `[expr $key - 1]` should be `[expr {$key - 1}]`, so add extra `{}`.

This is just one example. But this remark applies to _all_ `expr` in this script.

The reason is technical and complicated, see [here](https://wiki.tcl-lang.org/page/Brace+your+expr%2Dessions) for a detailed explanation. Summary: it's faster and it avoids surprises.

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,790 @@
...
+    variable a_files
+    foreach filename [array names a_files] {
+        set arrayname a_[file rootname $filename]2addr
+        variable $arrayname
+        warn "1fix_last_address: $arrayname"
+        fix_last_address $arrayname
+        fix_blank_spaces $arrayname
+    }
+    fix_blank_spaces addr2file
+    return
```
m9710797, line 422: Very much a detail. A `return` at the end of a proc is not needed.

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,790 @@
...
+}
+
+proc normal_glob {dir pattern} {
+    return [glob -nocomplain -type f -directory $dir $pattern]
```
m9710797: Also a detail: if the last line in a proc is `return [abc def]`, it's more idiomatic to leave out that `return` so just leave `abc def`.

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,790 @@
...
+}
+
+proc recursive_glob {dir pattern} {
+    set result {}
```
m9710797: Detail:
`set result {}` and `set result [list]` functionally do the same thing. The former creates an empty string, the latter an empty list, but an empty string is represented as an empty string.

So from a functional point of view there's no difference. But `set result [list]` gives a hint to the reader of the code that result will be used as a list. And it will run a little bit faster because the internal Tcl interpreter can avoid one more conversion between string and list (internally strings and lists DO have a different representation, it's only when we ask for the string-representation of the list that it becomes the same).

So I'd suggest to write this line as:  `set result [list]`

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,790 @@
...
+    foreach file [glob -nocomplain -directory $dir $pattern] {
+        lappend result $file
+    }
+    foreach subdir [glob -nocomplain -directory $dir *] {
+        if {[file isdirectory $subdir]} {
+            set sub_results [recursive_glob $subdir $pattern]
+            set result [concat $result $sub_results]
```
m9710797: Appending one list to another list is better done via:
```
lappend result {*}$sub_results
```
You could even directly do:
```
lappend result {*}[recursive_glob $subdir $pattern]
```

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,790 @@
...
+        }
+        set ${arrayname}($filename) $path
+    }
+}
+
+proc sdcdb_add {param {path {}}} {
+    variable initialized
+    if {$initialized} {
+        error "SDCDB already initialized"
+    }
```
m9710797: Is this correct? Isn't it allowed to add more paths after `sdcdb open`? IOW should all paths be added before the `.cdb` file is opened?

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,790 @@
...
+    add_files_to_database c_files $new_files
+    set new_files [$function $path *[ASM]]
+    warn "adding files: [join $new_files {, }]"
+    add_files_to_database a_files $new_files
+}
+
+proc sdcdb_list {args} {
+    set pattern1 {([^:]+):(\d+)(-(\d+))?}
+    set pattern2 {([^:]+):(\S+)}
+    set pattern3 {(\S+)}
```
m9710797, line 459: A comment explaining what the regexp is supposed to match (e.g. an example) could be useful.

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,790 @@
...
+    if {[file extension $file] eq ".c"} {
+        variable c_files
+        set record [array get c_files $file]
+        if {$record eq {}} {
+            if {$showerror} {
+                error "file '$file' not found in database, add a directory that contains such file with 'sdcdb add <dir>'"
+            } else {
+                output "$file: not found"
+            }
+            return
```
m9710797: Just a suggestion (for lines 481-488, but can be applied elsewhere as well):
```
if {$record eq {}} {
    if {$showerror} {
        error "file '$file' not found in database, add a directory that contains such file with 'sdcdb add <dir>'"
    }
    return "$file: not found"
}
```
* Simplified the control flow (drop `else` after `error`).
* Replaced `output` with `return`.

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,790 @@
...
+    # 10 lines by default
+    if {$end eq {}} { set end [expr $start + 9] }
+    while {[gets $fh line] >= 0} {
+        incr pos
+        if {$pos >= $start} {
+            if {$pos == $focus} {
+                output "[format %-5d $pos]*    $line"
+            } else {
+                output "[format %-5d $pos]:    $line"
+            }
```
m9710797: I was going to write: "isn't the then and else branch identical?". So it took me a while to spot the difference (`*` vs `:`). Maybe it's clearer to write it like:
```
output "[format %-5d $pos][expr {$pos == $focus ? "*" : ":"}]    $line"
```
(And in an earlier comment I suggested to replace `output` by returning a string.)

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,790 @@
...
+        }
+        if {$pos >= $end} { break }
+    }
+    close $fh
+}
+
+proc list_pc {{x0 -1} {x1 9}} {
+    variable addr2file
+    if {[array get addr2file [reg PC]] ne {}} {
```
m9710797: Would it be useful to allow to list an arbitrary address, not just the `PC`?
A proc like `list_pc` is fine of course, but maybe it should call a more general `list_addr` proc?

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,790 @@
...
+        debug break  ;# makes no sense to list a running program
```
m9710797: I'm not sure I agree with this. If the users requests that (list while running), then why not?

I'm also thinking about making this script (re-)usable from other Tcl procs (from within a larger Tcl proc).
You never know what crazy use cases our users come up with it. But making this policy decision (always break on list) limits the re-usability.

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,790 @@
...
+    variable addr2file
+    if {[array get addr2file $begin] ne {}} {
+        lassign $addr2file($begin) file begin
+        lassign $addr2file($end)   {}   end
+        if {$file ne $filename} {
+            error "function database '$funcname' not found in '$filename'"
+        }
+        return [list_file $file [expr $begin - 1] $end -1 1]
+    }
+    error "function database '$funcname' not found"
```
m9710797: Suggestion for coding style. Feel free to ignore if you don't like it. (That's applies to all my review comments of course).

Currently you have:
```
if {[array ...] ne {}} {
   ... a lot of code ...
   return [list_file ...]
}
error "... not found"
```
An alternative is:
```
if {[array ...] eq {}} {
    error "... not found"
}
... a lot of code ...
list_file ...
```
* This keeps the check and the corresponding error message close together
* The "main" code path is one level less indented.
* It allows to drop the final `return`, making the Tcl code more "idiomatic".

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,790 @@
...
+}
+
+# scan file database then global database for function name
+proc search_fun {filename funcname} {
+    set record {}
+    if {$filename ne {}} {
+        set arrayname [file rootname $filename]_func2addr
+        variable $arrayname
+        if {[info exists $arrayname] && [info exists $arrayname($funcname)]} {
```
m9710797: Can this be simplified to `if {[info exists $arrayname($funcname)]}`?
(I haven't tested this myself).

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,790 @@
...
+        # not found in file, search globally
+        variable g_func2addr
+        set record [lindex [array get g_func2addr $funcname] 1]
+    }
+    if {$record eq {}} {
+        error "file database for '$funcname' not found"
+    }
+    set result $record
+    return [eval list $result]
+}
```
m9710797: I don't understand this `eval` stuff. Isn't this identical to `return $record`? There's probably some list-escaping stuff that I'm missing??

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,790 @@
...
+            variable $arrayname
+            if {[info exists ${arrayname}($funcname)]} {
+                set record [lindex [array get $arrayname $line] 1]
+                break
+            }
+        }
+    }
+    set result $record
+    return [eval list $result]
+}
```
m9710797, line 569: This proc isn't used currently. Is it a left over? Can it be removed? Do you want to keep it for future development?

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,790 @@
...
+    set match [regexp -inline $pattern1 $pos]
+    if {[llength $match] == 3} {
+        lassign $match {} filename linenum
+        set record [search_file_array $filename $linenum]
+        if {$record ne {}} {
+            debug breakpoint create -address [lindex $record 1] -condition $cond -command $cmd
+            return
+        } else {
+            error "address not found"
+        }
```
m9710797: This is again a pattern where switching the then and else branch can simplify the code. (If you prefer that).

--

An idea for the future. I'm thinking about making this functionality reusable from other Tcl scripts:
* Currently `sdcdb_break` looks up the address and immediately creates an openMSX breakpoint. That's exactly what you want if you run `sdcdb break` interactively in the console.
* I can imagine that some Tcl proc only wants to translate a function and/or line number into an address, but then do something else with that address. So maybe be can split this function in two? And expose an helper function (meant for scripts, not necessarily for interactive use in the console).

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,790 @@
...
+
+proc find_source {address} {
+    variable addr2file
+    # get current source file
+    set record [lindex [array get addr2file [reg PC]] 1]
+    if {$record ne {}} {
+        warn "record: 0x[format %04X [reg PC]] -> $record"
+        return $record
+    }
+    return {}
```
m9710797: Detail, this can be just `return $record` (merge the 2 `return` statements into 1).

(If we don't need the `warn` then this proc simplifies a lot, but i understand it's useful for debugging).

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,790 @@
...
+        catch {unset addr2file}
+        variable g_func2addr
+        catch {unset g_func2addr}
+        variable cond
+        if {$cond ne {}} {
+            debug condition remove $cond
+            set cond {}
+        }
+        set initialized 0
+        output "done."
```
m9710797: Detail: output like this: "done." is not very useful. Not every command must print something. The user already sees when the command is done. And output like this sometimes makes it harder to reuse the functionality from within a larger script.

m9710797 (commented):

I added lots of small review remarks. Not all _must_ be addressed, some are only suggestions or questions.

share/scripts/sdcdb.tcl
pvmm: As it is right now, it is not because I process some stuff in read_cdb.

share/scripts/sdcdb.tcl
pvmm: Yes, in some building processes, if you rename the project name (binary name) and rerun it.

pvmm:
Changes:
* `sdcdb next` is working now;
* more efficient `sdcdb step` because it doesn't rely on a condition any more;
* `sdcdb map` function returns the memory position of file:line and functions just like a dry-run version of `sdcdb break`;
* `sdcdb laddr` (list address) returns source code from any region of memory;
* all the fixes from the raised issues (I guess);

m9710797:
Thanks for the updates!

I still have a question / suggestion to simplify the usage:
* Currently you must call `sdcdb add` one or more times, followed by `sdcdb open`. (Because of this order it might be better to also list `add` before `open` in the help text).
* Before you can call `add` or `open` again, you must first call `sdcdb quit`.  (Would `close` be a more logical name?)

Here's my proposal:
* Remove `sdcdb add`.
* Change `sdcdb open` so that it takes a variable number of arguments:
** If an argument is a directory it acts like the old `add` command.
** Make recursive scanning the default (it's probably fine to not have a non-recursive version?).
** We automatically scan all given directories for `.cdb` files. If there's exactly one such file, the user doesn't need to do anything else (we'll automatically use that as if it was specified on the old `open` command).
** Next to directories it's also possible to specify a `.cdb` file (I propose that we no NOT automatically add a missing `.cdb` file extension, that makes tab-completion for this `open` command simpler). If specified this overrules the automatic `.cdb` file scanning from the previous point (thus it's not an error anymore if more than one `.cdb` file is present).
* Remove the `sdcdb quit` (or `close`) command. It's no longer needed. To switch to a new project simply issue a new `open` command. (Explicitly freeing memory is probably not a useful/required feature on modern host systems).

So for a simple project, with all .c files within a single directory tree (subdirs are allowed), and only a single .cdb file. The usage becomes just: `sdcdb open <myProjectDir>`.

What do you think?

pvmm:
> Remove the `sdcdb quit` (or `close`) command. It's no longer needed. To switch to a new project simply issue a new `open` command. (Explicitly freeing memory is probably not a useful/required feature on modern host systems).

Agree. I guess it made more sense when I had to create a condition before in a previous version.

pvmm:
> ** Make recursive scanning the default (it's probably fine to not have a non-recursive version?).

I think it is necessary. There are some frameworks (like the popular MSXgl) that put the library source a level below project source code, with multiple projects at the same level. It can even affect the script negatively if the unnecessary files are replacing the correct ones because they have the same names.

pvmm:
For the reason above, I kept it and renamed it to `sdcdb free` because it is important for a new session to free the previous file references.

m9710797:
> For the reason above, I kept it and renamed it to `sdcdb free` because it is important for a new session to free the previous file references.

What I meant is: we can keep `free` as an internal helper function. But it will be called automatically at the start of a `open` command. There's no reason to expose it to the user.

pvmm:
OK, I removed it. Only change this time.

m9710797:
> > ** Make recursive scanning the default (it's probably fine to not have a non-recursive version?).
> 
> I think it is necessary ...

Do you mean it's necessary to (also) have a non-recursive version? Can you explain your example (using MSXgl) in more detail? Is it really not possible to solve this by specifying (multiple) recursive directories and possibly a .cdb file from a different location.

pvmm:
This is the file structure of MSXgl:
```
root (MSXgl)
├── engine
│   ├── (...)
│   └── src
├── projects
│   ├── my_current_project
│   ├── example
│   ├── samples
│   ├── shmup-project
├── tools
│   ├── (...)
│   └── sdcc
└── (...)
```
From your project path (`my_current_project`), if you just include `../..` you will scan the sources of all other projects in `projects`, including `example`, `samples` and `tools`, which is not MSX code, but x86 code.

OK, so you can include `../../engine` to make sure it doesn't recursively scan for `../../tools/sdcc` for instance, but you could have a specific setting (which MSXgl supports) that puts MSXgl inside your project instead (and I actually use it), like this:
```
my_project
├── arkanoid.c
├── arkanoid.h
├── build.sh
├── debug.c
├── debugdevice.tcl
├── emul
│   └── rom
│       └── arkanoid.rom
├── Makefile
├── msxgl (all MSXgl stuff here)
│   ├── engine
│   ├── tools
│   │   └── (...)
│   └── (...)
├── platform_defines.h
├── project_config.js
├── README.md
├── sprites.c
├── sprites.h
└── tools
    ├── converter.c
    ├── COPYING
    ├── Makefile
    └── README.md
```
Now you have MSX source code in root and MSXgl in a directory. You cannot add `arkanoid.c` without adding all of MSXgl, including the x86 tools. So I think a non recursive version of glob is kinda obligatory at this point.

m9710797:
Ok, thanks for clarifying.

One possibility is to also have an option to exclude a subtree. But I don't know if that's easier.
So let's keep it simple: let's not add an exclude option and let's not make recursive the default.

pvmm:
I changed from recursive by default to non recursive by default.

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,795 @@
...
+        flush $chan
+    }
+}
+
+proc sdcdb {args} {
+    if {[catch {set result [dispatcher {*}$args]} msg]} {
+        variable debug
+        if {$debug} {
+            puts stderr $::errorInfo
+        }
```
m9710797, line 160: Detail: can be simplified to `debug_out $::errorInfo`.

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,795 @@
...
+    variable c_files
+    variable a_files
+    puts "[array size c_files] C files added"
+    puts "[array size a_files] assembly files added"
+    puts "[array size c_files_count] C files references added"
+    puts "[array size a_files_count] assembly files references added"
+    puts "$c_count C source lines found"
+    puts "$a_count assembly source lines found"
+    puts "$gf_count global function(s) registered"
+    puts "$sf_count static function(s) registered"
```
m9710797, line 370: Question: should these remain `puts` or should these be returned as a (multi-line) string?

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,795 @@
...
+        default { error "Unknown command \"[lindex $args 0]\"." }
+    }
+}
+
+proc sdcdb_open {param args} {
+    set function normal_glob
+    variable initialized
+    if {$initialized} {
+        free
+    }
```
m9710797, line 195: Detail: `free` already does a check for `initialized`.

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,795 @@
...
+    variable a_files
+    foreach filename [array names a_files] {
+        set arrayname a_[file rootname $filename]2addr
+        variable $arrayname
+        fix_last_address $arrayname
+        fix_blank_spaces $arrayname
+    }
+    fix_blank_spaces addr2file
+    return
```
m9710797, line 422: You resolved this comment, but then forgot to remove the (useless) `return`.

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,795 @@
...
+        error "Unknown or unspecified file extension"
+    }
+    variable $files
+    set record [array get $files $file]
+    if {$record eq {}} {
+        if {$showerror} {
+            error "file '$file' not found in database, add a directory that contains such file with 'sdcdb add <dir>'"
+        }
+        puts "$file: not found"
+        return
```
m9710797, line 494: General question: should all `puts` be replaced by return? Both will (by default) end up in the console, but if desired a script can capture the output if it's returned as a string, not when it's directly outputted via `puts`.

So maybe change this to `return "$file not found"`?

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,795 @@
...
+    }
+    debug_out "opening file [lindex $record 1]..."
+    set fh [open [lindex $record 1] r]
+    set pos 0
+    # 10 lines by default
+    if {$end eq {}} { set end [expr {$begin + 9}] }
+    while {[gets $fh line] >= 0} {
+        incr pos
+        if {$pos >= $begin} {
+            puts "[format %-5d $pos][expr {$pos == $focus ? "*" : ":"}]    $line"
```
m9710797, line 504: Same here, should these be accumulated in a multi-line string, and then returned?

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,795 @@
...
+}
+
+proc list_address {address {x0 -1} {x1 9} {showerror 0}} {
+    variable addr2file
+    set address [expr {$address}]
+    if {![info exists addr2file($address)]} {
+        if {$showerror} {
+            error "database address not found for 0x[h $address]"
+        }
+        return [puts "no file"]
```
m9710797, line 522: Change to just return (drop the `puts`)?

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,795 @@
...
+    debug breakpoint create -address $start -condition $cond -command $cmd
+}
+
+proc check {} {
+    variable old_PC
+    return [reg PC] != $old_PC
+}
+
+proc print_status {file line} {
+    puts "file: $file:$line, position: 0x[h [reg PC]]"
```
m9710797, line 665: This is less clear. This command is meant to be used interactively (less useful to be used by a script). But who knows what crazy ideas our users have for scripts. Maybe it is useful after all to also capture this output in a script. And there's also no real downside to returning vs printing. Both end up in the console by default.

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,795 @@
...
+    # get current position in source code
+    set record [find_source [reg PC]]
+    lassign $record file line
+    lassign [search_file $file $line] begin end
+    debug_out "search_file: $file, $line -> [search_file $file $line]"
+    if {$begin ne {} && $end ne {}} {
+        variable current_file $file
+        variable current_line $line
+    } else {
+        puts "Possibly out of scope, skipping till we get back."
```
m9710797, line 706: Same here.

share/scripts/sdcdb.tcl
```diff
@@ -0,0 +1,795 @@
...
+    }
+    print_status $file $line
+}
+
+proc sdcdb_info {arg} {
+    if {$arg eq "-break"} {
+        debug break
+    }
+    if {[catch {print_info} fid]} {
+        puts $fid
```
m9710797, line 740: And also here?
BTW this only converts an error message into a non-error return string, is that right?

m9710797 (commented):
I did a second full review. Only some details remain. But I'll already merge this. (If needed) we can still fix these details later.

share/scripts/sdcdb.tcl
pvmm: It can return instead of print as long as `update_set` prints it, since `update_step` runs in background and cannot return anything.

m9710797: Good point, I missed that.

And such functions are not meaningful to be called from other scripts anyway, So directly using `puts` here seems appropriate.

PR #1856, opened 2025-02-11
WIP: D88 add.

polysoma:


src/fdc/D88DiskImage.cc
```diff
@@ -0,0 +1,350 @@
...
+static constexpr uint8_t D88_STATUS_NO_ADDRESS_MARK = 0xE0;
+static constexpr uint8_t D88_STATUS_NO_DATA_MARK = 0xF0;
+
+struct D88Header
+{
+  std::array<uint8_t, 17> diskName;
+  std::array<uint8_t, 9> reserved;
+  uint8_t writeProtected;
+  uint8_t diskType;
+  int32_t diskSize;
```
m9710797, line 30: Detail.
Big endian machines are becoming rare, though in principle openMSX still supports them.
I assume this `diskSize` field is stored on disk in little-endian format? So that will go wrong if we read this on a big endian machine.

We have an easy fix for this in openMSX: `Endian::L32` (defined in `src/utils/endian.hh`). See `src/sound/WavData.hh:72` for an example). So replace `int32_t` with `Endian::L32`, that's all. After that reading `diskSize` on a little-endian machine acts exactly the same as reading a regular `int32_t`. But on a big-endian machine, it will also swap the bytes.

Note: reading a `Endian::L32` actually gives you a `uint32_t` (not `int32_t`). But I think using an unsigned value is better here?

--

This comment also applies to all other non-`uint8_t` types in the structs below. And to the `trackOffsets` member variable.

src/fdc/D88DiskImage.cc
```diff
@@ -0,0 +1,350 @@
...
+		prevTrackOffs = trackOffs;
+    }
+    for(int32_t i=160; i<D88_TRACK_MAX; ++i)
+    {
+        if (trackOffsets[i] != 0)
+            return false;
+    }
+
+	return std::ranges::all_of(header.reserved, [](auto& r) { return r == 0; });
+}
```
m9710797, line 75: Should we also check that `diskSize == file.size()`?
(Because openMSX doesn't support disk images with multiple disks in the same file.)

src/fdc/D88DiskImage.cc
```diff
@@ -0,0 +1,350 @@
...
+D88DiskImage::D88DiskImage(const Filename& filename, std::shared_ptr<File> file_)
+    : Disk(DiskName(std::move(filename)))
+    , file(std::move(file_))
+	, trackData(8192)
+	, cachedTrackNo(-1)
+{
+    D88Header header;
+    file->seek(0);
+    file->read(std::span{&header, 1});
+    file->read(std::span{&trackOffsets, 1});
```
m9710797, line 86: If I read the [D88 File Format documentation](https://www.pc98.org/project/doc/d88.html) correctly, then older D88 disk images may only contain 160 track offsets (not 164). Should we support those? For me it's fine to not support those to keep the code simpler.

src/fdc/D88DiskImage.cc
```diff
@@ -0,0 +1,350 @@
...
+	if (prevTrackOffs != sizeof(D88Header) + 4 * D88_TRACK_MAX)
+		return false;
+    for(int32_t i=1; i<160; ++i)
+    {
+        auto trackOffs = trackOffsets[i];
+        auto trackSize = trackOffs - prevTrackOffs;
+        if (trackSize <= 128 || trackSize >= 0x4000)
+            return false;
+		prevTrackOffs = trackOffs;
+    }
```
m9710797, line 67: Is there a missing check for the last track?
Something like `diskSize - prevTrack` is between 128 and 0x4000.

src/fdc/D88DiskImage.cc
```diff
@@ -0,0 +1,350 @@
...
+        nextTrackOffs = diskSize;
+    auto trackSize = nextTrackOffs - trackOffs;
+
+    if (trackOffs==0 || trackSize<1)
+        return false;
+
+	if (trackData.size() < trackSize)
+	{
+		trackData.resize(trackSize);
+	}
```
m9710797, line 128: I think we always need to do `trackData.resize(trackSize)`, not only when `trackData.size() < trackSize`. (So also shrink instead of only grow).

The reason is that otherwise `file->read(std::span{trackData});` below may read more than `trackSize` number of bytes.

src/fdc/D88DiskImage.cc
```diff
@@ -0,0 +1,350 @@
...
+	trackSectors.clear();
+
+	/* Read TrackData */
+    file->seek(trackOffs);
+	file->read(std::span{ trackData });
+
+    uint16_t inTrackOffset = 0;
+    while(inTrackOffset < trackSize)
+    {
+		D88SectorHeader* pHeader = (D88SectorHeader*)(trackData.data() + inTrackOffset);
```
m9710797, line 138: We should protect (better) against corrupt D88 disk images:
* It could be that `inTrackOffset < trackSize` (so we enter this `while` loop), but `inTrackOffset + sizeof(D88Sectorheader) > trackSize`. (There's not enough room left for a D88SectorHeader).
* If that's the case then the code below (e.g. `pHeader->sectorDataSize`) will read past the end of the `trackData` buffer.

src/fdc/D88DiskImage.cc
```diff
@@ -0,0 +1,350 @@
...
+		int nn = pHeader->n & 3;
+		sector.c = pHeader->c;
+		sector.h = pHeader->h;
+		sector.r = pHeader->r;
+		sector.n = pHeader->n;
+		sector.deleteFlag = pHeader->deleteFlag;
+		sector.status = pHeader->status;
+		sector.sectorSize = 128 << nn;
+		sector.dataOffset = inTrackOffset + 16;
+		sector.dataSize = pHeader->sectorDataSize;
```
m9710797, line 150: And here we should also protect better:
* It could be that `inTrackOffset + 16 + sector.dataSize > trackSize`.
* If that's the case then later in `readTrack()`, `trackDataSpan.subspan(sct.dataOffset, sct.dataSize)` will be past the end of the `trackData` buffer.

--

The [D88 documentation](https://www.pc98.org/project/doc/d88.html) suggests to NOT use the `sectorDataSize` field (because too many D88 images have a corrupt value for that field). Instead it should be calculated like you do above for `sectorSize`.

src/fdc/D88DiskImage.cc
```diff
@@ -0,0 +1,350 @@
...
+	write( 1, 0xFC); //            (2)
+	write(50, 0x4E); // gap1
+
+	// sectors
+	for(const auto& sct: trackSectors)
+	{
+		if (sct.status == D88_STATUS_NO_ADDRESS_MARK)
+		{
+			continue;
+		}
```
m9710797, line 230: Above in the calculation of `trackLen` you assume that all sectors contribute `62 + 84 + sct.sectorSize` bytes. But because of this `continue` that's not correct.
* Maybe we should instead write 0x4E bytes???
* Or maybe this is fine??? Then there will be a larger gap at the end of the track ... though with uninitialized data ... that's not good.

--

There's a similar problem below, with the `continue` on "NO_DATA_MARK".

src/fdc/D88DiskImage.cc
```diff
@@ -0,0 +1,350 @@
...
+		if (sct.status == D88_STATUS_DATA_CRC_ERROR)
+		{
+			dataCrc = ~dataCrc;
+		}
+		output.write(idx++, narrow_cast<uint8_t>(dataCrc >> 8));   // CRC (high byte)
+		output.write(idx++, narrow_cast<uint8_t>(dataCrc & 0xff)); //     (low  byte)
+
+		write(84, 0x4E); // gap3
+	}
+	//write(182, 0x4E); // gap4b
```
m9710797, line 278: I think it's important to also write these last 182 bytes. We should not leave any uninitialized data in the track.
* With normal sector read commands you cannot read these bytes.
* But you can with the WD2793 "Read Track" command.

src/fdc/D88DiskImage.cc
```diff
@@ -0,0 +1,350 @@
...
+
+size_t D88DiskImage::getNbSectorsImpl() const
+{
+	size_t t = singleSided ? 80 : (2 * 80);
+	return t * (sectorsPerTrack == 8 ? 8 : 9);
+}
+
+bool D88DiskImage::isWriteProtectedImpl() const
+{
+	return writeProtected || file->isReadOnly();
```
m9710797, line 335: In the current implementation writing is not supported. That's fine.
But then this method should always return `true`.

src/fdc/D88DiskImage.cc
```diff
@@ -0,0 +1,350 @@
...
+	// } else {
+	// 	throw NoSuchSectorException("Sector not found");
+	// }
+
+	// writeTrack(track, side, rawTrack);
+}
+
+size_t D88DiskImage::getNbSectorsImpl() const
+{
+	size_t t = singleSided ? 80 : (2 * 80);
```
m9710797, line 329: Very much a detail:
Maybe it's clearer to write this as: `80 * (singleSided ? 1 : 2)`

src/fdc/D88DiskImage.cc
```diff
@@ -0,0 +1,350 @@
...
+}
+
+Sha1Sum D88DiskImage::getSha1SumImpl(FilePool& filePool)
+{
+	return filePool.getSha1Sum(*file);
+}
+
+void D88DiskImage::detectGeometryFallback()
+{
+	setSectorsPerTrack(9);
```
m9710797, line 345: * Above in `getNbSectorsImpl()` we set `sectorsPerTrack == 8 ? 8 : 9`.
* But here we always set `9`.

I don't know which of the two alternatives is best. But I think it should be consistent.

m9710797 (changes requested):
Thanks!

In general the code is looking good. But there are a few small bugs. And I also made some (small) suggestions for improvements.

Note that I have no experience with D88 disk images. So it's possible that some of my review comments are just wrong.

m9710797:
Thanks a lot of this PR!
I already reviewed your code. But I'd also like to discuss the non-technical part (non c++ code related).

First, can you give me some background info on the D88 file format:
* I found [this](https://www.pc98.org/project/doc/d88.html) documentation, is that correct?
* I've never seen D88 disk images used in the context of MSX. What's your motivation for wanting to support them in openMSX?
* Actually before today, I hadn't heard of D88 disk images at all. Can you point me to some (MSX related) D88 images, so that I can do some experiments?

There exist dozens (maybe hundreds) of different disk image file formats. But in openMSX we _intentionally_ only support a few formats:
* Raw disk images (.dsk, .di1, .di2, .fd1, .fd2). Because there are VERY common.
* DMK disk images (.dmk). For copy protected disks.
* XSA disk images (.xsa). These were added +20 years ago, but I think that was a mistake. We intent to drop these in one of the following releases.
* In the `Contrib` subdirectory we ship tools (source code) to convert between various non-directly-supported disk image formats, for example `der2dsk` (the blueMSX DiskERror file format), `svi2dmk` and `svicpm2dmk` (some svi related formats), `xsa2dsk` (because we want to drop XSA support).

I believe DMK is a strict superset of D88, that is:
* All D88 images can be converted to DMK without loss of information.
* But not all DMK images can be converted to D88.

So my personal preference is to NOT support D88 directly in openMSX. Instead we can add an external tool `d88_to_dmk` to convert D88 disk images to DMK format. (Bonus: we do already have support for writing to DMK images). 90% of the code you wrote for this PR is directly usable for such a `d88_to_dmk` tool.

But that's just my current idea. You can of course still try to convince me otherwise :)

polysoma:
sorry i misssed

polysoma:
It was an operational mistake.
I didn't intend to make it public as it was just a personal test.

m9710797:
Ah OK, no problem.
I hope my review comments are useful for you anyway :-)

PR #1857, opened 2025-02-14
remaining bugs found in sdcdb debugger

pvmm:
Remaining bugs and conflicts from #1854

### DIFF
--- a/share/scripts/sdcdb.tcl
+++ b/share/scripts/sdcdb.tcl
@@ -69,6 +69,7 @@ variable current_line   {}              ;# debugger line
 variable cond           {}              ;# breakpoint check condition
 variable old_PC         {}
 variable times_left     0
+variable out_of_scope   0               ;# running without debug info?
 proc ASM {} { return ".asm" }
 
 set_help_proc sdcdb [namespace code sdcdb_help]
@@ -154,10 +155,7 @@ proc debug_out {args} {
 
 proc sdcdb {args} {
     if {[catch {set result [dispatcher {*}$args]} msg]} {
-        variable debug
-        if {$debug} {
-            puts stderr $::errorInfo
-        }
+        debug_out stderr $::errorInfo
         error $msg
     }
     return $result
@@ -188,11 +186,8 @@ proc dispatcher {args} {
 }
 
 proc sdcdb_open {param args} {
+    free
     set function normal_glob
-    variable initialized
-    if {$initialized} {
-        free
-    }
     if {$param eq "-recursive"} {
         set function recursive_glob
     } else {
@@ -223,9 +218,11 @@ proc sdcdb_open {param args} {
         }
         set cdb_file $cdb_files
     }
+    variable initialized
     set initialized 1
-    read_cdb $cdb_file
+    set result [read_cdb $cdb_file]
     process_data
+    return $result
 }
 
 proc complete {arrayname name list} {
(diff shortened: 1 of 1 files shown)


share/scripts/sdcdb.tcl
```diff
@@ -156,7 +157,7 @@ proc sdcdb {args} {
     if {[catch {set result [dispatcher {*}$args]} msg]} {
         variable debug
         if {$debug} {
-            puts stderr $::errorInfo
+            debug_out stderr $::errorInfo
```
m9710797: What I meant:
you can leave out `variable debug` and `if ${debug} {`, that's the responsibility of `debug_out`.

share/scripts/sdcdb.tcl
```diff
@@ -360,14 +360,16 @@ proc read_cdb {fname} {
...
+    return [subst [join {
+        "[array size c_files] C files added"
+        "[array size a_files] assembly files added"
+        "[array size c_files_count] C files references added"
+        "[array size a_files_count] assembly files references added"
+        "$c_count C source lines found"
+        "$a_count assembly source lines found"
+        "$gf_count global function(s) registered"
+        "$sf_count static function(s) registered"
+    } "\n" ]]
```
m9710797: This is perfectly fine. No need to change (I mean pick what you like better).
My idea was to write it like:
```
set result ""
append result "....\n"
append result "....\n"
return $result
```

share/scripts/sdcdb.tcl
```diff
@@ -490,22 +491,23 @@ proc list_file {file begin {end {}} {focus 0} {showerror 0}} {
...
     while {[gets $fh line] >= 0} {
         incr pos
         if {$pos >= $begin} {
-            puts "[format %-5d $pos][expr {$pos == $focus ? "*" : ":"}]    $line"
+            lappend output "[format %-5d $pos][expr {$pos == $focus ? "*" : ":"}]    $line"
         }
         if {$pos >= $end} { break }
     }
     close $fh
+    join $output "\n"
```
m9710797: This is also fine (creating a list, and then in the end join that list with `\n`).
Above I used `append output "...\n"`.
Pick which you prefer best.

share/scripts/sdcdb.tcl
```diff
@@ -515,13 +517,14 @@ proc list_pc {{x0 -1} {x1 9}} {
...
     variable addr2file
     set address [expr {$address}]
-    if {![info exists addr2file($address)]} {
+    set record [lindex [array get addr2file $address] 1]
+    if {$record eq {}} {
         if {$showerror} {
             error "database address not found for 0x[h $address]"
         }
-        return [puts "no file"]
+        return "no file for address 0x[h $address]"
```
m9710797: Detail: currently the string is different for the `showerror` = `false` vs `true` case. Or maybe that's intentional?
Would it be better to calculate the common message into a variable, and then either `error` or `return` that string?

share/scripts/sdcdb.tcl
```diff
@@ -729,16 +729,22 @@ proc print_info {} {
...
+    if {$param eq "-break"} {
         debug break
     }
-    if {[catch {print_info} fid]} {
-        puts $fid
+    set result {}
+    if {[catch {set result [check_info]} msg]} {
+        variable debug
+        if {$debug} {
+            debug_out stderr $::errorInfo
```
m9710797: The check for `debug` is already part of `debug_out`.

(Maybe something to check (I'm not 100% sure it's true): `variable debug` should _only_ be used in the proc `debug_out`).

Issue #1284, opened 2020-08-27
hankaku/zenkaku language input key causes [ keypress


MBilderbeek:
Takamichi wrote on MRC:
> Another thing that has been since before 0.16: Pushing the hankaku/zenkaku language input key when openMSX is running causes [ character to be entered forever like [[[[[[[[[[. Pushing the language input key again stops that. I know that key exists only on Japanese keyboard

In the SDL2 scancodes I see:
`    SDL_SCANCODE_LEFTBRACKET = 47,`
and in openMSX src/events/Keys.cc:
```
        // Handle keys that don't have a key code but do have a scan code.
        if (key == K_UNKNOWN) {
                // Assume it is a Japanese keyboard and check
                // scancode to recognize a few japanese
                // specific keys for which SDL does not have an
                // SDL_Keycode keysym definition.
                switch (sdlscancode) {
                // Keys found on Japanese keyboards:
                case 49:
                        key = K_ZENKAKU_HENKAKU;
                        break;
                case 129:
                        key = K_HENKAN_MODE;
                        break;
```
Note that SDL_scancode.h also contains:
```
    SDL_SCANCODE_LANG1 = 144, /**< Hangul/English toggle */
    SDL_SCANCODE_LANG2 = 145, /**< Hanja conversion */
    SDL_SCANCODE_LANG3 = 146, /**< Katakana */
    SDL_SCANCODE_LANG4 = 147, /**< Hiragana */
    SDL_SCANCODE_LANG5 = 148, /**< Zenkaku/Hankaku */
```

Anyway, I don't know what this all means, but it would be great to see which scancode and/or keycode is sent by SDL. So I asked this:
> I understand this is not what you would expect to happen when pressing this key. But, and please don't take things wrong, what would you expect to happen when pressing it?
> 
> Furthermore the following data is useful to know;
> - which MSX are you emulating when seeing this phenomenon?
> - what happens on another MSX type, e.g. European vs Japanese?
> - did this also happen with openMSX 0.15.0, or were you talking about developer builds?
> - which keyboard mode are you using? See Menu -> Misc settings -> Keyboard mapping mode
> - to debug in detail (none of the developers have a Japanese keyboard...) you can do this:
> 1. open the console with F10
> 2. type the command: set kbd_trace_key_presses on
> 3. press F10 to close the console
> 4. hit the hankaku/zenkaku key. After 10 seconds, hit it again.
[...]

PR #1859, opened 2025-02-16
[Windows] disable IME on main window (Japanese keyboard fix)

uniskie:
for issue: https://github.com/openMSX/openMSX/issues/1284

add setting "kbd_disable_IME" (default: true)

Disable IME in main window.
(Subwindows are not suppressed.)

Basically, if the IME is turned on, there will be problems with key input, so the default is kbd_disable_IME=false.

In case you want to use IME in menus etc. in the future, we have added an option to enable IME.


MBilderbeek:
@m9710797 any comments on this?
I'm a bit unhappy about such a very Windows specific solution in the middle of mostly very portable code...

PR #1861, opened 2025-02-16
Fixed a problem that occurred when multiple host keys were bound to one MSX key.

uniskie:
Fixed issues such as:

1. Press the left shift key and right shift key on the host at the same time.
2. Release only one side
3. MSX will be in a state where the shift key is released

* I don't think serialization is necessary because cntKeyMatrix adds a counter related to the host keyboard state.


MBilderbeek:
Thanks! I see what you mean, but how is it a problem, exactly? So, in which scenario does this introduce unexpected behavior?
Just to get an idea why you started to work on this.

uniskie:
- The reason I discovered this is this scenario:
   In the case of Japanese, if you use software that automatically inserts a SHIFT event to input "_" when you press the "\_ろ" key, the SHIFT key will be automatically released even if you are holding it down.

- Unique features of Japanese keyboard:
  1. If you press only the "\_ろ" key, "￥" will be entered or nothing will be entered.
  2. There is another "￥" key, and it is usually used to input "￥".

  There are a certain number of people who use software that helps them input "_" when they press the "\_ろ" key.

- Other cases:

  When using blind touch to use left and right shift, if there was a timing when they were pressed at the same time, unintended input could occur.
  This has been happening on rare occasions before, but I was not aware of the cause.

In cases where the key scan is implemented, such as in games or other emulators, the problem of discrepancy with the physical press state does not occur.
There may be a better solution, but I thought using a counter would be easier if it was event-driven.

uniskie:
If there are people who are using the previous specifications, it might be a good idea to enable/disable it using settings.

MBilderbeek:
Just one question: why the `#ifdef` stuff?

uniskie:
Oops, I forgot to delete ifdef.
I removed the ifdef and committed.

src/input/Keyboard.hh
```diff
@@ -261,6 +261,10 @@ private:
 	/** Combination of 'cmdKeyMatrix', 'typeKeyMatrix' and 'userKeyMatrix'. */
 	mutable std::array<uint8_t, KeyMatrixPosition::NUM_ROWS> keyMatrix;
 
+	/** keyboard matrix press counter for multi-binded key. */
+	std::array<std::array<int8_t, KeyMatrixPosition::NUM_COLS>, KeyMatrixPosition::NUM_ROWS> cntKeyMatrix;
+	// e.g.) LSHIFT and RSHIFT are assigned to the same SHIT in the MSX key matrix.
```
MBilderbeek: Funny typo here :)

MBilderbeek:
Oops, I completely forgot about this patch, sorry.

> Fixed issues such as:
> 
>     1. Press the left shift key and right shift key on the host at the same time.
> 
>     2. Release only one side
> 
>     3. MSX will be in a state where the shift key is released
> 
> 
>     * I don't think serialization is necessary because cntKeyMatrix adds a counter related to the host keyboard state.

I just tried this scenario with the current code (not including your patch) and this is not a problem. If I release one side, the shift is still pressed in the MSX. Can you try this scenario again, please?

uniskie:
It's natural that when you're in the middle of a big revision project and busy, you tend to forget things.

I tested it again.

1. While holding down LSHIFT, press RSHIFT as well.
2. Without releasing RSHIFT, release LSHIFT only.
3. Without releasing RSHIFT, press the [G] key.

- with openMSX 20.0
  A lowercase "g" will appear.

- with openmsx-20.0-401-g8be333ee4-windows-vc-x64-bin.zip
  A lowercase "g" will appear.
  
* correct behavior
  A capital "G" must appear.

MBilderbeek:
> It's natural that when you're in the middle of a big revision project and busy, you tend to forget things.
> 
> I tested it again.
> 
>     1. While holding down LSHIFT, press RSHIFT as well.
> 
>     2. Without releasing RSHIFT, release LSHIFT only.
> 
>     3. Without releasing RSHIFT, press the [G] key.
> 
> 
>     * with openMSX 20.0
>       A lowercase "g" will appear.
> 
>     * with openmsx-20.0-401-g8be333ee4-windows-vc-x64-bin.zip
>       A lowercase "g" will appear.
> 
> 
>     * correct behavior
>       A capital "G" must appear.

Aha, interesting! However, in my case, nothing appears. The G keypress is ignored when I try it!
But strangely enough, the function key display is still showing the SHIFTed position! The Virtual Keyboard doesn't even show the G is pressed. 

After applying your patch, I still see the same behaviour. So I'm not sure we're talking about the same thing here.

uniskie:
I forgot to mention an important premise. This problem occurs when Keyboard Mapping Mode is POSITIONAL (or KEY). 

(There seems to be a different problem when using CHARACTER mode, but unfortunately I was not able to fix CHARACTER mode to a practical level in the Japanese environment, so I am not using it now.)

MBilderbeek:
OK, I enabled POSITIONAL mode, but I see the same thing: no key is pressed when I press G in your described scenario, with your patch :(

uniskie:
> OK, I enabled POSITIONAL mode, but I see the same thing: no key is pressed when I press G in your described scenario, with your patch :(

That may be a different problem...

No matter which keyboard mapping mode I used, the "G not being typed" problem did not occur.

My environment:

- OS: Windows 10
- Language setting: Japanese
- Keyboard: Topre Realforce R3SA13

MBilderbeek:
@m9710797 this patch helps users with Japanese keyboards on Windows at least, but for me it has no effect. Can you help to review it, at least technically?

uniskie:
Due to the specifications of SDL key events, problems occur when the WINDOW focus is moved while a key is being held down, so this update is withdrawn for the time being.

Additional measures are required.

1. While holding down a key, focus is moved from openMSX to another application: KEYUP does not occur.

2. While holding down a key, focus is moved from another application to openMSX: KEYUP occurs without KEYDOWN.

2 is not so much of a problem, but in any case it is not the desired behavior.

> [!NOTE]
> In Windows game application development, to deal with such problems, you basically either capture (occupy) key events or perform key scanning with periodic polling.
>
> (I don't know the standard for SDL application development.)

MBilderbeek:
@m9710797 can you say a bit more on how things are dealt with, that @uniskie mentioned in his last comment?

m9710797:
@MBilderbeek I have no idea. As in: same as you or uniskie, I'd have to run experiments to find out. (But i don't plan to investigate this myself).

MBilderbeek:
OK, I had hoped you remembered something of this, but never mind then.

uniskie:
The symptoms and solutions for keyboard event problems may vary depending on the OS.

There may also be changes with the upgrade to SDL3.

There are more issues involved than I thought, so it's not a problem that can be resolved with just one person's testing or information.

I'll organize the information about the phenomenon itself and post it in the issue section.

Thank you for your continued support.

uniskie:
> Due to the specifications of SDL key events, problems occur when the WINDOW focus is moved while a key is being held down, so this update is withdrawn for the time being.
> 
> Additional measures are required.
> 
> 1. While holding down a key, focus is moved from openMSX to another application: KEYUP does not occur.
> 2. While holding down a key, focus is moved from another application to openMSX: KEYUP occurs without KEYDOWN.
> 
> 2 is not so much of a problem, but in any case it is not the desired behavior.

After rebooting with Windows Update, this problem no longer occurred.
(Before rebooting Windows, the keyevent bug occurred even in versions that did not have this fix applied.)

There may have been some other cause for the problem.
This fix doesn't seem to cause any problems, so I'll re-open it.

uniskie:
The fix in the pull request addresses a problem specific to POSITIONAL.

The use of POSITIONAL is required in a Japanese environment, so this fix is ​​also necessary.

> [!Note]
> For CHARACTER and KEY, this problem manifests when looking at the key matrix for SHIFT alone, but the problem is hidden when using SHIFT + another key.

uniskie:
I'm sorry for the confusion.

Thank you for your continued support.

uniskie:
I determined that this approach is not ideal, as it complicates processing and increases the likelihood of conflicts with the key history mechanism.

I believe the best solution to this issue is to implement special handling for the Left and Right Shift keys on the host side using a flag. Therefore, I am withdrawing this pull request.

PR #1862, opened 2025-02-16
fix CAHACTER mode key_up checker

uniskie:
There are keys that do not notify the keycode, so use scancode to determine key_up key.

### DIFF
--- a/src/input/Keyboard.cc
+++ b/src/input/Keyboard.cc
@@ -898,7 +898,7 @@ void Keyboard::syncHostKeyMatrix(EmuTime::param time)
 	}
 	msxModifiers = 0xff;
 	msxKeyEventQueue.clear();
-	lastUnicodeForKeycode.clear();
+	lastUnicodeForScancode.clear();
 }
 
 uint8_t Keyboard::needsLockToggle(const UnicodeKeymap::KeyInfo& keyInfo) const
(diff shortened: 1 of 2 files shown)


MBilderbeek:
Right, so you propose to use scancode instead of keycode, as sometimes the keycode is not reported even for PRESSED events. Right? @m9710797 what do you think? It sounds like a good idea to me, but I'm not sure :)

uniskie:
> Right, so you propose to use scancode instead of keycode, as sometimes the keycode is not reported even for PRESSED events. Right?

Yes, that's right.

(If there is only one key that does not notify the keycode, like on a Japanese keyboard, the problem will not surface.)

m9710797:
Looks good to me.

src/input/Keyboard.cc
```diff
@@ -1194,7 +1194,8 @@ bool Keyboard::processKeyEvent(EmuTime::param time, bool down, const KeyEvent& k
 {
 	auto mode = keyboardSettings.getMappingMode();
 
-	auto keyCode  = keyEvent.getKeyCode();
+	auto keyCode = keyEvent.getKeyCode();
+	auto ScanCode = keyEvent.getScanCode();
```
MBilderbeek, line 1198: we prefer to start a local variable with lower case... but I don't mind to change this myself before merging if I find out how :)

MBilderbeek:
I merged it via  cca93f4a687090f788d77ff3954df1351b8ef6e4 with some minor edits. THANKS!

EDIT: Oops, I forgot to push the change. But it should be merged now.

PR #1863, opened 2025-02-17
Matching multiple files in SDCDB debugger

pvmm:
* now `sdcdb map` `list` and `break` may return multiple elements if they match. For instance: `sdcdb list myfunction` will return more than one function if for instance you have multiple C files each with a static function with the same name). This feature copies GDB behaviour.
* fixed a weird behaviour when a function has inline assembly it resets the C to address mapping returning the wrong position when you try to list those functions.


m9710797:
I didn't review all in detail. But at first sight it looked reasonable.

PR #1866, opened 2025-02-22
focus problem when SetCursorPos called without mnemonic labels

pvmm:
The Mnemonic column normally loses focus on hover where a popup should appear ("multiple possibilites, click to cycle..."), but if there are no mnemonic labels, there is no popup, so the column loses focus (SetCursorPos is called) for nothing.

### DIFF
--- a/src/imgui/ImGuiDisassembly.cc
+++ b/src/imgui/ImGuiDisassembly.cc
@@ -479,16 +479,11 @@ void ImGuiDisassembly::paint(MSXMotherBoard* motherBoard)
 								ImGui::TextUnformatted(mnemonic);
 							});
 							if (mnemonicAddr) {
-								ImGui::SetCursorPos(pos);
-								if (ImGui::InvisibleButton("##mnemonicButton", {-FLT_MIN, textSize})) {
-									if (!mnemonicLabels.empty()) {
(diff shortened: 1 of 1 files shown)


m9710797:
I agree with the code change, we shouldn't do unnecessary work, like create an invisible button with no function. So I've merged this PR. Thanks!

But what's not clear to me is the problem description. I don't see the connection between "losing focus" and "calling SetCursorPos()" (in this context "cursor" means the position where ImGui would draw the next widget, it's not related to the mouse cursor). Can you explain in more detail?

pvmm:
image
I found experimentally that because of the `SetCursorPos`, the line `ld hl,#0006` will not turn blue on mouseover (I have disabled `SetCursorPos` for this screenshot). So if there are no mnemonicLabels, we can avoid calling `SetCursorPos` to avoid this problem.

m9710797:
Thanks for clarifying...
... I also did an experiment. I left the call to `SetCursorPos()` and commented out the call to `InvisibleButton()`. That also fixes the problem. So I (now more strongly) believe that `SetCursorPos()` is harmless, and that instead `InvisibleButton()` was causing the problem.

Anyway, none of this changes anything to your PR. It's still a good improvement. Thanks again!

PR #1868, opened 2025-02-24
Toggles breakpoint in selected address

### LINKED ISSUE #1852: Operating the Disassembly widget by keyboard has issues
Tested using openMSX 20.0 on macOS 13.7.3:

Operating the Debugger Disassembly widget only by keyboard has some issues:

1) The selected assembly line doesn't stay selected after you click it. The only way to force it to stay selected is to press UP or DOWN on the keyboard after you clicked with the mouse.
2) Pressing the CMD+B (CTRL+B on Windows/Linux) to toggle a breakpoint operates on the current running line of code (highlighted in yellow), instead of the selected line of code (highlighted in blue). Because of this, it's not possible use the keyboard to toggle a series of breakpoints. The only way to do it is by using the mouse to click on the column "bp" of each desired line

pvmm:
Allow user to select a row by left-clicking on it in the Disassembly
widget until widget loses focus. The selected address changes with the
arrow keys. Pressing the breakpoint shortcut toggles breakpoints in the
selected address OR program counter if no address is selected.

This fixes #1852

### DIFF
--- a/src/imgui/ImGuiDisassembly.cc
+++ b/src/imgui/ImGuiDisassembly.cc
@@ -36,6 +36,8 @@ using namespace std::literals;
 
 namespace openmsx {
 
+ImGuiDisassembly* ImGuiDisassembly::lastWidget = nullptr;
+
 ImGuiDisassembly::ImGuiDisassembly(ImGuiManager& manager_, size_t index)
 	: ImGuiPart(manager_)
 	, symbolManager(manager.getReactor().getSymbolManager())
@@ -181,16 +183,16 @@ static void toggleBp(uint16_t addr, const BpLine& bpLine, std::span<BreakPoint> 
 }
 void ImGuiDisassembly::actionToggleBp(MSXMotherBoard& motherBoard)
 {
-	auto pc = motherBoard.getCPU().getRegisters().getPC();
 	auto& cpuInterface = motherBoard.getCPUInterface();
 	auto& debugger = motherBoard.getDebugger();
 	auto& bps = cpuInterface.getBreakPoints();
-
-	auto bpLine = examineBpLine(pc, bps, cpuInterface, debugger);
-
 	std::optional<BreakPoint> addBp;
 	std::optional<unsigned> removeBpId;
-	toggleBp(pc, bpLine, bps, cpuInterface, debugger, addBp, removeBpId);
+	auto addr = lastWidget && lastWidget->selectedAddr ? *lastWidget->selectedAddr : motherBoard.getCPU().getRegisters().getPC();
+
+	auto bpLine = examineBpLine(addr, bps, cpuInterface, debugger);
+
+	toggleBp(addr, bpLine, bps, cpuInterface, debugger, addBp, removeBpId);
 	if (addBp) {
 		cpuInterface.insertBreakPoint(std::move(*addBp));
 	}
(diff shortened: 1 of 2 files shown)


src/imgui/ImGuiDisassembly.cc
```diff
@@ -179,18 +179,25 @@ static void toggleBp(uint16_t addr, const BpLine& bpLine, std::span<BreakPoint>
 		addBp->setCondition(toTclExpression(getCurrentSlot(cpuInterface, debugger, addr)));
 	}
 }
+
+static std::optional<uint16_t> selectedAddr = {};
+
```
m9710797: This should be a member variable of the `ImGuiDisassembly` class. Not a global variable.

For example now if you have more than one disassembly view open at the same time, keyboard navigation doesn't work anymore (`selectedAddr` is always reset).

src/imgui/ImGuiDisassembly.cc
```diff
@@ -179,18 +179,25 @@ static void toggleBp(uint16_t addr, const BpLine& bpLine, std::span<BreakPoint>
...
 	std::optional<unsigned> removeBpId;
-	toggleBp(pc, bpLine, bps, cpuInterface, debugger, addBp, removeBpId);
+
+	if (selectedAddr) {
+		auto bpLine = examineBpLine(*selectedAddr, bps, cpuInterface, debugger);
+		toggleBp(*selectedAddr, bpLine, bps, cpuInterface, debugger, addBp, removeBpId);
+	} else {
+		auto bpLine = examineBpLine(pc, bps, cpuInterface, debugger);
+		toggleBp(pc, bpLine, bps, cpuInterface, debugger, addBp, removeBpId);
+	}
```
m9710797: This can be simplified to:
```
auto addr = selectedAddr? *selectedAddr : motherBoard.getCPU().getRegisters().getPC();
```
And then mostly the original code, with `pc` replaced by `addr`.

m9710797 (changes requested):
Thanks for your work! I finished the review: 1 bug, 1 suggestion for simplification.

src/imgui/ImGuiDebugger.cc
```diff
@@ -313,7 +313,9 @@ void ImGuiDebugger::checkShortcuts(MSXCPUInterface& cpuInterface, MSXMotherBoard
 	} else if (shortcuts.checkShortcut(DEBUGGER_STEP_BACK)) {
 		actionStepBack();
 	} else if (shortcuts.checkShortcut(DISASM_TOGGLE_BREAKPOINT)) {
-		ImGuiDisassembly::actionToggleBp(motherBoard);
+		for (auto& d : disassemblyViewers) {
+			d->actionToggleBp(motherBoard);
+		}
 	}
```
m9710797: UPDATE: I started typing this reply when the code still used `disassemblyViewers.back()`, later you changed that this this for-loop. So part of this reply is for this `.back()` approach, part is for the loop.

Hmm, indeed now it matters in which disassembly viewer we call `actionToggleBp()`.
We should not call it in all viewers. Because when there are an even number of viewers and none a row with keyboard focus, then we toggle a bp at the PC location an even number of times, so effectively doing nothing.

2 thing to keep in mind:
* It's possible `disassemblyViewers` is empty (no view has been opened yet since openMSX was started). Then calling `.back()` will crash. We should protect against that.
* This function`ImGuiDebugger::chekShortcuts()` can get called from 2 locations:
  * From `ImGuiDisassembly::paint()`, and then it would be logical to also use the `ImGuiDisassembly` object that did the call. So maybe `checkShortcuts()` should take an extra `ImGuiDisassembly*` pointer parameter? (And `ImGuiDisassembly::paint()` can then pass `this` as argument).
  * Or from `ImGuiDebugger::drawControl()`, but then it's not clear which `ImGuiDisassembly` object to pass. We can maybe pass `nullptr` and then in this function use `.back()` (the last opened view, like you do now)  but **only** when there is at least 1 element present.

PR #1876, opened 2025-03-13
Simplify parity calculation

seanyoung:
Rather than iterating over all the bits, use popcount. This should use the popcntl instruction on x86, and makes the code more readable, I think.


seanyoung:
I just spotted this is in `constexpr` so no runtime change.

PR #1877, opened 2025-03-13
When both laserdisc and cassette are present, autorun the cassette

seanyoung:
The MysteryDisc has both cassette and laserdisc. This means there is no program encoded on the right auto channel of the laserdisc and so CALLLD will not work.

### DIFF
--- a/src/laserdisc/LaserdiscPlayer.cc
+++ b/src/laserdisc/LaserdiscPlayer.cc
@@ -7,6 +7,7 @@
 #include "HardwareConfig.hh"
 #include "XMLElement.hh"
 #include "CassettePort.hh"
+#include "CassettePlayer.hh"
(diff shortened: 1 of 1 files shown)


src/laserdisc/LaserdiscPlayer.cc
```diff
@@ -679,6 +680,16 @@ void LaserdiscPlayer::autoRun()
 		return;
 	}
 
+	auto *player = motherBoard.getCassettePort().getCassettePlayer();
```
m9710797, line 683: In the general case `getCassettePlayer()` can return `nullptr`.
But I don't know if in this specific context that's possible. May I suggest to either:
* Change the line below to `if (player && player->...)`.
* Or add a `assert(player); // short reasoning for why this must be the case`.

PR #1879, opened 2025-03-23
Implemented a programmable device

pvmm:
The Programmable Device is based on Debug Device and allow developers to easily create prototypes on Tcl to access hardware ports. There are three TclCallbacks for read write and reset.

```
variable {Programmable Device ports} {6 7 242}  ;# read and write ports 6, 7 and 242.
variable {Programmable Device reset callback}  "reset"
variable {Programmable Device input callback}  "read_port"
variable {Programmable Device output callback} "write_port"
```

### DIFF
--- a/share/extensions/programmabledevice.xml
+++ b/share/extensions/programmabledevice.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" ?>
+<!DOCTYPE msxconfig SYSTEM 'msxconfig2.dtd'>
+<msxconfig>
+  <info>
+    <name>Programmable Device</name>
+    <manufacturer>PVM</manufacturer>
+    <code/>
+    <release_year>2025</release_year>
+    <description>Virtual device to read and write to specific hardware ports.</description>
+    <type>debug tool</type>
+  </info>
+  <devices>
+    <ProgrammableDevice id="Programmable Device">
+    </ProgrammableDevice>
+  </devices>
+</msxconfig>
--- a/src/ProgrammableDevice.cc
+++ b/src/ProgrammableDevice.cc
@@ -1,0 +1,109 @@
+#include "ProgrammableDevice.hh"
+
+#include "CommandController.hh"
+#include "MSXCliComm.hh"
+#include "MSXCPUInterface.hh"
+#include "MSXException.hh"
+#include "StringSetting.hh"
+#include "serialize.hh"
+
+#include "narrow.hh"
+
+namespace openmsx {
+
+static IterableBitSet<256> parseAllPorts(const Setting& setting);
+
+ProgrammableDevice::ProgrammableDevice(const DeviceConfig& config)
+	: MSXDevice(config)
+	, portListSetting(
+		getCommandController(), tmpStrCat(getName(), " ports"),
+		"List of ports that are added to the pool of I/O ports this extension scans for", "")
+	, resetCallback(getCommandController(),
+		getName() + " reset callback",
+		"Tcl proc to call when reset was detected, no parameters and return values are "
+		"used", "", Setting::Save::NO)
+	, inCallback(getCommandController(),
+		tmpStrCat(getName(), " input callback"),
+		"Defines a Tcl proc that receives a port number (byte) and returns a byte when "
+		"the MSX is reading from a port of the device", "", Setting::Save::NO)
+	, outCallback(getCommandController(),
+		tmpStrCat(getName(), " output callback"),
+		"Defines a Tcl proc that receives a port number (byte) and a byte when the "
+		"MSX is writing to a port from the device", "", Setting::Save::NO)
+{
+	portListSetting.attach(*this);
+	try {
+		outPorts = inPorts = parseAllPorts(portListSetting);
+	} catch (MSXException&) {
+		// silently ignore error in settings.xml
+		portListSetting.setString("");
+	}
+}
+
+ProgrammableDevice::~ProgrammableDevice()
+{
+	portListSetting.detach(*this);
+}
+
+static IterableBitSet<256> parseAllPorts(const Setting& setting)
+{
+	Interpreter& interp = setting.getInterpreter();
+	TclObject portList = setting.getValue();
+	IterableBitSet<256> ports;
+
+	for (unsigned n = 0; n < portList.getListLength(interp); n++) {
+		int tmp = portList.getListIndex(interp, n).getInt(interp);
+		if ((tmp < 0) || (tmp > 255)) {
+		        throw MSXException("outside range 0..255");
+		}
+		ports.set(narrow_cast<byte>(tmp));
+	}
+	return ports;
+}
+
+void ProgrammableDevice::update(const Setting& setting) noexcept
+{
+	try {
+		IterableBitSet<256> tmpPorts = parseAllPorts(setting);
+		unregisterPorts();
+		inPorts = tmpPorts;
+		outPorts = tmpPorts;
+		doRegisterPorts();
+	} catch (MSXException& e) {
+		getCliComm().printWarning(
+			"Parse error in \"",
+			portListSetting.getFullName(), "\": ", e.getMessage());
+	}
+}
+
+void ProgrammableDevice::writeIO(word port, byte value, EmuTime::param /*time*/)
+{
+	outCallback.execute(port, value);
+}
+
+byte ProgrammableDevice::readIO(word port, EmuTime::param /* time */)
+{
+	try {
+		auto obj = inCallback.execute(port);
+		int tmp = obj.getInt(getCommandController().getInterpreter());
+		if ((tmp < 0) || (tmp > 255)) {
+		        throw MSXException("outside range 0..255");
+		}
+		return narrow_cast<byte>(tmp);
+	} catch (MSXException& e) {
+		getCliComm().printWarning(
+			"Wrong result for callback function \"",
+			inCallback.getSetting().getFullName(),
+			"\": ", e.getMessage());
+		return 255;
+	}
+}
+
+void ProgrammableDevice::reset(EmuTime::param /*time*/)
+{
+	resetCallback.execute();
+}
+
+REGISTER_MSXDEVICE(ProgrammableDevice, "ProgrammableDevice");
+
+} // namespace openmsx
--- a/src/ProgrammableDevice.hh
+++ b/src/ProgrammableDevice.hh
@@ -1,0 +1,33 @@
+#ifndef PROGRAMMABLEDEVICE_HH
+#define PROGRAMMABLEDEVICE_HH
+
+#include "MSXDevice.hh"
+#include "StringSetting.hh"
+#include "TclCallback.hh"
+
+namespace openmsx {
+
+class ProgrammableDevice final : public MSXDevice
+                               , private Observer<Setting>
+{
+public:
+	explicit ProgrammableDevice(const DeviceConfig& config);
+	~ProgrammableDevice();
+
+	void writeIO(word port, byte value, EmuTime::param time) override;
+	[[nodiscard]] byte readIO(word port, EmuTime::param time) override;
+	void reset(EmuTime::param time) override;
+
+private:
+	void update(const Setting& setting) noexcept override;
+
+private:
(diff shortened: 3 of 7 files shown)


src/ProgrammableDevice.cc
```diff
@@ -0,0 +1,103 @@
...
+
+#include "narrow.hh"
+
+namespace openmsx {
+
+ProgrammableDevice::ProgrammableDevice(const DeviceConfig& config)
+	: MSXDevice(config)
+	, portLayoutSetting(
+		getCommandController(), tmpStrCat(getName(), " port layout"),
+		"List of {port num} tuples are added to the pool of IO ports", "")
```
m9710797: Question. What kind of use case do you have in mind? Will that typically require a large number of ports, or just a few ports? Maybe it's simpler to instead of list of (base, size)-pairs, just have a list of ports?

share/extensions/programmabledevice.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" ?>
+<!DOCTYPE msxconfig SYSTEM 'msxconfig2.dtd'>
+<msxconfig>
+  <info>
+    <name>Programmable Device</name>
+    <manufacturer>Pedro de Medeiros</manufacturer>
+    <code/>
+    <release_year>2025</release_year>
+    <description>Virtual device to read and write specific hardware ports, see the openMSX manual.</description>
```
m9710797: Here you refer to the manual. But that is not part of the PR. Is this still a work-in-progress PR?  (It's marked as ready (not as in-progress)).

src/ProgrammableDevice.cc
```diff
@@ -0,0 +1,103 @@
...
+		"Tcl proc to call when reset was detected",
+		"", Setting::Save::NO)
+	, readCallback(getCommandController(),
+		getName() + " read callback",
+		"Tcl proc to call when reading a port from the device",
+		"", Setting::Save::NO)
+	, writeCallback(getCommandController(),
+		getName() + " write callback",
+		"Tcl proc to call when writing to a port of the device",
+		"", Setting::Save::NO)
```
m9710797: For these 3 callback procs it would be good to document the parameters and the expected return value (if any). Maybe that will be part of the manual?

src/ProgrammableDevice.cc
```diff
@@ -0,0 +1,103 @@
...
+{
+	portLayoutSetting.attach(*this);
+	readAllPorts(portLayoutSetting);
+	reset(getCurrentTime());
```
m9710797: Just checking, why the call to reset() here? Do you want the reset callback to already trigger on power-on? If that's intentional then maybe also document that behavior in the manual.

src/ProgrammableDevice.cc
```diff
@@ -0,0 +1,103 @@
...
+
+	for (unsigned n = 0; n < ports.getListLength(interp); n++) {
+		TclObject item = ports.getListIndex(interp, n);
+		int port;
+		int num;
+		try {
+			port = item.getListIndex(interp, 0).getInt(interp);
+			num = item.getListIndex(interp, 1).getInt(interp);
+		} catch (...) {
+			continue;
```
m9710797: It's probably a good idea to give the user some kind of feedback when he made a mistake in the "port layout" settings (rather then ignoring it). And Apart from parse errors, we should probably also check the range 0..255.

Currently when there's a mistake in this list, we anyway apply a shorter (possibly empty) list. It would be good if making a mistake in this list would make no changes at all. That's possible by creating a helper function `parsePortLayout()` that returns a `IterableBitSet<256>`. And only when the full setting could be pared successfully, then call `unregisterPorts()`, assign this result to both `inPorts` and `outPorts`, and call `doRegisterPorts()`. In other words: when there's a mistake in the new setting value, then that new value gets rejected, and the situation remains unchanged.

Note that the current `MSXDevice` subclasses that do something with dynamic IO ports, don't manipulate `MSXDevice::inPorts` and `outPorts` directly, but instead they call `cpuInterface.register_IO_InOut()`. E.g. see devices like `RomArc`, `MSXMidi`, `DalSoRiR2`, ... In other words: they handle (un)registration of IO ports themselves instead of using this functionality from the base class. Maybe this device should do the same?

(See also my earlier question about list of (base,size)-pairs vs list of ports).

src/ProgrammableDevice.cc
```diff
@@ -0,0 +1,103 @@
...
+			}
+			result = narrow_cast<byte>(tmp);
+		}
+	} catch (MSXException& e) {
+		getCliComm().printWarning(
+			"Wrong result for callback function \"",
+			readCallback.getSetting().getFullName(),
+			"\": ", e.getMessage());
+	}
+	return result;
```
m9710797: Detail. I think the code becomes simpler with this structure:
```
try {
    ...
    return narrow_cast<byte>(tmp);
} catch (MSXException& e) {
    ...
    return 255;
}
```
So remove the local variable `result`, and have 2 instead of 1 `return` statements.

src/ProgrammableDevice.hh
```diff
@@ -0,0 +1,34 @@
...
+public:
+	explicit ProgrammableDevice(const DeviceConfig& config);
+	~ProgrammableDevice();
+
+	void writeIO(word port, byte value, EmuTime::param time) override;
+	[[nodiscard]] byte readIO(word port, EmuTime::param time) override;
+
+private:
+	void update(const Setting& setting) noexcept override;
+	void reset(EmuTime::param time) override;
```
m9710797: Technically this works. But `reset()` belongs to `readIO()` and `writeIO()`. So all 3 should be either `public` (my preference) or `private`.

m9710797 (changes requested):
Hi. I finished the review from a technical point of view.

But I'd also like to ask for the motivation and for an example (the example could be (part of) the motivation).

Note that using a Tcl callback can be _very_ slow when the port is accessed frequently. E.g. I don't think it's a good idea to use this approach for an IO port that can get accessed hundreds of times per second.

pvmm:
> But I'd also like to ask for the motivation and for an example (the example could be (part of) the motivation).

The motivation is that I am using this code to create UNAPI devices like a network adapter for OpenMSX. I am already using it, so I might as well make it available to anyone that finds it useful.

src/ProgrammableDevice.cc
```diff
@@ -0,0 +1,109 @@
...
+		portListSetting.setString("");
+	}
+}
+
+ProgrammableDevice::~ProgrammableDevice()
+{
+	portListSetting.detach(*this);
+}
+
+IterableBitSet<256> ProgrammableDevice::parseAllPorts(const Setting& setting)
```
m9710797: This can be a (static) non-member function. Currently the only things this function needs from `ProgrammableDevice` is `getCommandController()`, but that can also be obtained from the `setting` parameter. There's even a `Setting::getInterpreter()` method that can be used directly (no intermediate command controller needed).

src/ProgrammableDevice.cc
```diff
@@ -0,0 +1,109 @@
...
+{
+	Interpreter& interp = getCommandController().getInterpreter();
+	TclObject portStr(((StringSetting&)setting).getString());
```
m9710797: We disallow C-style casts in openMSX. (At one point we had warnings enabled when using C-style casts, but we may have disabled those again, because they trigger in 3rd-party code). Instead use (a combination of) `static_cast`, `dynamic_cast`, `const_cast`. Or even better our `checked_cast` (this is like a `static_cast`, but in debug build also checks that `dynamic_cast` would give the same result).

By using these (intentionally) more verbose casts you would have noticed that this cast accidentally(?) also removes `const`.

So I suggest to write this line like:
```
TclObject portList = checked_cast<const StringSetting&>(setting).getString();
```

UPDATE:
Ah no. `StringSetting::getString()` returns a `zstring_view`. And then you immediately convert that back into a `TclObject`. If you want to get the value from a `Setting` as a `TclObject`, you can call `Setting::getValue()`. So no need for a cast at all.
```
TclObject portList = setting.getValue();
```

src/ProgrammableDevice.cc
```diff
@@ -0,0 +1,109 @@
...
+	IterableBitSet<256> ports = {};
```
m9710797: This is not wrong. And in some coding styles this is even recommended. But in openMSX we'd usually write this just like:
```
IterableBitSet<256> ports;
```
No need to call the default constructor (more) explicitly.

src/ProgrammableDevice.cc
```diff
@@ -0,0 +1,109 @@
...
+	}
+	return ports;
+}
+
+void ProgrammableDevice::update(const Setting& setting) noexcept
+{
+	try {
+		IterableBitSet<256> tmpPorts = parseAllPorts(setting);
+		unregisterPorts();
+		inPorts = outPorts = tmpPorts;
```
m9710797: detail: coding style. Prefer to write this as two separate assignments. (I mean on two lines).

src/ProgrammableDevice.cc
```diff
@@ -0,0 +1,109 @@
...
+{
+	writeCallback.execute(port, value);
+}
+
+byte ProgrammableDevice::readIO(word port, EmuTime::param /* time */)
+{
+	try {
+		auto obj = readCallback.execute(port);
+		int tmp;
+		if (obj != TclObject()) {
```
m9710797: This is wrong. If this test is false, then we return an uninitialized `tmp` value. But shouldn't it be an error if `obj` is empty? Shouldn't the read-callback _always_ return a valid integer?

src/ProgrammableDevice.cc
```diff
@@ -0,0 +1,109 @@
...
+#include "serialize.hh"
+
+#include "narrow.hh"
+
+namespace openmsx {
+
+ProgrammableDevice::ProgrammableDevice(const DeviceConfig& config)
+	: MSXDevice(config)
+	, portListSetting(
+		getCommandController(), tmpStrCat(getName(), " ports"),
```
m9710797, line 19: Here you use `tmpStrCat(getName(), " ports")` to construct the setting name, but below you use `getName() + "..."` to construct the callback names. I checked: in both cases the constructor expects a `std::string_view` parameter. So both are correct, but we shouldn't mix both approaches for no reason. Pick one and use it for all 4 cases.

src/ProgrammableDevice.cc
```diff
@@ -0,0 +1,109 @@
...
+		getName() + " write callback",
+		"Defines a Tcl proc that receives a port number (byte) and a byte when the "
+		"MSX is writing a port from the device", "", Setting::Save::NO)
+{
+	portListSetting.attach(*this);
+	try {
+		outPorts = inPorts = parseAllPorts(portListSetting);
+	} catch (MSXException&) {
+		// silently ignore error in settings.xml
+		portListSetting.setString("");
```
m9710797, line 39: This only happens when a saved setting value is already wrong. Or when the user entered this setting, and only later inserted the `ProgrammableDevice` (in a running MSX instance).

Genuine question, i don't know the answer: is it better to silently clear the setting, or leave the current wrong value (maybe give a warning), so that the user can still query and correct the value?

m9710797 (commented):
I finished my 2nd review. Looks good. I only have some very detailed, minor remarks.

You did remove the reference to the manual from your PR. But I think it would be very good to have a section on this in the manual. Otherwise nobody but you will understand what the purpose of this device is, and how to use it. It could be very short. It should:
* Document the parameters and the return value for all 3 callbacks.
* Maybe this can be done via a (simple) example.

src/ProgrammableDevice.cc
pvmm: > Genuine question, i don't know the answer: is it better to silently clear the setting, or leave the current wrong value (maybe give a warning), so that the user can still query and correct the value?

If we leave it there, the command
```set {Programmable Device ports}```
will still display the wrong data, but OpenMSX will not use it, which I guess could be confusing for the user.

m9710797:
Thanks!
I merged this with minor edits:
* I made some minor mostly cosmetic changes.
* A few rewordings in the manual.
* The only functional change is mask the (16-bit) port number to 8-bit before passing it to the Tcl callback.

PR #1880, opened 2025-04-02
Second initialization using GLEW

brossaip:
In my system, glewInit does not work but it does the macro GLEW_VERSION_2_1. I've done a modification that if it failes glewInit it gives some information for debugging and tries to initialize using GLEW_VERSION_2_1.

It's also explained in [this post in the msx.org](https://msx.org/forum/msx-talk/openmsx/glew-libraries-needed-for-compilation#comment-474102).


brossaip (commented):
I've inserted the debugging c files by error. I've tried to remove them without success. They aren't needed. The only changed file needed is VisibleSurface.cc

m9710797:
Thanks for investigating and sending a PR!

Though I think your approach is a bit blunt. It ignores _all_ glew errors. I think it's better to check which specific error code you get (there's a list in glew.h) and then only allow the specific error code that you get. Similar to how we already allow `GLEW_ERROR_NO_GLX_DISPLAY`.

As a follow-up PR:
Once we know the specific error we can maybe also google for this problem and maybe come up with a better approach?

brossaip:
You are totally right. I was so excited that my openMSX was working again and that I had been able to pull a request that I didn't pay enough attention to the code.

I plan to do it during this weekend.

Thanks for your answer and for developing openMSX

brossaip:
I've found [this page](https://community.khronos.org/t/missing-gl-version-how-do-i-solve-it/75296/3?utm_source=chatgpt.com) that suggests to initialize a window using glfw3 before running glewInit, but I get the same error. Maybe the key is to know what the macro does that is different to glewInit

MBilderbeek:
I'm still very interested to know which glew error you get...

brossaip:
Oh, sorry, the error is GLEW_ERROR_NO_GL_VERSION (code 1)

brossaip:
Oh, sorry, the error is GLEW_ERROR_NO_GL_VERSION (code 1)fdsf

brossaip:
I've updated to the new development version and I don't have any problem. Maybe because openMSX has moved to epoxy? Or maybe because I had a missconfigured library and it has been fixed in one of the system udpates.

MBilderbeek:
openMSX has not (yet) moved to epoxy. But anyway, thanks for the update!

PR #1882, opened 2025-04-07
Allow user to call a Tcl proc on exit

pvmm:
This allow advanced users to mitigate some of the problems of the lack of a complete session management by giving them power to do things automatically when openMSX exits, such as saving breakpoints to a file. Obviously the user must provide their own Tcl script with session management procs, but this setting makes it possible.

### DIFF
--- a/src/GlobalSettings.cc
+++ b/src/GlobalSettings.cc
@@ -13,6 +13,9 @@ GlobalSettings::GlobalSettings(GlobalCommandController& commandController_)
 	        "turn power on/off", false, Setting::Save::NO)
 	, autoSaveSetting(commandController, "save_settings_on_exit",
 	        "automatically save settings when openMSX exits", true)
+	, exitCallBackSetting(commandController, "exit_callback",
+		"Tcl proc to call automatically when openMSX exits", "",
+		Setting::Save::YES)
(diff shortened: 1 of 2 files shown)


src/GlobalSettings.cc
```diff
@@ -37,6 +40,7 @@ GlobalSettings::GlobalSettings(GlobalCommandController& commandController_)
 GlobalSettings::~GlobalSettings()
 {
 	getPowerSetting().detach(*this);
+	exitCallBackSetting.execute();
```
m9710797, line 43: You activate the callback during the destructor of `GlobalSettings`. But at this point a lot of stuff is already being destroyed. E.g. all stuff that can potentiality use `GlobalSetting` (like the MSX machine, the debugger, etc) is already gone. So if in your callback you still want to save info from the machine or from the debugger, that info might no longer be available.

Instead the callback should trigger *before* we start destroying anything. We already have an event `QUIT` that gets send right before we exit the emulator. And with `after quit` you can already register (multiple) callbacks for that event. Can you check if the existing `after quit` stuff can replace this PR?

Today I would name this `after exit`. I would accept a PR that renames this subcommand from `quit` to `exit` (but also keep the old name for backwards compatibility).

PR #1886, opened 2025-04-08
better looking palette widget

pvmm:
Changed old palette widget from this:
image
into this:
image
* left and right border are now visible when first and last column buttons are selected;
* black button on black background has an outline now;


m9710797:
Looks a lot better now. Thanks!

PR #1887, opened 2025-04-09
Basic breakpoint session management

pvmm:
Should we proceed with watchpoints and conditions?

### DIFF
--- a/share/scripts/breakpoints.tcl
+++ b/share/scripts/breakpoints.tcl
@@ -1,0 +1,77 @@
+# Easily save breakpoints to and restore from a file
+
+namespace eval breakpoints {
+
+proc save {type filename} {
+	set all [dict values [debug $type list]]
+	set fh [open $filename "w"]
+	puts $fh [join [dict values [debug $type list]] \n]
+	close $fh
+}
+
+proc load {type filename} {
+	set current_bps [dict values [debug $type list]]
+	set fh [open $filename "r"]
+	# compare and create breakpoints if not found
+	foreach bp [split [read -nonewline $fh] \n] {
+		if {$bp in $current_bps} continue
+		debug $type create {*}$bp
+	}
+	close $fh
+}
+
+proc make_save_help {type name} {
+	set body "return \"save_$type <filename>\n\nSave all $type to a given file.\""
+	proc $name {args} $body
+}
+
+proc make_load_help {type name} {
+	set body "return \"load_$type <filename>\n\nLoad only new $type from a given file, while preserving pre-existing ones if any.\""
+	proc $name {args} $body
+}
+
+make_save_help breakpoints save_breakpoints_help
+set_help_proc save_breakpoints breakpoints::save_breakpoints_help
+
+make_save_help watchpoints save_watchpoints_help
+set_help_proc save_watchpoints breakpoints::save_watchpoints_help
+
+make_save_help conditions save_conditions_help
+set_help_proc save_conditions breakpoints::save_conditions_help
+
+make_load_help breakpoints load_breakpoints_help
+set_help_proc load_breakpoints breakpoints::load_breakpoints_help
+
+make_load_help watchpoints load_watchpoints_help
+set_help_proc load_watchpoints breakpoints::load_watchpoints_help
+
+make_load_help conditions load_conditions_help
+set_help_proc load_conditions breakpoints::load_conditions_help
+
+namespace export breakpoints
+
+}
+
+proc save_breakpoints {filename} {
+	breakpoints::save breakpoint $filename
+}
+
+proc save_watchpoints {filename} {
+	breakpoints::save watchpoint $filename
+}
+
+proc save_conditions {filename} {
+	breakpoints::save condition $filename
+}
+
+proc load_breakpoints {filename} {
+	breakpoints::load breakpoint $filename
+}
+
+proc load_watchpoints {filename} {
+	breakpoints::load watchpoint $filename
+}
+
+proc load_conditions {filename} {
+	breakpoints::load condition $filename
+}


share/scripts/breakpoints.tcl
```diff
@@ -0,0 +1,41 @@
...
+		foreach {_ data2} [debug breakpoint list] {
+			if {$data1 eq $data2} {
+				set found 1
+			}
+		}
+		if {!$found} {
+			debug breakpoint create {*}$data1
+		}
+	}
+}
```
m9710797, line 21: There are 2 or even 3 'reasonable' behaviors for `load_breakpoints`:
* Completely replace all current breakpoints with the loaded ones.
* Overwrite current breakpoints with the loaded ones, but leave currently existing (extra) breakpoints in tact.
* Only load breakpoints from the file that completely don't exist yet (leave all currently existing breakpoints unchanged).

This script implements the 3rd option. That's fine. But the 1st and 2nd can also be useful. It might be good to at least document the behavior a bit better.

m9710797 (commented):
In isolation this script is fine. And (like you asked) it might be a good idea to have something similar for watchpoints, conditions, probes, ...

But I don't think we should extend this so that we automatically save breakpoints on exit and load on startup. The reason is that in practice we've often seen that people upgrade to a newer openMSX release but (wrongly) KEEP the Tcl scripts from the older release in place. In the future we'd like to extend openMSX _itself_ to have proper session management for breakpoints. And then I'm afraid it will interfere with such (old) Tcl scripts  (e.g. I can already imagine the question: "Why do my breakpoints get restored when I explicitly disabled that feature" ... because you still have some old (now removed) Tcl script that does this).

pvmm:
This will not save breakpoints automatically on exit. There are just two functions: save_breakpoints and load_breakpoints.

m9710797:
> This will not save breakpoints automatically on exit. There are just two functions: save_breakpoints and load_breakpoints.

I understood. And I said "in isolation this script is fine ...". I only wanted to caution about (possibly in the future) extending this script further as a form of session management.

share/scripts/breakpoints.tcl
```diff
@@ -0,0 +1,41 @@
...
+	set all [gets $fh]
+	close $fh
+	# compare and create breakpoints if not found
+	foreach {_ data1} [eval list $all] {
+		set found 0
+		foreach {_ data2} [debug breakpoint list] {
+			if {$data1 eq $data2} {
+				set found 1
+			}
+		}
```
m9710797: I only thought about this now ... this implementation is correct, but might be slow when there are many breakpoints, it's an O(n^2) algorithm.

The following suggestion is still O(n^2), but probably (i haven't actually measured) already a lot faster because the inner loops is now moved from the Tcl script to a builtin Tcl function (implemented in C).

```
set existing_bps [dict values [debug breakpoint list]]
foreach {_ bp} [eval list $all] {
    if ($bp in $existing_bps} continue
    ...
}
```
The inner loop is now hidden inside the `in` operator.

share/scripts/breakpoints.tcl
```diff
@@ -0,0 +1,41 @@
...
+
+Load only new breakpoints from a given file, while preserving pre-existing ones if any.
+}}
+
+proc load_breakpoints {filename} {
+	set fh [open $filename "r"]
+	set all [gets $fh]
+	close $fh
+	# compare and create breakpoints if not found
+	foreach {_ data1} [eval list $all] {
```
m9710797: We completely ignore the name of the saved breakpoint. That's fine. But if we don't need that info, then maybe we also shouldn't save it?

I think with `[dict values ...]` you can ignore the keys from a key-value-pair list.

Caveat: I haven't actually tried any of this.

pvmm:
Updated to include watchpoints and conditions.

m9710797:
Looks good from a high level point of view. I like how breakpoints/watchpoints/conditions can reuse the same load/save function.
But I only had a few minutes to look at this today. I'll try to do a more detailed review tomorrow.

E.g. I'd like to check what the generated file looks like. It might be good to have one item per line (I'm not sure that's the case now).

pvmm:
> E.g. I'd like to check what the generated file looks like. It might be good to have one item per line (I'm not sure that's the case now).

It is now.

share/scripts/breakpoints.tcl
```diff
@@ -0,0 +1,80 @@
+# Easily save breakpoints to and restore from a file
+
+namespace eval breakpoints {
+
+proc save {type filename} {
+	set all [dict values [debug $type list]]
+	set fh [open $filename "w"]
+	foreach line [list {*}$all] {
+		puts $fh "$line"
+	}
```
m9710797: I think this loop can be simplified to:
```
puts $fh [join [dict values [debug $type list]] \n]
```

share/scripts/breakpoints.tcl
```diff
@@ -0,0 +1,80 @@
...
+proc load {type filename} {
+	set existing_bps [dict values [debug $type list]]
+	set fh [open $filename "r"]
+	# compare and create breakpoints if not found
+	while {[gets $fh line] >= 0} {
+		set bp [eval list $line]
+		if {$bp in $existing_bps} continue
+		debug $type create {*}$bp
+	}
+	close $fh
```
m9710797, line 20: And the loading loop can maybe be simplified to:
```
foreach line [split [read $fh] \n] {
    set item [split $line]
    if {$item in $existing_items} continue
    debug $type create {*}$item
}
```

EDIT: maybe this needs more explanation:
* This reads the whole file at once, then splits on newline, instead of reading line-by-line. That's a tradeoff, for small files (less than 100kB) this is likely faster, not for large files.
* But the more important change is replacing `[eval list $line]` with `[split $line]`.

m9710797 (commented):
Just some quick ideas (untested).
I still didn't have time to actually try your code.

m9710797:
Thanks!

PR #1888, opened 2025-04-09
VRAM access pointer should also be visible on MSX1

pvmm:
It is useful to know what the value of the VRAM access pointer is for debugging purposes, so it shouldn't just be connected to the MSX2 access register R14. I kept it inside the access registers fold because that is where it was before (principle of least surprise).

image

I know you can already read the VRAM pointer in a hex editor widget via the menu under `Debugger > Add hex editor > VRAM Pointer ...`, but this is kind of obscure for most users.


turbor:
Seems we tackled the same problem at the same time, probably since we were in the same discussion. :-) 
I just created a pull request with my solution to the same problem (see PR #1889 )
I have no predilection for the place were the info is added, but I would nevertheless  prefer it that it is made clear that it not a real/regular register.
I make the distinction mainly since the way to set this internal pointer is different from the regular way to set "real" VDP registers (and the data sheets do not list this address pointer under the register overview neither).

pvmm:
I think your solution is cleaner so I am closing this.

Issue #14, opened 2015-06-21
[Support] problem in exection [sf#3]


openMSX-import:
**Reported by *anonymous on 2003-04-01 13:47 UTC**

Hi 
can someone helpme with this:

[section1@nikita bin]$ ./openmsx -setting 
../.openMSX/share/settings.xml 
../.openMSX/share/machines/Philips_NMS_8250/hardwareconfig.xml 
mcop warning: user defined signal handler found for 
SIG_PIPE, overriding
Power LED OFF
Caps LED OFF
Kana LED OFF
Pause LED OFF
Turbo LED OFF
FDD LED OFF
Couldn't load SRAM nms_8250.cmos
Power LED ON
Caps LED ON
Caps LED OFF
Kana LED ON
Kana LED OFF
openmsx: SpriteChecker.hh:228: int 
SpriteChecker::getSprites(int, 
SpriteChecker::SpriteInfo*&amp;amp;): Assertion `spriteCount[line] 
!= -1' failed.
Abort

or 
[section1@nikita bin]$ ./openmsx 
mcop warning: user defined signal handler found for 
SIG_PIPE, overriding
Power LED OFF
Caps LED OFF
Kana LED OFF
Pause LED OFF
Turbo LED OFF
FDD LED OFF
Couldn't load SRAM RTC.SAVE
Power LED ON
Caps LED ON
Caps LED OFF
openmsx: SpriteChecker.hh:228: int 
SpriteChecker::getSprites(int, 
SpriteChecker::SpriteInfo*&amp;amp;): Assertion `spriteCount[line] 
!= -1' failed.
Abort

whats is wrong here???
thanks
see ya.-

openMSX-import:
**Commented by manuelbi on 2003-05-04 21:41 UTC**
Logged In: YES 
user_id=78178

Can you recheck this with the latest CVS version?

openMSX-import:
**Commented by manuelbi on 2003-08-02 15:10 UTC**
Logged In: YES 
user_id=78178

Since there is no way to contact the submitter of this
request (PLEASE LOG IN!), and I get no response on my
question, I am closing this item.

openMSX-import:
**Updated by manuelbi on 2003-08-02 15:10 UTC**
- **assigned_to**: nobody --> mthuurne
- **status**: open --> closed

PR #1889, opened 2025-04-10
Show internal vram pointer

turbor:
A developer on discord remarked that for MSX2 or higher he could see the vram pointer of the VDP but not if he was working on an MSX1 machine. Turned out that he was talking about the fact that on a V99x8 reg #14 showed the extra info of the VRAM pointer as extra info. 
Of course, since a TMS chip has no reg #14 this info wasn't displayed.
This commit displays the extra info of this 14-bits internal pointer/register under the table base registers.


m9710797:
But that's very weird, no? Showing the VRAM pointer under the table base registers. It doesn't belong there. And especially not because on a V99x8 the info _is_ shown in the logical place (so for MSX1 you'd expect to find it in the same location).

I like the look of pvmm's solution better. But not all of the implementation details.
I'll try to create a mix of both and merge that.

PR #1891, opened 2025-04-12
GUI to handle breakpoint session

pvmm:
image
* Added "Save on exit" option, opt-in;
* Added "Load on startup" option, opt-in;

### DIFF
--- a/src/imgui/ImGuiDebugger.cc
+++ b/src/imgui/ImGuiDebugger.cc
@@ -16,6 +16,7 @@
 
 #include "Debuggable.hh"
 #include "Debugger.hh"
+#include "FileContext.hh"
 #include "MSXCPU.hh"
 #include "MSXCPUInterface.hh"
 #include "MSXMemoryMapperBase.hh"
@@ -24,18 +25,22 @@
 #include "RomBlockDebuggable.hh"
 #include "RomInfo.hh"
 
+#include "foreach_file.hh"
 #include "narrow.hh"
 #include "stl.hh"
 #include "strCat.hh"
 #include "StringOp.hh"
 
 #include <imgui.h>
+#include <imgui_stdlib.h>
 
 #include <algorithm>
 
 using namespace std::literals;
 
 namespace openmsx {
+
+static constexpr std::string_view BREAKPOINT_DIR = "breakpoints";
 
 ImGuiDebugger::ImGuiDebugger(ImGuiManager& manager_)
 	: ImGuiPart(manager_)
@@ -105,6 +110,13 @@ void ImGuiDebugger::signalContinue()
 	showChangesFrameCounter = 10;
 }
 
+void ImGuiDebugger::signalQuit()
+{
+	if (saveOnExit && !breakpointFile.empty()) {
+		manager.tclHelper(makeTclList("save_breakpoints", breakpointFile));
+	}
+}
+
 template<typename T>
 static void openOrCreate(ImGuiManager& manager, std::vector<std::unique_ptr<T>>& viewers)
 {
@@ -149,6 +161,13 @@ void ImGuiDebugger::loadStart()
 	hexEditors.clear();
 }
 
+void ImGuiDebugger::loadEnd()
+{
+	if (loadOnStartup && !breakpointFile.empty()) {
+		manager.tclHelper(makeTclList("load_breakpoints", breakpointFile));
+	}
+}
+
 void ImGuiDebugger::loadLine(std::string_view name, zstring_view value)
 {
 	static constexpr std::string_view hexEditorPrefix = "hexEditor.";
@@ -186,6 +205,8 @@ void ImGuiDebugger::loadLine(std::string_view name, zstring_view value)
 
 void ImGuiDebugger::showMenu(MSXMotherBoard* motherBoard)
 {
+	bool openConfirmPopup = false;
+
 	auto createHexEditor = [&](const std::string& name) {
 		// prefer to reuse a previously closed editor
 		auto [b, e] = std::ranges::equal_range(hexEditors, name, {}, &DebuggableEditor::getDebuggableName);
@@ -221,7 +242,62 @@ void ImGuiDebugger::showMenu(MSXMotherBoard* motherBoard)
 			}
 		}
 		ImGui::Separator();
-		ImGui::MenuItem("Breakpoints", nullptr, &manager.breakPoints->show);
+		auto getBreakpointFiles = [] {
+			std::vector<std::string> names;
+			for (auto context = userDataFileContext(BREAKPOINT_DIR);
+			     const auto& path : context.getPaths()) {
+				foreach_file(path, [&](const std::string& fullName) {
+					names.emplace_back(fullName);
+				});
+			}
+			std::ranges::sort(names, StringOp::caseless{});
+			return names;
+		};
+		im::Menu("Breakpoints", [&]{
+			auto listFiles = [&](const std::vector<std::string>& names) {
+				std::optional<std::string> selectedFile;
+				im::ListBox("##select-bp", [&]{
+					for (const auto& name : names) {
+						auto displayName = FileOperations::getFilename(name.c_str());
+						if (ImGui::Selectable(displayName.data())) { // TODO replace data() with c_str()
+							selectedFile = name;
+						}
+						im::PopupContextItem([&]{
+							if (ImGui::MenuItem("delete")) {
+								confirmText = strCat("Delete breakpoint file: ", name);
+								confirmAction = [name]{ FileOperations::unlink(name); };
+								openConfirmPopup = true;
+							}
+						});
+					}
+				});
+				return selectedFile;
+			};
+			ImGui::MenuItem("Editor", nullptr, &manager.breakPoints->show);
+			im::Menu("Save to file", [&]{
+				ImGui::TextUnformatted("Enter name:"sv);
+				ImGui::InputText("##save-bps-name", &breakpointFile);
+				ImGui::SameLine();
+				im::Disabled(breakpointFile.empty(), [&]{
+					if (ImGui::Button("Save as")) {
+						manager.tclHelper(makeTclList("save_breakpoints", breakpointFile));
+						ImGui::CloseCurrentPopup();
+					}
+				});
+			});
+			im::Menu("Load from file", [&]{
+				ImGui::TextUnformatted("Select breakpoint file"sv);
+				auto names = getBreakpointFiles();
+				if (auto selectedFile = listFiles(names)) {
+					breakpointFile = *selectedFile;
+					manager.tclHelper(makeTclList("load_breakpoints", breakpointFile));
+					ImGui::CloseCurrentPopup();
+				}
+			});
+			ImGui::Separator();
+			ImGui::MenuItem("Save on exit", nullptr, &saveOnExit);
+			ImGui::MenuItem("Load on startup", nullptr, &loadOnStartup);
+		});
 		ImGui::MenuItem("Symbol manager", nullptr, &manager.symbols->show);
 		ImGui::MenuItem("Watch expression", nullptr, &manager.watchExpr->show);
 		ImGui::Separator();
@@ -247,6 +323,26 @@ void ImGuiDebugger::showMenu(MSXMotherBoard* motherBoard)
 				}
 			}
 		});
+	});
+
+	const auto* confirmTitle = "Confirm##debugger";
+	if (openConfirmPopup) {
+		ImGui::OpenPopup(confirmTitle);
+	}
+	im::PopupModal(confirmTitle, nullptr, ImGuiWindowFlags_AlwaysAutoResize, [&]{
+		ImGui::TextUnformatted(confirmText);
+
+		bool close = false;
+		if (ImGui::Button("Ok")) {
+			confirmAction();
+			close = true;
+		}
+		ImGui::SameLine();
+		close |= ImGui::Button("Cancel");
+		if (close) {
+			ImGui::CloseCurrentPopup();
+			confirmAction = {};
+		}
 	});
 }
 
--- a/src/imgui/ImGuiDebugger.hh
+++ b/src/imgui/ImGuiDebugger.hh
@@ -41,12 +41,14 @@ public:
 	[[nodiscard]] zstring_view iniName() const override { return "debugger"; }
 	void save(ImGuiTextBuffer& buf) override;
 	void loadStart() override;
+	void loadEnd() override;
 	void loadLine(std::string_view name, zstring_view value) override;
 	void showMenu(MSXMotherBoard* motherBoard) override;
 	void paint(MSXMotherBoard* motherBoard) override;
 
 	void signalBreak();
 	void signalContinue();
+	void signalQuit();
 	void setGotoTarget(uint16_t target);
 	void checkShortcuts(MSXCPUInterface& cpuInterface, MSXMotherBoard& motherBoard,
 	                    ImGuiDisassembly* disassembly);
@@ -92,6 +94,12 @@ private:
 	bool showFlags = false;
 	bool showXYFlags = false;
 	int flagsLayout = 1;
+	std::string breakpointFile;
+	bool loadOnStartup = false;
+	bool saveOnExit = false;
+
+        std::string confirmText;
+        std::function<void()> confirmAction;
 
 	static constexpr auto persistentElements = std::tuple{
 		PersistentElement{"showControl",     &ImGuiDebugger::showControl},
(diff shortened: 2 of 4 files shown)


m9710797:
I support the general idea. But I have two main objections (missing functionality):
* It only does "breakpoints", not "watchpoints" and "conditions". You add this option in "main menu bar > Debugger > Breakpoints". And then move the current "Debugger > Breakpoints" window down under "Debugger > Breakpoints > Edit ...". It would be very strange if the "edit" window does also handle "watchpoints" and "conditions", but the "save/load" functionality does not.
  *  It would also be strange to have 3 separate files for breakpoints/watchpoints/conditions. So maybe we should also (first?) rework the current Tcl scripts so that they can (also?) save/load these 3 into/from a single file. What do you think?
* The UI for saving is acceptable it's not ideal, but it's the same (or similar) like what we currently have for saving savestates/replays/screenshots. But loading is not nice. You need to type the name. So you need to remember it from the previous session, there no possibility to get a list of files or browse for a file. I see two possible solutions:
  * Like savestates/replays. Pick a fixed directory where these files are stored. Then offer a list box where the user can pick the file.
  * Like media (disk, rom, tape). Offer a file dialog where the user can browse for the file. Though if we do this for load, then we should probably also do this for save.

pvmm:
> The UI for saving is acceptable it's not ideal, but it's the same (or similar) like what we currently have for saving savestates/replays/screenshots. But loading is not nice. You need to type the name. So you need to remember it from the previous session, there no possibility to get a list of files or browse for a file. I see two possible solutions:

The idea is to save the breakpoint file in `imgui.ini`, so you should only type it once. Next time you open openMSX, the breakpoint file will be already filled in for you in both save and load menu items. Do you still think it needs improvement?

m9710797:
Caveat: this feedback is only based on the comments and on the screenshot you added (thanks the screenshot is very useful). I have not yet looked at the latest code.

> The idea is to save the breakpoint file in `imgui.ini`, so you should only type it once. Next time you open openMSX, the breakpoint file will be already filled in for you in both save and load menu items. Do you still think it needs improvement?

That's a good idea, and it will cover the most common use cases.

But if you offer an option to type a file name, I think we should also offer an option to re-discover previously saved files. Consider this hypothetical scenario: most of the time I'm working on the same project. When I exit and restart openMSX my breakpoints are automatically restored. But now I'm doing a different experiment, so I change the filename. The next day I want to continue my main work (or some other day I want to continue this experiment), but I've forgotten the original filename (in my case that would be a realistic scenario). There's no way in openMSX itself to re-discover this filename. The only way is to know where these files are stored and browse there with an external tool.

I think it would be good to handle this like we handle loading replay files. So store the files in a fixed directory (a sibling directory of "screenshots", "replays", "savestates", maybe "debugger", or "breakpoints"?). And on load show a list of all available breakpoint-files (that also eliminates the possibility of trying to load a non-existing file) (I think with right-click you can delete files you no longer need).

I think you can mostly copy/paste the replay-loading GUI code (see `src/imgui/ImGuiReverseBar.cc` lines 270-302). You can also simplify it (no need for a table, a sorted list of filenames should be fine).

---

Thanks a lot for working on this. And I'll accept this PR once it ready.

But your work is making me think ... we also have requests to save/load "symbols" and "watch expressions". Should we add similar load/save functionality for those? Or should we just have a single option to save/load these all together. So the full(?) debugger state.

src/imgui/ImGuiDebugger.hh
```diff
@@ -92,17 +93,22 @@ private:
...
+		, PersistentElement{"showRegisters",   &ImGuiDebugger::showRegisters}
+		, PersistentElement{"showSlots",       &ImGuiDebugger::showSlots}
+		, PersistentElement{"showStack",       &ImGuiDebugger::showStack}
+		, PersistentElement{"showFlags",       &ImGuiDebugger::showFlags}
+		, PersistentElement{"showXYFlags",     &ImGuiDebugger::showXYFlags}
+		, PersistentElementMax{"flagsLayout",  &ImGuiDebugger::flagsLayout, 2}
+		, PersistentElement{"showChanges",     &ImGuiDebugger::showChanges}
+		, PersistentElement{"changesColor",    &ImGuiDebugger::changesColor}
+		, PersistentElement{"breakpointFile",  &ImGuiDebugger::breakpointFile}
+		, PersistentElement{"saveOnExit",      &ImGuiDebugger::saveOnExit}
```
m9710797: Very much a detail: I assume you changed the layout (move the comma from the end of the line to the start of the next line) so that in the future adding a new item is a pure single line addition in the diff (not also changing the previous line)? We already use that style for constructor initializer lists. Though for initializer lists the first element is already special (it starts with `:` instead of `,`). But here you now made the first element special while it was not special before. As I said: very much a detail, I understand why you made this change, but it's different from the style we use everywhere else in openMSX. So unless you feel strongly about this, my preference would be to revert this style change.

More important. You've chosen to handle `loadOnStartup` manually (i mean not via this `persistentElements` list). That is because you want to do something extra after loading that field. Though there's a better way for that: in `ImGuiPartInterface` there's a method `loadEnd()` that gets called after everything is loaded. You can override that method and in there check whether you need to load the breakpoints. This makes the code also independent of the order of the fields (e.g. currently you require that the `breakpointFile` fields gets loaded before `loadOnStartup`. (Sure that is the case today, but it's always good to be more robust, e.g. maybe the user manually edited the `imgui.ini` file, or maybe in a future version we do swap these fields (forgetting there was this undocumented ordering constraint)). In the `loadEnd()` method you know that _all_ fields have been loaded already, so the loading order no longer matters.

src/imgui/ImGuiDebugger.cc
```diff
@@ -105,6 +107,15 @@ void ImGuiDebugger::signalContinue()
 	showChangesFrameCounter = 10;
 }
 
+void ImGuiDebugger::signalQuit()
+{
+	try {
+		if (saveOnExit) manager.getInterpreter().execute(tmpStrCat("save_breakpoints ", breakpointFile));
```
m9710797: Here you construct the temporary string `save_breakpoints<space><filename>`. You pass that (as a string) to `execute()`. (One of) the first things the Tcl interpreter does is interpret that string as a list, so (more or less) split on spaces. Now think about what happens if `<filename>` contains internal spaces ... it will see this as if the `save_breakpoints` command is called with more than 2 arguments. That's not what we want.

We could solve that via escaping the filename. But that's not trivial (in theory the filename itself can also contain characters like`"`, `{` or `}`).

A better way is to directly create a Tcl list via `makeTclList()`. Then the interpreter doesn't need to split on spaces anymore (so no risk of splitting-up `<filename>`). And as a bonus it's also faster: we no longer construct the full string and immediately after split it again. The full statement then becomes something like:
```
manager.execute(makeTclList("save_breakpoints", breakpointFile));
```
Thus also eliminating the need for `getInterpreter()` and `#include "Interpreter.hh"`.

src/imgui/ImGuiDebugger.cc
```diff
@@ -121,6 +136,7 @@ void ImGuiDebugger::save(ImGuiTextBuffer& buf)
 {
 	savePersistent(buf, *this, persistentElements);
 
+	buf.appendf("loadOnStartup=%d\n", narrow<int>(loadOnStartup));
```
m9710797: This is a leftover. It's no longer needed.

src/imgui/ImGuiDebugger.cc
```diff
@@ -221,7 +248,71 @@ void ImGuiDebugger::showMenu(MSXMotherBoard* motherBoard)
 			}
 		}
 		ImGui::Separator();
-		ImGui::MenuItem("Breakpoints", nullptr, &manager.breakPoints->show);
+		auto getDirFiles = [] {
```
m9710797: I don't understand the name `getDirFiles`. Is it a function that returns a list of all filenames in a directory? If so, then probably this function should take that directory as a parameter, and we should move it to a more general location (maybe somewhere in `FileOperations`?) and try to reuse it in many more functions.

Or we can just give this current function a better name. Maybe something line `getBreakpointFiles()`?

src/imgui/ImGuiDebugger.cc
```diff
@@ -221,7 +248,71 @@ void ImGuiDebugger::showMenu(MSXMotherBoard* motherBoard)
...
+				foreach_file(path, [&](const std::string& fullName) {
+					names.emplace_back(fullName);
+				});
+			}
+			std::ranges::sort(names, StringOp::caseless{});
+			return names;
+		};
+		im::Menu("Breakpoints", [&]{
+			auto listFiles = [&](const std::vector<std::string>& names) {
+				std::optional<std::pair<std::string, std::string>> selectedFile;
```
m9710797: This function returns a (optional) pair of strings. Though the only caller of this function only uses one of the elements of this pair. So better simplify this to just return a single string.

Instead of returning an `optional<string>` you can also return an empty string to indicate no selection (because filenames can never be empty). Either is fine for me. Choose what you like best.

src/imgui/ImGuiDebugger.cc
```diff
@@ -221,7 +248,71 @@ void ImGuiDebugger::showMenu(MSXMotherBoard* motherBoard)
...
+				im::ListBox("##select-bp", [&]{
+					for (const auto& name : names) {
+						auto displayName = FileOperations::getFilename(name);
+						if (ImGui::Selectable(displayName.data())) {
```
m9710797: Using `data()` works fine in practice, but semantically it's a bit misleading. Better use `c_str()`.

`ImGui::Selectable()` takes a `const char*` as the first parameter. This is a zero-terminated C-style string. In the past there was a difference between `data()` and `c_str()`, only the latter was guaranteed to return a pointer to a zero-terminated string (technically the zero-terminator character could be lazily added, and that was only done by `c_str()`). Starting from c++11 both methods offer this guarantee. But nevertheless it's still better to be specific. Use `data()` if you only care about the actual string characters, and use `c_str()` if you also care about the zero-terminating character.

src/imgui/ImGuiManager.cc
```diff
@@ -397,6 +397,10 @@ bool ImGuiManager::signalEvent(const Event& event)
 		}
 	} else {
 		switch (getType(event)) {
+		case EventType::QUIT: {
+			debugger->signalQuit();
+			break;
+		}
```
m9710797: Detail: these braces `{}` aren't needed here. Typically we don't write them when not needed. But I see that in the `IMGUI_DELAYED_ACTION` case below we also have redundant braces, and you probably copy/pasted the style from there?

src/imgui/ImGuiDebugger.cc
```diff
@@ -248,6 +339,26 @@ void ImGuiDebugger::showMenu(MSXMotherBoard* motherBoard)
...
+			confirmAction();
+			close = true;
+		}
+		ImGui::SameLine();
+		close |= ImGui::Button("Cancel");
+		if (close) {
+			ImGui::CloseCurrentPopup();
+			confirmAction = {};
+		}
+	});
```
m9710797, line 346: Maybe something for a future PR...

This code block for a confirm dialog is copy/pasted in multiple locations already. Now one more location. It would be nice to create a common helper proc for this. (Or a small struct that also holds the text and action).

src/imgui/ImGuiDebugger.cc
```diff
@@ -221,7 +248,71 @@ void ImGuiDebugger::showMenu(MSXMotherBoard* motherBoard)
...
+					breakpointFile = name;
+					try {
+						manager.execute(makeTclList("load_breakpoints", breakpointFile));
+					} catch (MSXException& e) {
+						manager.getCliComm().printWarning(e.getMessage());
+					}
+					ImGui::CloseCurrentPopup();
+				}
+			});
+			ImGui::MenuItem("Editor", nullptr, &manager.breakPoints->show);
```
m9710797: Detail. I'm not sure what's best.

Currently the order is: "Save", "Load", "Editor". I was wondering if we should instead place "Editor" in front.

I tried to come up with some reasoning, but I'm not sure if these reasons make sense. And maybe you do have a good reasoning for your current order?
* If I look at how I currently use openMSX, then I *think* I'll personally use "Editor" more often than "Save" and "Load". So then it makes sense to place that option first. But I can understand that for different people this may be different.
* If you're new to openMSX and discovering the functionality, then you should first try the "Editor" before "Save" or "Load" make sense.

src/imgui/ImGuiDebugger.cc
```diff
@@ -16,6 +16,8 @@
 
 #include "Debuggable.hh"
 #include "Debugger.hh"
+#include "FileContext.hh"
+#include "Interpreter.hh"
```
m9710797: Detail: this #include is probably no longer needed?

src/imgui/ImGuiDebugger.cc
```diff
@@ -105,6 +111,15 @@ void ImGuiDebugger::signalContinue()
 	showChangesFrameCounter = 10;
 }
 
+void ImGuiDebugger::signalQuit()
+{
+	try {
+		if (saveOnExit) manager.execute(makeTclList("save_breakpoints", breakpointFile));
+	} catch (MSXException& e) {
+		// ignore
+	}
```
m9710797: This pattern: a try-catch block with building-and-executing a Tcl command is repeated 4 times in the code. Maybe it can be factored out into a helper function? I'm not sure, this is something I'd have to try and see how it looks before I can form a good opinion.

The 4 occurrences do have slight variations:
* Loading vs saving (maybe this can become a parameter?)
* Sometimes the catch block is empty, sometimes it prints a warning. Shouldn't we always print that warning?
  * On startup when previously saved breakpoints can't be reloaded, you might want to see a warning.
  * On exit when the breakpoints can't be saved you also might want to see the warning (though it will only be visible in the terminal from which openMSX was started. not in openMSX itself because the openMSX window will be gone soon).
* When `breakpointFile` is empty we shouldn't attempt to load or save. If this pattern is factored out we only need to check this once instead of 4 times. I understand that by ignoring the error in an empty catch block it happens to work fine anyway (but see previous point, maybe we don't want to ignore the error). And currently when you do print the warning you already ensure that the string can't be empty.
* Sometimes you check `saveOnExit` or `loadOnStartup`. But those checks can be done outside this new helper function.

m9710797 (changes requested):
Looks good. Most of my review comments are about details. Or are just questions, or suggestions for future improvements.

I am still wondering about symbols and watch-expressions. Should those also be saved/loaded together with these breakpoints (and then probably in the same file?). That's something we should decide before the next release. But let's already merge this PR now (as an intermediate step).

src/imgui/ImGuiDebugger.cc
pvmm: But `displayName` is a `std::string_view`, not a `std::string`, so it doesn't have a `c_str()` function.

m9710797: Aha, I didn't see that. But then it's more dangerous than I thought. Because a `std::string_view` does not guarantee zero-termination.

Looking in more detail:
* `FileOperations::getFilename()` both takes a `std::string_view` as input and returns a `std::string_view`.
* In general a `std::string_view` is not required to be zero-terminated.
* `FileOperations::getFilename()` returns the last part of the input.
* So if the input is not zero-terminated, then the output also won't be.
* And similarly if the input happens to be zero-terminated anyway, then the output will also be.

We actually pass a `std::string`, that one is zero-terminated. So we're lucky and it works out fine. But that's not clear from only looking at the type-system.

For this PR can it's sufficient if you add a comment like "TODO replace data() with c_str()". But this is something I'd like to fix in a better way: ensure via the type-system that this is safe.

My idea is the following: we create two overloaded versions of `FileOperations::getFilename():
```
std::string_view getFilename(std::string_view)
zstring_view     getFilename(zstring_view)
```
So if we pass a zero-terminated string (`zstring_view`) as input, the result remains zero-terminated. But if we pass a string without that guarantee (a plain `string_view`) then the result also doesn't have that guarantee.

At least that's the idea. I still have to work out some details. (I suspect there might be some problems related to ambiguous overloads that I need to resolve).

m9710797:
I merged this with some small edits.
Mostly the `tclHelper` function that you added. But I probably explained it badly.

pvmm:
> But your work is making me think ... we also have requests to save/load "symbols" and "watch expressions". Should we add similar load/save functionality for those? Or should we just have a single option to save/load these all together. So the full(?) debugger state.

I understand creating new symbols in the symbol table that were not imported from .noi or .sym files, but aren't expressions already automatically saved? My imgui.ini file:
```
[openmsx][watch expr]
show=0
watch={peek at fixed address} {[peek 0xfcaf]} screen=%d
watch={VDP command executing} {[debug read "VDP status regs" 2] & 1} {}
watch={PSG enable-channel status} {[debug read "PSG regs" 7]} %08b
watch={The following 2 require an appropriate symbol file} {} {}
watch={value of 'myLabel'} {$sym(myLabel)} 0x%04x
watch={peek at symbolic address} {[peek16 $sym(numItems)]} {%d items}
```

m9710797:
> I understand creating new symbols in the symbol table that were not imported from .noi or .sym files, but aren't expressions already automatically saved? My imgui.ini file:

You're right about watch expressions (I admit I forgot about this). But still, wouldn't it be more logical if these were instead saved together with the other debugger state?

For symbols we only need to save the list of loaded symbol files. We don't support creating new symbols inside openMSX itself (and IMHO we should keep it that way). I'm not 100% sure, but I think querying this list and re-loading symbol files is already possible in Tcl. So it should be possible to extend the load/save_breakpoints script ... if you agree that's a good idea (I'm not 100% sure myself).

m9710797:
Hi pvmm,

I did some refactoring: it should now be easier to add file-selection lists in the GUI.
This refactoring also touched on the GUI-breakpoint-code that you recently added.

So my question: can you help test this. I did some test myself of course, and all still seems to work fine. But it never hurts to double-check.

If you're interested, and if you have time, then a code review would also be welcome.

Thanks.

PR #1892, opened 2025-04-13
save_breakpoint saves all to single file

pvmm:
GUI command "Breakpoints > Save to file" (#1891) expects a single file to save breakpoints, watchpoints and conditions. This change makes it possible to use a single save_breakpoints function to accommodate this behaviour.

### DIFF
--- a/share/scripts/breakpoints.tcl
+++ b/share/scripts/breakpoints.tcl
@@ -1,77 +1,43 @@
 # Easily save breakpoints to and restore from a file
 
-namespace eval breakpoints {
-
-proc save {type filename} {
-	set all [dict values [debug $type list]]
+proc save_breakpoints {filename} {
 	set fh [open $filename "w"]
-	puts $fh [join [dict values [debug $type list]] \n]
-	close $fh
-}
-
-proc load {type filename} {
-	set current_bps [dict values [debug $type list]]
-	set fh [open $filename "r"]
-	# compare and create breakpoints if not found
-	foreach bp [split [read -nonewline $fh] \n] {
-		if {$bp in $current_bps} continue
-		debug $type create {*}$bp
+	foreach type {breakpoint watchpoint condition} {
+		puts $fh "\[$type\]\n[join [dict values [debug $type list]] \n]\n"
 	}
 	close $fh
 }
 
-proc make_save_help {type name} {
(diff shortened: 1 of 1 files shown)


m9710797:
Sorry for only telling you now (I've only started to think about this after I saw this PR) ...

Currently you reconstruct the type (bp, wp or cond) from the presence of specific properties in the line. That's an heuristic. And (today) that works of course.

But maybe it's cleaner to be more explicit? Both to make the file more readable/descriptive for humans. And maybe also to make it more robust for future developments.

Maybe a format like this:

```
[breakpoint]
-address 0x1234 -command {debug break} ...
-address 0xabcd ...
...

[watchpoint]
-type read_io ...
...

[condition]
...
```

Writing such a file is more or less the current code in a loop (over bp, wp and cond), and print a header in front and an empty line after. (You can choose to skip empty sections, or include them, either is fine for me. Skipping is more compact, including empty sections indicates what's possible).

Reading is a bit more complex:
* Loop over all lines. Ignore empty lines.
* First check for a heading `[breakpoint]`, `[watchpoint]` or `[condition]`, and remember that type. (Either check for exactly these three, or check for lines that start/end with `[`/`]` and take the word inside as the current type).
* If it's not a header then process the line in the same way as we do now. And use the type of the last header.

What do you think?

--

Currently you have 3 procs `load_breakpoints`, `load_watchpoints`, `load_conditions` that all 3 do the exact same thing. Maybe that's an indication we only need a single proc? But what would be a good name? Just `load_breakpoints` would be acceptable IMHO (if we document it handles the 3 types), it's consistent with what we have in the GUI. But if you know a better name then please discuss.

And for symmetry and simplicity maybe we also only need 1 `save` command. Or do you think it's useful to be able to only save 1 type, not always all 3 (E.g. for the GUI it seems sufficient to always do all three, but sometimes we do allow more with the console commands).

--

UPDATE:
Maybe in the future(*1) we'll add "debug probes" as a 4th type (in the openMSX console type `help debug probe`. It would be good if these can be supported in the same file format.

(*1) Before we can do this, it would be good to first refactor the `debug probe` command to be more compatible with the syntax of `debug breakpoint`.

m9710797:
Thanks!

PR #1893, opened 2025-04-15
Give breakpoints file an extension

pvmm:
* New functions are hidden in a namespace now;
* Save and load files at `$(OPENMSX_HOME)/breakpoints`;
* Create `$(OPENMSX_HOME)/breakpoints` if it doesn't exist.

### DIFF
--- a/share/scripts/breakpoints.tcl
+++ b/share/scripts/breakpoints.tcl
@@ -1,30 +1,28 @@
 # Easily save breakpoints to and restore from a file
 
-proc save_breakpoints {filename} {
-	set fh [open $filename "w"]
-	foreach type {breakpoint watchpoint condition} {
-		puts $fh "\[$type\]\n[join [dict values [debug $type list]] \n]\n"
-	}
-	close $fh
+proc get_breakpoints_dir {} {
+	set path [file normalize $::env(OPENMSX_USER_DATA)/../breakpoints]
+	file mkdir $path
+	return $path
 }
 
-proc load_breakpoints {filename} {
-	foreach type {breakpoint watchpoint condition} {
-		set existing_bps($type) [dict values [debug $type list]]
-	}
-	set section {}
-	set fh [open $filename "r"]
-	foreach line [split [read -nonewline $fh] \n] {
-		set line [string trim $line]
-		# load valid file section only
-		if {[regexp {^\[([^\]]+)\]$} $line -> section]} continue
-		if {!($section in {breakpoint watchpoint condition})} continue
-		# compare and create breakpoints if not found
-		if {$line eq {} || $line in $existing_bps($section)} continue
-		debug $section create {*}$line
-	}
-	close $fh
+namespace eval breakpoints {
+
+proc fix_extension {filename} {
+	expr {[string match *.breakpoints $filename] ? $filename : "$filename.breakpoints"}
 }
+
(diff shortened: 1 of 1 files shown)


share/scripts/breakpoints.tcl
```diff
@@ -1,7 +1,50 @@
 # Easily save breakpoints to and restore from a file
 
+namespace eval breakpoints {
+
+proc create_directory {dir} {
+    if {![file isdirectory $dir]} {
+        file mkdir $dir
```
m9710797: I checked the documentation of `file mkdir`. It's not an error if `dir` is already an existing directory. So the `if` is not needed, and then `create_directory` is just the same as `file mkdir`, so no need for this proc.

share/scripts/breakpoints.tcl
```diff
@@ -1,7 +1,50 @@
...
+    }
+}
+
+proc has_path {filename} {
+    return [regexp {[/\\]} $filename]
```
m9710797: `regexp` is very powerful, but also relatively slow. I think this one is simple enough that `string match {*[/\\]} $filename` could work as well.
Caveat: I have not tested this, and I have not measured the speed of either.

Detail: for simple one-line procs it's common practice to leave out the `return` keyword. A Tcl proc automatically returns the result of the last executed statement. Some people like that style, others don't.

share/scripts/breakpoints.tcl
```diff
@@ -1,7 +1,50 @@
...
+	return [expr {[string match *.breakpoints $filename] ? $filename : "$filename.breakpoints"}]
+}
+
+proc fix_filename {filename} {
+	if {![has_path $filename]} {
+		set path [file normalize $::env(OPENMSX_USER_DATA)/../breakpoints]
+		create_directory $path
+		set filename [file join $path $filename]
+	}
+	return [fix_extension $filename]
```
m9710797: I have mixed feelings about this:
* I do understand that this makes using `load/save_breakpoints` easier.
* But it does introduce an irregularity:
  * You can pass an arbitrary path, and it will then use that path ...
  * ... except when that path is "too simple" (contains just a file without a directory). Then it will transform the path.
* This behavior is fine for interactive use, but it does make it difficult to use this from within another script. You can in general not really pass a path because some paths get transformed and others are not.

There are various different solutions:
* Split this into 2 procs, one proc that only adjusts the path (no loading or saving, just transforming the path), and another proc that just takes the path and uses it without transformation. Main disadvantage is that this is now a lot harder to use interactively, using it in scripts is easy now.
* You could expose two procs, one that takes a path and uses it as is, and another that takes a single word and unconditionally prepends the directory and appends the extension.
* We could choose to not support different directories. Then the proc will always, unconditionally, prepend the directory and append the extension.

Maybe the 3rd option is good enough for us?

m9710797 (commented):
I'm not really sure about this PR. I understand it's convenient for interactive use. But it also introduces some complexity (especially for use in other scripts). More details in the review comments below.

And maybe we don't really need this. Maybe it's sufficient to just always load/save in the openMSX `breakpoints` directory?

Feel free to disagree.

m9710797:
I forgot to say: we now always add the `.breakpoints` extension. And (once PR#1891 is merged) we will show that extension in the GUI. But it would be nice if the GUI omits that extension.

PR #1894, opened 2025-04-15
 Fix SIGABRT caused by empty listbox item 

pvmm:
When the same file name is detected, ImGui hides the menu and displays an error message that is impossible to read because it is too fast (second image was captured by video recording).
image
image
It's possible to find two files with the same name because one is in `$PWD` and the other is in `$(OPENMSX_HOME)/breakpoints`.

### DIFF
--- a/src/imgui/ImGuiDebugger.cc
+++ b/src/imgui/ImGuiDebugger.cc
@@ -240,14 +240,12 @@ void ImGuiDebugger::showMenu(MSXMotherBoard* motherBoard)
 		ImGui::Separator();
 		auto getBreakpointFiles = [] {
 			std::vector<std::string> names;
-			for (auto context = userDataFileContext(BREAKPOINT_DIR);
-			     const auto& path : context.getPaths()) {
-				foreach_file(path, [&](const std::string& fullName) {
+			foreach_file(FileOperations::join(openmsx::FileOperations::getUserOpenMSXDir(), "breakpoints"),
+				[&](const std::string& fullName) {
 					if (fullName.ends_with(".breakpoints")) {
 						names.emplace_back(fullName);
 					}
 				});
-			}
 			std::ranges::sort(names, StringOp::caseless{});
 			return names;
 		};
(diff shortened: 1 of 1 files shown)


pvmm:
* Changed getBreakpointFiles() to use im:ID() to list box item because a file named ".breakpoints" can cause SIGABRT;
* Changed getBreakpointFiles() to scan only `~/.openMSX/breakpoints` instead of array of paths.

src/imgui/ImGuiDebugger.cc
```diff
@@ -240,24 +240,22 @@ void ImGuiDebugger::showMenu(MSXMotherBoard* motherBoard)
...
 		im::Menu("Breakpoints", [&]{
 			auto listFiles = [&](const std::vector<std::string>& names) {
 				std::optional<std::string> selectedFile;
 				im::ListBox("##select-bp", [&]{
-					for (const auto& name : names) {
-						auto displayName = std::string(FileOperations::stripExtension(FileOperations::getFilename(name)));
-						if (ImGui::Selectable(displayName.c_str())) {
+					for (const auto& [id, name] : enumerate(names)) {
+						if (auto displayName = std::string(FileOperations::stripExtension(FileOperations::getFilename(name)));
+						    ImGui::Selectable(tmpStrCat(displayName, "##", id).c_str())) {
```
m9710797: That certainly works. And I think we still use this approach in various other locations in openMSX.

But in new code I prefer the approach with `ImGui::PushID()` and `ImGui::PopID()`, or better our c++ wrappers around that e.g. `im::ID(int)`.

For the combination of a for-loop and in each iteration a different ID, we also have `im::ListClipperID()`. Maybe not super important for this specific case, but that immediately also optimized drawing lists with many items (where only part of the items are visible at the same time).


Here's some more info on the Dear ImGui ID system:
https://github.com/ocornut/imgui/blob/master/docs/FAQ.md#q-about-the-id-stack-system

m9710797:
Thanks!

PR #1895, opened 2025-04-16
rename loadOnStartup to reloadBreakpoints

pvmm:
Doesn't make sense to keep both saveOnExit and loadOnStartup since the only purpose is to restore session breakpoints. A single variable is enough.


m9710797:
Fully agree. I also missed this during my review.

PR #1896, opened 2025-04-16
Remove .breakpoints from file when showing them in ListBox

pvmm:


m9710797:
Thanks.

src/imgui/ImGuiDebugger.cc
```diff
@@ -254,7 +256,7 @@ void ImGuiDebugger::showMenu(MSXMotherBoard* motherBoard)
 				std::optional<std::string> selectedFile;
 				im::ListBox("##select-bp", [&]{
 					for (const auto& name : names) {
-						auto displayName = FileOperations::getFilename(name);
+						auto displayName = std::string(FileOperations::stripExtension(FileOperations::getFilename(name)));
```
m9710797, line 259: The combination `getFilename()` + `stripExtension()` is often called `stem()`, e.g. see here: https://en.cppreference.com/w/cpp/filesystem/path/stem

Maybe we should add such a function in our `FileOperations` utility functions (and then check where else in the code we use this combination, and simplify that code).

In the long term we should think about switching to `std::filesystem::path` for most of our path handling.

PR #1897, opened 2025-04-16
Fix SIGABRT caused by empty list item

pvmm:
How to reproduce this error:
* Go to directory ~/.openMSX/layouts and create a file called ".ini";
* Execute openMSX and open the layout file window in Settings > GUI > Restore layout
* openMSX will crash with the message:
```
openmsx: src/3rdparty/imgui/imgui.cc:11308: bool ImGui::ItemAdd(const ImRect&, ImGuiID, const ImRect*, ImGuiItemFlags): Assertion `id != window->ID && "Cannot have an empty ID at the root of a window. If you need an empty label, use ## and read the FAQ about how the ID Stack works!"' failed.

Thread 1 "openmsx" received signal SIGABRT, Aborted.
0x00007ffff7399b54 in __pthread_kill_implementation () from /lib64/libc.so.6
```

### DIFF
--- a/src/imgui/ImGuiSettings.cc
+++ b/src/imgui/ImGuiSettings.cc
@@ -310,25 +310,23 @@ void ImGuiSettings::showMenu(MSXMotherBoard* motherBoard)
 		im::Menu("GUI", [&]{
 			auto getExistingLayouts = [] {
 				std::vector<std::string> names;
-				for (auto context = userDataFileContext("layouts");
-				     const auto& path : context.getPaths()) {
-					foreach_file(path, [&](const std::string& fullName, std::string_view name) {
+				foreach_file(FileOperations::join(openmsx::FileOperations::getUserOpenMSXDir(), "layouts"),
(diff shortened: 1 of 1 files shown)


pvmm:
Changed PR to only look for files in the `~/.openMSX/layouts` directory.

src/imgui/ImGuiSettings.cc
```diff
@@ -374,7 +374,7 @@ void ImGuiSettings::showMenu(MSXMotherBoard* motherBoard)
 				ImGui::TextUnformatted("Select layout"sv);
 				auto names = getExistingLayouts();
 				if (auto selectedLayout = listExistingLayouts(names)) {
-					const auto& [name, displayName] = *selectedLayout;
+					const auto& name = *selectedLayout;
```
m9710797: (I only read this PR very quickly, I did not test anything).

Is this correct? If I read the code correctly, then this now uses the `displayName`. But I think `manager.loadIniFile` should be the `fullName`, no?

PR #1898, opened 2025-04-16
Store stem of the file only

pvmm:
Now that save_breakpoints is confined to ~/.openMSX/breakpoints, we don't need to store the complete path to the file.

This was missing in #1894

### DIFF
--- a/src/imgui/ImGuiDebugger.cc
+++ b/src/imgui/ImGuiDebugger.cc
@@ -243,7 +243,7 @@ void ImGuiDebugger::showMenu(MSXMotherBoard* motherBoard)
 			foreach_file(FileOperations::join(openmsx::FileOperations::getUserOpenMSXDir(), "breakpoints"),
 				[&](const std::string& fullName) {
 					if (fullName.ends_with(".breakpoints")) {
-						names.emplace_back(fullName);
+						names.emplace_back(FileOperations::stripExtension(FileOperations::getFilename(fullName)));
 					}
 				});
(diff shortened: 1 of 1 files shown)


src/imgui/ImGuiDebugger.cc
```diff
@@ -257,7 +257,7 @@ void ImGuiDebugger::showMenu(MSXMotherBoard* motherBoard)
 						im::ID(id, [&]{
 							if (auto displayName = std::string(FileOperations::stripExtension(FileOperations::getFilename(name)));
 							    ImGui::Selectable(displayName.c_str())) {
-								selectedFile = name;
+								selectedFile = displayName;
```
m9710797: May I suggest to do this stem-calculation already in `getBreakpointFiles()`.
The small advantage is that then here we don't (again) need to convert the stem from `std::string_view` to `std::string` (`getBreakPointFiles()` already needs to do that to be able to store the string).

m9710797:
Note: for showing the list of savestates or replay files, we only scan the filesystem once when the sub-menu is first (re-)opened. But here for breakpoint files we rescan the filesystem on each frame. If there are only a few files that's acceptable. If there are tens or hundreds of files, then maybe that's something we should avoid.

Not urgent (I don't expect there will be many breakpoint files), but maybe something we should address in the future. Hopefully by factoring out some common code between listing breakpoints/savestates/replays.

m9710797:
Thanks.

PR #1899, opened 2025-04-18
Context menu that copies pattern and color to clipboard

pvmm:
image
Also:
* formatToString and rawToString moved from DebuggableEditor to ImGuiUtils.

### DIFF
--- a/src/imgui/DebuggableEditor.cc
+++ b/src/imgui/DebuggableEditor.cc
@@ -137,47 +137,6 @@ void DebuggableEditor::paint(MSXMotherBoard* motherBoard)
 	});
 }
 
-
-[[nodiscard]] static std::string formatToString(
-	Debuggable& debuggable, unsigned begin, unsigned end,
-	std::string_view prefix, unsigned columns, std::string_view suffix,
-	std::string_view formatStr, Interpreter& interp)
-{
-	std::string result;
-	unsigned col = 0;
-	for (unsigned addr = begin; addr <= end; ++addr) {
-		if (col == 0) strAppend(result, prefix);
-
-		auto val = debuggable.read(addr);
-		auto cmd = makeTclList("format", formatStr, val);
-		auto formatted = cmd.executeCommand(interp); // may throw
-		strAppend(result, formatted.getString());
-
-		if (++col == columns) {
-			col = 0;
-			strAppend(result, suffix, '\n');
-		} else if (addr != end) {
-			strAppend(result, ", ");
-		}
-	}
-	if (col != 0) {
-		strAppend(result, suffix, '\n');
-	}
-	return result;
-}
-
-[[nodiscard]] static std::string rawToString(
-	Debuggable& debuggable, unsigned begin, unsigned end)
-{
-	std::string result;
-	result.reserve(end - begin + 1);
-	for (unsigned addr = begin; addr <= end; ++addr) {
-		auto val = debuggable.read(addr);
-		result += static_cast<char>(val);
-	}
-	return result;
-}
-
 void DebuggableEditor::drawExport(const Sizes& s, Debuggable& debuggable)
 {
 	auto& interp = manager.getInterpreter();
@@ -336,12 +295,12 @@ void DebuggableEditor::drawExport(const Sizes& s, Debuggable& debuggable)
 			  (exportDestination != OUTPUT_FILE || !exportFilename.empty());
 		im::Disabled(!ok, [&]{
 			if (ImGui::Button("Export")) {
+				auto fetch = [&](unsigned address) { return debuggable.read(address); };
 				try {
 					auto output = (exportFormatted == EXPORT_FORMATTED)
-						? formatToString(
-							debuggable, *begin, *end, exportPrefix, exportColumns,
-							exportSuffix, formatStr, interp)
-						: rawToString(debuggable, *begin, *end);
+						? formatToString(fetch, *begin, *end, exportPrefix, exportColumns, exportSuffix,
+								formatStr, interp)
+						: rawToString(fetch, *begin, *end);
 					if (exportDestination == OUTPUT_CLIPBOARD) {
 						manager.getReactor().getDisplay().getVideoSystem().setClipboardText(output);
 					} else {
--- a/src/imgui/ImGuiCharacter.cc
+++ b/src/imgui/ImGuiCharacter.cc
@@ -412,6 +412,46 @@ void ImGuiCharacter::paint(MSXMotherBoard* motherBoard)
 				printAddressRange8("Color", colTable.getAddress((mode == SCR1) ? (pattern / 8) : (8 * pattern)));
 			}
 		};
+		auto formatBinaryData = [&](uint16_t address) {
+			return formatToString([&](unsigned addr){ return vram[addr]; }, address, address + 7, {}, 1, {}, "%02X", manager.getInterpreter());
+		};
+		auto getPatternFromGrid = [&]() {
+			return gridPosition.x + 32 * gridPosition.y;
+		};
+		auto copyPatternColorPopup = [&]() {
+			bool popup = ImGui::BeginPopupContextWindow("PatternCopyPopup");
+			if (popup) {
+				bool hasColorData = mode == one_of(SCR1, SCR2);
+				auto caption = hasColorData
+					? "Copy pattern and color data to clipboard"
+					: "Copy pattern data to clipboard";
+				if (ImGui::MenuItem(caption)) {
+					auto pattern = getPatternFromGrid();
+					auto text = strCat("Pattern data\n", formatBinaryData(patTable.getAddress(8 * pattern)));
+					if (hasColorData) {
+						auto cp = (mode == SCR1) ? (pattern / 8) : (8 * pattern);
+						strAppend(text, "Color data\n", formatBinaryData(colTable.getAddress(cp)));
+					}
+					ImGui::SetClipboardText(text.c_str());
+				}
+				ImGui::EndPopup();
+			}
+			return popup;
+		};
+		auto drawGrid = [&]() {
+			auto pat = getPatternFromGrid();
+			printPatternNr(pat);
+			auto uv1 = gl::vec2(gridPosition) * recipPatTexChars;
+			auto uv2 = uv1 + recipPatTexChars;
+			auto pos2 = ImGui::GetCursorPos();
+			ImGui::Image(patternTex.getImGui(), zoomCharSize, uv1, uv2);
+			if (grid) {
+				ImGui::SetCursorPos(pos2);
+				ImGui::Image(gridTex.getImGui(),
+					zoomCharSize, {}, charSize);
+			}
+			printPatternColorAddress(pat);
+		};
 
 		ImGui::Separator();
 		im::TreeNode("Pattern Table", ImGuiTreeNodeFlags_DefaultOpen, [&]{
@@ -425,20 +465,12 @@ void ImGuiCharacter::paint(MSXMotherBoard* motherBoard)
 				bool hovered = ImGui::IsItemHovered() && (mode != OTHER);
 				ImGui::SameLine();
 				im::Group([&]{
-					if (hovered) {
-						auto gridPos = trunc((gl::vec2(ImGui::GetIO().MousePos) - scrnPos) / charZoom);
-						auto pat = gridPos.x + 32 * gridPos.y;
-						printPatternNr(pat);
-						auto uv1 = gl::vec2(gridPos) * recipPatTexChars;
-						auto uv2 = uv1 + recipPatTexChars;
-						auto pos2 = ImGui::GetCursorPos();
-						ImGui::Image(patternTex.getImGui(), zoomCharSize, uv1, uv2);
-						if (grid) {
-							ImGui::SetCursorPos(pos2);
-							ImGui::Image(gridTex.getImGui(),
-								zoomCharSize, {}, charSize);
-						}
-						printPatternColorAddress(pat);
+					if (copyPatternColorPopup()) {
+						drawGrid();
+					} else if (hovered) {
+						// store last selected position before popup
+						gridPosition = trunc((gl::vec2(ImGui::GetIO().MousePos) - scrnPos) / charZoom);
+						drawGrid();
 					} else {
 						ImGui::Dummy(zoomCharSize);
 					}
--- a/src/imgui/ImGuiCharacter.hh
+++ b/src/imgui/ImGuiCharacter.hh
@@ -63,6 +63,7 @@ private:
 	int manualNamBase = 0;
 	int manualRows = 0;
 	int manualColor0 = 16;
+	gl::vecN<2, int> gridPosition;
 
 	gl::Texture patternTex{gl::Null{}}; // TODO also deallocate when needed
 	gl::Texture gridTex   {gl::Null{}};
--- a/src/imgui/ImGuiUtils.cc
+++ b/src/imgui/ImGuiUtils.cc
@@ -335,4 +335,44 @@ void setColors(int style)
 	imColors[size_t(KEY_NOT_ACTIVE)] = 0x80'00'00'00;
 }
 
+std::string formatToString(function_ref<uint8_t(unsigned)> fetch, unsigned begin, unsigned end, std::string_view prefix,
+	unsigned columns, std::string_view suffix, std::string_view formatStr, Interpreter& interp)
+{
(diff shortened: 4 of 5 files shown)


src/imgui/ImGuiCharacter.cc
```diff
@@ -412,6 +413,24 @@ void ImGuiCharacter::paint(MSXMotherBoard* motherBoard)
 				printAddressRange8("Color", colTable.getAddress((mode == SCR1) ? (pattern / 8) : (8 * pattern)));
 			}
 		};
+		auto formatBinaryData = [&](Debuggable& debuggable, uint16_t address) {
+			return formatToString(debuggable, address, address + 7, "", 1, "", "%02X", manager.getInterpreter());
+		};
+		auto copyPatternColorData = [&](byte pattern) {
+			auto* debuggable = motherBoard->getDebugger().findDebuggable("VRAM");
```
m9710797: In theory it's possible that there is no debuggable with the name `VRAM` (should never happen in practice, but in theory it can happen when the user creates his own machine config files, and doesn't use our "standard" names).

The above is rare, so maybe it's acceptable that the copy-to-clipboard functionality doesn't work in that case. But it should not crash! So at least add a: `if (!debuggable) return;`

A better solution would be to (directly) use the same `vram` object we obtain on line 8 in `ImGuiCharacter::paint()` (then the behavior of the viewer itself is in sync with the copy-to-clipboard functionality). At the moment that won't work because the `formatToString()` utility function needs a `Debuggable` (not a `std::span`). But the only thing that `formatToString()` needs is to be able to read a byte from an address. So instead of passing a `Debuggable` parameter we could pass a `std::function<uint8_t(unsigned)> fetch` parameter.

src/imgui/ImGuiCharacter.cc
```diff
@@ -412,6 +413,24 @@ void ImGuiCharacter::paint(MSXMotherBoard* motherBoard)
...
+			auto patData = formatBinaryData(*debuggable, patTable.getAddress(8 * pattern));
+			auto colData = mode == one_of(SCR1, SCR2, SCR4)
+				? formatBinaryData(*debuggable, colTable.getAddress((mode == SCR1) ? (pattern / 8) : (8 * pattern)))
+				: "";
```
m9710797: If I read the code correctly, then you now _always_ (on each frame) format the pattern and color data (or at least when some pattern is hovered). Even if you don't activate the copy-to-clipboard function.

This isn't too bad (it's relatively fast), but is it's also very easy to avoid. So please move this calculation closer to where it's actually used.

src/imgui/ImGuiCharacter.cc
```diff
@@ -412,6 +413,24 @@ void ImGuiCharacter::paint(MSXMotherBoard* motherBoard)
...
+			if ((openCopyPopup = ImGui::BeginPopupContextWindow("PatternCopyPopup"))) {
```
m9710797: Technically this is 100% correct. But I don't like this style: an assignment (to an existing variable) inside an if-statement. And you did have to type double parenthesis to suppress the compiler warning.

Better write this in two statements:
```
openCopyPopup = ...;
if (openCopyPopup) { ...
```
---
I should clarify, in other places we _do_ an assignment inside an if-statement. So why is it acceptable there and not here. Some examples:
```
if (auto pos = myString.find("bla"); pos != std::string::npos) { ...
if (auto* debuggable = debugger.find("VRAM")) { ...
```
In these cases we're assigning to a _new_ variable. And then this syntax helps to limit the scope of that new variable. Also visually it's a good indication that the variable really only makes sense inside that if statement.

But above the assignment is to an existing variable with a scope that lives outside the if statement. So IMHO it's a good style to also visually indicate that the effect of that assignment will last till after the if statement. In other words: the assignment has an effect also outside the if, so put it outside.

For example the following has the same meaning, but I would also NOT use that (because it again hints that the assignment is local in the if, but it is not):
```
if (openCopyPopup = ...;    // don't use this style
    openCopyPopup) { ...
```

src/imgui/ImGuiCharacter.cc
```diff
@@ -438,8 +459,11 @@ void ImGuiCharacter::paint(MSXMotherBoard* motherBoard)
 							ImGui::Image(gridTex.getImGui(),
 								zoomCharSize, {}, charSize);
 						}
-						printPatternColorAddress(pat);
+						printPatternColorAddress(currentPattern);
+						copyPatternColorData(currentPattern);
 					} else {
+						// the popup turns hovered off, so we check if it's still here.
```
m9710797: This is a very useful comment! Thanks.

So if I understand correctly, the only(?) purpose of the `openCopyPopup` variable is to "fix" the behavior of `ImGui::IsItemHovered()` (this calculates the `hovered` variable), right?

Did you experiment with passing some flags to `ImGui::IsItemHovered()`? I have not tried myself, but `ImGuiHoveredFlags_AllowWhenBlockedByPopup` sounds promising (see `imgui.h` around line 1366).

Also because currently when I right-click the preview disappears. But it would be a good visual indication that I indeed right-clicked on the correct item.

src/imgui/ImGuiCharacter.hh
```diff
@@ -3,6 +3,7 @@
 
 #include "ImGuiPart.hh"
 
+#include "MSXDevice.hh"
```
m9710797: Is this #include a left-over from some experiment? I don't think it's still needed.

src/imgui/ImGuiCharacter.hh
```diff
@@ -63,6 +64,8 @@ private:
 	int manualNamBase = 0;
 	int manualRows = 0;
 	int manualColor0 = 16;
+	byte currentPattern = 0;
+	bool openCopyPopup = false;
```
m9710797: In general it's a good idea to minimize the state that's stored between two frames. So I was wondering do we really need these 2 new variables?

For `currentPattern` I think the answer is "yes". Because while the popup is open, we want to keep the preview-pattern fixed, right?

But I'm not sure about `openCopyPopup`. Isn't that variable overwritten in each frame? If so, can't it be a local variable? Or in other words: do you really need the `openCopyPopup` value from the previous frame, or is the value of the current frame sufficient? (Possibly you need to swap the order of some stuff to make this sufficient???)

src/imgui/ImGuiCharacter.cc
```diff
@@ -6,6 +6,7 @@
 #include "ImGuiUtils.hh"
 
 #include "DisplayMode.hh"
+#include "Debugger.hh"
```
m9710797: See comment about using `std::span vram`. Hopefully this #include is then no longer needed.

src/imgui/ImGuiCharacter.hh
```diff
@@ -63,6 +64,8 @@ private:
 	int manualNamBase = 0;
 	int manualRows = 0;
 	int manualColor0 = 16;
+	byte currentPattern = 0;
```
m9710797: Detail: I'm (slowly) trying to move away from the type `byte`, and instead use `uint8_t`. This transition was (mostly) already done in `src/imgui`, so we shouldn't introduce new uses.

src/imgui/ImGuiUtils.cc
```diff
@@ -335,4 +335,30 @@ void setColors(int style)
 	imColors[size_t(KEY_NOT_ACTIVE)] = 0x80'00'00'00;
 }
 
+std::string formatToString(Debuggable& debuggable, unsigned begin, unsigned end, std::string_view prefix, unsigned columns, std::string_view suffix, std::string_view formatStr, Interpreter& interp)
```
m9710797: Detail: in openMSX we don't have a hard limit on the line length. But maybe this line is a bit too long?

src/imgui/ImGuiUtils.hh
```diff
@@ -3,6 +3,7 @@
 
 #include "ImGuiCpp.hh"
 
+#include "Debuggable.hh"
```
m9710797: This #include should be moved to the `ImGuiUtils.cc` file. In this `ImGuiUtils.hh` file a forward declaration like `class Debuggable;` is sufficient (because now all users of `ImGuiUtils.hh` also indirectly `#include Debuggable.hh`, and we should minimize the number of #includes).

---

But above I suggested to instead of a `Debuggable& debuggable` parameter to pass a `std::function` parameter. Even better would be `function_ref<uint8_t(unsigned)> fetch` parameter.

About the difference between `std::function` and `function_ref`: this is similar to the difference between `std::string` and `std::string_view`. The former owns the string (or function), the latter only refers to an existing string (or function). So `std::function` is needed when you e.g. want to store the parameter for later (e.g. it's a callback that you'll only execute later after this function has already returned). But in this case inside `formatToString()` we'll only call `fetch` (we don't store it for later) and then a `function_ref` is sufficient (and it's cheaper).

[std::function_ref](https://en.cppreference.com/w/cpp/utility/functional/function_ref) will be part of c++26, but for now we have our own implementation.

src/imgui/ImGuiCharacter.cc
```diff
@@ -412,6 +413,24 @@ void ImGuiCharacter::paint(MSXMotherBoard* motherBoard)
 				printAddressRange8("Color", colTable.getAddress((mode == SCR1) ? (pattern / 8) : (8 * pattern)));
 			}
 		};
+		auto formatBinaryData = [&](Debuggable& debuggable, uint16_t address) {
+			return formatToString(debuggable, address, address + 7, "", 1, "", "%02X", manager.getInterpreter());
```
m9710797: Very much a detail: instead of passing empty strings as `""` is a tiny bit better to pass `{}`.

The former calls the constructor that takes a `const char*`, that constructor calls `strlen()` and figures out it's an empty string. The latter calls the default constructor (the constructor without parameters) and that one directly constructs an empty string.

Compilers like gcc and clang optimize the call to strlen() away, so both variants generate the same code. But msvc is less smart and there using `{}` does generate better code.

m9710797 (changes requested):
Thanks a lot for working on this! This was indeed functionality that was present in the old debugger but that I (intentionally) did not port to the new debugger. Because it would IMHO clutter the UI too much. But your solution solves that objection.

Several review comments, most are details, only a few bigger ones.

src/imgui/ImGuiCharacter.cc
pvmm: > Also because currently when I right-click the preview disappears. But it would be a good visual indication that I indeed right-clicked on the correct item.

It doesn't work like that. When the mouse hovers the context menu, it stills set `hovered` to `false`. It only set `hovered` to `true` when the mouse is outside the context menu. I don't think it is useful to change the zoomed tile on the right when the selected one is under the context menu.

pvmm: I also moved rawToString to ImGuiUtils.

src/imgui/DebuggableEditor.cc
```diff
@@ -338,10 +297,10 @@ void DebuggableEditor::drawExport(const Sizes& s, Debuggable& debuggable)
 			if (ImGui::Button("Export")) {
 				try {
 					auto output = (exportFormatted == EXPORT_FORMATTED)
-						? formatToString(
-							debuggable, *begin, *end, exportPrefix, exportColumns,
+						? formatToString([&](unsigned address) { return debuggable.read(address); },
+							*begin, *end, exportPrefix, exportColumns,
 							exportSuffix, formatStr, interp)
-						: rawToString(debuggable, *begin, *end);
+						: rawToString([&](unsigned address) { return debuggable.read(address); }, *begin, *end);
```
m9710797: Maybe a detail: factor out the (now repeated) fetch function. And pass this to `formatToString()` and `rawToString()`.
```
auto fetch = [&](unsigned addr) { return debuggable.read(addr); };
...
? formatToString(fetch, ...);
: rawToString(fetch, ...);
```

src/imgui/ImGuiUtils.hh
```diff
@@ -63,6 +63,7 @@ inline void RightAlignText(std::string_view text, std::string_view maxWidthText)
 namespace openmsx {
 
 class BooleanSetting;
+class Debuggable;
```
m9710797: Detail: this forward declaration is not needed (now that we use a `function_ref` rather than a `Debuggable`).

src/imgui/ImGuiCharacter.cc
```diff
@@ -412,6 +412,22 @@ void ImGuiCharacter::paint(MSXMotherBoard* motherBoard)
 				printAddressRange8("Color", colTable.getAddress((mode == SCR1) ? (pattern / 8) : (8 * pattern)));
 			}
 		};
+		auto formatBinaryData = [&](uint16_t address) {
+			return formatToString([&](unsigned addr){ return vram[addr]; }, address, address + 7, {}, 1, {}, "%02X", manager.getInterpreter());
+		};
+		auto copyPatternColorPopup = [&](uint8_t pattern) {
+			if (ImGui::BeginPopupContextWindow("PatternCopyPopup")) {
+				if (ImGui::MenuItem("copy pattern and color data to clipboard")) {
```
m9710797: Detail: start with a capital letter: "**C**opy pattern ...".

m9710797 (changes requested):
Thanks, you've address all my previous review remark. Only some (new) very small details remain.

I've now tested your PR. Before merging I'd still like to improve the behavior of the preview and the context menu. Now if you open the context menu, the preview disappears. I would really like to see the preview when I click the text "Copy pattern and color data to clipboard". Maybe it's possible with something like this:
* Move the call to `copyPatternColorPopup()` higher, till before the `if (hovered) {`.
* This means `copyPatternColorPopup()` uses `currentPattern` from the previous frame. That should be fine, no?
* Return a `bool` from `copyPatternColorPopup()`, indicating whether or not the context menu is currently open (this return value is the `openCopyPopup` variable from your previous version).
  * `copyPatternColorPopup()` is only called once. So alternatively you may also inline that code here. And then instead of a return value you have a local variable.
* Then replace `if (hovered) {` with if (hovered || contextMenuOpen) {`

I have not tested this, but I think this should work. Maybe not exactly this, but *something* like this.

src/imgui/ImGuiCharacter.cc
```diff
@@ -412,6 +412,45 @@ void ImGuiCharacter::paint(MSXMotherBoard* motherBoard)
...
+		};
+		auto getPatternFromGrid = [&]() {
+			return gridPosition.x + 32 * gridPosition.y;
+		};
+		auto copyPatternColorPopup = [&]() {
+			bool popup = ImGui::BeginPopupContextWindow("PatternCopyPopup");
+			if (popup) {
+				auto caption = mode == one_of(TEXT40, TEXT80)
+					? "Copy pattern data to clipboard"
+					: "Copy pattern and color data to clipboard";
```
m9710797: The `caption` thing is a nice tweak. Though it's inconsistent with the check below (line 430). `SCR3` is missing here. Maybe it's good to calculate this condition only once (to avoid inconsistencies):
```bool hasColorData = mode == one_of(SCR1, SCR2);```
Note that I skipped the test for `SCR4` (see line 138).

File: src/imgui/ImGuiCharacter.cc [hunk 16]
```diff
@@ -412,6 +412,45 @@ void ImGuiCharacter::paint(MSXMotherBoard* motherBoard)
...
+				if (ImGui::MenuItem(caption)) {
+					auto pattern = getPatternFromGrid();
+					auto patData = formatBinaryData(patTable.getAddress(8 * pattern));
+					auto colData = mode == one_of(SCR1, SCR2, SCR4)
+						? formatBinaryData(colTable.getAddress((mode == SCR1) ? (pattern / 8) : (8 * pattern)))
+						: "";
+					ImGui::SetClipboardText(tmpStrCat("Pattern Data\n", patData,
+						(colData.empty() ? "" : "Color Data\n"), colData).c_str());
```
m9710797: Very much a detail, personal coding style. Feel free to keep your version, or pick the one below. Whichever you like best:
```
auto text = strCat("Pattern data\n, formatBinaryData(patTable.getAddress(8 * pattern)));
if (hasColorData) {
    auto cp = (mode == SCR1) ? (pattern / 8) : (8 * pattern);
    strAppend(text, "Color data\n", formatBinaryData(colTable.getAddress(cp)));
}
ImGui::SetClipboardText(text.c_str());
```

m9710797 (commented):
Thanks again. Looks very good. I've also tested it now, and it behaves as I'd expect.

Only one small detail about the caption text in `SCR3`.

m9710797:
Merged. Thanks for working on this!

PR #1902, opened 2025-04-21
Context menu copies sprite pattern to clipboard

pvmm:
Mirrors the context menu that copies tile patterns and colours to the clipboard.
image

### DIFF
--- a/src/imgui/ImGuiSpriteViewer.cc
+++ b/src/imgui/ImGuiSpriteViewer.cc
@@ -327,6 +327,20 @@ void ImGuiSpriteViewer::paint(MSXMotherBoard* motherBoard)
 		});
 		ImGui::Separator();
 
+		auto copySpriteDataPopup = [&](std::string text, function_ref<std::string()> supplier) {
+			bool popup = ImGui::BeginPopupContextWindow("CopySpriteDataPopup");
+			if (popup) {
+				if (ImGui::MenuItem(text.c_str())) {
+					ImGui::SetClipboardText(supplier().c_str());
+				}
+				ImGui::EndPopup();
+			}
+			return popup;
+		};
+		auto formatClipboardData = [&](uint16_t address, size_t size) {
+			return formatToString([&](unsigned addr) { return vram[addr]; }, address, address + size - 1, {}, 1, {}, "%02X", manager.getInterpreter());
+		};
+
 		int mode = manMode ? manualMode : vdpMode;
 		int size = manSize ? manualSize : vdpSize;
 		int mag  = manMag ? manualMag  : vdpMag;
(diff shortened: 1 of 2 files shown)


m9710797:
It's still a draft, but I did a review anyway. Looks good, no remarks.
Thanks!

pvmm:
Should we add a context menu to capture each line colour from sprite attributes, even if the side panel already displays them? I think it should, but it doesn't make sense when using mode 1 (single colour) sprites.

m9710797:
You mean right-click when hovered over a sprite in the sprite-attribute section, shows a context menu with "Copy sprite colors to clipboard"? (Or something like that).

IMHO it's not super important, but I don't mind at all if you add that. And indeed only for sprite-mode 2.

src/imgui/ImGuiSpriteViewer.cc
```diff
@@ -327,6 +327,16 @@ void ImGuiSpriteViewer::paint(MSXMotherBoard* motherBoard)
...
 
+		auto copySpriteDataPopup = [&](std::string text, function_ref<std::string()> supplier) {
+			bool popup = ImGui::BeginPopupContextWindow("CopySpriteDataPopup");
+			if (popup) {
+				if (ImGui::MenuItem(text.c_str())) {
+					ImGui::SetClipboardText(supplier().c_str());
+				}
+				ImGui::EndPopup();
+			}
+			return popup;
```
m9710797, line 338: Typically instead of directly using `ImGui::BeginPopupContextWindow()` and `ImGui::EndPopup()` we use our wrappers in `ImGuiCpp.hh`, in this case that would be the function `im::PopupContextWIndow()`.

Though that function returns `void` while in this case you do need the return value from `ImGui::BeginPopupContextWindow()`. So feel free to extend our `im::PopupContextWindow()` functions (all overloads) with this `bool` return value  (e.g. we already did the same for `im::Menu()`).

I think this would be a good change, but it's not required to do this as part of this PR:
* You can include the change of cource.
* Or we (either you or me) can do it in a separate patch (either before or after this PR).
* And if we do it in a separate patch, then maybe we can consider giving many more of our wrapper functions a `bool` return type.

src/imgui/ImGuiSpriteViewer.cc
```diff
@@ -459,6 +476,9 @@ void ImGuiSpriteViewer::paint(MSXMotherBoard* motherBoard)
 			auto zoomSize = float(zm * size);
 			auto fullSize = zoomSize * gl::vec2(8, 4);
 			im::Child("##attrib", {0, fullSize.y}, 0, ImGuiWindowFlags_HorizontalScrollbar, [&]{
+				auto formatColorData = [&](uint16_t address) {
+					return formatToString([&](unsigned addr){ return vram[addr]; }, address, address + size - 1, {}, 1, {}, "%02X", manager.getInterpreter());
+				};
```
m9710797: This `formatColorData()` function is similar to the `formatPatternData()` function above. Maybe we can unify them into a single function that takes both an `address` and a `size` parameter?

src/imgui/ImGuiSpriteViewer.hh
```diff
@@ -60,6 +60,8 @@ private:
 	gl::Texture zoomGridTex{gl::Null{}};
 	gl::Texture checkerTex {gl::Null{}};
 	gl::Texture renderTex  {gl::Null{}};
+	gl::vecN<2, int> patternGridPosition;
+	gl::vecN<2, int> colorGridPosition;
```
m9710797: I didn't have time yet for a detailed review. I don't know if the following suggestion would work, or even if it is a good idea.

My thinking was: `patternGridPosition` and `colorGridPosition` are never both in use at the same time. So maybe having a single variable `gridPosition` is sufficient?

m9710797 (commented):
I haven't tested this, but at first sight your code looks correct.

I do have some suggestions for small cleanups. Feel free to implement those, otherwise I could merge this PR and do those cleanups myself.

m9710797:
Merged. Thanks!

I'll still work on simplifying `ImGui::BeginPopupContextWindow()` (maybe later today?)

m9710797:
Ok, I've pushed the `ImGui::BeginPopupContextWindow()` simplifications.

PR #1900, opened 2025-04-18
Make openMSX compile and run with tcl 9

retrodood:
Fedora Linux 42 is out now, shipping with tcl 9. Make openMSX compile and run with tcl 9.
Note that these changes are not backward compatible so this patch cannot be used with tcl 8 systems.

PR #1904, opened 2025-04-22
replace int with Tcl_Size parameter

pvmm:
Fix compilation errors.

When Tcl 9 is used, Tcl_Size is defined as ptrdiff_t (long int) instead of size_t (int).

### DIFF
--- a/src/commands/TclObject.cc
+++ b/src/commands/TclObject.cc
@@ -3,6 +3,8 @@
 #include "CommandException.hh"
 #include "narrow.hh"
 
+#include <tcl.h>
+
 namespace openmsx {
 
 [[noreturn]] static void throwException(Tcl_Interp* interp)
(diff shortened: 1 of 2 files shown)


m9710797:
See the discussion here: https://github.com/openMSX/openMSX/pull/1900
That PR seems a more evolved version of this PR, so I'll close this one.

At the moment openMSX is meant to use Tcl8, not (yet) Tcl9. Both are not fully compatible, you cannot "just" switch between them.
_Eventually_ we do want to switch to Tcl9, but only when Tcl9 is more widely available. For now we stick to Tcl8, but we can start to prepare our code for Tcl9, e.g. change it so that it compiles with both versions.

Unfortunately the `typedef` `Tcl_Size` is not available in Tcl8. So this PR breaks compilation with Tcl8.
My plan was to use some construction using `#if TCL_VERSION == 8`.

Also note that e.g. in `getListLength()` this PR changes `int` to `Tcl_Size` (so 32-bit to 64-bit), but the function still returns a `unsigned` (32-bit). That truncates the value. Sure that compiles, but it also defeats the purpose of the change. A better fix would be to also change the return type to `size_t`. And then also adapt all callers of this function (that might be a lot of work, I haven't checked yet).

PR #1905, opened 2025-04-22
Modify Tcl_ChannelType so that openMSX can be compiled either with Tcl8 or Tcl9

retrodood:
Modify Tcl_ChannelType using preprocessor directives so that openMSX can be compiled either with Tcl8 or Tcl9.

### DIFF
--- a/src/commands/Interpreter.cc
+++ b/src/commands/Interpreter.cc
@@ -36,7 +36,11 @@ static std::vector<Trace> traces; // sorted on id
 static uintptr_t traceCount = 0;
 
 
+#if TCL_MAJOR_VERSION < 9
 static int dummyClose(ClientData /*instanceData*/, Tcl_Interp* /*interp*/)
+#else
+static int dummyClose(void* unused1, Tcl_Interp* unused2, int unused3)
+#endif
 {
 	return 0;
 }
@@ -55,8 +59,13 @@ static int dummyGetHandle(ClientData /*instanceData*/, int /*direction*/,
 }
 Tcl_ChannelType Interpreter::channelType = {
 	"openMSX console",	 // Type name
+#if TCL_MAJOR_VERSION < 9
 	nullptr,		 // Always non-blocking
 	dummyClose,		 // Close proc
+#else
+	TCL_CHANNEL_VERSION_5,
+	nullptr,		 // Not used with Tcl9
+#endif
 	dummyInput,		 // Input proc
 	Interpreter::outputProc, // Output proc
 	nullptr,		 // Seek proc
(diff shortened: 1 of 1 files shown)


MBilderbeek:
So, with this patch, the `staticbindist` remains on Tcl8.... that makes it easier to already support Tcl9. It will become interesting if there's a conflict, i.e. if something will work in either Tcl8 or Tcl9....

retrodood:
> So, with this patch, the `staticbindist` remains on Tcl8.... that makes it easier to already support Tcl9. It will become interesting if there's a conflict, i.e. if something will work in either Tcl8 or Tcl9....

Aargh, yeah, sorry. I forgot the staticbindist case was in the previous Pull Request. It is up to you folks to decide what is best for it. It is very easy to make it use Tcl9 if that's the way to go. Just let me know.

MBilderbeek:
Well, it's not that easy if you also want the Windows build to succeed. But by all means, go for it if you want to :) it's doable.

m9710797:
@retrodood Thanks again for working on this!

Maybe I wasn't clear in my messages yesterday (sorry about that). But I was also still working on this part.

I was reading Tcl documentation, and I believe it should be possible to fix this in a way that works on both Tcl8 and Tcl9 without using pre-processor `#if / #else / #endif`. Tcl8.6 should already support `TCL_CHANNEL_VERSION_5`. So basically the solution will be your patch, but only keep the Tcl9 parts. I tried this yesterday, and at first sight it seemed to work. But I still needed to do more tests and cleanup the code.

I'm at the office now. This evening when I'm back at home I'll try to finish this.

---

Question: with (only) this PR openMSX now compiles/runs again when linked with Tcl9 instead of Tcl8?

From our side we still need to do more testing before we can "officially" switch to Tcl9. And personally I'd like to wait with this switch until several more Linux distributions ship with libtcl9. But it's good that we've already prepared the code for it. Thanks for your help. And when that time comes we'll have to work on the build system-part, maybe using your changes (from your earlier PR) as a starting point, also thanks for that.

retrodood:
> Well, it's not that easy if you also want the Windows build to succeed. But by all means, go for it if you want to :) it's doable.

I only meant the Linux build. I know nothing of Windows as I have not used it in 25 years now. The last Windows I had was Windows 2000. I have sent another Pull Request about staticbindist.

retrodood:
Wouter, with the #if directives, openMSX master branch built fine on Fedora 42 using g++ v15. 

I did not test with Tcl 8, but your build farm on Github seemed to build it okay on Linux.

m9710797:
I've merged a simplified version of this PR. Thanks again for working on this.
I'm closing this PR now.

I think the current openMSX code should now compile fine with both Tcl8 (this remains the default for now) and Tcl9.
Please re-open if there's still something missing.

PR #1910, opened 2025-04-26
Get rid of GCC -Wstringop-overflow false positive warnings

retrodood:
Get rid of GCC -Wstringop-overflow false positive warnings.


m9710797:
Thanks!
I've also re-analyzed the code. I'm 99% sure it's correct, so indeed a false-positive gcc warning.

PR #1911, opened 2025-04-27
Move from GLEW to epoxy

### LINKED ISSUE #1906: Migrate from GLEW to libepoxy or GLAD
Hi,

we are currently trying to get rid of GLEW in Fedora as it isn't [maintained](https://github.com/nigels-com/glew/issues/416#issuecomment-2646570907) anymore. Also GLEW with the last release only supports GLX and not EGL.

There are two ways moving forward:
* Use [libepoxy](https://github.com/anholt/libepoxy)
* Use [GLAD](https://github.com/Dav1dde/glad/), or check out the [webservice for glad2](https://glad.sh/) to generate the files you need!

With both of them migration should be straight forward.

Thanks for considering!

MBilderbeek:
Closes #1906 .


retrodood:
Hello! I just tested this branch on Fedora 42. The opt build produced a working openMSX executable linked to libepoxy and not libGLEW. Seems good.

retrodood:
The opt build worked fine on FreeBSD 14.2 too!

retrodood:
Ubuntu 24.04.02 LTS verified to work as well. I hope this branch can be merged to master soon.

retrodood:
Debian GNU/Linux 12 builds a binary executable that is correctly linked with libepoxy.

However, the binary will not start because this branch does not have the Tcl 8 and Tcl 9 compatibility fixes that the master branch has. Running openMSX produces the following error:

channel type openMSX console must define closeProc
Aborted

This is because the tcl version is 8.16.13 on Debian 12. So nothing to worry about, the error will go away when the Tcl patches are applied.

MBilderbeek:
Yeah, I fixed that in master a few days ago.

Op wo 30 apr. 2025 23:11 schreef retrodood ***@***.***>:

Vampier:
image

2 tests performed by different user of windows resulted in this. openMSX window briefly pops up and then exits with this error.

MacOSX compile (Apple Silicon) works as expected.

Vampier:
https://github.com/anholt/libepoxy/blob/master/src/dispatch_common.c

Line 229 is throwing the error.

`if (!library_initialized) {
        fputs("Attempting to dlopen() while in the dynamic linker.\n", stderr);
        abort();}`

retrodood:
> https://github.com/anholt/libepoxy/blob/master/src/dispatch_common.c
> 
> Line 229 is throwing the error.
> 
> `if (!library_initialized) { fputs("Attempting to dlopen() while in the dynamic linker.\n", stderr); abort();}`

The C code with all that CONSTRUCTOR stuff looks very hairy indeed.

Chatting briefly with ChatGPT about this, the problem might be in the static library. Yes, I know ChatGPT is not always trustworthy, but it seems to suggest that libepoxy core code might assume it will live inside a dynamically linked library. So it might be worthwhile to try building a DLL and linking to it.

Unfortunately I know nothing about it, but maybe some Windows gurus can help?

MBilderbeek:
> image
> 
> 2 tests performed by different user of windows resulted in this. openMSX window briefly pops up and then exits with this error.
> 
> MacOSX compile (Apple Silicon) works as expected.

That first message, about OpenGL,  is coming from ImGUI. It looks like it has its own OpenGL loader, see `src/3rdparty/imgui/imgui_impl_opengl3_loader.h`

I'm not sure what's going wrong here and why ImGUI would suddenly also fail to load OpenGL (just like epoxy is failing apparently).

cmitu:
@retrodood said
> ...
> Chatting briefly with ChatGPT about this, the problem might be in the static library. Yes, I know ChatGPT is not always trustworthy, but it seems to suggest that libepoxy core code might assume it will live inside a dynamically linked library. So it might be worthwhile to try building a DLL and linking to it.

This (static linkage not working for _libepoxy) seems to be diagnosed in detail at https://github.com/anholt/libepoxy/issues/200#issuecomment-760119583 and a PR is opened to fix is at https://github.com/anholt/libepoxy/pull/304.

Perhaps the PR code can be added as a local patch to the Win32 build steps until upstream decides to include it.

MBilderbeek:
> @retrodood said
> 
> > ...
> > Chatting briefly with ChatGPT about this, the problem might be in the static library. Yes, I know ChatGPT is not always trustworthy, but it seems to suggest that libepoxy core code might assume it will live inside a dynamically linked library. So it might be worthwhile to try building a DLL and linking to it.
> 
> This (static linkage not working for _libepoxy) seems to be diagnosed in detail at [anholt/libepoxy#200 (comment)](https://github.com/anholt/libepoxy/issues/200#issuecomment-760119583) and a PR is opened to fix is at [anholt/libepoxy#304](https://github.com/anholt/libepoxy/pull/304).
> 
> Perhaps the PR code can be added as a local patch to the Win32 build steps until upstream decides to include it.

That's possible, but are we sure the patch is good? There is some open review comment on it, isn't there?

MBilderbeek:
Can a Windows user please try out the latest build from this branch? See GitHub Actions.

f-fix:
a black window opens and then immediately vanishes. when i start it from CMD it does this:
image
i downlaoded it from https://github.com/openMSX/openMSX/actions/runs/27167816767#artifacts

MBilderbeek:
@cmitu I guess that didn't work out then, the behaviour is unchanged.

cmitu:
@MBilderbeek it would seem so, but I think it might be a problem for the VC build - looking at the log I see it restored a previous cache and it might not have re-build `libepoxy.a`.
The _mingw_ build seems to be working though (I checked the action logs and it applied the patch):

image

Is there a way to 'clear' the build/action cache for the VC build and retry the build ?

MBilderbeek:
@cmitu good point. If possible, please try with the latest build. (Also asked @f-fix again.)

f-fix:
> @cmitu good point. If possible, please try with the latest build. (Also asked @f-fix again.)

With the build from https://github.com/openMSX/openMSX/actions/runs/27283500086#artifacts it still crashes right away. still get the same error message too
image

cmitu:
I get the same error on the VC artifact (openmsx-21.0-383-gb5568ed79-windows-vc-x64-bin.zip), so my suggestion doesn't work. Apologies for the noise, the static build patch seemed a promising solution to the runtime error.

MBilderbeek:
> I get the same error on the VC artifact (openmsx-21.0-383-gb5568ed79-windows-vc-x64-bin.zip), so my suggestion doesn't work. Apologies for the noise, the static build patch seemed a promising solution to the runtime error.

No problem, thanks for the suggestion. If you find out more or have more ideas, please speak up (now or later).

andete:
Heya all. With my latest fix I can confirm the epoxy build works on Windows both with the VC build and then MINGW build. It also still works on OSX ARM.

Just to be clear: I also tested _running_ them on windows and Mac laptops.

build/3rdparty/3rdparty.sln
```diff
@@ -3,8 +3,6 @@ Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
 VisualStudioVersion = 15.0.26228.4
 MinimumVisualStudioVersion = 10.0.40219.1
-Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "libepoxy", "libepoxy.vcxproj", "{E78E1412-E9F8-475B-9504-72031C8FBAEA}"
```
MBilderbeek: This doesn't look right, does it?? Can you explain? Also the other changes in this file.

build/3rdparty/libepoxy-1.5.10.diff
```diff
@@ -94,3 +94,24 @@ diff -ur libepoxy-1.5.10.orig/src/dispatch_wgl.c libepoxy-1.5.10/src/dispatch_wg
...
+@@ -295,12 +295,12 @@ get_dlopen_handle(void **handle, const char *lib_name, bool exit_on_fail, bool l
+     if (*handle)
+         return true;
+ 
+-    if (!library_initialized) {
+-        fputs("Attempting to dlopen() while in the dynamic linker.\n", stderr);
+-        abort();
+-    }
+-
+ #ifdef _WIN32
```
MBilderbeek, line 115: As per your commit message, shouldn't this check specifically on MSVC and not Windows in general?

MBilderbeek:
> Heya all. With my latest fix I can confirm the epoxy build works on Windows both with the VC build and then MINGW build. It also still works on OSX ARM.
> 
> Just to be clear: I also tested _running_ them on windows and Mac laptops.

Thanks a lot for reviving this PR! Looks like it's going in the right direction now. I did have some questions though. And a request: can you remove the commits that were a mistake? No problem to force push this branch to do that.

andete:
> > Heya all. With my latest fix I can confirm the epoxy build works on Windows both with the VC build and then MINGW build. It also still works on OSX ARM.
> > Just to be clear: I also tested _running_ them on windows and Mac laptops.
> 
> Thanks a lot for reviving this PR! Looks like it's going in the right direction now. I did have some questions though. And a request: can you remove the commits that were a mistake? No problem to force push this branch to do that.

Yeah, I was not sure it was okay to rewrite history as this is your branch. Will do.

build/3rdparty/3rdparty.sln
andete: As it is build by meson in the CI and not buildable by VC it seemed more honest not to pretend to ;)
It is still on my TODO list to actually install the windows toolchain and see if I can fix that properly.
So far I've relied on the CI and used windows only for testing the running.

build/3rdparty/libepoxy-1.5.10.diff
andete: I've tightened the fix, downloaded the MINGW and VC artifacts again and both still work.

build/3rdparty/3rdparty.sln
MBilderbeek: Right, I had completely forgotten we are using meson now for this library.

It was kind of a hack or workaround indeed, it is now the odd one out, which is the biggest issue. But yeah, as it does work now, we have made a huge step.
Did we already document in the compilation guide that also for the VC build, meson is now required?

build/3rdparty/libepoxy-1.5.10.diff
MBilderbeek: Thanks. Fine to fold that in the original commit as well.

build/3rdparty/3rdparty.sln
andete: No time today, but I first want to fully understand the VC build before I fiddle with it further.

MBilderbeek: I wonder if we aren't causing this `dlopen()` issue by ourselves, because we simply have some bug in the meson build.

andete:
With these final changes the MSVC build no longer requires meson. 
Also added some clarification about the whole dlopen issue.
I've not further squashed commits together as it gets tricky with the merge in between.

Next I'll see into fixing the mingw, linux and Mac builds to also not require meson and ninja for the 3rd party build.

andete:
That last commit makes epoxy build via Make instead of meson on all platforms except MSVC.
I tested the CI Mac build here locally and it works fine.

andete:
I've tested this on MacOSX and windows and both work fine.
Perhaps @MBilderbeek you want to do a quick check on linux as well.
For me it seems fine to merge in as is.

MBilderbeek:
I just tested, but just like the SDL3 branch, it doesn't seem to support ViewPorts :( So, all windows only show inside the main window, until I use `SDL_VIDEO_DRIVER=x11` before the `openmsx` command.... I can't remember that I saw this earlier. It seems it has to do with Wayland... but I have no idea why this is now also showing with libepoxy.

```
manuel     22672   22264  0 18:45 ?        00:00:03 /usr/bin/Xwayland :0 -rootless -noreset -accessx -core -auth /run/user/1000/.mutter-Xwaylandauth.A74EV3 -listenfd 4 -listenfd 5 -displayfd 6 -initfd 7 -byteswappedclients -enable-ei-portal
```
See also #2062 .

MBilderbeek:
@ocornut any idea why the ViewPorts aren't working with epoxy, whilst they worked fine with GLEW? Still on SDL2. And when using Wayland.

ocornut:
Dear imgui has nothing to do with neither GLEW or Epoxy.

Viewports never worked on Wayland.

MBilderbeek:
> Dear imgui has nothing to do with neither GLEW or Epoxy.
> 
> Viewports never worked on Wayland.

Thanks, but somehow they have always worked for me, and I didn't know it was special, so now I'm looking at some kind of regression when moving to epoxy. 
Well, I don't get it, but I guess it is what it is. Thanks anyway, @ocornut.

MBilderbeek:
Did some more investigation: when I use SDL2 + GLEW, SDL2 selects the x11 backend and not wayland. That explains why viewports work (via XWayland). With epoxy or SDL3, the wayland backend is selected, so viewports do not work.

I have no idea why this selection depends on GLEW vs epoxy though....

MBilderbeek:
I did a test. When I use `glewInit()` in my program, SDL2 selects x11. When I don't, it selects `wayland`. Not sure why, as I do `glewInit` /after/ initializing SDL2!

MBilderbeek:
If I only force to link against GLEW, I already get SDL2 to select x11 as backend and not wayland. Very odd. So, no need to call `glewInit` even.

The same binary selects x11 instead of wayland, if I LD_PRELOAD the GLEW library only....

MBilderbeek:
For some reason, SDL2 is selecting x11 because it detects XWayland:
```
DISPLAY=:0
WAYLAND_DISPLAY=wayland-0
INFO: App name: test5
INFO: App version: <unspecified>
INFO: App ID: <unspecified>
INFO: SDL revision: SDL-release-3.4.16-0-gfa2c02bb6 (Debian 3.4.16+ds-1)
INFO: Detected XWayland
DEBUG: SDL chose video backend 'x11'
```
But on my system, strangely enough:
```
$ strings /usr/lib/x86_64-linux-gnu/libSDL2-2.0.so.0 | grep -i -C 3 'Detected XWayland'
$ rg 'Detected XWayland' /usr/lib/x86_64-linux-gnu/*
/usr/lib/x86_64-linux-gnu/libSDL3.so: binary file matches (found "\0" byte around offset 7)
/usr/lib/x86_64-linux-gnu/libSDL3.so.0: binary file matches (found "\0" byte around offset 7)
/usr/lib/x86_64-linux-gnu/libSDL3.so.0.4.16: binary file matches (found "\0" byte around offset 7)
/usr/lib/x86_64-linux-gnu/libSDL3.a: binary file matches (found "\0" byte around offset 68)
```
Yet:
```
$ ldd ./test5 | grep SDL
	libSDL2-2.0.so.0 => /usr/lib/x86_64-linux-gnu/libSDL2-2.0.so.0 (0x00007f16fb79c000)
```
Very confusing.

Test program I used:
```c
#include <stdio.h>
#include <SDL.h>
#include <GL/glew.h>

int main()
{
    printf("Before SDL_Init:\n");

    for (int i = 0; i < SDL_GetNumVideoDrivers(); ++i)
        printf("  %d: %s\n", i, SDL_GetVideoDriver(i));

    printf("1 Selected driver: %s\n", SDL_GetCurrentVideoDriver());


    SDL_LogSetAllPriority(SDL_LOG_PRIORITY_VERBOSE);

    printf("DISPLAY=%s\n", getenv("DISPLAY"));
    printf("WAYLAND_DISPLAY=%s\n", getenv("WAYLAND_DISPLAY"));


    printf("SDL_Init = %d\n", SDL_Init(SDL_INIT_VIDEO));
    printf("2 Selected driver: %s\n", SDL_GetCurrentVideoDriver());

    SDL_Window *window = SDL_CreateWindow(
        "test",
        SDL_WINDOWPOS_UNDEFINED,
        SDL_WINDOWPOS_UNDEFINED,
        640, 480,
        SDL_WINDOW_OPENGL
    );
    printf("3 Selected driver: %s\n", SDL_GetCurrentVideoDriver());

    if (!window) {
        printf("SDL_CreateWindow: %s\n", SDL_GetError());
        return 1;
    }

    SDL_GLContext context = SDL_GL_CreateContext(window);

    printf("4 Selected driver: %s\n", SDL_GetCurrentVideoDriver());
    if (!context) {
        printf("SDL_GL_CreateContext: %s\n", SDL_GetError());
        return 1;
    }

//    GLenum err = glewInit();
    printf("5 Selected driver: %s\n", SDL_GetCurrentVideoDriver());

//    printf("glewInit = %d (%s)\n",
//           err, glewGetErrorString(err));

//    printf("GL_VERSION = %s\n", glGetString(GL_VERSION));

    SDL_DestroyWindow(window);
    SDL_Quit();
}
```
compiled like this: 
```
$ gcc /tmp/test5.c $(sdl2-config --cflags --libs) -lGLEW -lGL -o test5
```

MBilderbeek:
OK, it seems I'm actually using sdl2-compat (Debian used a trick to replace the normal SDL2 with the compat version, and made the classic SDL2 available under a different package name). And SDL2 compat explicitly forces x11 if it detects GLEW, see a.o.:
https://discourse.libsdl.org/t/sdl2-compat-added-initial-shot-at-sdl2-compat-framework/40530 (search for GLEW)

MBilderbeek:
Summarized by ChatGPT:


    Your Debian libsdl2-2.0-0 is the SDL2 compatibility layer over SDL3.

    That layer detects GLEW by the presence of glxewInit.

    When GLEW is present, it forces the X11 backend because GLEW/GLX-based applications can have compatibility problems with native Wayland.

    Thus your old GLEW + SDL2 setup was actually:
    Wayland desktop → XWayland → SDL X11 → ImGui X11 viewports.

    Switching to epoxy removes the glxewInit detection, so SDL2-compat lets SDL3 select native Wayland.

    On native Wayland, the Dear ImGui SDL backend doesn't support platform viewports, hence PlatformHasViewports: NO.

    SDL_VIDEODRIVER=x11 reproduces the old behavior because it explicitly puts you back on the X11/XWayland path.

cmitu:
> OK, it seems I'm actually using sdl2-compat (Debian used a trick to replace the normal SDL2 with the compat version, and made the classic SDL2 available under a different package name). And SDL2 compat explicitly forces x11 if it detects GLEW, see a.o.:

FWIW, this behavior happens also on SDL2 (it's not just a `-compat` thing), forcing `x11` if it detects that the binary has X11 linkage, see:

https://github.com/libsdl-org/SDL/blob/b90ac95029d801c5abc59472ba8e2200dff31e1e/src/video/SDL_video.c#L498-L518

PR #1921, opened 2025-05-31
serialize_core: Support std::optional serialisation.

grauw:
Adding support for serializing std::optional.

### DIFF
--- a/src/serialize_core.hh
+++ b/src/serialize_core.hh
@@ -10,6 +10,7 @@
 #include <cassert>
 #include <initializer_list>
 #include <memory>
+#include <optional>
 #include <string>
 #include <type_traits>
 #include <variant>
@@ -62,6 +63,28 @@ void serialize(Archive& ar, std::pair<T1, T2>& p, unsigned /*version*/)
 	             "second", p.second);
 }
 template<typename T1, typename T2> struct SerializeClassVersion<std::pair<T1, T2>>
+{
+	static constexpr unsigned value = 0;
+};
+
+template<typename Archive, typename T>
+void serialize(Archive& ar, std::optional<T>& o, unsigned /*version*/)
+{
+	bool hasValue = o.has_value();
+	ar.attribute("hasValue", hasValue);
+	if (hasValue) {
+		if (o.has_value()) {
+			ar.serialize("value", *o);
+		} else {
+			T value{};
+			ar.serialize("value", value);
+			o = std::move(value);
+		}
+	} else {
+		o.reset();
+	}
+}
+template<typename T> struct SerializeClassVersion<std::optional<T>>
 {
 	static constexpr unsigned value = 0;
 };


m9710797:
I've reviewed the code, looks good.

But before merging, could you give some more context information? Why do you need this? Are you working on something?

E.g. we do already have several `std::optional` member variables in openMSX, and some are being serialized. Though not the "optional-ness" part itself. A very simple example is `src/RenShaTurbo.cc:31`. In the constructor we either initialize this `optional`, or leave it empty, afterwards this never changes. So in `serialize()` we can check whether this `optional` needs to be included or not.

grauw:
Thanks @m9710797! The change was part of a larger patch series for flash timing improvement that I had laying around since last year, which I wanted to start to push at least partially before it bitrots too much.

I’d thunk I could get this one out of the way since it’s a useful addition by itself, and whether my use of optional is appropriate can then be reviewed when I submit that series for review. But I can also put this change back into that (forthcoming) patch series.

m9710797:
OK thanks. I'm looking forward to that future flash timing stuff.
Let's maybe wait with integrating this until the full patch series is ready.

The reason why I'm now hesitating to already merge this is the following (I only came to this realization after my review):
* In this patch you set `SerializeClassVersion` for `optional<T>` to `value = 0` (same as is done for `std::pair`).
* Version=0 means we're committed to never update the version. This means we can omit the version info from the savestate. But then if we want to keep backwards compatibility with older savestate files, this means we can never change the format of how this struct is serialized.
* For a `std::pair` that's fine (how else can you serialize a pair). But for a `std::optional` I (now) see multiple possibilities.

Let's discuss via an example:
```
std::optional<int> myInt;
...
ar.serialize("myInt", myInt);
```
In openMSX we have two output formats for serialization:
* An in-memory binary format. This is used for reverse-snapshots.
* As XML files. This is used for savestates or for replays.

For the binary format the result will be something like:
```
00                // 1 byte,  for an empty optonal
  or
01 aa bb cc dd    // 5 bytes, for an optional containing the value 0xddccbbaa
```
For "XML" the result will currently look like this:
```
<myInt>
  <has_value>false</has_value>         for an empty optional
</myInt>
  or
<myInt>
  <has_value>true</has_value>          for an optional with a value
  <value>3721182122<value>
</myInt>
```
But a compacter alternative could be this:
```
<myInt/>                         for an empty optional
  or
<myInt>
  <value>3721182122<value>       for an optional with a value
</myInt>
  or maybe even
<myInt>3721182122</myInt>        another alternative  (too cute?   ambiguous for std::optional<std::string>)
```
In other words: use the presence or absence of the `<value>` tag instead of an explicit `<has_value>` tag.

 I'm not saying it's necessarily a good idea to do this. But there is precedence for such XML representation tuning. Search for `CAN_HAVE_OPTIONAL_ATTRIBUTES` or `CAN_COUNT_CHILDREN`.

It's a minor thing, and it shouldn't matter much (we won't serialize that many optionals I think). But because of the backwards compatibility stuff, I'd like to postpone this decision a little longer.

grauw:
Ok!

Btw, the current output is:

```
<myInt hasValue="0"/>                  for an empty optional
  or
<myInt hasValue="1">
  <value>3721182122<value>             for an optional with a value
</myInt>
```

Just now notice it’s saying 0 / 1 instead of true / false. I guess acceptable, hrm.

Value tag could be omitted.

grauw:
It’s been many moons since I wrote this (last summer), but perhaps my rationale for having the `value` tag was that there would be no clash with a potential `hasValue` attribute on whatever is inside the optional. Maybe that’s an unneccessary concern, perhaps with a careful pick for the attribute name.

PR #1922, opened 2025-05-31
AmdFlash: Improvements and timing emulation.

grauw:
This is a patch series that I worked on last year that I’ve updated to latest openMSX.

It improves the FlashROM emulation, including emulating timing so that operations no longer complete instantly.

While I complete my checks, I’d like to invite feedback @m9710797.

### DIFF
--- a/src/memory/AmdFlash.cc
+++ b/src/memory/AmdFlash.cc
@@ -575,7 +575,7 @@ void AmdFlash::write(size_t address, uint8_t value, EmuTime::param time)
 			cmd.clear();
 		}
 	} else if (state == State::PROGRAM) {
-		if (false) {
+		if (checkCommandSuspend(time)) {
 			// do nothing, we're still matching a command, but it is not complete yet
 		} else {
 			cmd.clear();
@@ -668,7 +668,7 @@ bool AmdFlash::checkCommandContinuityCheck()
 bool AmdFlash::checkCommandEraseSector(EmuTime::param time)
 {
 	static constexpr std::array<uint8_t, 5> cmdSeq = {0xaa, 0x55, 0x80, 0xaa, 0x55};
-	if (!statusRegister.eraseSuspend && partialMatch(cmdSeq)) {
+	if (!statusRegister.eraseSuspend && !statusRegister.programSuspend && partialMatch(cmdSeq)) {
 		if (cmd.size() < 6) return true;
 		if (cmd[5].value == 0x30) {
 			clearStatus();
@@ -708,7 +708,7 @@ bool AmdFlash::checkCommandEraseAdditionalSector(EmuTime::param time)
 bool AmdFlash::checkCommandEraseChip(EmuTime::param time)
 {
 	static constexpr std::array<uint8_t, 5> cmdSeq = {0xaa, 0x55, 0x80, 0xaa, 0x55};
-	if (!statusRegister.eraseSuspend && partialMatch(cmdSeq)) {
+	if (!statusRegister.eraseSuspend && !statusRegister.programSuspend && partialMatch(cmdSeq)) {
 		if (cmd.size() < 6) return true;
 		if (cmd[5].value == 0x10) {
 			clearStatus();
@@ -809,6 +809,15 @@ bool AmdFlash::checkCommandSuspend(EmuTime::param time)
 				}
 			}
 		}
+	} else if (state == State::PROGRAM) {
+		if (chip.program.suspend) {
+			if ((cmd[0].value == 0xB0 || (chip.program.enhancedSuspend && cmd[0].value == 0x51))) {
+				if (!syncSuspend.pendingSyncPoint()) {
+					syncOperation.removeSyncPoint();
+					syncSuspend.setSyncPoint(time + chip.program.suspendDuration);
+				}
+			}
+		}
 	} else {
 		UNREACHABLE;
 	}
@@ -823,6 +832,10 @@ void AmdFlash::execSuspend(EmuTime::param /*time*/)
 		status.dataPolling = true;
 		status.eraseTimer = true;
 		setState(State::READ);
+	} else if (state == State::PROGRAM) {
+		statusRegister.ready = true;
+		statusRegister.programSuspend = true;
+		setState(State::READ);
 	} else {
 		UNREACHABLE;
 	}
@@ -830,7 +843,14 @@ void AmdFlash::execSuspend(EmuTime::param /*time*/)
 
 bool AmdFlash::checkCommandResume(EmuTime::param time)
 {
-	if (statusRegister.eraseSuspend) {
+	if (statusRegister.programSuspend) {
+		if ((cmd[0].value == 0x30 || (chip.program.enhancedSuspend && cmd[0].value == 0x50))) {
+			statusRegister.ready = false;
+			statusRegister.programSuspend = false;
+			setState(State::PROGRAM);
+			scheduleProgramOperation(time);
+		}
+	} else if (statusRegister.eraseSuspend) {
 		if (cmd[0].value == 0x30) {
 			statusRegister.ready = false;
 			statusRegister.eraseSuspend = false;
--- a/src/memory/AmdFlash.cc
+++ b/src/memory/AmdFlash.cc
@@ -44,7 +44,7 @@ AmdFlash::AmdFlash(const std::string& name, const ValidatedChip& validatedChip,
 	: motherBoard(config.getMotherBoard())
 	, chip(validatedChip.chip)
 {
-	cmd.reserve(5 + chip.program.pageSize); // longest command is BufferProgram
+	cmd.reserve(5 + chip.program.bufferPageSize); // longest command is BufferProgram
 
 	assert(writeProtectSectors.size() <= chip.geometry.sectorCount);
 	sectors.reserve(chip.geometry.sectorCount);
@@ -391,7 +391,7 @@ uint16_t AmdFlash::peekCFI(size_t address) const
 	case 0x29:
 		return narrow<uint8_t>(std::to_underlying(chip.geometry.deviceInterface) >> 8);
 	case 0x2A:
-		return chip.program.pageSize.exponent;
+		return std::max(chip.program.fastPageSize.exponent, chip.program.bufferPageSize.exponent);
 	case 0x2B:
 		return 0x00;
 	case 0x2C:
@@ -673,31 +673,32 @@ bool AmdFlash::checkCommandProgram()
 bool AmdFlash::checkCommandDoubleByteProgram()
 {
 	static constexpr std::array<uint8_t, 1> cmdSeq = {0x50};
-	return chip.program.fastCommand && checkCommandProgramHelper(2, cmdSeq);
+	return checkCommandProgramHelper(2, cmdSeq);
 }
 
 bool AmdFlash::checkCommandQuadrupleByteProgram()
 {
 	static constexpr std::array<uint8_t, 1> cmdSeq = {0x56};
-	return chip.program.fastCommand && checkCommandProgramHelper(4, cmdSeq);
+	return checkCommandProgramHelper(4, cmdSeq);
 }
 
 bool AmdFlash::checkCommandProgramHelper(size_t numBytes, std::span<const uint8_t> cmdSeq)
 {
-	if (numBytes <= chip.program.pageSize && partialMatch(cmdSeq)) {
+	if (numBytes <= chip.program.fastPageSize && partialMatch(cmdSeq)) {
 		if (cmd.size() == cmdSeq.size()) clearStatus();
 		if (cmd.size() <= cmdSeq.size()) return true;
 		status.dataPolling = ~cmd.back().value & 0x80;
 		if (cmd.size() < (cmdSeq.size() + numBytes)) return true;
-		for (auto i : xrange(cmdSeq.size(), cmdSeq.size() + numBytes)) {
-			const Sector& sector = getSector(cmd[i].addr);
-			if (isWritable(sector)) {
-				auto ramAddr = sector.writeAddress + cmd[i].addr - sector.address;
+		const Sector& sector = getSector(cmd.back().addr);
+		if (isWritable(sector)) {
+			const size_t pageMask = chip.program.fastPageSize - 1;
+			for (auto i : xrange(cmdSeq.size(), cmdSeq.size() + numBytes)) {
+				auto ramAddr = sector.writeAddress + (cmd.back().addr & ~pageMask) + (cmd[i].addr & pageMask) - sector.address;
 				uint8_t ramValue = (*ram)[ramAddr] & cmd[i].value;
 				ram->write(ramAddr, ramValue);
-			} else {
-				statusRegister.sectorLock = true;
 			}
+		} else {
+			statusRegister.sectorLock = true;
 		}
 		status.dataPolling = !status.dataPolling; // immediate completion
 	}
@@ -707,17 +708,17 @@ bool AmdFlash::checkCommandProgramHelper(size_t numBytes, std::span<const uint8_
 bool AmdFlash::checkCommandBufferProgram()
 {
 	static constexpr std::array<uint8_t, 2> cmdSeq = {0xaa, 0x55};
-	if (chip.program.bufferCommand && partialMatch(cmdSeq) && (cmd.size() <= 2 || cmd[2].value == 0x25)) {
+	if (chip.program.bufferPageSize > 1 && partialMatch(cmdSeq) && (cmd.size() <= 2 || cmd[2].value == 0x25)) {
 		if (cmd.size() <= 3) {
 			if (cmd.size() == 3) clearStatus();
 			return true;
 		} else if (cmd.size() <= 4) {
-			if (cmd[3].value < chip.program.pageSize && getSector(cmd[3].addr) == getSector(cmd[2].addr)) {
+			if (cmd[3].value < chip.program.bufferPageSize && getSector(cmd[3].addr) == getSector(cmd[2].addr)) {
 				return true;
 			}
 		} else if (cmd.size() <= size_t(5 + cmd[3].value)) {
-			const size_t pageMask = ~(chip.program.pageSize - 1);
-			if ((cmd.back().addr & pageMask) == (cmd[4].addr & pageMask)) {
+			const size_t pageMask = chip.program.bufferPageSize - 1;
+			if ((cmd.back().addr & ~pageMask) == (cmd[4].addr & ~pageMask)) {
 				status.dataPolling = ~cmd.back().value & 0x80;
 				return true;
 			}
--- a/src/memory/AmdFlash.cc
+++ b/src/memory/AmdFlash.cc
@@ -193,6 +193,7 @@ void AmdFlash::reset()
 	setState(State::READ);
 	syncOperation.removeSyncPoint();
 	erase.reset();
+	program.reset();
 }
 
 void AmdFlash::softReset()
@@ -229,7 +230,7 @@ uint8_t AmdFlash::read(size_t address, EmuTime::param time)
 	const uint8_t value = peek(address, time);
 	if (state == State::STATUS) {
 		setState(State::READ);
-	} else if (state == State::ERROR) {
+	} else if (state == one_of(State::PROGRAM, State::ERROR)) {
 		status.busyToggle = !status.busyToggle;
 	} else if (state == State::ERASE_SECTOR) {
 		status.busyToggle = !status.busyToggle;
@@ -273,7 +274,7 @@ uint8_t AmdFlash::peek(size_t address, EmuTime::param /*time*/) const
 		}
 	} else if (state == State::STATUS) {
 		return statusRegister.raw;
-	} else if (state == one_of(State::ERASE_SECTOR, State::ERASE_CHIP, State::ERROR)) {
+	} else if (state == one_of(State::ERASE_SECTOR, State::ERASE_CHIP, State::PROGRAM, State::ERROR)) {
 		return status.raw;
 	} else {
 		UNREACHABLE;
@@ -503,10 +504,10 @@ void AmdFlash::write(size_t address, uint8_t value, EmuTime::param time)
 		if (checkCommandAutoSelect() ||
 		    checkCommandEraseSector(time) ||
 		    checkCommandEraseChip(time) ||
-		    checkCommandProgram() ||
-		    checkCommandDoubleByteProgram() ||
-		    checkCommandQuadrupleByteProgram() ||
-		    checkCommandBufferProgram() ||
+		    checkCommandProgram(time) ||
+		    checkCommandDoubleByteProgram(time) ||
+		    checkCommandQuadrupleByteProgram(time) ||
+		    checkCommandBufferProgram(time) ||
 		    checkCommandCFIQuery() ||
 		    checkCommandContinuityCheck() ||
 		    checkCommandStatusRead() ||
@@ -555,6 +556,12 @@ void AmdFlash::write(size_t address, uint8_t value, EmuTime::param time)
 			cmd.clear();
 		}
 	} else if (state == State::ERASE_CHIP) {
+		if (false) {
+			// do nothing, we're still matching a command, but it is not complete yet
+		} else {
+			cmd.clear();
+		}
+	} else if (state == State::PROGRAM) {
 		if (false) {
 			// do nothing, we're still matching a command, but it is not complete yet
 		} else {
@@ -773,25 +780,25 @@ void AmdFlash::EraseOperation::reset()
 	buffer.clear();
 }
 
-bool AmdFlash::checkCommandProgram()
+bool AmdFlash::checkCommandProgram(EmuTime::param time)
 {
 	static constexpr std::array<uint8_t, 3> cmdSeq = {0xaa, 0x55, 0xa0};
-	return checkCommandProgramHelper(1, cmdSeq);
+	return checkCommandProgramHelper(1, cmdSeq, time);
 }
 
-bool AmdFlash::checkCommandDoubleByteProgram()
+bool AmdFlash::checkCommandDoubleByteProgram(EmuTime::param time)
 {
 	static constexpr std::array<uint8_t, 1> cmdSeq = {0x50};
-	return checkCommandProgramHelper(2, cmdSeq);
+	return checkCommandProgramHelper(2, cmdSeq, time);
 }
 
-bool AmdFlash::checkCommandQuadrupleByteProgram()
+bool AmdFlash::checkCommandQuadrupleByteProgram(EmuTime::param time)
 {
 	static constexpr std::array<uint8_t, 1> cmdSeq = {0x56};
-	return checkCommandProgramHelper(4, cmdSeq);
+	return checkCommandProgramHelper(4, cmdSeq, time);
 }
 
-bool AmdFlash::checkCommandProgramHelper(size_t numBytes, std::span<const uint8_t> cmdSeq)
+bool AmdFlash::checkCommandProgramHelper(size_t numBytes, std::span<const uint8_t> cmdSeq, EmuTime::param time)
 {
 	if (numBytes <= chip.program.fastPageSize && partialMatch(cmdSeq)) {
 		if (cmd.size() == cmdSeq.size()) clearStatus();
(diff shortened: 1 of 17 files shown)


src/memory/AmdFlash.cc
```diff
@@ -700,31 +700,32 @@ bool AmdFlash::checkCommandProgram()
...
 		if (cmd.size() < (cmdSeq.size() + numBytes)) return true;
-		for (auto i : xrange(cmdSeq.size(), cmdSeq.size() + numBytes)) {
-			const Sector& sector = getSector(cmd[i].addr);
-			if (isWritable(sector)) {
-				auto ramAddr = sector.writeAddress + cmd[i].addr - sector.address;
+		const Sector& sector = getSector(cmd.back().addr);
+		if (isWritable(sector)) {
+			const size_t pageMask = chip.program.fastPageSize - 1;
+			for (auto i : xrange(cmdSeq.size(), cmdSeq.size() + numBytes)) {
+				auto ramAddr = sector.writeAddress + (cmd.back().addr & ~pageMask) + (cmd[i].addr & pageMask) - sector.address;
```
m9710797: Minor. The subexpression `sector.writeAddress + (cmd.back().addr &~pageMask) - sector.address` is loop invariant. We could factor it out and calculate it only once outside the loop.

The motivation is not performance (there's a good chance that the compiler will do this transformation automatically anyway). But to make it more clear to human readers that only the `cmd[i].add & pageMask` varies within the loop. So make it more clear what you say in the commit message: all writes go to the same page.

src/memory/AmdFlash.cc
```diff
@@ -719,11 +719,25 @@ bool AmdFlash::checkCommandProgramHelper(size_t numBytes, std::span<const uint8_
...
+
+			// For fast program commands, if you write 0x7F and then 0xF7 to the same empty
+			// flash memory location within the same command, the final value is 0x77.
+			// Tested on: M29W640GB.
 			for (auto i : xrange(cmdSeq.size(), cmdSeq.size() + numBytes)) {
-				auto ramAddr = sector.writeAddress + (cmd.back().addr & ~pageMask) + (cmd[i].addr & pageMask) - sector.address;
-				uint8_t ramValue = (*ram)[ramAddr] & cmd[i].value;
-				ram->write(ramAddr, ramValue);
+				auto& value = programBuffer[cmd[i].addr & pageMask];
+				value = value ? cmd[i].value & *value : cmd[i].value;
```
m9710797: Not sure if this is more readable:
```
value = cmd[i].value & value.value_or(255);
```

src/memory/AmdFlash.cc
```diff
@@ -758,12 +772,25 @@ bool AmdFlash::checkCommandBufferProgram()
...
 			if (cmd.back().value == 0x29 && getSector(cmd.back().addr) == sector) {
 				if (isWritable(sector)) {
-					// TODO de-duplicate same-address writes to the last one
-					for (auto i : xrange<size_t>(4, cmd.size() - 1)) {
-						auto ramAddr = sector.writeAddress + cmd[i].addr - sector.address;
-						uint8_t ramValue = (*ram)[ramAddr] & cmd[i].value;
-						ram->write(ramAddr, ramValue);
+					const size_t pageMask = chip.program.bufferPageSize - 1;
+					const size_t programAddress = cmd[4].addr & ~pageMask;
+					programBuffer.assign(chip.program.bufferPageSize, std::nullopt);
```
m9710797: If I understand correctly `programBuffer` will be sparse, mostly filled with `nullopt`, right?
Though the maximum value for `bufferPageSize` seems to be 256? And then iterating over a (small) sparse array (below on line 786) may still be perfectly fine.

m9710797 (commented):
Looking at this patch in isolation I'd say `programBuffer` could be a local variable (it doesn't retain state outside the scope where it's assigned and cleared again).

But when I look at the final version I see `programBuffer` is used in a more complex way. So I assume this patch is just some intermediate step?

src/memory/AmdFlash.cc
```diff
@@ -563,6 +573,18 @@ void AmdFlash::write(size_t address, uint8_t value, EmuTime::param /*time*/)
...
 			cmd.clear();
 		}
+	} else if (state == State::ERASE_SECTOR) {
+		if (checkCommandEraseAdditionalSector(time)) {
+			// do nothing, we're still matching a command, but it is not complete yet
+		} else {
+			cmd.clear();
+		}
+	} else if (state == State::ERASE_CHIP) {
+		if (false) {
```
m9710797, line 572: This `if (false)` is unexpected (and it's still like this in the final code).
Is this an oversight? If not then maybe add an extra comment for why it's written like this.

src/memory/AmdFlash.hh
```diff
@@ -262,14 +278,18 @@ public:
...
+		std::optional<bool> erasePending = {};
 		ptrdiff_t writeAddress = -1;
 		const uint8_t* readAddress = nullptr;
 
 		std::weak_ordering operator<=>(const Sector& sector) const { return address <=> sector.address; }
 		bool operator==(const Sector& sector) const { return address == sector.address; }
+
+		template<typename Archive>
+		void serialize(Archive& ar, unsigned version);
 	};
```
m9710797: Not sure if this is a good idea, but something to consider.

Most of this `struct Sector` is fixed (set once during the constructor and then never changes again).
In fact I think only the `erasePending` field still changes after the constructor, right?

Would it make sense to split this struct into a mutable and immutable part? And have two parallel `sectors` vectors (with two different names of course). That would also simplify the serialization code for (the mutable part of) `sectors`.

src/memory/AmdFlash.hh
```diff
@@ -262,14 +278,18 @@ public:
...
 
 	struct Sector {
-		size_t address;
-		power_of_two<size_t> size;
-		bool writeProtect;
+		size_t address = 0;
+		power_of_two<size_t> size = 1;
+		bool writeProtect = false;
+		std::optional<bool> erasePending = {};
```
m9710797: The meaning of this `optional<bool>` wasn't immediately clear to me.
I think you want the following 3 states, right?
* No erase operation is running at all.
* Erase is still running (or has not yet started).
* Erase operation on this sector has finished.

We should at least document more clearly which of these 3 maps on which state of the `optional<bool>`.
But I also wonder if an `enum` isn't better suited for this.

src/memory/AmdFlash.hh
```diff
@@ -3,9 +3,13 @@
 
 #include "serialize_meta.hh"
 
+#include "EmuDuration.hh"
 #include "EmuTime.hh"
+#include "MemBuffer.hh"
```
m9710797: This #include doesn't seem to be used (anymore?).

src/memory/AmdFlash.cc
grauw: For normal program it is sized according to the `fastPageSize` definition, which is usually 1 but on the M29W640G it is 8. Here, single-byte programming will probably be the most common in practice, and it will usually be sparse.

For buffer program it is sized according to the `bufferPageSize` definition, which is 256 on the S29GL064S and 32 on the S29GL064N and M29W640G. Here, full buffer programming will probably be the most common, that is the most efficient after all, and it usually won’t be sparse.

src/memory/AmdFlash.cc
```diff
@@ -648,49 +670,115 @@ bool AmdFlash::checkCommandContinuityCheck()
...
 	}
 	return false;
 }
 
+void AmdFlash::scheduleEraseAlgorithm(EmuTime::param time)
+{
+	if (state == State::ERASE_SECTOR && !status.eraseTimer) {
+		syncAlgorithm.removeSyncPoint();
+		syncAlgorithm.setSyncPoint(time + chip.erase.timerDuration);
+	} else if (std::ranges::any_of(sectors, [](auto& s) { return s.erasePending.value_or(false); })) {
```
m9710797: Very much a detail: I'd expected `const auto&` parameter here.

And in some other place (not in this patch) I now see that you pass `auto` (instead of `const auto&`).

Not that this will make any difference in the generated code, but it's good to express the intention to human readers of the code.

src/memory/AmdFlash.hh
```diff
@@ -463,7 +475,10 @@ namespace AmdFlashChip
 			}, 2
 		},
 		.erase{.duration = EmuDuration{0.5}},
-		.program{.shortAbortReset = true, .fastPageSize = 8, .bufferPageSize = 32},
+		.program{
+			.shortAbortReset = true, .fastPageSize = 8, .bufferPageSize = 32,
+			.duration = EmuDuration{0.00001}, .bufferDuration = EmuDuration{0.000005}
```
m9710797: Detail: EmuDuration has static methods like this (a bit like named-constructors):
```
static constexpr EmuDuration usec(unsigned x);
```
That would allow to write:
```
.duration = EmuDuration::usec(10),
```
Maybe that's clearer?

And feel free to extend this set of functions. Maybe also one for nanoseconds, and for second that takes a `double` (instead of a `unsigned`) parameter?

src/memory/AmdFlash.cc
grauw: Just didn’t want to break the pattern of `if (checkCommandX()) { /* do nothing */ } else { cmd.clear }`.

m9710797:
I did a 1st quick review pass over the code.  If I have time tomorrow morning I might do a 2nd pass. Or I'll wait until you've processed my current remarks.

src/memory/AmdFlash.hh
```diff
@@ -235,6 +235,31 @@ public:
...
+		bool eraseStatus    = false; // SR(5)
+		bool programStatus  = false; // SR(4)
+		bool bufferAbort    = false; // SR(3)
+		bool programSuspend = false; // SR(2)
+		bool sectorLock     = false; // SR(1)
+		bool continuity     = false; // SR(0)
+
+		template<typename Archive>
+		void serialize(Archive& ar, unsigned version);
+	};
```
m9710797: No need to change this, this is perfectly fine.

But this reminded me of bit-fields. However the problem with bit-fields is that the c++ standard does not specify a specific layout. Thus in portable code you cannot use them to match a specific layout in a hardware register. (In practice it may work fine if you stick to a specific (version of a) compiler).

Anyway, at work we recently needed "portable bit-fields", and we came up with this small utility class. See here for an example: https://godbolt.org/z/W4816EGvo
* Lines 1-20 define the utility class
* Lines 25-50 define "operation status" and "status register"
* Lines 55-115 show some example usages. And on the right you can see the generated x86 code, seem pretty efficient (optimal?).

Again, no need to change. I just wanted to show this alternative.

grauw:
Thanks for the reviews @m9710797, I just got back from visiting my brother in Groningen and I’m pretty knackered from the 4 hours total of driving, so I’ll follow up more during the week!

src/memory/AmdFlash.cc
```diff
@@ -894,6 +897,40 @@ bool AmdFlash::checkCommandBufferProgram()
...
+{
+	assert(!syncAlgorithm.pendingSyncPoint());
+	syncAlgorithm.setSyncPoint(time + chip.program.duration);
+}
+
+void AmdFlash::execProgramAlgorithm(EmuTime::param time)
+{
+	assert(state == State::PROGRAM);
+	const Sector& sector = getSector(programAddress);
+	for (size_t i : xrange(programBuffer.size())) {
```
grauw: For optimisation, I can make this index a member variable (no need to serialize), and replace the ranges check later with a comparison.

src/memory/AmdFlash.hh
grauw: Couple of thoughts:

1. If we have separate arrays, then we may have to rely more on indexing instead of those nice range-fors (guess we could use `zip`), and also do more checks to ensure they have the same size (e.g. post serialization).

2. It may be interesting to explore extracting an `EraseCommand { index; buffer { mark[] } }` and `ProgramCommand { index; address; buffer { opt<value>[] } }` class.

3. In the old [sparse persistence](https://github.com/openMSX/openMSX/pull/1629#discussion_r1609524490) pull request (that I do want to revisit sometime), I am also storing a mutable `modified` flag on sectors. Though if we would follow my idea in the second comment there, we would track address ranges separately.

src/memory/AmdFlash.hh
grauw: If I were to use regular bitfields, for the normal access they can optimise well, though it’s a bit spotty:

https://godbolt.org/z/x69d1csfh

And then I could serialise the integer representation of the bitfield (using `getAB` / `setAB` in the example), which would allow reuse of the current status field serialisation without worrying about back compat, and permit adding more status bits in the future without changing the serialization format as well.

I notice btw that I hadn’t yet dealt with the status field serialisation back compat here.

grauw: But I also like your approach, just wanted to highlight another option. Either way I should probably store status as an (or two) integer field(s) rather than serialise the member fields separately, it will make back compat easier.

src/memory/AmdFlash.hh
m9710797: It was just an idea (triggered by the serialization). But if you say it makes other code more complex, then it's not a good trade-off.

src/memory/AmdFlash.hh
m9710797: Yes, using builtin-bit-fields would be ideal. Unfortunately the c++ standard does not guarantee a specific layout for bitfields (e.g. it may allocate from MSB to LSB or from LSB to MSB. Or it may insert extra padding bits, or it maybe ignore the request and allocate each bit in a different word anyway).

For your specific godbolt example all compilers do seem to pick the same layout, so in practice it does often work fine, but that's not guaranteed.

So unfortunately in portable code, builtin-bit-fields cannot be used to match a specific bit-layout from some external specification (e.g. like a hardware status register in the flash-rom-specification).

grauw: It is portable though! Not a union or cast in sight. Just may not result in optimal code... And it's kinda boilerplaty.

grauw: Can you add `BitField` to `utils`? I could do it too, but it’s not my code to credit.

m9710797: Feel free to commit this utility (if you mention me in the commit message (or not) that's more than sufficient).
Because it's a bit weird to already commit this utility without (in the next few commits) not also having a use for it.

I did write some comments. So if you want to add it, here's a good starting point:
```
#ifndef BITFIELD_HH
#define BITFIELD_HH

#include <cstddef>

// "Portable" bitfield class
//
// This is inspired by:
//   https://blog.codef00.com/2014/12/06/portable-bitfields-using-c11
// (and stripped down to the bare minimum).
//
// In theory C++ bitfields are not suited(*1) to replicate an exact
// bit-layout (e.g. of some hardware register or of some binary file-format).
// The C++ standard leaves many details of bitfield layout up to the
// implementation, including things like:
//   * layout from LSB to MSB, or the other way around.
//   * whether fields can cross byte or word boundaries (how large
//     is a single word, whether padding bits are inserted, ...).
//   * ...
// Fortunately in C++, with a relatively simple utility like this, we can take
// full control. Use it like this:
//
// template<size_t POS, size_t WIDTH = 1>
// using BF8 = BitField<uint8_t, POS, WIDTH>;
//
// union MyRegister { // (*2)
//     uint8_t raw = 0;  // the full raw 8 bit value
//     BF8<0> bit0;      // a single bit at bit0
//     BF8<1, 2> bit12;  // a 2-bit field covering bits 1 and 2
//     BF8<3, 5> bit37;  // a 6-bit field covering bits 3 through 7
//     BF8<0, 3> bit012; // a 3-bit field, overlapping with 'bit0' and 'bit12'
// };
// ...
// MyRegister reg;
// ...
// reg.bit0 = true;
// reg.bit12 = 2;
// int a = reg.bit37;
//
// (*1) If you stick to a specific compiler like only gcc, then you can often
//      make it work. But this relies on implementation-specific behavior, and
//      is not guaranteed by the C++ standard.
//
// (*2) Using a union, and then reading/writing fields in arbitrary order may
//      also seem like undefined behavior, normally you're only allowed to read
//      the 'active' member (=last written field). Though here it's fine: all
//      members are trivially layout-compatible and begin with the same
//      underlying type (e.g., uint8_t), satisfying the "common initial
//      sequence" rule.

template<typename T, size_t POS, size_t WIDTH>
class BitField {
    static constexpr T Mask = (T(1) << WIDTH) - T(1);

public:
    template<typename T2>
    BitField &operator=(T2 newVal) {
        value = (value & ~(Mask << POS)) | ((newVal & Mask) << POS);
        return *this;
    }

    operator T() const {
        return (value >> POS) & Mask;
    }

private:
    T value;
};

#endif
```

grauw: I’m making the following modifications:

1. To avoid `warning: implicit conversion loses integer precision: 'int' to 'unsigned char'`:

```diff
-        value = (value & ~(Mask << POS)) | ((newVal & Mask) << POS);
+        value = T((value & ~(Mask << POS)) | ((newVal & Mask) << POS));
```

2. To avoid unexpected results when treating fields as boolean, e.g. `status.dataPolling = ramValue & 0x80`, allow explicitly specifying T2 type (cf. standard bitfields):

```diff
-template<typename T, size_t POS, size_t WIDTH>
+template<typename T, size_t POS, size_t WIDTH, typename T2 = T>
 class BitField {
```

```diff
-    template<typename T2>
     BitField &operator=(T2 newVal) {
```

m9710797: OK. If you want you could also remove the `T2` parameter completely.

grauw:
Updated patch series;

* Use `BitField` for status.
* Use `EmuDuration` `sec`, `msec`, `usec` helper methods.
* Introduce `EraseOperation` struct, `Sector` no longer tracks pending state but gained an `index` field.
* Introduce `ProgramOperation` struct, with `index` field.
* Fixed serialisation.
* Some other bits & bobs.

Marked as draft because I still need to do thorough testing and final checks before merging. Also, the std::optional serialisation (#1921) remains an open question.

src/memory/AmdFlash.cc
```diff
@@ -652,18 +696,82 @@ bool AmdFlash::checkCommandEraseChip()
...
+			setState(State::IDLE);
+		} else {
+			syncOperation.setSyncPoint(time + chip.erase.duration);
+		}
+	}
+}
+
+bool AmdFlash::EraseOperation::isPending(const Sector& sector) const
+{
+	return sector.index >= index && sector.index < buffer.size() && buffer[sector.index];
```
m9710797: Detail:  this is the pattern:
```
x >= low && x < high
```
I prefer to write that like:
```
low <= x && x < high
```
So both comparisons are "in the same direction". That's closer to what you'd write mathematically. And there are proposals to allow this notation in a future c++ version, but unfortunately not yet c++26 :(
```
low <= x < high
```

src/memory/AmdFlash.cc
```diff
@@ -652,18 +696,82 @@ bool AmdFlash::checkCommandEraseChip()
...
+}
+
+void AmdFlash::EraseOperation::markPending(const Sector& sector)
+{
+	for (size_t i = buffer.size(); i <= sector.index; ++i) {
+		buffer.push_back(false);
+	}
```
m9710797, line 786: Can this loop be replaced with something like `buffer.resize(sector.index + 1, false)`?
Possibly with some check to only enlarge the buffer (in case `sector.index < buffer.size(), or maybe this function has a precondition that this won't happen?).

src/memory/AmdFlash.hh
```diff
@@ -85,6 +88,16 @@ public:
 		}
 	};
 
+	struct Erase final {
```
m9710797: Is `Erase` meant to be used in some class hierarchy? If not then the `final` annotation is confusing (to me). It's not wrong, but it also doesn't bring any benefits.

Same remark for `EraseOperation` below.

src/memory/AmdFlash.hh
```diff
@@ -316,8 +337,29 @@ private:
...
+		friend class AmdFlash;
+		explicit SyncOperation(Scheduler& s) : Schedulable(s) {}
+		void executeUntil(EmuTime::param time) override {
+			auto& outer = OUTER(AmdFlash, syncOperation);
+			outer.execOperation(time);
+		}
+	} syncOperation;
 };
 SERIALIZE_CLASS_VERSION(AmdFlash, 4);
+SERIALIZE_CLASS_VERSION(AmdFlash::EraseOperation, 1);
```
m9710797: Version=1 is the default. It's fine to state it explicitly, but you may also omit it.

src/memory/AmdFlash.cc
```diff
@@ -798,13 +906,23 @@ bool AmdFlash::partialMatch(std::span<const uint8_t> dataSeq) const
 	});
 }
 
+void AmdFlash::execOperation(EmuTime::param time)
+{
+	if (state == one_of(State::ERASE_SECTOR, State::ERASE_CHIP)) {
+		execEraseOperation(time);
+	} else {
+		UNREACHABLE;
+	}
```
m9710797, line 1040: This is fine. Though a more direct way to express this is:
```
assert(state == one_of(State::ERASE_SECTOR, State::ERASE_CHIP));
execEraseOperation(time);
```

UPDATE:
I see that in a follow-up commit you add an `else if`. In that case the pattern with `UNREACHABLE` makes sense.

src/memory/AmdFlash.cc
```diff
@@ -43,14 +44,16 @@ AmdFlash::AmdFlash(const std::string& name, const ValidatedChip& validatedChip,
...
 	, chip(validatedChip.chip)
+	, syncOperation(motherBoard.getScheduler())
 {
 	cmd.reserve(5 + chip.program.bufferPageSize); // longest command is BufferProgram
 
 	assert(writeProtectSectors.size() <= chip.geometry.sectorCount);
 	sectors.reserve(chip.geometry.sectorCount);
 	for (size_t address = 0; const Region& region : chip.geometry.regions) {
-		for (size_t regionSector = 0; regionSector < region.count; ++regionSector, address += region.size) {
+		for (size_t regionSector = 0, index = 0; regionSector < region.count; ++regionSector, address += region.size, ++index) {
```
m9710797: Technically nothing wrong with this.
Personally I'm not a big fan of for loops that do multiple things (via the comma-operator) in the "next"-expression. Here you already do 3 different things. I don't think that's super readable.
But it's your code, pick the style that you prefer.


`regionSector` and `index` are always equal. (Almost certainly the compiler will figure this out and only use one variable). Do you think the intention of the code is more clearly expressed by using two variables? Or just one?

src/memory/AmdFlash.cc
```diff
@@ -8,6 +8,7 @@
 #include "MSXCliComm.hh"
 #include "MSXException.hh"
 
+#include "enumerate.hh"
```
m9710797: Maybe I overlooked it, but this patch doesn't use `enumerate`. Is this a left over from some experiment?

src/memory/AmdFlash.hh
```diff
@@ -316,8 +337,29 @@ private:
 		BF8<1> sectorLock;
 		BF8<0> continuity;
 	} statusRegister;
+
+	struct EraseOperation final {
+		size_t index = 0;
+		std::vector<uint8_t> buffer;
```
m9710797: Is this conceptually meant as a `vector<bool>`? I know that `vector<bool>` is weird in c++, and that you might want to avoid it for that reason. But maybe still document that as such?
```
std::vector</*bool*/uint8_t> buffer;
```

grauw:
SonarCloud [doesn’t appreciate](https://sonarcloud.io/project/issues?pullRequest=1922&open=AZdQ3UMswwbE3kvy2jr-&id=openMSX_openMSX) the BitField.

src/memory/AmdFlash.cc
```diff
@@ -652,18 +696,82 @@ bool AmdFlash::checkCommandEraseChip()
...
 	}
 	return false;
 }
 
+void AmdFlash::scheduleEraseOperation(EmuTime::param time)
+{
+	if (state == State::ERASE_SECTOR && !status.eraseTimer) {
+		syncOperation.removeSyncPoint();
+		syncOperation.setSyncPoint(time + chip.erase.timerDuration);
+	} else if (std::ranges::any_of(erase.buffer, std::identity{})) {
```
m9710797, line 738: If my assumption below is correct (that logically this is a `vector<bool>`), you could write this using our `contains()` helper function in `stl.hh`.
```
contains(erase.buffer, true)
```
I'm not sure which is more clear. Pick whichever you like best.

m9710797:
> SonarCloud [doesn’t appreciate](https://sonarcloud.io/project/issues?pullRequest=1922&open=AZdQ3UMswwbE3kvy2jr-&id=openMSX_openMSX) the BitField.

Yeah, in general it's not allowed to write to one field of a union (=make it the active member) and then read from another field in the same union. In this specific case it is fine because all fields are struct with the same initial sequence (that sequence is just a single `uint8_t` member).

But it is tricky. good that sonar cloud warns for such things, but I'm fairly sure it is a false positive in this case.

src/memory/AmdFlash.cc
```diff
@@ -780,6 +792,55 @@ void AmdFlash::EraseOperation::reset()
...
+void AmdFlash::execSuspend(EmuTime::param /*time*/)
+{
+	if (state == State::ERASE_SECTOR) {
+		statusRegister.ready = true;
+		statusRegister.eraseSuspend = true;
+		status.dataPolling = true;
+		status.eraseTimer = true;
+		setState(State::IDLE);
+	} else {
+		UNREACHABLE;
```
m9710797, line 840: This is fine. But isn't it simpler to use `assert(state == State::ERASE_SECTOR);`?
And similar for some other places where you use `UNREACHABLE`.

UPDATE: OK, I now see that in a follow-up commit you added an `else if`. In that case the pattern with `UNREACHABLE` makes sense.

src/memory/AmdFlash.cc
grauw: Ai, I made a mistake here, `index` is supposed to be assigned to 0 in the top level. I want the total sector index.

src/memory/AmdFlash.hh
grauw: The serialisation doesn’t like `vector<bool>`.

src/memory/AmdFlash.cc
grauw: Will do, didn’t realise `resize` had a default value parameter, I thought all values would default-initialise.

src/memory/AmdFlash.hh
grauw: For me `final` indicates exactly that it’s not meant to be used in a class hierarchy, preventing such use…

src/memory/AmdFlash.cc
grauw: Will get rid of `index` and use `.index = sectors.size()` instead.

src/memory/AmdFlash.hh
m9710797: OK fine. Keep it if you think it's important to explicitly indicate that.

OTOH, by far most structs/classes are not meant to be used in a class hierarchy. And c++ is already verbose enough.

grauw: I’ll remove it. I thought it would help give the compiler cues for optimisation, but the absence of virtual pretty much already does that.

grauw:
Addressed comments and some tweaks based on testing. E.g.:

* Fixed some indexing errors when erasing (additional) sectors.
* My test suite showed that even if a bit raise error occurs the memory value is still written.
* I removed the `ProgramOperation::buffered` flag since it unnecessarily complicated things without solid testing basis to confirm such timing details.

src/memory/AmdFlash.cc
```diff
@@ -728,7 +728,7 @@ static constexpr std::initializer_list<enum_string<AmdFlash::State>> stateInfo =
 	{ "IDENT",  AmdFlash::State::IDENT },
 	{ "CFI",    AmdFlash::State::CFI },
 	{ "STATUS", AmdFlash::State::STATUS },
-	{ "PRGERR", AmdFlash::State::PRGERR }
+	{ "ERROR",  AmdFlash::State::ERROR }
```
m9710797: I fully agree that bw-compat with old savestates is not needed for this (very unlikely this state occurs in a savestate).

So just FYI, if you do want this for some other thing in the future. It's fine to have multiple strings in this list that map to the same enum-value. So not removing the old line is all you need to do (and add a comment that it's for bw-compat).

src/memory/AmdFlash.cc
```diff
@@ -692,11 +692,27 @@ bool AmdFlash::checkCommandProgramHelper(size_t numBytes, std::span<const uint8_
...
+			// For fast program commands, if you write 0x7F and then 0xF7 to the same empty
+			// flash memory location within the same command, the final value is 0x77.
+			// Tested on: M29W640GB.
+			for (size_t i : xrange(cmdSeq.size(), cmdSeq.size() + numBytes)) {
+				auto& value = programBuffer[cmd[i].addr & pageMask];
+				value = cmd[i].value & value.value_or(0xff);
 			}
+
+			size_t ramAddr = programAddress - sector.address + sector.writeAddress;
+			for (const auto value : programBuffer) {
```
m9710797: Detail: for `optional<uint8_t>` it technically doesn't matter (the current code is also correct). But I still wonder if this is a typo or not.

My expectation for declaring the type in a range-for loop is either:
* `auto`: if you intentionally want to make a copy
* `const auto&`: if you just want to use the value (not modify it)
* `auto&`: if you want to (use and) modify it.
* (less common) an explicit type: for an intentional type-conversion

But `const auto` is a mix between the 1st and 2nd. Therefor I wonder:
* (Most likely) did you forget the `&`?
* Did you intend to make a copy? Then `const` is not wrong, but also doesn't contribute much.

src/memory/AmdFlash.cc
grauw: Oh nice!

src/memory/AmdFlash.cc
grauw: I thought I wouldn’t need `&` since it’s a “small value” that easily fits in registers, and adding the `&` might force it to use pointers instead. Probably am thinking too hard on this.

PR #1925, opened 2025-06-05
Extend EmuDuration static methods

grauw:
To include `nsec` and floating point versions of `sec`, `msec`, `usec` and `hz`. Using concepts to prevent ambiguity errors.

Also remove the `double` overload of the constructor, which was interpreted very differently from the `uint64` version. Use `EmuDuration::sec()` instead.

I resisted the urge to rename `sec`, `msec` and `usec` to `s`, `ms` and `us` respectively.

I considered using `std::unsigned_integral` for the static methods, however that’s less ergonomic since you have to add `U` suffixes to all integer literals that you pass in, and negative floating point numbers can’t be prevented either anyway.

### DIFF
--- a/src/EmuDuration.hh
+++ b/src/EmuDuration.hh
@@ -39,14 +39,36 @@ public:
 	constexpr explicit EmuDuration(double duration)
 		: time(uint64_t(duration * MAIN_FREQ + 0.5)) {}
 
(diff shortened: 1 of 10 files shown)


m9710797:
Thanks!

PR #1927, opened 2025-06-06
macOS: Fix CI build for pull requests.

grauw:
Code sign identity is now specified in a GitHub Actions repository variable. This will allow forks to do their own code signing. If not configured, then the code is not signed (and can’t be distributed) but the build will succeed.

See 244eabd for the previous fix, which was lost in 2d060e7 / a281e5d.

I configured the following secrets:

afbeelding

And the following variables:

afbeelding

In the current openMSX repository the secrets will already have been configured, but for this change the `CODE_SIGN_IDENTITY` variable should be added with the value `748867M822`.

### DIFF
--- a/.github/workflows/build_release_from_tag_via_tarball.yml
+++ b/.github/workflows/build_release_from_tag_via_tarball.yml
@@ -108,8 +108,8 @@ jobs:
       with:
         os: ${{ matrix.os }}
     - name: Perform build on extracted tarball
-      env: 
-        CODE_SIGN_IDENTITY: 748867M822
+      env:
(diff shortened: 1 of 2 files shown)


andete:
Looks good. I indeed missed that your change served the purpose of enabling building unsigned branch builds.

grauw:
Looks like it’s always skipping the signing now, did you perform this change to the GitHub Actions configuration for the openMSX repo?

> for this change the CODE_SIGN_IDENTITY variable should be added with the value 748867M822.

The remark came after all the screenshots so I reckon it was easy to miss…

andete:
Looks like I added it as a secret and not as a variable. Should be fixed now.

Issue #1062, opened 2017-04-22
screenshot -raw should autodetect whether to use -doublesize option


MBilderbeek:
When using screenshot -raw to get (indeed) raw screenshots, the user now has to think about whether to use 320x240 of 640x480 screenshots. That's not very comfortable if you want to put this behind a hotkey. 

Why can't openMSX find out itself when to use -doublesize? (Use it automatically for screen TEXT80, 6 and 7.)

MBilderbeek:
Or perhaps use something like this:
`bind PRINT {if {[lsearch {TEXT80 6 7} [get_screen_mode]] == -1} { screenshot -raw -guess-name} else { screenshot -raw -doublesize -guess-name}}`

sdsnatcher:
Agreed. But it would be nice to be able to force the -doublesize if necessary.

MBilderbeek:
You can always force it by using the -doublesize parameter, of course. But perhaps that's not what you meant?

sdsnatcher:
I thought that once the autodetection would be implemented, the -doublesize could be gone. I wanted to make sure that it would stay.      :)

MBilderbeek:
This solution only changes the hotkey. But we could also implement something similar for the Tools -> Capture -> screenshot menu item. Maybe add an 'auto' option, that does this and which is selected by default?
What do you think @m9710797 ?

m9710797:
Good idea. I also think it should remain possible to override the auto detection, in *both* directions.
Maybe we can add an option like: `-size [auto|320|640]`
And the (old) `-doublesize` option then does the equivalent of `-size 640`.

For V9990 we could also support sizes `960` and `1280`. But let's leave that for a future extension (not bound to this ticket). It's also low priority IMHO.

One corner case is screen splits between modes of different widths. Actually now that I think of it (but without having looked at the code, I might be miss remembering), create a *raw* screenshot operates on `RawFrame` objects. These remember per line how wide it is (320, 640, ...). So we could calculate the maximum width of all lines. And that might actually be easier than figuring out the screen mode (because e.g. screen mode from which VDP?).

MBilderbeek:
Yes, this is about raw mode only. So, what would it look like if we use that info instead?

PR #1930, opened 2025-06-09
Create auto option for raw screenshot size

MBilderbeek:
See #1062.

### DIFF
--- a/src/imgui/ImGuiTools.hh
+++ b/src/imgui/ImGuiTools.hh
@@ -42,8 +42,8 @@ private:
 	std::string screenshotName;
 	enum class SsType : int { RENDERED, MSX, NUM };
 	int screenshotType = static_cast<int>(SsType::RENDERED);
-	enum class SsSize : int { S_320, S_640, NUM };
-	int screenshotSize = static_cast<int>(SsSize::S_320);
+	enum class SsSize : int { S_320, S_640, AUTO, NUM }; // keep order for bw compat of persisted element
+	int screenshotSize = static_cast<int>(SsSize::AUTO);
 	bool screenshotWithOsd = false;
 	bool screenshotHideSprites = false;
 
--- a/src/video/Display.cc
+++ b/src/video/Display.cc
@@ -30,6 +30,7 @@
 #include "stl.hh"
 #include "unreachable.hh"
 #include "xrange.hh"
+#include "join.hh"
 
 #include "build-info.hh"
 
@@ -429,11 +430,13 @@ void Display::ScreenShotCmd::execute(std::span<const TclObject> tokens, TclObjec
 	bool rawShot = false;
 	bool doubleSize = false;
 	bool withOsd = false;
+	std::string size;
 	std::array info = {
 		valueArg("-prefix", prefix),
 		flagArg("-raw", rawShot),
-		flagArg("-doublesize", doubleSize),
-		flagArg("-with-osd", withOsd)
+		flagArg("-doublesize", doubleSize), // bwcompat, alias for -size 640
+		flagArg("-with-osd", withOsd),
+		valueArg("-size", size)
 	};
 	auto arguments = parseTclArgs(getInterpreter(), tokens.subspan(1), info);
 
@@ -445,6 +448,28 @@ void Display::ScreenShotCmd::execute(std::span<const TclObject> tokens, TclObjec
 	if (rawShot && withOsd) {
 		throw CommandException("-with-osd cannot be used in "
 		                       "combination with -raw");
+	}
+	if (!size.empty()) {
+		if (!rawShot) {
+			throw CommandException("-size option can only be used in "
+			                       "combination with -raw");
+		}
+		static constexpr std::array<std::string_view, 3> validSizes = { "auto", "320", "640" };
+		if (!contains(validSizes, size)) {
+			throw CommandException(strCat("-size option must specify one of: ", join(validSizes, ", ")));
+		}
+		if (doubleSize) {
+			throw CommandException("Only specify either -size or -doublesize");
+		}
+	}
+
+	// backwards compatiblity
+	if (doubleSize) {
+		size = "640";
+	}
+	// default size for raw
+	if (rawShot && size.empty()) {
+		size = "auto";
 	}
 
 	std::string_view fname;
@@ -476,7 +501,7 @@ void Display::ScreenShotCmd::execute(std::span<const TclObject> tokens, TclObjec
 			throw CommandException(
 				"Current renderer doesn't support taking screenshots.");
 		}
-		unsigned height = doubleSize ? 480 : 240;
+		std::optional<unsigned> height = size == "auto" ? std::nullopt : size == "640" ? std::optional(480) : std::optional(240);
 		try {
 			videoLayer->takeRawScreenShot(height, filename);
 		} catch (MSXException& e) {
--- a/src/video/PostProcessor.cc
+++ b/src/video/PostProcessor.cc
@@ -159,16 +159,23 @@ static void getScaledFrame(const FrameSource& paintFrame,
 	}
 }
 
-void PostProcessor::takeRawScreenShot(unsigned height2, const std::string& filename)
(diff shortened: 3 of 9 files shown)


src/video/Display.cc
```diff
@@ -429,11 +430,13 @@ void Display::ScreenShotCmd::execute(std::span<const TclObject> tokens, TclObjec
 	bool rawShot = false;
 	bool doubleSize = false;
 	bool withOsd = false;
+	std::string size = "";
```
m9710797: Minor: initialize an empty string via the default constructor, so leave out the `= ""` part.

Functionally this means exactly the same, and the `= ""` syntax does make it more explicit what you mean. But unfortunately compilers don't (always) optimize away the extra code (= figure out that `""` does the same as the default constructor). Gcc usually can do this, but for msvc and clang I've seen that they aren't always able to do this.

src/video/Display.cc
```diff
@@ -442,10 +445,24 @@ void Display::ScreenShotCmd::execute(std::span<const TclObject> tokens, TclObjec
...
 	if (rawShot && withOsd) {
 		throw CommandException("-with-osd cannot be used in "
 		                       "combination with -raw");
 	}
+	if (!size.empty() && !contains(validSizes, size)) {
+		throw CommandException(strCat("-size option must specify one of: ", join(validSizes, ", ")));
+	}
+	if (!size.empty() && doubleSize) {
+		throw CommandException("Only specify either -size or -doublesize");
+	}
```
m9710797, line 464: Detail: would it make sense to group the 3 checks that start with `!size.empty()`?
```
if (!size.empty()) {
    if (!rawShot) ...
    if (!contains(...)) ...
    if (doubleSize) ...
}
```
Not for efficiency, but to make it clear that these are all checks for when the `-size` option was specified.

src/video/Display.hh
```diff
@@ -117,6 +117,7 @@ private:
 		explicit ScreenShotCmd(CommandController& commandController);
 		void execute(std::span<const TclObject> tokens, TclObject& result) override;
 		[[nodiscard]] std::string help(std::span<const TclObject> tokens) const override;
+		static constexpr std::array<std::string, 3> validSizes = { "auto", "320", "640" };
```
m9710797: `validSizes` is only used inside `ScreenShotCmd::execute()`, so better move it there.

Also the type `std::string` can be changed to `std::string_view`.

src/video/PostProcessor.cc
```diff
@@ -159,16 +159,25 @@ static void getScaledFrame(const FrameSource& paintFrame,
...
+void PostProcessor::takeRawScreenShot(std::optional<unsigned> desiredHeight, const std::string& filename)
 {
 	if (!paintFrame) {
 		throw CommandException("TODO");
 	}
 
-	inplace_buffer<const FrameSource::Pixel*, 480> lines(uninitialized_tag{}, height2);
+	auto getAutoHeight = [this] {
+		auto maxLineWidth = getLineWidth(paintFrame, 0, paintFrame->getHeight());
+		assert(maxLineWidth == one_of(320u, 640u));
```
m9710797: This `assert` can trigger in case of V9990.

src/video/PostProcessor.cc
```diff
@@ -159,16 +159,25 @@ static void getScaledFrame(const FrameSource& paintFrame,
...
 
-void PostProcessor::takeRawScreenShot(unsigned height2, const std::string& filename)
+void PostProcessor::takeRawScreenShot(std::optional<unsigned> desiredHeight, const std::string& filename)
 {
 	if (!paintFrame) {
 		throw CommandException("TODO");
 	}
 
-	inplace_buffer<const FrameSource::Pixel*, 480> lines(uninitialized_tag{}, height2);
+	auto getAutoHeight = [this] {
```
m9710797: Usually in openMSX (when the lambda object is not stored, or passed around) we don't bother specifying the exact capture list. We simply use `[&]`. But `[this]` is also correct, and more explicit.

src/video/PostProcessor.cc
```diff
@@ -159,16 +159,25 @@ static void getScaledFrame(const FrameSource& paintFrame,
...
+		auto maxLineWidth = getLineWidth(paintFrame, 0, paintFrame->getHeight());
+		assert(maxLineWidth == one_of(320u, 640u));
+		if (maxLineWidth == 320) return 240;
+		return 480;
+	};
+
+	const unsigned targetHeight = desiredHeight ? *desiredHeight : getAutoHeight();
```
m9710797: An alternative could be:
```
unsigned targetHeight = [&]{
    if (desiredHeight) return *desiredHeight;
    auto maxLineWidth = getLineWidth(paintFrame, 0, paintFrame->getHeight());
    return (maxLineWidth == 320) ? 240u : 480u;
}();
```
So using an immediately invoked lambda (not giving it a name).
Pick what you like best: your version is more explicit, this version is more compact.
That's more compact

src/video/VideoLayer.hh
```diff
@@ -34,10 +35,13 @@ public:
...
-	 * parameter should be either '240' or '480'. The current image will be
-	 * scaled to '320x240' or '640x480' and written to a png file. */
+	 * parameter should be either '240' or '480' if specified. If not
+	 * specified, the height will be determined based on the available
+	 * widths in the raw frame. The result will be scaled to either
+	 * '320x240' or '640x480' and written to a png file.
+	 */
 	virtual void takeRawScreenShot(
-		unsigned height, const std::string& filename) = 0;
+		std::optional<unsigned> height, const std::string& filename) = 0;
```
m9710797, line 44: Using `std::optional` for `height` is perfectly fine.

In this specific case only a few specific values are valid (`240` and `480`). And then using `0` to mean `auto` is also fine (0 is never a valid height). (And not changing `unsigned` -> `std::optional<unsigned` in several (6?) places makes the patch quite a bit smaller).

Pick what you like best.

src/imgui/ImGuiTools.hh
```diff
@@ -42,8 +42,8 @@ private:
 	std::string screenshotName;
 	enum class SsType : int { RENDERED, MSX, NUM };
 	int screenshotType = static_cast<int>(SsType::RENDERED);
-	enum class SsSize : int { S_320, S_640, NUM };
-	int screenshotSize = static_cast<int>(SsSize::S_320);
+	enum class SsSize : int { S_320, S_640, AUTO, NUM }; // keep order for bw compat of persisted element
```
m9710797, line 45: No need to change anything, just some background info.

Till now we never cared about bw compat for GUI settings.
And we know that we want to rework this in the future (e.g. split "recently used" from "GUI layout" etc).

src/video/Display.cc
MBilderbeek: Good idea.

src/video/Display.cc
MBilderbeek: OK, will change it nevertheless.

src/video/Display.hh
MBilderbeek: OK, will do. Seems to be needed to compile on Mac.

Also, I intended to use it also in `tabCompletion`.... oops.

MBilderbeek: Ah, but we don't implement `tabCompletion` in this "internal" command. Hmm.

src/video/PostProcessor.cc
MBilderbeek: Oops, just removed it.

src/video/PostProcessor.cc
MBilderbeek: OK

src/video/PostProcessor.cc
MBilderbeek: OK, will use this instead.

src/video/VideoLayer.hh
MBilderbeek: I prefer the optional, it's more explicit and I've done the changes already anyway.

src/imgui/ImGuiTools.hh
MBilderbeek: You're right about that. When we change it, we can remove that line :)

PR #1931, opened 2025-06-10
add support Japanese Gryphs (JIS level 1&2 kanji) in imGui

uniskie:
before
image

after
image

add setting "gui_font_japanese_gryph" (boolean)


m9710797:
Hi uniskie, thanks for working on this.

Let me start with some background information:
Text rendering in 'Dear ImGui' is done via a texture-atlas. That is: there is one texture that contains all possible glyphs, and then during text rendering, parts of this texture are used to compose the result (glyph per glyph). In the current version of 'Dear ImGui' there's a limitation that this texture-atlas must be create when the program starts (I'm simplifying a bit). So we must up-front already decide which glyphs will be needed.

I once tried just loading all glpyhs, but unfortunately that results in a much too large texture (and it also caused a noticeable delay when starting openMSX). That's why currently openMSX makes a selection of only the most basic unicode characters (lines 77-92 in ImGuiManager.cc). Unfortunately, as you demonstrated, that's not sufficient for full Japanese support :(

Your PR does extend support for all(?) Japanese glyphs. But for other languages the problem remains. And it's also weird that, as an openMSX user, you should select which languages to support in the font (I don't know any other programs that requires this).

So this is a well known limitation in 'Dear ImGui'. And luckily upstream in the 'Dear ImGui' project they are working on a solution:
In a future version of `Dear ImGui`, when a glyph is missing from the texture-atlas, it will dynamically be added:
* So it's no longer needed to up-front specify the list of glyphs.
* And only the glyphs that are actually needed are loaded in memory.

As I said: this is not yet possible today, but it will be in a future 'Dear ImGui' version. See here for more details:
    https://github.com/ocornut/imgui/issues/8465
Import quote from that page:
```
2025/06/06: pushed enough changes that I am hopeful this could be merged this month. Feedback welcome.
```
So hopefully we can soon use that new 'Dear ImGui' version in openMSX.

--

With the above in mind. I'd like to wait with merging this PR. Hopefully soon (this month?) there is a better solution for this problem.

But thanks again for your PR.

ocornut:
I actually just merged this to main branches now :)

m9710797:
> I actually just merged this to main branches now :)

Thanks!

uniskie:
Thank you for your reply.

I understand that it is best to treat this as a temporary solution limited to Japan and put it on hold rather than integrating it into the main branch.

So imgui's excellent dynamic font cache mechanism will be reflected in openMSX.
I'm looking forward to the future.

m9710797:
@uniskie I just pushed a78532f7fe8ba245b55b8ed3e5043f4f1b1e5ea6. That should resolve this issue (in a language agnostic way, and without any user settings). Please confirm.

@ocornut Thanks a lot for the dynamic font feature. It seems to be working well. In `CHANGELOG.txt` you ask for feedback, so here it is:

The upgrade from v19191 to v19198 went fairly smooth:

1) I first had to resolve some compilation errors:
* `{Push,Pop}TextureID()` was renamed.
* `GetCharAdvance()` is moved from `ImFont` to `ImFontBaked`.
* Calls to `ImGui_ImplOpenGL3_{Create,Destroy}FontsTexture()` are no longer needed.

The instructions in `CHANGELOG.txt` were clear on how to resolve these.
Suggestion: it would be even simpler if you could use the c++ features `[[deprecated("reason")]]` and `=delete; // reason`. See here for [an example](https://godbolt.org/z/Eh9r49hTG): 

I was a little bit surprised that passing a range of glyphs to `AddFontFromMemoryTTF()` continued to compile fine. In my mind specifying such a range would no longer be needed. But then when I read the code I found that it's only used for some legacy backends (not for SDL2 / openGL3, right?). So I replaced that range with `nullptr` and it still works fine.

2) Once the code compiled again, there were now a few places where the behavior had changed (is now wrong).

First some background info: in openMSX we mostly use a proportional font, but some part of the GUI use a mono-spaced font. For both we allow the user to select a custom font and font-size. Typically the user will select the same size for both. But in my tests I also explicitly test with two different sizes.

So in various places in our code we call `ImGui::PushFont(font)`. That code continues to compile, but the behavior is slightly changed. Before, switching to a different font also switched font size (a font was loaded for one specific size). But now the same font can be rendered at different sizes, and switching fonts with `ImGui::PushFont(font)` **keeps** the previous size. The fix is very easy: `ImGui::PushFont(font, font->LegacySize)`. But it took me a few minutes to figure this out. I found out by reading `CHANGELOG.txt`, it's clearly described there. I agree this new behavior makes sense, but from a backwards-compatibility point of view, I would have preferred a compilation error rather than a silent change in behavior.

Note: I'm using `font->LegacySize` because it's available, and it required the least amount of changes in the code. But I wouldn't mind if this field disappears in some future version (maybe together with the `float size_pixels` parameter from the `AddFontFrom{XXX}TTF()` family of functions).

After fixing `ImGui::PushFont()` I still found that some text was drawn with the wrong size. I'm not fully sure, but I think this was because the initial font size was still `13`? Anyway, it's fixed after I added another `ImGui::PushFont(font, font->LegacySize)` near the start of the frame (soon after `ImGui::NewFrame()`).

uniskie:
I ran a snapshot build of the MASTER branch and confirmed that the IMGUI Kanji display issue has been resolved.
Thank you for your great response.

PR #1936, opened 2025-06-16
Make sure SCC is also enabled in the Konami mapper of MegaFlashROM SCC+

### LINKED ISSUE #1033: MegaFlashROM SCC+ (and SD version) do not have standard SCC mapper
The changes I did in 4930272267a05f885ea58bc96ed3c34e9f2c993c for KUC are also valid for MFRSCC+ and MFRSCC+SD, according to Pazos.

However, these 2 devices are a lot more complex, so I didn't immediately try to fix it and thus I made this issue to not forget about it.

Some relevant quotes:

> The explanation is that the bit 5 only changes the address range of the mapper (Konami or Konami SCC) but the SCC is always available.
>     This feature is inherit from MFC SCC+ subslot mode because all subslots share the same mapper type. So to make Konami combinations (i.e.: Konami ROM with Konami SCC ROM) a “common mapper” is needed and the SCC must be available. So I made a “Konami mapper” with SCC.
> 
> In fact, all Konami and Konami SCC ROMs should work with “Konami” mapper in KUC.

and

> Bit 5 changes the addresses checked for switching banks.
> So in SCC mode the range will be the same used in Konami SCC mapper (i.e.: Bank 2 = #7000-#7FFF)
> Since Konami mapper overlaps and shares the same range (#6000-#7FFF), all ROMs should work in Konami mode.
> But KUC uses both mappers (Konami and Konami SCC) because I did not remember this feature when I made KUC software.

and

> It is the same in MFR SCC+ SD, MFR SCC+ and KUC.

MBilderbeek:
Fixes #1033

### DIFF
--- a/src/memory/MegaFlashRomSCCPlus.cc
+++ b/src/memory/MegaFlashRomSCCPlus.cc
@@ -449,6 +449,7 @@ void MegaFlashRomSCCPlus::writeMem(word addr, byte value, EmuTime::param time)
 			if ((addr < 0x5000) || ((0x5800 <= addr) && (addr < 0x6000))) break; // only SCC range works
 			byte mask = (configReg & 0x01) ? 0x1F : 0x7F;
 			bankRegs[subslot][page8kB] = value & mask;
+			sccBanks[page8kB] = value;
 			invalidateDeviceRCache(0x4000 + 0x2000 * page8kB, 0x2000);
 			break;
 		}
--- a/src/memory/MegaFlashRomSCCPlus.cc
+++ b/src/memory/MegaFlashRomSCCPlus.cc
@@ -62,15 +62,22 @@ Main features:
 [REGISTERS]
 
  -#7FFE or #7FFF = CONFIGURATION REGISTER:
-    7 mapper mode 1: \ %00 = SCC,  %01 = 64K
+    7 mapper mode 1: \ %00 = Konami/Konami-SCC,  %01 = 64K
     6 mapper mode 0: / %10 = ASC8, %11 = ASC16
-    5 mapper mode  :   Select Konami mapper (0 = SCC, 1 = normal)
+    5 mapper mode  :   Select Konami mapper (0 = SCC, 1 = normal) [1]
     4 Enable subslot mode and register #FFFF (1 = Enable, 0 = Disable)
     3 Disable #4000-#5FFF mapper in Konami mode (0 = Enable, 1 = Disable)
     2 Disable configuration register (1 = Disabled)
     1 Disable mapper registers (0 = Enable, 1 = Disable)
     0 Enable 512K mapper limit in SCC mapper or 256K limit in Konami mapper
 (1 = Enable, 0 = Disable)
+
+[1] bit 5 only changes the address range of the mapper (Konami or Konami SCC)
+but the SCC is always available.  This is done for the subslot mode because all
+subslots share the same mapper type. So to make Konami combinations (i.e.:
+Konami ROM with Konami SCC ROM) a "common mapper" is needed and the SCC must be
+available. So I made a "Konami mapper" with SCC.  In fact, all Konami and
+Konami SCC ROMs should work with "Konami" mapper.
 
 
 --------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


MBilderbeek:
@m9710797 What am I still doing something wrong here? I still don't hear the SCC when flashing the Konami SCC ROM with Konami mapper mode with these 2 patches.

m9710797:
To me this looks fine to make the SCC register available in the Konami mapper (non-scc-konami mapper).

If it doesn't work, then we'd have to investigate what the requirements of the MSX software are. Because isn't it strange that the MSX software would be using the non-scc mapper behavior (e.g. bank switch addresses) while still writing to SCC registers?

MBilderbeek:
I'm not sure, but I've seen a video where this worked.... 

This is how I test: `$ openmsx -machine Philips_NMS_8250 -ext MegaFlashROM_SCC+ -diska test.dsk.gz`

[test.dsk.gz](https://github.com/user-attachments/files/20784419/test.dsk.gz)

and then in DOS do: `opfx hamaraja.rom /x`
You'll see it selects the Konami mapper and no SCC is played.

But on real hardware it is working:
https://drive.google.com/file/d/1Oe9i75NCGWKQAYVd0v2SPdKp8Pw9oJD8/view

MBilderbeek:
@m9710797 except for the commit with the logging, do you think this is acceptable and likely how to the hardware works?

src/memory/MegaFlashRomSCCPlus.cc
```diff
@@ -428,25 +450,30 @@ void MegaFlashRomSCCPlus::writeMem(word addr, byte value, EmuTime::param time)
...
 				break;
 			}
-			// Making of the mapper bits is done on
+			std::cerr << "Write Konami mapper: 0x" << std::hex << int(value) << " at switch address: 0x" << int(addr) << std::endl;
+			// Masking of the mapper bits is done on
 			// write (and only in Konami(-scc) mode)
 			if ((addr < 0x5000) || ((0x5800 <= addr) && (addr < 0x6000))) break; // only SCC range works
 			byte mask = (configReg & 0x01) ? 0x1F : 0x7F;
 			bankRegs[subslot][page8kB] = value & mask;
+			sccBanks[page8kB] = value;
```
m9710797, line 452: You added "write to `sccBanks[]`" for non-scc-konami mapper. Just a guess, but I think there's a good chance it happens always. I mean for bank-switches in any mapper-mode. So whenever `bankRegs[][]` gets written, we may also need to write to `sccBanks[]`.

Side note: I would naively have guessed that in real hardware there is no separate set of `sccBanks[]` and `bankRegs[][]` registers. But that's wrong, it's contradicted by this comment:
```
Masking of the mapper bits is done on write (and only in Konami(-scc) mode).
```
(Because this masking removes bit 7, and that bit IS needed for the test in `getSCCMode()`: `sccBanks[3] & 0x80`. This implies the hardware must store this information elsewhere.)

Anyway, this is a minor detail. I doubt there's any existing software that depends on this. To observe the difference the MSX software would have to:
* Set a non-Konami mapper mode.
* Select the SCC mapper page (see `getSCCMode()`, e.g. make `sccBanks[3] & 0x80`).
* Then switch to Konami mapper mode.
* And then, without changing the mapper pages, write to the SCC, and expect it to be active.

m9710797:
> @m9710797 except for the commit with the logging, do you think this is acceptable and likely how to the hardware works?

I added a comment about writes to `sccBanks`. It might not yet be 100% like the real hardware(*1). But the current code is maybe already "good enough". It's for sure an improvement over the current code in `master`. So feel free to already push this (and also consider this finished).

(*1) To really get all the details right it would be useful to get the original VHDL  (or verilog?) code. Because now we're mostly guessing and fixing symptoms.

src/memory/MegaFlashRomSCCPlus.cc
```diff
@@ -270,14 +277,21 @@ unsigned MegaFlashRomSCCPlus::getFlashAddr(unsigned addr) const
 	return ((0x40000 * subslot) + tmp) & 0xFFFFF; // wrap at 1MB
 }
 
+bool MegaFlashRomSCCPlus::isSCCAccessEnabled() const
+{
+	// In both Konami and Konami SCC mappers the SCC is available (see note
+	// above)
+	return (configReg & 0xE0) == one_of(0x00, 0x20);
```
m9710797, line 284: Didn't you have `(configReg & 0x80) == 0x00` at some point? That seems closer to what Manuel Pazos mentioned in his email? But no need to change this. We can wait till we get the VHDL to really fix all details. (And not testing bit 6 is also a bit weird).

Note: the current code `(configReg & 0xE0) == one_of(0x00, 0x20)` can be simplified to `(configReg & 0xC0) == 0x00` (test that bit 6 and 7 are both 0). But totally fine to write it like this (with `one_of`). I verified that the compiler optimizes both versions equally well (it optimizes both to: `configReg <= 63`).

MBilderbeek:
As agreed in chat: let's merge this now and see whether we can improve details later.

PR #1937, opened 2025-06-25
Use cassette file type indication from filename...

MBilderbeek:
...if we can't derive it from the contents. Better than nothing!

### DIFF
--- a/src/cassette/CassetteImage.cc
+++ b/src/cassette/CassetteImage.cc
@@ -1,4 +1,6 @@
 #include "CassetteImage.hh"
+#include "FileOperations.hh"
+#include "Filename.hh"
 #include <cassert>
 
 namespace openmsx {
@@ -13,6 +15,32 @@ std::string CassetteImage::getFirstFileTypeAsString() const
 		return "BASIC";
 	} else {
 		return "unknown";
+	}
+}
+
+void CassetteImage::setFirstFileType(FileType type, const Filename& fileName) {
+	if (type == UNKNOWN) {
+		// see if there is a hint in the filename
+		auto file = fileName.getResolved();
+		auto fileStem = FileOperations::stem(file);
+
+		auto containsInstructionAndCAS = [&](std::string_view instruction) {
+			auto pos = fileStem.find(instruction);
+			if (pos == std::string_view::npos) return false;
+			return fileStem.find("CAS", pos + instruction.size()) != std::string_view::npos;
+		};
+
+		if (containsInstructionAndCAS("BLOAD")) {
+			firstFileType = BINARY;
+		} else if (fileStem.contains("CLOAD")) {
+			// note: probably better to first check on CLOAD and then on RUN. Some filenames
+			// contain hints like CLOAD + RUN.
+			firstFileType = BASIC;
+		} else if (containsInstructionAndCAS("RUN")) {
+			firstFileType = ASCII;
+		}
+	} else {
+		firstFileType = type;
 	}
 }
 
--- a/src/cassette/CassetteImage.hh
+++ b/src/cassette/CassetteImage.hh
@@ -10,6 +10,8 @@
 
 namespace openmsx {
 
+class Filename;
+
 class CassetteImage
 {
 public:
@@ -36,7 +38,9 @@ public:
 
 protected:
 	CassetteImage() = default;
-	void setFirstFileType(FileType type) { firstFileType = type; }
+	// Please make sure this method is called from the constructor of each
+	// subclass! (And only from there.)
+	void setFirstFileType(FileType type, const Filename& fileName);
 	void setSha1Sum(const Sha1Sum& sha1sum);
 
 private:
--- a/src/cassette/WavImage.cc
+++ b/src/cassette/WavImage.cc
@@ -107,13 +107,14 @@ void WavImageCache::release(const WavData* wav)
 }
 
 
-// Note: type detection not implemented yet for WAV images
 WavImage::WavImage(const Filename& filename, FilePool& filePool)
 {
 	const auto& entry = WavImageCache::instance().get(filename, filePool);
 	wav = &entry.wav;
 	setSha1Sum(entry.sum);
 	clock.setFreq(wav->getFreq());
+	// Note: type detection not implemented yet for WAV images
+	setFirstFileType(FileType::UNKNOWN, filename);
 }
 
 WavImage::~WavImage()
--- a/doc/manual/commands.html
+++ b/doc/manual/commands.html
@@ -486,7 +486,7 @@
     <tr>
       <td><code>cassetteplayer insert &lt;tape image&gt;</code></td>
 
-      <td>Insert tape image (WAV or CAS format) in the cassette player</td>
+      <td>Insert tape image (WAV, CAS or TSX format) in the cassette player</td>
     </tr>
 
     <tr>
--- a/doc/manual/commands.html
+++ b/doc/manual/commands.html
@@ -2722,7 +2722,7 @@
   </table>
 
   <div class="note">
-    Note: Autorun is only supported for cassette images in the CAS format.
+    Note: Autorun works practically always for cassette images in the CAS and TSX formats. If that fails or for WAV files, openMSX will try to find a hint of the loading instruction (in upper case) in the filename.
   </div>
 
   <h3><a id="autorunlaserdisc">autorunlaserdisc</a></h3>
--- a/doc/manual/user.html
+++ b/doc/manual/user.html
@@ -59,7 +59,7 @@
 			</li>
 			<li><a class="internal" href="#tape">5.3 Running Tape Software</a>
 				<ol class="toc">
-					<li><a class="internal" href="#cas">5.3.1 Using CAS files</a></li>
+					<li><a class="internal" href="#cas">5.3.1 Using CAS/TSX files</a></li>
 				</ol>
 			</li>
 
@@ -942,7 +942,7 @@ software cannot be used on these machines.
 
 <p>
 Cassette/tape image file formats that are supported are WAV files (raw digitized
-recordings of real tapes) and CAS files. Differences are explained below.
+recordings of real tapes) and CAS/TSX files. Differences are explained below.
 </p>
 
 <p>
@@ -986,9 +986,9 @@ command again (or the Record button with the circle icon in the Tape Deck
 window).
 </p>
 
-<h4><a id="cas">5.3.1 Using CAS files</a></h4>
+<h4><a id="cas">5.3.1 Using CAS and TSX files</a></h4>
 <p>
-You can also use the so-called CAS files. Use them exactly as you would use WAV
+You can also use the so-called CAS files and also TSX files. Use them exactly as you would use WAV
 files, described in the previous section.
 </p>
 
@@ -996,10 +996,10 @@ files, described in the previous section.
 We don't support using CAS files by patching a BIOS natively, because it is not
 really something we want: we prefer a more authentic emulation without hacks
 like this.
-So, the CAS files are automatically converted to WAV files, internally. Note
+So, the CAS (and also TSX) files are automatically converted to WAV files, internally. Note
 that the loading time is drastically longer this way (but the <span class="menu-reference">Main menu bar &rarr; Settings &rarr; Speed &rarr; Go full speed when loading</span> <a class="external" href="commands.html#fullspeedwhenloading">setting</a>
 will help a lot). On the other hand, you will be able to hear the cassette
-sounds now also with the CAS files... Admit it, using cassettes with an MSX
+sounds also with these file formats... Admit it, using cassettes with an MSX
 without those characteristic sounds just wouldn't be the same.
 </p>
 
--- a/doc/manual/user.html
+++ b/doc/manual/user.html
@@ -1004,12 +1004,12 @@ without those characteristic sounds just wouldn't be the same.
 </p>
 
 <p>
-To make it even more comfortable to run software from CAS images, check the try to enable the <span class="menu-reference">Main menu bar &rarr; Media &rarr; Tape Deck &rarr; Controls &rarr; (try to) Auto Run</span> <a class="external" href="commands.html#autoruncassettes">setting</a>, that will attempt to
+To make it even more comfortable to run software from cassette tapes, check the <span class="menu-reference">Main menu bar &rarr; Media &rarr; Tape Deck &rarr; Controls &rarr; (try to) Auto Run</span> <a class="external" href="commands.html#autoruncassettes">setting</a>, that will attempt to
 type the loading instruction for you after the MSX has started up.
 </p>
 
 <p>
-Note that saving to CAS files (new or existing ones) is not possible; you can
+Note that saving to existing cassette images is not possible; you can
 only save to new cassette images in WAV format.
 </p>
 
--- a/src/cassette/CasImage.cc
+++ b/src/cassette/CasImage.cc
@@ -283,7 +283,7 @@ CasImage::Data CasImage::init(const Filename& filename, FilePool& filePool, CliC
 			return MSX_CAS::convert(cas, filename.getOriginal(), cliComm, fileType);
 		}
 	}();
-	setFirstFileType(fileType);
+	setFirstFileType(fileType, filename);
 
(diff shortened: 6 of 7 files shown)


src/cassette/CassetteImage.cc
```diff
@@ -16,6 +23,24 @@ std::string CassetteImage::getFirstFileTypeAsString() const
 	}
 }
 
+void CassetteImage::setFirstFileType(FileType type) {
+	if (type == FileType::UNKNOWN) {
+		// see if there is a hint in the filename
+		if (fileStem.find("BLOAD") != std::string::npos) {
```
m9710797: (With c++23) this can be simplified to:
```
if (fileStem.contains("BLOAD")) { ...
```
And 2x more in the lines below.

src/cassette/CassetteImage.cc
```diff
@@ -16,6 +23,24 @@ std::string CassetteImage::getFirstFileTypeAsString() const
...
+			std::cerr << "BL";
+			firstFileType = FileType::BINARY;
+		} else if (fileStem.find("CLOAD") != std::string::npos) {
+			firstFileType = FileType::BASIC;
+		} else if (fileStem.find("RUN") != std::string::npos) {
+			firstFileType = FileType::ASCII;
+		}
+		std::cerr << "OVerride to " << getFirstFileTypeAsString() << ", based on " << fileStem << std::endl;
```
m9710797: I know this is WIP, but I'll mention it anyway. Remove these debug-prints before merging.

Detail: `<< std::endl` is equivalent to `<< '\n' << std::flush`. But that flush is redundant on `std::cerr`: it already does a flush after every component.
And more in general (e.g. also for `std::cout` or any other stream) usually the extra flush is not needed. So most often you want to use just `'\n'` instead of `std::endl`.

src/cassette/CassetteImage.hh
```diff
@@ -35,12 +37,13 @@ public:
...
 protected:
-	CassetteImage() = default;
-	void setFirstFileType(FileType type) { firstFileType = type; }
+	explicit CassetteImage(const Filename& fileName);
+	void setFirstFileType(FileType type);
 	void setSha1Sum(const Sha1Sum& sha1sum);
 
 private:
 	FileType firstFileType = UNKNOWN;
+	const std::string fileStem;
```
m9710797: We shouldn't make `fileStem` a member variable.

It's only used inside `setFirstFileType()`, and that method is only called from the subclass constructors. Those constructors all have a `fileName` parameter. So we can pass that parameter as a 2nd parameter to `setFirstFileType()`.

In other words, after the object has been constructed, this member variable isn't needed anymore. That's a good indication it shouldn't be a member variable.

--

This means we can also revert the changes to the `CassetteImage` constructor, and the corresponding changes in the subclass constructors.

src/cassette/WavImage.cc
```diff
@@ -107,13 +107,15 @@ void WavImageCache::release(const WavData* wav)
...
-// Note: type detection not implemented yet for WAV images
 WavImage::WavImage(const Filename& filename, FilePool& filePool)
+	: CassetteImage(filename)
 {
 	const auto& entry = WavImageCache::instance().get(filename, filePool);
 	wav = &entry.wav;
 	setSha1Sum(entry.sum);
 	clock.setFreq(wav->getFreq());
+	// Note: type detection not implemented yet for WAV images
+	setFirstFileType(FileType::UNKNOWN);
```
m9710797: It's OK to make this explicit. Though it's also redundant because `FileType::UNKNOWN` is already the default.

EDIT:
I was wrong. You need this because of the filename parsing. (And `filename` becomes a parameter of `setFirstFileType()`, then also makes it clearer what's going on).

src/cassette/CassetteImage.cc
```diff
@@ -1,8 +1,15 @@
 #include "CassetteImage.hh"
+#include "Filename.hh"
 #include <cassert>
+#include <filesystem>
 
 namespace openmsx {
 
+CassetteImage::CassetteImage(const Filename& fileName)
+	: fileStem(std::filesystem::path(fileName.getResolved()).stem().string())
```
m9710797: The calculation of `fileStem` can be moved into `setFirstFileType()` (and inside the branch `type == UNKNOWN`).

But I also think this calculation might go wrong, at least on windows:
* On Linux both native paths and std::filesystem::path use UTF8 encoding. So constructing a `fs::path` from a `string` just uses that string.
* But on Windows constructing a `fs::path` from a `string` will interpret that string as an 8-bit encoded path, so via the windows code page stuff. Or at least that's my current understanding.
* But in openMSX we have the convention that all our internal paths are already UTF8 encoded (we only convert at 'external' boundaries). Thus `Filename` does contain a UTF-8 encoded string. We should not `fs::path` interpret that as a "code page" string.

A possible solution could be to use `fs::path` for all internal paths in openMSX. Though that will be a lot of work. A simpler solution could be to use our `FileOperations::stem()` function. That operates on UTF-8 strings.

src/cassette/CassetteImage.cc
```diff
@@ -16,6 +23,24 @@ std::string CassetteImage::getFirstFileTypeAsString() const
 	}
 }
 
+void CassetteImage::setFirstFileType(FileType type) {
```
m9710797: Detail: consider adding `using enum FileType;` as the first line in this function. And then drop this prefix in several places below.

m9710797 (changes requested):
Good idea. But the implementation needs some tweaks.

src/cassette/CassetteImage.cc
```diff
@@ -16,6 +23,24 @@ std::string CassetteImage::getFirstFileTypeAsString() const
...
+	if (type == FileType::UNKNOWN) {
+		// see if there is a hint in the filename
+		if (fileStem.find("BLOAD") != std::string::npos) {
+			std::cerr << "BL";
+			firstFileType = FileType::BINARY;
+		} else if (fileStem.find("CLOAD") != std::string::npos) {
+			firstFileType = FileType::BASIC;
+		} else if (fileStem.find("RUN") != std::string::npos) {
```
m9710797: Should we search for more specific sub-strings? Because `RUN` may also occur in filenames like `OUTRUN.CAS`.

src/cassette/CassetteImage.cc
MBilderbeek: Ah, always the shizzle with string operations in C++! But good to know. I didn't even know for sure we were already fully on C++23.

src/cassette/CassetteImage.cc
MBilderbeek: I just find it easier to type (the even longer) `std::endl` :) So whenever I make debug prints, I just use that.

src/cassette/CassetteImage.hh
MBilderbeek: To me it was very surprising to see this construction anyway. But I think it is because the value of the first file type isn't clear yet when the superclass constructor is called, so we can't pass it to the superclass constructor and instead have to use that `setFirstFileType` method. Is that correct? 
The problem is of course that it is totally not clear (nor guaranteed) that this method is ever called.

But my implementation does rely on having that setter called for sure (to calculate the alternative instead of use the default, see below).

src/cassette/WavImage.cc
MBilderbeek: Indeed, it *must* be called now. And currently this is not enforced, alas. See previous comment.

src/cassette/CassetteImage.cc
MBilderbeek: OK, I didn't realize that. 

Wouldn't it be nice to integrally move to `std::filesystem` soonish? It's a really nice API and very rich. At first sight it seems to contain everything we need (and have in `FileOperations.hh` and other classes now).

src/cassette/CassetteImage.cc
MBilderbeek: Yeah, I also thought about that, but I wanted to keep it simple. For the typical CAS file (this specific example) the `CasImage` will already give the right type of course. I could check whether the substring `CAS` is also in it (and even *after* `RUN`) to make it more robust.
Worst thing that can happen is that instead of no auto-run, a *wrong* auto-run will happen. I think it will be rare, but I suppose it's best to do a reasonable effort to avoid that.

MBilderbeek:
@m9710797 about that debug message... it gets also printed when I close openMSX.... any idea why? It's like the tape is reinserted briefly before closing down?

src/cassette/CassetteImage.hh
m9710797: > To me it was very surprising to see this construction anyway. But I think it is because the value of the first file type isn't clear yet when the superclass constructor is called, so we can't pass it to the superclass constructor and instead have to use that `setFirstFileType` method. Is that correct?

That's correct:
In general, to be able to determine the first file type, you need to parse the content of the cassette image. That is only done during the construction of the derived class. And that constructor only runs after the constructor of the base class. So we cannot yet pass this information to the base class constructor.

m9710797:
> @m9710797 about that debug message... it gets also printed when I close openMSX.... any idea why? It's like the tape is reinserted briefly before closing down?

No idea. But would be interesting to investigate.

src/cassette/CassetteImage.cc
```diff
@@ -16,6 +18,26 @@ std::string CassetteImage::getFirstFileTypeAsString() const
 	}
 }
 
+void CassetteImage::setFirstFileType(FileType type, const Filename& fileName) {
+	if (type == UNKNOWN) {
+		// see if there is a hint in the filename
+		auto fileStem = FileOperations::stem(fileName.getResolved());
```
m9710797: Today this line is correct. Nevertheless I would suggest to rewrite it:
* `FileOperations::stem()` both takes a `std::string_view` as input and also returns a `std::string_view`. (It returns part of the input, without making a copy).
* _If_ `fileName.getResolved()` were to return a temporary, then `stem()` returns a view into that temporary, and `fileStem` becomes invalid when that temporary gets destoyed. Thus at the end of this statement.

The fix is to split this over two statements. Then `file` stays alive for the whole duration of the then-branch:
```
auto file = fileName.getResolved();
auto stem = FileOperations::stem(file);
```

Today this line happens to work fine anyway:
* `Filename::getResolved()` has two overloads. One returns a `string&` the other a `string` (by value).
* Today the 1st overload is picked. And it returns a reference to an object with a long enough lifetime.
* But this relies on implementation details in `Filename`. I think it's better to make this more robust and split it over two statements.

src/cassette/CassetteImage.cc
```diff
@@ -16,6 +18,26 @@ std::string CassetteImage::getFirstFileTypeAsString() const
...
+		if (fileStem.contains("BLOAD")) {
+			firstFileType = BINARY;
+		} else if (fileStem.contains("CLOAD")) {
+			firstFileType = BASIC;
+		} else if (size_t runPos = fileStem.find("RUN"); runPos != std::string::npos) {
+			if (fileStem.find("CAS", runPos) != std::string::npos) {
+				firstFileType = ASCII;
+			}
```
m9710797: I agree `BLOAD` and `CLOAD` are less likely to accidentally occur as part of the filename. But for reasons of symmetry maybe we should still also check those in a more robust way. Maybe like this? (untested)
```
auto stem = ...;
auto containsInstruction = [&](std::string_view instruction) {
    auto pos = stem.find(instruction);
    if (pos == std::string_view::npos) return false;
    return stem.find("CAS", pos) != std::string_view::npos;
};
if (containsInstruction("BLOAD") {
    firstFileType = BINARY;
} else if (containsInstruction("CLOAD") {
    firstFileType = BASIC;
} else if (containsInstruction("RUN") {
    firstFileType = ASCII;
}
```

Ah, I only notice this now: `std::string_view::npos`, not `std::string::npos`  (in practice this happens to work fine anyway).

src/cassette/CassetteImage.cc
m9710797: Yes, it would be nice. And I've already done some (small) experiments. Though (so far) I've always abandoned those experiments.

The reason is that it's too easy to (accidentally) convert a UTF8 encoded `std::string` into a `std::filesystem::path`. That fine on Linux, but not on Windows, see above. (Almost) all strings in openMSX are UTF8 encoded.

The "standard" (safe) way to solve this, is to use `std::u8string` and `std::u8string_view` for UTF8 encoded strings. Though that requires even more changes in our code base (we have ~5000 occurrences of the word `string{_view}` in our code. 

And most (all) libraries (SDL, Dear ImGui, FreeType, Tcl, ...) expect `char` (not `char8_t`) as input. It is OK to cast between these two types, but it would make each interaction with those libraries more complex. And we have a lot of string-interaction with Dear ImGui.

src/cassette/CassetteImage.hh
MBilderbeek: That's quite unfortunate, but it explains the current construction. Unfortunately, it's also not possible to force a method to be called from the subclass constructor or something. So I'll add some comments to make it clear.

src/cassette/CassetteImage.cc
MBilderbeek: OK, I wasn't aware of this complication. I thought such libraries would reduce the platform specific worries one has to have.... but not yet enough, as it seems.

MBilderbeek:
> > @m9710797 about that debug message... it gets also printed when I close openMSX.... any idea why? It's like the tape is reinserted briefly before closing down?
> 
> No idea. But would be interesting to investigate.

Hahah, it is the auto-save of the setup! From storeAsSetup, the media is reinserted in the copy of the machine :)

src/cassette/CassetteImage.cc
MBilderbeek: I used your idea still for 2 of the 3 options.

src/cassette/CassetteImage.cc
MBilderbeek: OK, subtle... I would never ever have thought about that... isn't it a bit of a boobytrap that we should try to avoid?

src/cassette/CassetteImage.cc
MBilderbeek: Now it's time to remove this debug print...

src/cassette/CassetteImage.cc
```diff
@@ -16,6 +18,32 @@ std::string CassetteImage::getFirstFileTypeAsString() const
...
+void CassetteImage::setFirstFileType(FileType type, const Filename& fileName) {
+	if (type == UNKNOWN) {
+		// see if there is a hint in the filename
+		auto file = fileName.getResolved();
+		auto fileStem = FileOperations::stem(file);
+
+		auto containsInstructionAndCAS = [&](std::string_view instruction) {
+			auto pos = fileStem.find(instruction);
+			if (pos == std::string::npos) return false;
+			return fileStem.find("CAS", pos) != std::string::npos;
```
m9710797: Detail, no need to change (doesn't matter in the current code).

We should check that `CAS` _follows_, the load instruction, so we should start searching _after_ that instruction:
```
return fileStem.find("CAS", pos + instruction.size()) != std::string_view::npos;
```
```

File: src/cassette/CassetteImage.cc [hunk 11]
```diff
@@ -16,6 +18,32 @@ std::string CassetteImage::getFirstFileTypeAsString() const
...
 
+void CassetteImage::setFirstFileType(FileType type, const Filename& fileName) {
+	if (type == UNKNOWN) {
+		// see if there is a hint in the filename
+		auto file = fileName.getResolved();
+		auto fileStem = FileOperations::stem(file);
+
+		auto containsInstructionAndCAS = [&](std::string_view instruction) {
+			auto pos = fileStem.find(instruction);
+			if (pos == std::string::npos) return false;
```
m9710797: This should be `std::string_view::npos` (not `std::string::npos`). Same in the line below.

m9710797 (approved):
A few details left. After that ready to merge.

src/cassette/CassetteImage.cc
MBilderbeek: Fixed. Although it doesn't matter much, it's making more sense of course.

src/cassette/CassetteImage.cc
MBilderbeek: Fixed.

m9710797 (approved):

PR #1944, opened 2025-08-18
Detect Z80 instruction fetch cycle

pvmm:
Contrary to CPU tracing that sends everything to stdout, instruction tracing is a potentially more useful probe because set_bp can invoke Tcl code exactly before a z80 instruction is executed.

### DIFF
--- a/src/cpu/CPUCore.cc
+++ b/src/cpu/CPUCore.cc
@@ -308,6 +308,9 @@ template<typename T> CPUCore<T>::CPUCore(
 	            "This probe is only useful to set a breakpoint on (the value "
 		    "return by read is meaningless). The breakpoint gets triggered "
 		    "right after the CPU accepted an IRQ.")
+	, m1Cycle(motherboard.getDebugger(), name + ".m1Cycle",
+	            "This probe is used for debugging instruction fetch cycle ("
+	            "M1 cycle).")
 	, freqLocked(
 		motherboard.getCommandController(), tmpStrCat(name, "_freq_locked"),
 	        "real (locked) or custom (unlocked) CPU frequency",
@@ -608,6 +611,7 @@ template<typename T> template<unsigned PC_OFFSET> ALWAYS_INLINE uint8_t CPUCore<
 	// deviation. Apart from that functional aspect it also turns out to be
 	// faster to only update PC once per instruction instead of after each
 	// fetch.
+	if (this->isM1Cycle(getPC())) m1Cycle.signal();
 	unsigned address = narrow_cast<uint16_t>(getPC() + PC_OFFSET);
 	return RDMEM_impl<false, false>(address, cc);
 }
--- a/src/cpu/CPUCore.hh
+++ b/src/cpu/CPUCore.hh
@@ -147,6 +147,7 @@ private:
 
 	Probe<int> IRQStatus;
 	Probe<void> IRQAccept;
+	Probe<void> m1Cycle;
 
 	// dynamic freq
 	BooleanSetting freqLocked;


m9710797:
Hi, thanks for your PR.

Can you talk more about the goal of this feature? I read the full PR (it's very short), and I do understand each individual line. But the overall goal is not clear to me.

Is this a re-implementation of the existing "debug condition" stuff? Debug condition: execute a Tcl callback after each instruction.

EDIT:
Maybe I should clarify: a debug condition with condition "true" (trigger unconditionally after every instruction) with a custom action (the default is "debug break", but you can change that) should be equivalent to the proposed feature.

pvmm:
I changed what this does based on Tcl script testing to do what I originally intended and didn't know yet how to proceed: to detect and separate Z80 operating modes. This is useful for the code analyser script that I am writing (#1707) because it will make it easier to separate code that detect parameters from code that detects opcodes and OpenMSX kinda already does that, this just exposes that information to Tcl.

m9710797:
Can you explain what functionality you need?

I should also say that I've spend A LOT of effort in making CPU emulation fast. This patch adds extra code on the CPU-instruction-emulation path. So unless there's a very good reason to do that I will reject this PR. I mean setting just a bool variable (twice) per instruction does already significantly increase the cost of emulating instructions.

For example look at the isM1Cycle() function, that one also tries to detect the start of instructions. But without introducing ANY overhead.

So if you can explain what information you need exactly, then maybe we can come up with a cheaper way to achieve this.

pvmm:
> Can you explain what functionality you need?

I want to be able to tell from a watchpoint that `read_mem` was done in M1 cycle or not. This could be used to tell apart what is instruction parameter and what is instruction fetch and that could make de **codeanalyser.tcl** simpler to program. At first I was doing this:

```
debug watchpoint create -type read_mem -address {0x0000 0xFFFF} -command {read_mem $::wp_last_address}

proc read_mem {addr} {
        puts stderr "watchpoint@[format %04x $addr] called when IFH=[reg IFH]"
}
```
But now I changed the code to create a `Probe<void>` called `m1Cycle` that is signaled when isM1Cycle() returns true, and now I can change that code to:
```
variable fetch 0

proc read_mem {addr} {
        variable fetch
        puts stderr "watchpoint@[format %04x $addr] called when IFH=$fetch"
        set fetch 0
}

proc ifetch {} {
        variable fetch 1
}

debug watchpoint create -type read_mem -address {0x0000 0xFFFF} -command {read_mem $::wp_last_address}
debug probe set_bp z80.m1Cycle {} {ifetch}
```
> So if you can explain what information you need exactly, then maybe we can come up with a cheaper way to achieve this.

I think I just found a cheaper way in this last push.

m9710797:
Aha, it's for the code analyzer. But if I remember correctly, there you needed to know if a (memory read) watchpoint triggered for opcode reads of for data reads. But then don't you want to know ALL opcode reads, not just those that are part of an M1 cycle??? E.g. for a 3 bytes instruction like `CALL nn`, don't you want to classify all 3 bytes as opcodes rather than just the first byte of this instruction? Or for a 4-byte instruction like `LD IX,nn`, the first 2 have an M1 cycle, the last 2 have not (so M1 is not a good detector for the start of all instructions (if you wanted that)).

So can you  explain in more detail why you specifically want M1?  Don't you just want opcode vs data?

And today, with only a small helper, you can already distinguish between opcode and data:
* You query the PC(*1), and calculate the length of the instruction(*2) at that location.
* If the address that triggered the watchpoint is inside `[PC, PC+length)` then it's an opcode fetch, otherwise it's a data fetch.
* Maybe one exception is when an instruction reads it's own bytes. But that's very rare. And if it does happen, then your code analyzer _should_ still classify that address as instruction rather than as data IMHO.

(*1) This does rely on the fact that openMSX only updates the PC once per instruction.
(*2) Today you can already indirectly query the length of an instruction via the existing `debug disasm` command. The length of the returned list corresponds to the length of the instruction. But we could also expose a faster version that more directly returns this length (and doesn't needlessly produce the mnemonic string). This can be based on the `instructionLength()` function is `Dasm.hh`. (But that's only a speed optimization, it may not be needed at all).

m9710797:
Also your updated PR actually adds MORE overhead than your first version. It adds an extra `if` in *each* `RDMEM_OPCODE`, so possibly multiple times per instruction. We don't want _any_ extra code in the CPU emulation path (not setting booleans, and certainly also not extra tests).

What I meant with the `isM1Cycle()` function in my earlier post. Is that this is a property that's fully _passive_. I mean we don't _actively_ update some state to track M1, instead we can _lazily_ calculate this property only in the (infrequent) case that we need it.

And a property like "is this read part of an instruction or is it a data read" is also something that can be calculated _passively_ (see my previous post). So without _any_ overhead in case we don't need this information.

PR #1946, opened 2025-08-19
Add WonderKid mapper.

MBilderbeek:
For newly discovered (proto?) version of Adventure Kid, called Wonder Kid.


src/memory/RomWonderKid.hh
```diff
@@ -0,0 +1,22 @@
...
+
+class RomWonderKid final : public Rom16kBBlocks
+{
+public:
+	RomWonderKid(const DeviceConfig& config, Rom&& rom);
+
+	void reset(EmuTime time) override;
+        [[nodiscard]] byte peekMem(uint16_t address, EmuTime time) const override;
+        [[nodiscard]] byte readMem(uint16_t address, EmuTime time) override;
+	void writeMem(uint16_t address, byte value, EmuTime time) override;
```
m9710797, line 16: This uses a mix of "4 spaces" and "1 tab(?)" for indentation. Usually in openMSX we use tabs for indentation.
Same in the file `RomWonderKid.cc`, although there it does seem consistent but uses spaces instead of tabs.

I'm not opposed to switching to 4 tabs for indentation, that seems to be to most common convention nowadays. But then maybe we should do that for the whole openMSX code base?

And for sure within a single file we shouldn't use a mix.

m9710797 (approved):
Technically it all seems fine. I assume you also tested it and it works?

Just one minor nit about formatting. Feel free to merge after that.

src/memory/RomWonderKid.hh
MBilderbeek: That was unintentional (I was lazy and copy pasted these methods from another file, but I didn't notice there were spaces left instead of tabs).

MBilderbeek:
> Technically it all seems fine. I assume you also tested it and it works?
> 
> Just one minor nit about formatting. Feel free to merge after that.

Yes, the game seems to run fine with it.

f-fix:
i neglected to check the thread, Bock did indeed answer the questions there! see https://www.smspower.org/forums/20603-WonderKidGGProto2TaiwaneseVariants#135939 and https://www.smspower.org/forums/20603-WonderKidGGProto2TaiwaneseVariants#135975

My summary:
- mapper max size is 128KB, page numbers past that wrap around to the beginning again
- initial mapping for 0x8000-0xBFFF is equivalent to page number 2
- PCB pics!

PR #1963, opened 2025-09-03
MSXPi Extension added

costarc:
This change implements a new MSXPiDevice for an extension named MSXPi.

MSXPi is a hardware & software interface I developed and is hosted in https://github.com/costarc/MSXPi

The openMSX implementation of the interface allows mostly the same features without he need of a Raspberry Pi. The openMSX device interfaces with the same MSXPi server-side component used in the physical project, which is a Python program running in the local host.


costarc:
Github checks failed - seems the code does not compile for the platforms in the failed checks.
I have opened two inssues in my fork to fix that asap.

m9710797:
Hi,

First of all, thanks a lot for this PR!
But I do have some questions, some technical, some non-technical:

* I wanted to find out more about this project, so I followed the link to https://github.com/costarc/MSXPi. But then the first thing I read on that page is:
> This project is discontinued ...

* Also in the description of this PR you mention "... a python program running on the local host ...". I couldn't find that program anywhere (within a couple of minutes searching the `MSXPi` project page).

So can you clarify the status of the project. I'm also interested in the protocol between the MSX and that python program. It's not needed to understand the implementation of the `MSXPiDevice` in openMSX (I think). But I'm curious. Is there any documentation available?

Next, you've already noticed this yourself. This code doesn't build on Windows. All code in openMSX needs to compile on Linux, macos and Windows. (In some cases it's fine to disable some code on some of the platforms). At first sight the reason for the failure is the network communication. We do have some abstraction for that already in openMSX. I didn't check in detail, but maybe it's already sufficient for you. Check the files `src/events/Socket.hh` and `src/events/Socket.cc`.

In `MSXPi.xml` you have this line:
```
<io base="0x56" num="8"/>
```
This registers 8 IO ports: 0x5a..0x61. But in the implementation of the device I see you only need ports 0x56 and 0x5a. Any reason why you register 8 ports instead of just these two?

Then more about the implementation itself. You have a helper thread that handles both read and write via *busy waiting* (poll every 1ms). That should be avoided (it wastes CPU time). One possible solution is to perform writes in the main thread and only handle reading in the helper thread (via a blocking read now). You can get inspiration from `src/serial/RS232Tester.cc` that implements a `Pluggable` (more or less comparable to a `MSXDevice`) that also both reads and writes. Writes are done directly in the main thread, reads are done in a helper thread. Though the difference is that `RS232Tester` reads and writes files, while you read/write a network socket. Or you can look at `src/events/CliConnection.cc`, that does read/write a network socket, again write from main thread, read via a helper thread.

I do have more minor suggestions for tweaks in the c++ code, but none are super important. We'll discuss those once the above points are resolved.

Thanks.

Wouter

costarc:
@m9710797:

I am back maintaining the MSXPi project. I am in the process of merging the changes I did to support openMSX into my master branch and then will update the status in my repository.

Thbe server-side runs on Python. I was not sure I could add non-openMSX specific code in my PR. Maybe this is a good opportunity to ask if I can, should or must, and then where? Which folders in openMSX repository should I add the server code, configuration files, maybe the BIOS (ROM) as well ? Let me know - hopefully you are able to give all these answers, are you a maintainer of the project ?

Yes, there are documentation. Have a look at the "MSXPi Users and Developers Guide.pdf" here: https://github.com/costarc/MSXPi/tree/master/documents

Regarding the compilation error: I am aware of the issue and what needs to be done, just afraid I may not be able to do it all alone timely (that is, prepare both a Mac and Windows envivonment where I can compile the project). Help wanted!

Regarding the I/O ports, they are actually 8 on range 0x56 to 0x5A. For historical reasons, I reserved 8 but ended up using 3 only, being 0x5A non-consecutive.
0x56 is Status Register
0x57 is Version Register
0x5A Data Input/Output

I could change that, surely, and maybe I will, to two I/O ports only. My problem now is that all the physical interfaces out there have these ports configured in the CPLD - some users may not be happy when they have to grab a USB-Blaster to reconfigure the CPLD!

The implementation: All done by CoPilot and ChatGPT, as I know nothing about C++ and knew nothing about openMSX architecture before I started this about 5 days ago!! I hoped there was not much to be concerned as the code is pretty simple and small, but if the polling thread is probably going to be an overhead, I defintely want to change and optimize that. Probably there are a lot to optimize! Again, help wanted.

MBilderbeek:
> Thbe server-side runs on Python. I was not sure I could add non-openMSX specific code in my PR. Maybe this is a good opportunity to ask if I can, should or must, and then where? Which folders in openMSX repository should I add the server code, configuration files, maybe the BIOS (ROM) as well ? Let me know - hopefully you are able to give all these answers, are you a maintainer of the project ?

Haha, @m9710797 is the main developer of openMSX! He didn't ask you to add the Python script to the project, but he asked you where that Python script can be found in *your* project.

> Regarding the I/O ports, they are actually 8 on range 0x56 to 0x5A. For historical reasons, I reserved 8 but ended up using 3 only, being 0x5A non-consecutive. 0x56 is Status Register 0x57 is Version Register 0x5A Data Input/Output
> 
> I could change that, surely, and maybe I will, to two I/O ports only. My problem now is that all the physical interfaces out there have these ports configured in the CPLD - some users may not be happy when they have to grab a USB-Blaster to reconfigure the CPLD!

The suggestion was not to change the ports, but to only register the used ports in the xml file that describes the MSX hardware.

m9710797:
Thanks for the quick update.

Note: at this point I'm still mostly trying to get to know/understand your PR. I'm making suggestions for changes, but maybe not all of these make sense.

E.g. if you say the real hardware uses 8 IO ports, then that's a good reason to do the same in openMSX. For sure you don't have to make any changes in the real hardware. (If I look at the openMSX code then registering 8 ports was unexpected).

About linux/macos/windows, there's no need for you to setup build environments on 3 systems. You can make a reasonable attempt on one system, push that, and then the automatic builds will run as github actions on the 3 systems.

Thanks for trying to add windows network code. But you misunderstood my suggestion. The idea was to USE our `Socket.hh` file (not recreate it). For example: instead of using `close()` on linux and `closesocket()` on windows, both use the `sock_close()` function from our `Socket.hh` file. Ideally there should be no platform specific code in `MSXPiDevice`, all platform differences should (already) be handled in the existing `Socket` file.

costarc:
> Thanks for the quick update.
> 
> Note: at this point I'm still mostly trying to get to know/understand your PR. I'm making suggestions for changes, but maybe not all of these make sense.
> 
> E.g. if you say the real hardware uses 8 IO ports, then that's a good reason to do the same in openMSX. For sure you don't have to make any changes in the real hardware. (If I look at the openMSX code then registering 8 ports was unexpected).
> 
> About linux/macos/windows, there's no need for you to setup build environments on 3 systems. You can make a reasonable attempt on one system, push that, and then the automatic builds will run as github actions on the 3 systems.
> 
> Thanks for trying to add windows network code. But you misunderstood my suggestion. The idea was to USE our `Socket.hh` file (not recreate it). For example: instead of using `close()` on linux and `closesocket()` on windows, both use the `sock_close()` function from our `Socket.hh` file. Ideally there should be no platform specific code in `MSXPiDevice`, all platform differences should (already) be handled in the existing `Socket` file.


The msxpi server is here:
https://github.com/costarc/MSXPi/tree/release/v1.2_candidate/software/Server/Python/src

The msxpibios.rom is here:
https://github.com/costarc/MSXPi/tree/release/v1.2_candidate/software/openMSX/share/systemroms/extensions

MSX-DOS commands for MSXPi here:
https://github.com/costarc/MSXPi/tree/release/v1.2_candidate/software/target

BASIC programs here:
https://github.com/costarc/MSXPi/tree/release/v1.2_candidate/software/target/BASIC

Documentation here:
https://github.com/costarc/MSXPi/tree/release/v1.2_candidate/documents

m9710797:
Hi. From reading the code I had no idea you were not familiar with c++. Good job.

Thanks for the update. I had a more detailed look at the code and then decided to give some help (also because you're new to c++). See the attached two files (remove the `.txt` extension):
[MSXPiDevice.cc.txt](https://github.com/user-attachments/files/22142829/MSXPiDevice.cc.txt)
[MSXPiDevice.hh.txt](https://github.com/user-attachments/files/22142830/MSXPiDevice.hh.txt)

Here are some highlights of the changes I made:
* Could it be you made a mistake in your last update, and accidentally changed IO port 0x5a to 0x57? I changed that back to 0x5a, openMSX should match the existing hardware. 
* No more `#include` of platform specific files. All is handled via our `Socket.hh` abstraction (for the network stuff) and via `Timer.hh` (for the `sleep()` stuff).
* In fact most of the `#ifdef _WIN32` stuff is gone. (Almost) all differences between Linux and windows should be hidden inside `Socket.hh` and `Timer.hh`.
* I renamed `serverThread` to `readLoop`. The stuff in `MSXPiDevice` is not really a server.
* I created a helper function `close()` to close the socket and set the socket-variable to the invalid value. Because this sequence was repeated several time.
* I've renamed `sockfd` to `sock` (`socket` would be a better name, but that conflicts with the function with this name). because on Linux it's indeed a file descriptor. But on Windows it's a SOCKET handle. Our file `Socket.hh` also hides these differences. In our code we can just use `SOCKET` as the type, and don't worry about the platform differences.
* I've removed `txQueue`. Now that we handle sending from the main thread this queue is no longer needed.
* I've simplified the implementation of `readIO()` and `peekIO()` a bit. Handling of port 0x56 was identical between both, so I don't repeat that code. I tried to take the lock (`mtx`) for smaller sections in the code, only when really doing stuff with `rxQueue`. Because `readIO()` and `peekIO()` are so closely related, I've moved them closer together in the code.
* Before we had both the variables `serverAvailable` and `sockfd` (now renamed to `sock`). But these two both indicated the same thing: `serverAvailable` should always be equal to `sock != OPENMSX_INVALID_SOCKET`. So I removed `serverAvailable`.
* Ah, I did make one functional change on reading IO port 0x5a: before when no data was available in `rxQueue` the variable `readRequested` was NOT set to `false`. Now it's always set to false after reading port 0x5a. I thought that was more logical, but please carefully review this change. Is my change correct?
* I've made similar simplifications in `writeIO()`.
* I've made major changes in the helper thread (in `readLoop()):
  * The windows socket startup/cleanup code is all handled via `SocketActivator` (part of `Socket.hh`). This also takes care of not initializing this subsystem twice when e.g. some other part of openMSX already initialized it.
  * I no longer use `inet_pton()` to parse the fixed string `127.0.0.1` instead I directly initialize the address with `INADDR_LOOPBACK`. I also removed the member variables `serverIP` and `serverPort`. If in the future we want to make the port configurable(??) we may want to re-add this. (I don't think we should make the IP-address configurable).
  * You use `select()` to be able to specify a timeout. But in a way that's still busy-looping. I replace that with a *blocking* read call. And to abstract some minor differences between Linux and Window, we replace `recv()` with `sock_recv()` from our `Socket.hh` file.
  * In the previous point I say we do a blocking read, that's only partially true. On Windows it's true, there `sock_recv()` really blocks until there is data available. On Linux we instead use another openMSX utility `Poller`. We call `poller.pol()`, that one blocks until there is data available, and only then we read that data via `sock_recv()`.
  * The reason for this Window/Linux difference is related to stopping the thread. On Windows when the main thread calls `sock_close()` the blocking call to `sock_revc()` does return with an error. In other words: it unblocks that call. On Linux (or more general on Unix platforms) that behavior is not guaranteed, and there we need a more complex mechanism. But that's all hidden in the `poller.poll()` call. That one does get unblocked when the main thread calls `poller.abort()`.
  * I left one TODO in the code: after we've received our data via `sock_recv()` we put that data in `rxQueue`. But currently that queue can grow unbounded in size. Probably we should give it an upper limit, so that the python server cannot arbitrarily keep on filling memory in the openMSX process.
* I've removed the variables `statusReg`, `controlReg` and `dataReg`. They were not used.
* I've renamed `running` to `shouldStop`. That better reflects the intend of this variable (request the helper thread to stop).
* I've changed the `rxQueue` type from `std::queue` to `cb_queue`. This is an openMSX specific class (so CoPilot likely doesn't know about it). It's a queue based on a circular buffer, it's a bit faster than `std::queue` (which is internally based on a `std::deque`), and in a few cases it offers a more convenient interface (e.g. it offers `clear()` and a `pop_front()` which in one call both returns the front value and removes it from the queue).

BIG CAVEAT:
I made lots of changes. But I ONLY tested that the code still compiles fine on Linux. I did not test in any way that it still works. I'm counting on you to carefully review my changes and test it.

One more minor thing: I see you've added the files in a new directory `src/MSXPi`. That's not wrong, but typically we don't add new directories for every new device. Instead just directly add the files in `src/`.

costarc:
m9710797, many thanks for the changes to the code.
I accepted and implemented also the other suggestions:
- Changed I/O data port back to 0x5A so it stays compatible with all MSXPi out there
- Moved the code from src/MSXPi/ to src/
- Implemented a limit to the queue size (I set to 16KB for now )
Data is transfered in blocks, and the protocol assures the server waits for the MSX to be ready before sending next block, so this should not be a problem.

Compilation on Linux was clean, and functional tests passed, it all looks good. 
Really appreciated the help, many thanks.

build/msvc/openmsx.vcxproj
```diff
@@ -1661,6 +1662,7 @@ popd</Command>
   </ItemGroup>
   <ItemGroup>
     <ClInclude Include="$(OpenMSXSrcDir)\sound\fake_windows.h" />
+	<ClInclude Include="$(OpenMSXSrcDir)\MSXPi\MSXPiDevice.hh" />
```
m9710797: Files have moved, the location should be changed here as well.

share/extensions/MSXPi.xml
```diff
@@ -0,0 +1,28 @@
...
+    <code/>
+    <release_year/>
+    <description>MSXPi is an extension for openMSX that emulates the MSXPi hardware interface, enabling an MSX computer to interact with modern systems. Unlike the physical version, the emulated interface does not require a Raspberry Pi — it communicates with a Python-based server component that provides access to web resources, files, REST APIs, and other external services.</description>
+    <type>pluggable</type>
+  </info>
+  <devices>
+    <primary slot="any">
+      <secondary slot="any">
+        <MSXPiDevice id="MSXPi">
+          <io base="0x56" num="8"/>
```
m9710797: 0x56..0x5a that's 5 ports, right? Not 8.
But even better would be to really only register the required ports:
```
<io base="0x56" num="1"/>
<io base="0x5a" num="1"/>
```
Unless you expect that future versions will use more ports than just 0x56 and 0x5a? You know the hardware and your future plans better, you decide.

src/MSXPiDevice.cc
```diff
@@ -0,0 +1,149 @@
...
+		if (n < 0) { // error
+			close();
+			continue;
+		}
+		std::lock_guard lock(mtx);
+		for (ssize_t i = 0; i < n; ++i) {
+			if (rxQueue.size() < MAX_QUEUE_SIZE) {
+				rxQueue.push_back(buf[i]);
+			} else {
+				break; // skip this byte
```
m9710797: Is it better to skip adding new bytes, or to drop the oldest bytes? I'm not sure, you know the protocol better. (Normally this shouldn't happen, but it's anyway better to make this robust).

m9710797 (commented):
Thanks for the update.
Here are some more small suggestions.

src/MSXPiDevice.cc
costarc: dropping older bytes may cause interruption in healthy blocks that are being ingested by MSX, so I choose to stop accepting new bytes instead.

share/extensions/MSXPi.xml
costarc: That (8 ports) reflects the current state of the physical hardware documentation, as originally designed. Does not make difference if I change the extension to 5, since the ports 0x5B-0x5D are not used anyway. But same is true to ports 0x57-0x59, reason I originally accepted you sugestion to change ports assignment and use only 0x56 and 0x57.

I now wnt to keep the design (both extension and hardware) as is, but then I think would be better to have the MSXPi.xml configured to use only two ports:
<io>
  <port>0x56</port>
  <port>0x5A</port>
</io>
I suppose that syntax is accepted by openMSX?

Is this gap going to be a problem other than just inconvenience ? If there are no objections, I would proceed with this change.

costarc: Yes, that is the best option:

> 0x56..0x5a that's 5 ports, right? Not 8. But even better would be to really only register the required ports:
> 
> ```
> <io base="0x56" num="1"/>
> <io base="0x5a" num="1"/>
> ```
> 
> Unless you expect that future versions will use more ports than just 0x56 and 0x5a? You know the hardware and your future plans better, you decide.

The allocation of 8 ports reflected the current state of the physical hardware documentation, as originally designed. Does not make difference if I change the extension to 2 ports, since the ports 0x57-0x59 and 0x5B-0x5D are not used anyway. Its the best choice indeed, will do it.

build/msvc/openmsx.vcxproj
costarc: Ok, will do - I do not even recall making this update in first place - thanks for pointing out.

src/MSXPiDevice.cc
```diff
@@ -0,0 +1,149 @@
...
+#endif
+		std::array<char, 64> buf;
+		auto n = sock_recv(sock, buf.data(), buf.size());
+		if (n < 0) { // error
+			close();
+			continue;
+		}
+		std::lock_guard lock(mtx);
+		for (ssize_t i = 0; i < n; ++i) {
+			if (rxQueue.size() < MAX_QUEUE_SIZE) {
```
m9710797: Instead of testing the size in each iteration of the loop, we can calculate how many bytes will fit just once outside the loop. See the attached patch.
[tweak.patch.txt](https://github.com/user-attachments/files/22175861/tweak.patch.txt)

I've also moved the definition of `MAX_QUEUE_SIZE` from the `.hh` file to here: it's a local internal implementation detail, no need to expose it in the header file.

m9710797 (commented):
Thanks again. One final(?) tweak. 

The code looks (close to?) ready to me. Though I must say I didn't find any time to actually try to run it myself yet. I'll try to make time for that this weekend.

It would be nice if others could help test as well. Maybe it'll help if you could post somewhere clear/concise instructions on how to use this:
* Where to download the `.rom` file and the python program.
* How to start the python program (and when, before openMSX starts, after, doesn't matter?).
* Some examples of things to try once openMSX is booted with the `MSXPi` extension.

I know that you've already pointed to (a lot) of documentation. But that's maybe too much at once. Few people will want to read +40 pages to help test this.

MBilderbeek:
I'm a bit reluctant to try and run this on my Linux desktop PC. The server script contains several dangerous commands with `sudo` in front of them....

src/MSXPiDevice.cc
costarc: Thank you. Applying the patch now.
Will surely do a quick-start gudie, very short. Will also address concerns about the server using "sudo" - please read details on my answer to [MBilderbeek] below.

costarc:
> I'm a bit reluctant to try and run this on my Linux desktop PC. The server script contains several dangerous commands with `sudo` in front of them....

Understandable.
I am prioritizing the openMSX extension, and have not done any effort to fully port the msxpi-server for other platforms.

Keep in mind that the MSXPi needs a plugged Raspberry Pi Zero W to the interface, MSX needs full control, reason the "sudo" in the server component, which are there for commands that require elevated privileges such as for WiFi configuration, etc.

The way the msxpiserver works is:
Each MSX command has a corresponding method in the msxpi-server.

For example:
pcd.com -> def pcd()
prun.com -> def prun(cmd = ''):

and so on.

If you do not want to wait for me to clean the server component (which I will), just delete the python methods that include "sudo", which are only three:

def preboot():
def pshut():
def pwifi():

You do not need those to test.

src/MSXPiDevice.cc
costarc: Also one final comment: The build for Windows-VC x64 refuses to compile. I see this as a big issue as many users are on Windows. Hopefully some Win dev can step in and help fix those last issues and get a clean build for Windows.

src/DeviceFactory.cc
```diff
@@ -95,7 +96,7 @@
 #include "components.hh"
 #include "one_of.hh"
 #include <memory>
-
+
```
MBilderbeek: Why change this whitespace?

src/MSXPiDevice.cc
MBilderbeek: Looks like it's choking on `ssize_t`:
```
         D:\a\openMSX\openMSX\src\MSXPiDevice.cc(137,8): error C2065: 'ssize_t': undeclared identifier [D:\a\openMSX\openMSX\build\msvc\openmsx.vcxproj]
```
I have never heard of that type... perhaps `auto` is enough here?

m9710797: `ssize_t` is part of POSIX (not c++), so it works in Linux, but not Windows.

MBilberbeek's suggestion to use `auto` will work. Like on line 131: `auto n = sock_rec(...);` there we also let the compiler deduce the type of `n` instead of explicitly naming it ourselves.

In my suggested tweak (with `for (auto i : xrange(...))`) I also (unintentionally) avoid the problem.

src/DeviceFactory.cc
costarc: some mistake. I will revert that change.

> Why change this whitespace?

costarc:
I added a README inside Contrib folder. It has instructions to start up with the MSXPi in openMSX, and can be seen here in this PR, or else head to my repo to get it: https://github.com/costarc/MSXPi/tree/master/software/openMSX

patch
m9710797, line 1: I assume that adding this file was a mistake?

costarc: > I assume that adding this file was a mistake?

Yes, a mistake unfortunately. Removed with a commit already.

costarc:
Being tested extensively over last days.
I had to make some decisions regarding the implementation, to assure compatibility with physical interface.
I will continue the tests and will be using the builds from these checks here (very helpful) to continue.

MBilderbeek:
There are builds being made automatically, right?
Also, the branch has conflicts, did you notice? These need to be resolved.

costarc:
> There are builds being made automatically, right? Also, the branch has conflicts, did you notice? These need to be resolved.

No, what i see in my PR is:
**_No conflicts with base branch
Changes can be cleanly merged._**

costarc:
> There are builds being made automatically, right? Also, the branch has conflicts, did you notice? These need to be resolved.

Yes, I got confused with the "2 checks passes" flagged, and though the builds had not been done.
Forgive these mistakes, this is literraly the 1st PR I am using this nightly build feature.

MBilderbeek:
> > There are builds being made automatically, right? Also, the branch has conflicts, did you notice? These need to be resolved.
> 
> No, what i see in my PR is: **_No conflicts with base branch Changes can be cleanly merged._**

In this PR page I clearly see 
image

costarc:
> > > There are builds being made automatically, right? Also, the branch has conflicts, did you notice? These need to be resolved.
> > 
> > 
> > No, what i see in my PR is: **_No conflicts with base branch Changes can be cleanly merged._**
> 
> In this PR page I clearly see image

Are we looking in different places? :)

This is what I see:

image

MBilderbeek:
> > > > There are builds being made automatically, right? Also, the branch has conflicts, did you notice? These need to be resolved.
> > > 
> > > 
> > > No, what i see in my PR is: **_No conflicts with base branch Changes can be cleanly merged._**
> > 
> > 
> > In this PR page I clearly see image
> 
> Are we looking in different places? :)
> 
> This is what I see:
> image

Strange, just looking at the bottom part of this page, which is https://github.com/openMSX/openMSX/pull/1963

image

costarc:
@m9710797 @MBilderbeek 
@pvmm and myself cleaned up the commits.
Debug messages to console also removed from the code.
The extension is in a stable and working condition, verified to work with the same software used in the actual physical interface.
Going forward, all changes to MSXPi software will take into consideration this extension, to carry on campatibility.

Let me know if there is something else I need to do to have the extension accepted and merged to openMSX code.

m9710797:
I seem to recall a discussion about the python script containing stuff like `sudo reboot` etc.
Has that been addressed?

That python server is the main (only) purpose for this extension, right? As long as the script contains stuff like that, we can't really recommend using it in combination with openMSX.

Apart from this, I believe, all other technical concerns have been addressed, right?

costarc:
> I seem to recall a discussion about the python script containing stuff like `sudo reboot` etc. Has that been addressed?
> 
> That python server is the main (only) purpose for this extension, right? As long as the script contains stuff like that, we can't really recommend using it in combination with openMSX.
> 
> Apart from this, I believe, all other technical concerns have been addressed, right?

I added a msxpi-server-generic version for non-raspberrypi hosts with all sudo commands removed. Commands to reboot and shutdown the hosts was also removed.

The original version that runs on Raspberry Pi needs the sudo on the commands that control the raspberry pi, since it is inegral part of the interface and its only purpose is to process MSX commands. That cannoe be removed otherwise MSX cannot configur wifi, reboot the PI or shutdown via MSX-DOS commands.


Edit: Technical issues has been solved. The extension works similar to the physical counter part.

m9710797:
Thanks. But may I propose a different solution for the following reasons:
* Having two (almost the same) versions of the script makes it more difficult for you to maintain both.
* The documentation still refers to the non-generic version of the script. But even if the documentation is fixed, users can then still accidentally download/run the wrong version.

Would it be a good idea to instead:
* Only have a single version.
* Can the script somehow detect whether it's running on a generic host or on a Raspberry Pi connected to a MSX? E.g. detect this at startup and remember for later.
* The above check must be robust/reliable, otherwise this proposal doesn't work.
* Around the "dangerous" commands add an `if` and then either do execute the command on the Raspberry Pi, or print a message containing what command would have been executed if we were running on a Raspberry Pi.

Maybe it's also a good idea to do the following:
* When the python script starts, print a message like this (please phrase this in a better way, you're much more familiar with the possibilities than I am):
```
This server process is meant to handle communication with an MSX computer. It allows the MSX to:
* List/read (any) file from this computer or network (via the the PDIR/PCOPY commands).
* Execute arbitrary(!) shell commands (via the PRUN command).
* Configure the WiFi settings (via the PSET/PWIFI commands).
...
```
* Of course this message should be adapted for the generic and the Raspberry Pi version. (But don't write two completely separate text blocks. Just omit printing selected lines depending on the host-test).
* I assume that stdout of the MSXPi won't be directly visible to the users? But that's OK. Even in that case I would still just print it, if only to keep both versions of the script more similar.
* But especially for the generic version it would be good to explicitly list _all_ the capabilities that this script gives to the MSX.
* E.g. you did disable the `sudo reboot` stuff. But in principle the `PRUN` command can do exactly the same.
* That's not a problem per se. In fact the whole point of this server is to allow the MSX access to the host machine. But I think it's a good idea to explicitly inform the users about this capability. E.g. users may not realize how powerful the `PRUN` command is.

costarc:
> Thanks. But may I propose a different solution for the following reasons:
> 
> * Having two (almost the same) versions of the script makes it more difficult for you to maintain both.
> * The documentation still refers to the non-generic version of the script. But even if the documentation is fixed, users can then still accidentally download/run the wrong version.


Yes, I agree. I have this already in place, in the code, but thought you guys still did not like the fact that the sudo is there, reason I split creating one version completelly stripped off all commands that require sudo.
I rather have one code - its not like its compiled or obscured - its python, very readable - anyone concerned about security could just delete the code in python commands they don't like. But I also agree in adding checks for platform and ksipping those commands - more below.

> 
> Would it be a good idea to instead:
> 
> * Only have a single version.
> * Can the script somehow detect whether it's running on a generic host or on a Raspberry Pi connected to a MSX? E.g. detect this at startup and remember for later.

Yes, that's how the server was configured. I added the function detect_host() and use it thorughtout the program to make decisions based on the platform. This function is there for some time already.

> * The above check must be robust/reliable, otherwise this proposal doesn't work.
> * Around the "dangerous" commands add an `if` and then either do execute the command on the Raspberry Pi, or print a message containing what command would have been executed if we were running on a Raspberry Pi.

Yes, done for for example for "pwifi set" where sudo is needed on Raspberry Pi. 
In other platforms, it returns a message "Not supported in this platform"


> 
> Maybe it's also a good idea to do the following:
> 
> * When the python script starts, print a message like this (please phrase this in a better way, you're much more familiar with the possibilities than I am):
> 
> ```
> This server process is meant to handle communication with an MSX computer. It allows the MSX to:
> * List/read (any) file from this computer or network (via the the PDIR/PCOPY commands).
> * Execute arbitrary(!) shell commands (via the PRUN command).
> * Configure the WiFi settings (via the PSET/PWIFI commands).
> ...
> ```
> 


Sure, can do that - Raspberry Pi users won't see that, but its a good idea for other platforms.
Will add, very easy work.

> * Of course this message should be adapted for the generic and the Raspberry Pi version. (But don't write two completely separate text blocks. Just omit printing selected lines depending on the host-test).
> * I assume that stdout of the MSXPi won't be directly visible to the users? But that's OK. Even in that case I would still just print it, if only to keep both versions of the script more similar.
> * But especially for the generic version it would be good to explicitly list _all_ the capabilities that this script gives to the MSX.
> * E.g. you did disable the `sudo reboot` stuff. But in principle the `PRUN` command can do exactly the same.
> * That's not a problem per se. In fact the whole point of this server is to allow the MSX access to the host machine. But I think it's a good idea to explicitly inform the users about this capability. E.g. users may not realize how powerful the `PRUN` command is.

Yes, prun have full access - and you can pass sudo as parameter, so does not matter how much gates you put in the code if prun can still bypass them.

So lets agree on this: enhance and update the server to help user to avoid mistakes - skipping elevation of privileges via MSXPi native commands (such as pwifi, preboot, pshut) on non-rasppberry pi platforms only - tht is, the commands stay but they detect when it is not Raspberry Pi and alert the user. PRUN won't be changed. I will add a disclaimer to the server upon starting, so user knows about the need of sudo on raspberry pi and how it's not used in the other platforms.

m9710797:
@costarc I'm sorry if we made you take out the "sudo" stuff if it was already technically disabled. It wasn't clear (to me) that there were already sufficient checks in place that prevented these dangerous commands from ever executing on a non-MSXPi host.

So the summary is:
* When running on something else than MSXPi, the server should not execute commands that "obviously" don't make sense.
* But a simple check for commands like `PRESET`, `PREBOOT`, ... is sufficient. It's not needed (nor doable) to filter arbitrary commands that can be passed via `PRUN`.
* And it would be good to inform the user better about what possibilities the server enables for the MSX (like reading arbitrary files, executing arbitrary commands, ...).

costarc:
> @costarc I'm sorry if we made you take out the "sudo" stuff if it was already technically disabled. It wasn't clear (to me) that there were already sufficient checks in place that prevented these dangerous commands from ever executing on a non-MSXPi host.
> 
> So the summary is:
> 
> * When running on something else than MSXPi, the server should not execute commands that "obviously" don't make sense.
> * But a simple check for commands like `PRESET`, `PREBOOT`, ... is sufficient. It's not needed (nor doable) to filter arbitrary commands that can be passed via `PRUN`.
> * And it would be good to inform the user better about what possibilities the server enables for the MSX (like reading arbitrary files, executing arbitrary commands, ...).

Yes, that's its implemented. Commands specifically for RPi have a check:
def prestart():
    if hostType == "pi":
        print("Restarting MSXPi Server")
        exitDueToSyncError()
    else:
        print("Command not supported in this platform")
        
def preboot():
    if hostType == "pi":
        print("Rebooting Raspberry Pi")
        os.system("sudo reboot")
    else:
        print("Command not supported in this platform")
        
def pshut():
    if hostType == "pi":
        print("Shutting down Raspberry Pi")
        os.system("sudo shutdown -h now")
    else:
        print("Command not supported in this platform")
        
  
And also in pwifi command where sudo is required to configure the wifi network (pwifi set), we got this:
    if (cmd[:1] == "s" or cmd[:1] == "S"):
        if hostType == "pi":
            wifisetcmd = 'sudo nmcli device wifi connect "' + wifissid + '" password "' + wifipasss + '"'
            prun(wifisetcmd)
        else:
            sendmultiblock(b'Parameter not supported in this platform', BLKSIZE, RC_SUCCESS)


And finally, I added a disclaimer that is shown everytime the server runs - it is sent to console or all platforms, even RPi:

** Starting MSXPi Server Version 1.2 Build 20250910.790 **


=================================================================================
This server process is meant to handle communication with a MSX computer.
It allows the MSX to:

 - List/read (any) file from this computer or network (via the the PDIR/PCOPY commands).

 - Execute arbitrary(!) shell commands (via the PRUN command).

 - Configure the WiFi settings (via the PSET/PWIFI commands). 
<br>Some of the commands in Raspberry Pi require elevation of privileges using sudo.
The server will not execute such commands in the other non-Raspberry Pi platforms,
and in some cases when supported by the command, a message will be returned to the MSX
informing that the command is not supported.
=================================================================================

[MSXPi Server Non-Raspberry Pi Platform:win] Listening on 0.0.0.0:5000...

Contrib/README.MSXPi
```diff
@@ -0,0 +1,155 @@
...
+   copy to you openMSX/share/extensions/ folder.
+
+5. Use a DIR as DISK with openMSX to facilitate the tests.
+   Create a directory named FloppyA
+   Copy MSXDOS.SYS and COMMAND.COM to FloppyA
+   Make sure openMSX can boot into MSX-DOS using this folder as a drive:
+   
+   openmsx -machine Panasonic_FS-A1WSX -diska FloppyA
+   
+6. Dowload the MSXPi commands you want to test from:
```
m9710797: Typo: Dowload -> download

I noticed this by accident, I did not do a full spelling check ;-)
But maybe it's a good idea to do that.

share/extensions/MSXPi.xml
```diff
@@ -0,0 +1,31 @@
...
+        <MSXPiDevice id="MSXPi">
+          <io base="0x56" num="1"/>
+          <io base="0x57" num="1"/>
+          <io base="0x5A" num="1"/>
+        </MSXPiDevice>
+        <ROM id="MSXPi BIOS">
+          <mem base="0x4000" size="0x4000"/>
+          <rom>
+            <sha1>be90b05395ccd5435e3eace188d3858a1b99d56a</sha1>
+			<sha1>1b36118b5f436e42d653f0ec8b23f67aa33de858</sha1>
```
m9710797: Typically when we list multiple sha1 tags we also add a small comment explain what they are. Something like:
```
<sha1>1234...</sha1>  <!-- version 1.0 -->
<sha1>abcd...</sha1>  <!-- version 1.1 -->
```

src/MSXPiDevice.cc
```diff
@@ -0,0 +1,154 @@
...
+		return 0xff;
+	}
+}
+
+void MSXPiDevice::writeIO(uint16_t port, byte value, EmuTime /*time*/)
+{
+	switch (port & 0xff) {
+	case 0x56: // control
+		if (value == 0xFF) {
+			reset(openmsx::EmuTime::dummy());
```
m9710797: Instead of passing an explicit dummy time, it's easier to just pass the `time` parameter. So uncomment it on line 87.  That's also more robust for when in the future you do need a meaningful time in `reset()`.

EDIT:
Ah, and should sending `0xff` also immediately set `readRequested = true`? My personal guess is not, but you know the hardware better.

m9710797 (commented):
I did another quick review. Only some minor comments.

---

I did also download the `msxpi-server.py` script. But when I run it it prints (also) this line:
```
 * Configure the WiFi settings (via the PSET/PWIFI commands).
```
That's unexpected to me. I understand that's useful when running on MSXPi. But I probably don't want that the MSX can change the WiFi settings of desktop/laptop running openMSX. Should this still also be disabled?

costarc:
> I did another quick review. Only some minor comments.
> 
> I did also download the `msxpi-server.py` script. But when I run it it prints (also) this line:
> 
> ```
>  * Configure the WiFi settings (via the PSET/PWIFI commands).
> ```
> 

I see.  That is true only on RPi platform. I may add a "(on Raspberry Pi only)" to that line,  to keep the same generic text, since this message is now sent to console on all platforms (although on RPi users will only see if they run the server in foregound).


> That's unexpected to me. I understand that's useful when running on MSXPi. But I probably don't want that the MSX can change the WiFi settings of desktop/laptop running openMSX. Should this still also be disabled?

pwifi will run on PC, but will only display IP settings - that means the command is still useful on PCs.
However if the user try "pwifi set", it will diaplay the "not supported" message.

Contrib/README.MSXPi
costarc: Certainly a good idea.
Will do.

share/extensions/MSXPi.xml
costarc: Got it.
Will do. 
But now I do not see the need for both ROMs, as they are both v1.2, and in the instructions I say the users to get the latets MSXPi.xml which will match the ROM.
I will keep the sha for the ROM in the release files.

src/MSXPiDevice.cc
costarc: > Instead of passing an explicit dummy time, it's easier to just pass the `time` parameter. So uncomment it on line 87. That's also more robust for when in the future you do need a meaningful time in `reset()`.
> 
Chaned that, compilation fine so will push this change.

> EDIT: Ah, and should sending `0xff` also immediately set `readRequested = true`? My personal guess is not, but you know the hardware better.

0x55 is a reset for the SPI in the physical interface, and it is not meant to also signal a read request.
Therefore the software implementation match the interface behaviour.

Contrib/README.MSXPi
```diff
@@ -0,0 +1,162 @@
...
+   https://github.com/costarc/openMSX/releases
+   
+   Always use the latest release.
+   
+2. Download msxpi-server.py from:
+
+   https://github.com/costarc/MSXPi/tree/master/software/Server/Python/src
+
+
+   Note: the msxpi-server component has some commands that requires elevation of privileges
```
pvmm: Note: the msxpi-server component has some commands that **require** elevation of privileges

pvmm (changes requested):
just a small typo.

src/MSXPiDevice.cc
m9710797: Ok, that matches my expectation.
BUT the current implementation WILL do first a `reset` AND then continue with setting `readRequested = true`.
I think the code structure should instead be something like:
```
if (value == 0xFF) {
    reset(time);
} else if (sock != OPENMSX_INVALID_SOCKET) {
    readRequested = true;
}
```

m9710797:
> > ```
> >  * Configure the WiFi settings (via the PSET/PWIFI commands).
> > ```
> 
> I see. That is true only on RPi platform. I may add a "(on Raspberry Pi only)" to that line, to keep the same generic text, since this message is now sent to console on all platforms (although on RPi users will only see if they run the server in foregound).
> 
> > That's unexpected to me. I understand that's useful when running on MSXPi. But I probably don't want that the MSX can change the WiFi settings of desktop/laptop running openMSX. Should this still also be disabled?
> 
> pwifi will run on PC, but will only display IP settings - that means the command is still useful on PCs. However if the user try "pwifi set", it will diaplay the "not supported" message.

I think it will be good if the text closely matches the actual behavior. Because now it gives the wrong impression that also on PC the MSX can make changes to the WiFi settings.  (And I'm ignoring the `PRUN` command .. that can do anything).

That means you can not have a single text that's used for both MSXPi and PC. I already suggested this in an earlier comment: instead of having two completely separate texts, I suggested to have a if statements to select between individual lines. E.g. completely omit printing a line, or select between two versions of a line. I think this structure will increase the chance that the text remains up-to-date (=gets adapted) after future changes.

I'm sorry to stress this point so much. But I think it's important to clearly document what running this server implies. Because it does give MSX programs unexpected capabilities. Unexpected: at least to me, and I assume also to many other users.

costarc:
> > > ```
> > >  * Configure the WiFi settings (via the PSET/PWIFI commands).
> > > ```
> > 
> > 
> > I see. That is true only on RPi platform. I may add a "(on Raspberry Pi only)" to that line, to keep the same generic text, since this message is now sent to console on all platforms (although on RPi users will only see if they run the server in foregound).
> > > That's unexpected to me. I understand that's useful when running on MSXPi. But I probably don't want that the MSX can change the WiFi settings of desktop/laptop running openMSX. Should this still also be disabled?
> > 
> > 
> > pwifi will run on PC, but will only display IP settings - that means the command is still useful on PCs. However if the user try "pwifi set", it will diaplay the "not supported" message.
> 
> I think it will be good if the text closely matches the actual behavior. Because now it gives the wrong impression that also on PC the MSX can make changes to the WiFi settings. (And I'm ignoring the `PRUN` command .. that can do anything).
> 
> That means you can not have a single text that's used for both MSXPi and PC. I already suggested this in an earlier comment: instead of having two completely separate texts, I suggested to have a if statements to select between individual lines. E.g. completely omit printing a line, or select between two versions of a line. I think this structure will increase the chance that the text remains up-to-date (=gets adapted) after future changes.
> 
> I'm sorry to stress this point so much. But I think it's important to clearly document what running this server implies. Because it does give MSX programs unexpected capabilities. Unexpected: at least to me, and I assume also to many other users.

This feels a bit like a rabbit hole.
We may need to settle to something, so I am suggesting these changes below - see how it is displayed once the server is started, and of course give your suggestions so we can settle to a final text:

(ignore the formatting - I simply cannot get a blank line after the asterisk!!!)

** Starting MSXPi Server Version 1.2 Build 20250910.790 **

==================================================================================
This server process is meant to handle communication with a MSX computer.
It allows the MSX to:

* List/read (any) file from this computer or network (via the the PDIR/PCOPY commands).
* Execute arbitrary(!) shell commands (via the PRUN command).
Some very few commands designed specifically for Raspberry Pi requires elevation of
privileges using sudo - these commands will not be executed in the PC platforms and
when possible, a message will be returned to the MSX informing that the command is
not supported.
However notice that using PRUN, the MSX user can execute any commands in the host,
bypassing the controls in the native MSXPi commands.
==================================================================================
[MSXPi Server on Windows] Listening on 0.0.0.0:5000...

Contrib/README.MSXPi
costarc: Fixed. Will come with next update on this PR.

costarc:
@m9710797 @MBilderbeek 
Is there anything else I can do to help you decide about merging this change?

m9710797:
Ah, I still saw a lot of activity the last few days (I admit I didn't look at that in detail), so I had the impression you were still actively working on this PR.

I had another quick look just now. The openMSX side looks fine to me.
There was still some discussion on the server script. Is that side also ready?

costarc:
> Ah, I still saw a lot of activity the last few days (I admit I didn't look at that in detail), so I had the impression you were still actively working on this PR.
> 
> I had another quick look just now. The openMSX side looks fine to me. There was still some discussion on the server script. Is that side also ready?

I am happy with the current state of the extension, is behaves exactly like the actual interface. Of course if I detect differences and bugs or add improvements, I will submit new PRs in the future.

As for the server side, I thought our discussed was very fruitful and beneficial, so I implemented both the disclosure at the startup of the server, and the checks in the code.

Contrib/README.MSXPi
pvmm: Note: The msxpi-server component has some commands that **require** elevated privileges on the Raspberry Pi platform using "sudo".

pvmm (approved):

PR #1969, opened 2025-09-11
Add fullscreen horizontal screen stretching option

TitanX101:
### Summary

This PR introduces a new video setting that enables horizontal screen stretching when the application is running in fullscreen mode. The feature allows the image to scale dynamically to match the monitor's width, improving visual coverage without altering the vertical resolution.

### Implementation Details

- A new **fullscreen-only horizontal stretch** option was added to the video settings menu, allowing users to toggle the behavior.
- To support this feature, three new methods were added to the `Display` class:
  - `setFullScreen(bool flag)`: Tracks whether the application is currently in fullscreen mode.
  - `getDisplayWidth()`: Retrieves and caches the current monitor width using `SDL_GetDesktopDisplayMode`.
  - `getDisplayHeight()`: Retrieves and caches the current monitor height similarly.

These additions allow the renderer to access accurate display dimensions and apply horizontal scaling only when appropriate.

### Notes

- The stretching logic is isolated to fullscreen mode to avoid unintended distortion in windowed configurations.
- Display dimensions are cached to potentially reduce redundant system calls and improve responsiveness.

Feedback on edge cases or integration with other scaling modes is welcome.

### DIFF
--- a/src/video/PostProcessor.cc
+++ b/src/video/PostProcessor.cc
@@ -305,7 +305,13 @@ void PostProcessor::paint(OutputSurface& /*output*/)
 	renderedFrame.tex.bind();
 	auto [x, y] = screen.getViewOffset();
 	auto [w, h] = screen.getViewSize();
-	glViewport(x, y, w, h);
+
+	bool isFullStretching = renderSettings.getFullscreenHorizontalStretch() && display.isFullScreen();
+
+	if (isFullStretching)
+		glViewport(0, y, display.getDisplayWidth(), h);
+	else
+		glViewport(x, y, w, h);
 
 	if (deform == RenderSettings::DisplayDeform::_3D) {
 		drawMonitor3D();
--- a/src/video/PostProcessor.cc
+++ b/src/video/PostProcessor.cc
@@ -303,15 +303,18 @@ void PostProcessor::paint(OutputSurface& /*output*/)
 
 	renderedFrame.fbo.pop();
 	renderedFrame.tex.bind();
-	auto [x, y] = screen.getViewOffset();
-	auto [w, h] = screen.getViewSize();
-
-	bool isFullStretching = renderSettings.getFullscreenHorizontalStretch() && display.isFullScreen();
-
-	if (isFullStretching)
-		glViewport(0, y, display.getDisplayWidth(), h);
-	else
+	
+	if (renderSettings.getFullStretch())
+	{
+		auto [w, h] = screen.getPhysicalSize();
+		glViewport(0, 0, w, h);
+	}
+	else 
+	{
+		auto [x, y] = screen.getViewOffset();
+		auto [w, h] = screen.getViewSize();
 		glViewport(x, y, w, h);
+	}
 
 	if (deform == RenderSettings::DisplayDeform::_3D) {
 		drawMonitor3D();
--- a/.gitignore
+++ b/.gitignore
@@ -13,5 +13,7 @@ core
 *.VC.db
 *.VC.VC.opendb
 .vscode/
+.vs/
+*.vcxproj.user
 __pycache__/
 .DS_Store
--- a/doc/manual/commands.html
+++ b/doc/manual/commands.html
@@ -113,6 +113,7 @@
         <li><a class="internal" href="#gamma">gamma</a></li>
         <li><a class="internal" href="#glow">glow</a></li>
         <li><a class="internal" href="#grabinput">grabinput</a></li>
+        <li><a class="internal" href="#full_stretch">full_stretch</a></li>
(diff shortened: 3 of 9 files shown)


src/video/PostProcessor.cc
```diff
@@ -305,7 +305,13 @@ void PostProcessor::paint(OutputSurface& /*output*/)
 	renderedFrame.tex.bind();
 	auto [x, y] = screen.getViewOffset();
 	auto [w, h] = screen.getViewSize();
-	glViewport(x, y, w, h);
+
+	bool isFullStretching = renderSettings.getFullscreenHorizontalStretch() && display.isFullScreen();
+
+	if (isFullStretching)
+		glViewport(0, y, display.getDisplayWidth(), h);
```
m9710797: It's been a while since I last looked at this code. So for me part of it is also figuring out how exactly this stuff works again. (So for sure it's possible I made mistakes in the stuff below).

I wonder if this new function `Display::getDisplayWidth()` is needed ...

A few lines above we query `screen.getViewOffset()` and `getViewSize()`. This 'offset' (`x` and `y`) is actually responsible for the black borders. But if `screen` can calculate the size of those borders it must also know the dimensions of the full screen. Looking a bit closer: `screen` has also a method `OutputSurface::getPhysicalSize()`, doesn't that already give the information you need?

---

And if the above is true, then maybe the call to `display.isFullScreen()` also isn't needed?

In windowed mode `offset` will always be `(0, 0)` (there are no black borders in windowed mode). I also think in windowed mode `OutputSurface::getPhysicalSize()` and `OutputSurface::getLogicalSize()` will be the same. 
 That means there won't be a difference between `glViewPort(0, 0, phys.x, phys.y)` and `glViewPort(x, y, w, h)`.

And thus the check for `display.isFullScreen()` is maybe not needed?

---

If both points above are true, then that would allow to simplify your PR (remove your changes to `Display.hh` and `Display.cc`).

m9710797 (commented):
Thanks a lot for this PR!

In general the code looks good. But maybe it can be done simpler? See comments below.

src/video/PostProcessor.cc
TitanX101: Thanks for pointing that out! After re-testing with OutputSurface::getPhysicalSize(), I confirmed it already provides the correct dimensions for stretching, both in fullscreen and windowed modes. I must have overlooked it earlier while experimenting with other OutputSurface members.

Based on your feedback, I simplified the implementation by removing the extra methods I had added to Display.hh/cc. The stretch logic now relies directly on getPhysicalSize(), which makes the code much cleaner.

Additionally, I renamed the setting in the ImGui menu from Fullscreen Horizontal Stretch to Full Stretch for better clarity, since the behavior is not tied strictly to fullscreen anymore.

Let me know if you think further adjustments are needed!

m9710797:
Ok, I finally found time to fully review this PR.

Two minor issues:
* The code generates the following compiler warning (gcc). Should be an easy fix (use the same order in the initializer list (in `RenderSettings.cc`) as the order in which the settings member variables are declared (in `RenderSettings.hh`)):
```
In file included from src/video/RenderSettings.cc:1:
src/video/RenderSettings.hh: In constructor ‘openmsx::RenderSettings::RenderSettings(openmsx::CommandController&)’:
src/video/RenderSettings.hh:248:24: warning: ‘openmsx::RenderSettings::fullStretchSetting’ will be initialized after [-Wreorder]
  248 |         BooleanSetting fullStretchSetting;
      |                        ^~~~~~~~~~~~~~~~~~
src/video/RenderSettings.hh:245:22: warning:   ‘openmsx::FloatSetting openmsx::RenderSettings::horizontalStretchSetting’ [-Wreorder]
  245 |         FloatSetting horizontalStretchSetting;
      |                      ^~~~~~~~~~~~~~~~~~~~~~~~
src/video/RenderSettings.cc:44:1: warning:   when initialized here [-Wreorder]
   44 | RenderSettings::RenderSettings(CommandController& commandController)
      | ^~~~~~~~~~~~~~
```
* Can you also add this new setting in the manual `doc/manual/commands.html` (e.g. search for `horizontal_stretch` and add something similar for `full_stretch`, don't forget the table of content). Maybe add a sentence about 4:3 vs 16:9 display ratio (why would the user want to enable this setting).

TitanX101:
Allright, fixed the GCC compiler warning by correcting the initialization order between fullStretchSetting and horizontalStretchSetting. Also added complete documentation for the new full_stretch setting in the manual.

m9710797:
Merged, see 50d17b58cf1d6d6c79abad92495b5d5d1a04cc82.

With some minor edits (typo / code layout. No functional changes). Maybe that's why it isn't automatically marked as merged in github? So I'll manually close this.

PR #1970, opened 2025-09-11
Visual Studio Code complains about C++23ism

pvmm:
"3uz" is a size_t (or ptrdiff_t) suffix introduced in C++23 which VC doesn't recognize.

### DIFF
--- a/src/debugger/DasmTables.hh
+++ b/src/debugger/DasmTables.hh
@@ -12,7 +12,7 @@ extern const std::array<const char*, 256> mnemonic_ed;
 extern const std::array<const char*, 256> mnemonic_xx;
 extern const std::array<const char*, 256> mnemonic_main;
 
-extern const std::array<uint8_t, 3uz * 256> instr_len_tab;
+extern const std::array<uint8_t, 3u * 256> instr_len_tab;
 
 } // namespace openmsx
 


m9710797:
Do you mean a visual studio warning or an error?
Our code compile fine in the github actions with visual studio.

I intentionally wrote that `uz` suffix, here and in a dozen other places.  You only fixed one place, was that only an example? Do you plan to fix the other places as well?  (But then my preference is that your fix doesn't change `uz` to `u` because that also changes the type).

Could it be you're using a too old version of visual studio that doesn't recognize this yet (and github is using a newer version).

pvmm:
> Do you mean a visual studio warning or an error?

Error. It seems that I used msbuild (building from terminal) before and some object files were reused. But if I rebuild after "clean solution" from the Visual Studio GUI, I get 11 errors. Maybe the parameters used in msbuild are not the same used when "build solution" is invoked, for instance a `--std=c++23` or equivalent may be missing when I build from the GUI.

> Could it be you're using a too old version of visual studio that doesn't recognize this yet (and github is using a newer version).

I've just updated to the latest version of VS and build tools (Aug 2025) and it still happens.

MBilderbeek:
As the code builds fine on GitHub Actions, there must be some local problem or a different way of building. Please check with the yml file to see how we build.
Because it builds fine on Actions, I believe no changes should be needed in the code. We use C++-23 indeed.

m9710797:
See this link:  https://godbolt.org/z/ffYj45MWP
It shows that "x86 msvc v19.43 VS 17.13" can compile this fine.
But "x86 msvc v19.42 VS 17.12" cannot.
So it really needs a recent visual studio version.

And it does seem this version is available on github.  I did not realize we required (almost) the latest available version.

pvmm:
I am closing this PR. Visual Studio seems erratic to me: sometimes it complains and sometimes it doesn't and I don't know why.

PR #1981, opened 2025-09-28
MSXPi: Rebased & resolved conflitcs - MSXPi Device (original PR: #1963)

costarc:
Latest changes in the openMSX master resulted in conflicts to my fork. 
I rebased manually, seems the original pull request was closed due to the way I rebased.

Anyway, this is the clean pull request to merge my MSXPi Device to openMSX master branch.
Refer to previous (closed) PR for discussion about impact and implementation decisions:~#
https://github.com/openMSX/openMSX/pull/1963

-- Post in the original PR: --

This change implements a new MSXPiDevice for an extension named MSXPi.

MSXPi is a hardware & software interface I developed and is hosted in https://github.com/costarc/MSXPi

The openMSX implementation of the interface allows mostly the same features without he need of a Raspberry Pi. The openMSX device interfaces with the same MSXPi server-side component used in the physical project, which is a Python program running in the local host.


MBilderbeek:
@costarc the extension that was added to openMSX expects a ROM with sha1sum c429077a5f58b006125247a67783ea4ea6fb7389 - where can someone who is interested in this extension get that exact ROM image? (Obviously it cannot be dumped from someone's original hardware ;-))

costarc:
> @costarc the extension that was added to openMSX expects a ROM with sha1sum c429077a5f58b006125247a67783ea4ea6fb7389 - where can someone who is interested in this extension get that exact ROM image? (Obviously it cannot be dumped from someone's original hardware ;-))

The latest MSXPi.xml and ROM can be download from the MSXPi reposritory. 
The xml is here   : https://github.com/costarc/MSXPi/tree/master/software/openMSX/share/extensions/
The ROM is here: https://github.com/costarc/MSXPi/tree/master/software/openMSX/share/systemroms/extensions/

Do you really need ROM with sha12sum c429077a5f58b006125247a67783ea4ea6fb7389 ? If so its here: https://github.com/costarc/MSXPi/blob/master/software/openMSX/share/systemroms/extensions/msxpidos.c429077a5f58.rom

MBilderbeek:
@costarc the problem is that the extension has that sha1sum. So, the extension doesn't work without having that specific ROM.  And people will complain that the extension isn't working because the ROM isn't available anywhere... 

Apparently you are still developing that ROM? What would be the best way to manage this? (Also @m9710797 any ideas?)

costarc:
> @costarc the problem is that the extension has that sha1sum. So, the extension doesn't work without having that specific ROM. And people will complain that the extension isn't working because the ROM isn't available anywhere...
> 
> Apparently you are still developing that ROM? What would be the best way to manage this? (Also @m9710797 any ideas?)

That specific rom is here: https://github.com/costarc/MSXPi/blob/master/software/openMSX/share/systemroms/extensions/msxpidos.c429077a5f58.rom

If you like, I can submit a PR to update both XML and ROM.

MBilderbeek:
Sure, but how often is it going to change and what would be a good strategy to cope with that?

costarc:
> Sure, but how often is it going to change and what would be a good strategy to cope with that?

Let's do this: I submit a PR to get openMSX updated with lthis atest MSXPi config and ROM.

Going forward, I’ll update the build process so that the ROM and config are only changed when I introduce a breaking change to the ROM. When that happens, I’ll submit a PR once the new ROM is stable.

Is that fine?

MBilderbeek:
How do you keep the released ROMs available? E.g. for older openMSX releases.

costarc:
I expected that the ROM was shipped along with openMSX, but I seems it is not. In such case, the next best thing is that  users follow the Contrib/README.msxpi steps:
3. Download msxpibios.rom from:
   https://github.com/costarc/MSXPi/tree/master/software/openMSX/share/systemroms/extensions
   
   Copy to you openMSX/share/systemroms/extensions/ folder.
   
4. Download the matching MSXPi.xml from:
   https://github.com/costarc/MSXPi/tree/master/software/openMSX/share/extensions
   
   copy to you openMSX/share/extensions/ folder.

Follwing these steps, users will always have the most up to date MSXPi config and rom.

MBilderbeek:
Yeah, but this isn't really ideal.... the idea is that you get openMSX, get the ROMs and have a working up-to-date package (depending on when it was released). 

We could ship the ROM if the source code license is GPL2 compatible and available. Then it's like C-BIOS.

costarc:
> Yeah, but this isn't really ideal.... the idea is that you get openMSX, get the ROMs and have a working up-to-date package (depending on when it was released).
> 
> We could ship the ROM if the source code license is GPL2 compatible and available. Then it's like C-BIOS.

I agree.
I made that specific ROM available in the MSXPi repository.
If the user follows the documentation steps I mentioned previously, they can just grab the msxpibios.rom from my repository and it will work with the current MSXPi.xml in the openMSX.

m9710797:
For savestates/replays (that were made when using a specific version of the ROM) it's important that that specific ROM remains available. (Otherwise that savestate/replay doesn't load anymore).

So I'd like to ask that you also keep older major versions of the ROM available for download.

Most of the hardware we emulate in openMSX is no longer under active development, so we don't often have to deal with extensions with a changing ROM.

costarc:
> For savestates/replays (that were made when using a specific version of the ROM) it's important that that specific ROM remains available. (Otherwise that savestate/replay doesn't load anymore).
> 
> So I'd like to ask that you also keep older major versions of the ROM available for download.
> 
> Most of the hardware we emulate in openMSX is no longer under active development, so we don't often have to deal with extensions with a changing ROM.

Understood, noted and I agree.
Will keep ROMs acessible and available for download when they are replaced by newer versions.

MBilderbeek:
@m9710797 so what do we do? Some options:
1. add a readme.md with the extension of where users can find the required ROM(s) for MSXPi
2. add the matching ROM to Contrib and install it for the user (this will be a maintenance burden, but I suppose @costarc could do that with PR's)
3. don't do anything
4. something else?

m9710797:
All seem good options. I don't have a real preference.
Maybe option 1?

costarc:
> All seem good options. I don't have a real preference. Maybe option 1?

Option 1 also.
Since MSXPi ROM is not (yet) a fully a static or stale ROM, user may want to keep up with newer releases and features, so this options gives them information on where to look out for that when they want.

MBilderbeek:
@costarc can you add that MSXPi.md file next to the file (and possibly mention it or a hint about the ROM availability in the XML meta data's description tag) with the proper content via a PR?

costarc:
@MBilderbeek 
I could update the MSXPi.xml "Description" field to: 
<description>MSXPi is an extension for openMSX that emulates the MSXPi hardware interface, enabling an MSX computer to interact with modern systems. Unlike the physical version, the emulated interface does not require a Raspberry Pi — it communicates with a Python-based server component that provides access to web resources, files, REST APIs, and other external services. Newer versions of the msxpibios.rom can be downloaded from the MSXPi repository at https://github.com/costarc/MSXPi/tree/master/software/openMSX/share/systemroms/extensions.</description>

About the MSXPi.md, not 100% what you would like to have ther. Would be something on this line?
MSXPi.md: "For the latest MSXPi BIOS updates, refer to the main repository at https://github.com/costarc/MSXPi. The direct download link for the MSXPi BIOS is https://github.com/costarc/MSXPi/tree/master/software/openMSX/share/systemroms/extensions."

Feel free to sugegst the text. 
thanks.

m9710797:
Here's a suggestion for the config file of the extension. We do allow to specify multiple values for the `<sha1>` tag of the ROM, and openMSX tries them in-order. So if there are multiple releases of the ROM I'd suggest to not replace the `sha1` value, but add another possible value. At least for as long as those ROMs remain compatible with the openMSX emulation. Also for you during development of the ROM it's annoying to always have to update the extension with the correct `sha1`, you can work around that by also specifying a `filename` tag. Then if openMSX doesn't find any ROM with matching sha1, it will use the rom with the corresponding filename (and it must be located next to the extension IIRC). This will print a warning (about no matching sha1 found), but you as a developer understand and can ignore the warning.

So in summary I'd suggest to use something like this in the extension:
```
<rom id=...>
    <sha1>123456...</sha1>  <!-- version 1.2, newest version first-->
    <sha1>abcdef...</sha1>  <!-- version 1.1-->
    <sha1>a1b2c3...</sha1>  <!-- version 1.0 -->
    <!-- <sha1>fedcba...</sha1>  version 0.5 no longer compatible -->
    <filename>my.rom</filename>
</rom>
```

MBilderbeek:
Ah good point, @m9710797! Let's do that as well, especially when you add a new ROM version to the public, @costarc . (Of course only if that ROM is still compatible with the extension - which @m9710797 also put in his example.)
Maybe you can even add a direct link to the ROM behind these sha1sums, if available?

costarc:
Sure can do that. Will prepare a PR when I get some proper time with this initial updates. Will also assure future updates follow this pattern.

costarc:
I submitted a PR.

Build is failing due to non related issue with VC:

[Build Snapshots / Windows-VC x64 (pull_request)]Download 3rd party packages for Windows VC build

PR #1985, opened 2025-10-03
tcl_bridge: allow guest machine to execute Tcl commands

pvmm:
This Tcl script also requires the tcl.com program available in https://github.com/pvmm/openmsx-tcl-bridge to work.

Tcl side of the mechanism to connect guest machine to OpenMSX through Tcl environment. This program uses a watchpoint to trigger execution of the Tcl code and it sends a response back from Tcl to the MSX.

image

### DIFF
--- a/share/scripts/_tcl_bridge.tcl
+++ b/share/scripts/_tcl_bridge.tcl
@@ -1,0 +1,131 @@
+# tcl_bridge v1.0 (c) 2025 Pedro de Medeiros
+# There is an example program that consumes this script at:
+# https://github.com/pvmm/openmsx-tcl-bridge
+#
+# To execute a Tcl command from the MSX side, you need to perform the following
+# steps:
+#
+# 1) Make sure the 'tcl_bridge' is active.
+#
+# To do this execute the command 'tcl_bridge start' in the console.
+# The `tcl_bridge' is *intentionally* not active by default.
+#
+# 2) The MSX program needs to setup a control block.
+#
+# That control block has 6 fields
+#
+#   cmd_addr:      +0 (output, 2 bytes): pointer to the command string (plain string, no terminator needed)
+#   cmd_len:       +2 (output, 2 bytes): length of the command string
+#   res_addr:      +4 (output, 2 bytes): pointer to the result buffer
+#   res_max_len:   +6 (output, 2 bytes): length of the result buffer
+#
+#   res_real_len:  +8 (input,  2 bytes): actual length of the result
+#   status:       +10 (input,  1 byte ): success (0) or failure (1)
+#
+# Only the first 4 fields need to be filled in by the MSX program, the last 2
+# fields are filled in by the 'tcl_bridge' (see below).
+#
+# 3) Next the MSX program needs to send the address (2 bytes) of this control
+# block to I/O port '6'. This will trigger the 'tcl_bridge'.
+#
+# 4) The 'tcl_bridge' then executes the given Tcl command (located at 'cmd_addr').
+#
+# If the command was executed successful, a '0' is written to the 'status' field
+# in the control block'. If there was an error a '1' is written.
+#
+# The result string (successful result or the error message) is written into the
+# result buffer located at 'res_addr'.
+#
+# The 'tcl_bridge' only writes at most 'res_max_len' bytes into that buffer. So
+# if the result buffer is too small the result gets truncated. The result is
+# written WITHOUT terminator characters. The full (non-truncated) length is
+# written to the 'res_real_len' field of the control block.
+#
+# 5) The MSX program should then check the result.
+#
+# It should check success / failure by examining the 'status' field. Possibly
+# check for overflow by comparing the actual length (field 'res_real_len') with
+# the given buffer length (field 'res_max_len').
+#
+# If the MSX want to detect whether the 'tcl_bridge' is active, it can e.g.
+# write 0xff to the 'status' field and then check if it was changed into '0'
+# or '1'.
+#
+namespace eval tcl_bridge {
+
+variable wp {}
+variable latch 0
+variable address
+
+set_help_text tcl_bridge \
+{tcl_bridge -- allows guest/host communication through a MSX-to-Tcl bridge.
+
+Usage:
+  tcl_bridge start      Create communication channel
+  tcl_bridge stop       Destroy communication channel
+}
+
+proc tcl_bridge {cmd} {
+    switch -- $cmd {
+        start   { tcl_bridge::start }
+        stop    { tcl_bridge::stop }
+        default { error "Unknown sub-command \"$cmd\"." }
+    }
+}
+
+proc start {} {
+    variable wp
+    if {$wp eq {}} {
+        set wp [debug watchpoint create -type write_io -address 6 -command {tcl_bridge::read_command $::wp_last_value}]
+    }
+}
+
+proc stop {} {
+    variable wp
+    if {$wp ne {}} {
+        debug watchpoint remove $wp
+        set wp {}
+    }
+}
+
+proc min {x y} {
+    expr {$x < $y ? $x : $y}
+}
+
+proc execute_command {base} {
+    set cmd_addr     [peek16 $base]
+    set cmd_len      [peek16 [expr {$base + 2}]]
+    set res_addr     [peek16 [expr {$base + 4}]]
+    set res_max_len  [peek16 [expr {$base + 6}]]
+    set res_real_len [expr {$base + 8}]
+    set status       [expr {$base + 10}]
+    set cmd [debug read_block memory $cmd_addr $cmd_len]
+    # execute command
+    set response [catch {set output [eval $cmd]} output]
+    # write command result and output data
+    debug write_block memory $res_real_len [binary format s [min [string length $output] 0xFFFF]]
+    debug write_block memory $res_addr [string range $output 0 [min [string length $output] $res_max_len]-1]]
+    debug write memory $status $response
+}
+
+proc read_command {byte} {
+    variable latch
+    variable address
+    switch -- $latch {
+        0 {
+            set address $byte
+            set latch 1
+        }
+        1 {
+            incr address [expr {$byte << 8}]
+            set latch 0
+            execute_command $address
+        }
+    }
+}
+
+namespace export tcl_bridge
+
+} ;# namespace tcl_bridge
+
+namespace import tcl_bridge::*
--- a/share/scripts/lazy.tcl
+++ b/share/scripts/lazy.tcl
@@ -82,3 +82,4 @@ register_lazy "_vdp_access_test.tcl" toggle_vdp_access_test
 register_lazy "_vdp_busy.tcl" toggle_vdp_busy
 register_lazy "_vdrive.tcl" vdrive
 register_lazy "_vgmrecorder.tcl" {vgm_rec vgm_rec_next vgm_rec_end}
+register_lazy "_tcl_bridge.tcl" tcl_bridge


share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,113 @@
...
+        1 {
+            incr address [expr $byte << 8]
+            incr latch -1
+            read_string $address
+        }
+    }
+}
+
+proc write_command {port} {
+    if {$port ne 7} { return }
```
m9710797: Change this to `{$port != 7}`.
* `eq` and `ne` are for string comparisons.
* `==` and `!=` are for numerical comparisons.

Often both give the same result. But a counter example is `"0x7" == "7"` while `"0x7" ne "7"`.

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,113 @@
...
+}
+
+proc read_command {port byte} {
+    if {$port ne 6} { return }
+    variable latch
+    variable address
+    switch -- $latch {
+        0 {
+            set address $byte
+            incr latch
```
m9710797: Let's simplify this to `set latch 1`. And the line below to `set latch 0`.
Ah, or maybe you think in a later revision you'll need more than 2 stages? Then this is a good approach (if you can factor out the `incr latch` statement).

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,113 @@
...
+    }
+}
+
+proc read_string {base} {
+    set addr $base
+    set s ""
+    for {set c {}} {[set c [peek $addr]] ne 0} { incr addr } {
+        append s [format %c $c]
+        # detect overflows and runaways
+        if {$addr < $base || $addr > $base + 256} { break }
```
m9710797: This does always peek 256 bytes, right?
Maybe stop at the first `0` character? (And for safety still refuse to read more than 256 chars).

m9710797 (commented):
I'm sorry. I didn't have time yet for a detailed review. I'll try to do that later today.
My biggest "problem" with this PR is the missing return path to the MSX.

share/scripts/tcl_bridge.tcl
pvmm: But `[set c [peek $addr]] != 0` in the `for` above already stops at the first `0` character.

pvmm:
I removed the dependency of Programmable Device.

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,102 @@
+# tcl_bridge v1.0 (c) 2025 Pedro de Medeiros
+namespace eval tcl_bridge {
+
+variable wp {}
+variable latch 0
+variable address
+variable result
```
m9710797: `variable result` is no longer needed?

pvmm:
It also sends the message back to tcl.com app.

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,102 @@
...
+
+set_help_text tcl_bridge \
+{tcl_bridge -- allows guest/host communication through a MSX-DOS-to-Tcl bridge.
```
m9710797: Detail: the Tcl-bridge is not limited to MSX-DOS, right?
But you can of course mention your MSX-DOS tool as an example.

It would also be good to document the exact protocol that's used. Just in comments in this script is fine.
But we can do that after we've finished all other stuff in the script.

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,102 @@
...
+  tcl_bridge stop       Destroy communication channel
+}
+
+proc tcl_bridge {args} {
+    if {[catch {set result [tcl_bridge::dispatcher {*}$args]} msg]} {
+        puts stderr $::errorInfo
+        error $msg
+    }
+    return $result
+}
```
m9710797: The (only?) purpose of this proc is to catch a possible error, then print it on stderr and reissue the same error, right?

Usually we don't do such things. The Tcl error itself is sufficient, by default it already gets printed in the console (so now it'll get printed twice I think). Also when you call this script from some other script it's nicer if errors can't get printed automatically.

So I would drop this proc and then rename `dispatcher` to `tcl_bridge`.

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,102 @@
...
+
+proc dispatcher {args} {
+    set params "[lrange $args 1 end]"
```
m9710797: I'm not sure, but are these extra quotes (`""`) around the `lrange` command needed? Maybe it's even wrong?

But neither the `start` nor the `stop` sub-command takes extra parameters, so maybe we shouldn't bother with this at all?

Maybe for now it's sufficient if this proc takes a single `cmd` parameter (not an `args` parameter list ; a parameter with the name `args` is treated special in Tcl)

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,102 @@
...
+    set cmd [lindex $args 0]
+    switch -- $cmd {
+        start   { return [tcl_bridge::start {*}$params] }
+        stop    { return [tcl_bridge::stop  {*}$params] }
+        default { error "Unknown command \"[lindex $args 0]\"." }
```
m9710797: Detail: maybe it's clearer to say `Unknown sub-command $cmd`.
* **Sub**command instead of command.
* Reuse `$cmd` instead of recalculating `[lindex $args 0]`.

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,102 @@
...
+    }
+}
+
+proc start {} {
+    variable wp
+    if {$wp eq {}} {
+        puts "tcl_bridge v1.0 (c) 2025 Pedro de Medeiros."
```
m9710797: Usually the openMSX scripts don't print copyright stuff on usage. But it is perfectly fine to have it in comments in this script itself. Like you already have on line 1.

If you insist on printing this, then maybe we can move your script from the `share/scripts` to the `Contrib` directory.

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,102 @@
...
+        set wp [debug watchpoint create -type write_io -address 6 -command {tcl_bridge::read_command $::wp_last_value}]
+        set init 1
```
m9710797: This `init` stuff is no longer needed?

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,102 @@
...
+        puts "tcl_bridge initialized"
+    } else {
+        puts "tcl_bridge already running"
```
m9710797: Personally I'm not a big fan of prints like these:
* Instead of `puts`, `return` the string. Then if you call this command from within another script, that script can examine the result (and then choose itself to print it or not).
* Is it an error to call `tcl_bridge start` when it was already started? Currently it's not and I think that's good. But is it then useful to return a different message? Why not always return `tcl_bridge initialized`. (I mean for a user, it is really useful to distinguish between both cases?)
* But then you can also say: if this sub-command always returns `tcl_bridge initialized`, what's the value of that message. Isn't it cleaner to not return anything?

So my personal preference is not return anything (technically: return an empty string). But not all our scripts follow that convention (different people have different preferences).
For example most unix shell commands also don't return anything on success. E.g. `rm myfile.txt` returns nothing on success, only when something went wrong it shows extra info.

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,102 @@
...
+}
+
+proc stop {} {
+    variable wp
+    if {$wp ne {}} {
+        debug watchpoint remove $wp
+        set wp {}
+        puts "tcl_bridge stopped"
+    } else {
+        puts "tcl_bridge not running"
```
m9710797: Same remark here. Is this message really useful?
Personally also here I wouldn't return anything (in both cases).

In other words:
* I don't consider it an error to stop the tcl_bridge if it wasn't running (never started or was already stopped).
* I don't think it's useful to know if this was the command that stopped tcl_bridge or that this command did nothing.

But again, my personal opinion, other people have different opinion, so not all openMSX scripts follow this convention.

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,102 @@
...
+    }
+}
+
+proc read_data {base} {
```
m9710797: Maybe a better name for this proc is now `execute_command`?

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,102 @@
...
+    set iaddr [peek16 $base]            ;# input address
+    set oaddr [peek16 [incr base 2]]    ;# output address
+    set osize [peek16 [incr base 2]]    ;# output size
+    set s ""
+    for {set addr $iaddr} {[set c [peek $addr]] != 0} {incr addr} {
+        append s [format %c $c]
+        # detect overflows and runaways
+        if {$addr < $iaddr || $addr > $iaddr + 256} { break }
+    }
```
m9710797: During a previous review I misunderstood this loop, I failed to notice that you _did_ check for zero-characters in the loop.

I think it would be more natural to use a counting-for-loop and have the check for 0-char as a break inside the loop.

BTW you have a check for `$addr < $iaddr`, but that can only trigger when `addr` overflows, right? But that won't ever happen Tcl integers are arbitrary precision (e.g. they don't wrap at 16-bit).

```
for {set i 0} {$i < 256} {incr i} {
    set c [peek [expr {$iaddr + $i}]]
    if {$c == 0} break
    append s [format %c $c]
}
```

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,102 @@
...
+    if {$c == 0} {
+        # execute command
+        if {[catch {eval $s} msg]} {
+            puts "Caught error: $msg"
```
m9710797: These `puts` statements will trigger by an action initiated by the MSX. I don't think `puts` is a good idea here. Instead we should return this `msg` to the MSX side (and you already partially do).

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,102 @@
...
+        } else {
+            puts OK
+            set msg "OK"
+        }
+    } else {
+        set msg "Parse error"
+    }
+    debug write_block memory $oaddr [concat [string range $msg 0 [expr {$osize - 2}]]$]
```
m9710797: I do have some questions about this:
* The `concat` statement appends a `$` at the end? But what if the string I want to return contains embedded `$` characters? Probably it's better to add a zero-terminator?
* But also what if I return a binary string, that can have embedded zero-characters.
* Maybe it's best to explicitly return the length of the returned string?
* E.g. we could overwrite the 16-bit value at `$base + 4` with the length? (And for convenience, and if there's space left in the buffer, we could also add a zero-terminator, or even fill the full remaining buffer with zeros)

What if the user provided a buffer that's too small to hold the result?
* For sure we should not write outside the provided user buffer.
* But we should also indicated that the buffer was too small.
* Maybe by writing the full length that the result would have to `$base + 4`?
* Detail, in principle the result could be larger than 0xffff characters, that would not even fit in the 16-bit value at `$base + 4`, but it should be fine to clip this value at 0xffff.

I'm also not sure why you have `$osize - 2` was that to account for the `$` terminating character? But then why `-2` and not `-1`? But then you run into problems when the MSX passed `osize` equal to 0 (or 1).

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,102 @@
...
+    set osize [peek16 [incr base 2]]    ;# output size
+    set s ""
+    for {set addr $iaddr} {[set c [peek $addr]] != 0} {incr addr} {
+        append s [format %c $c]
+        # detect overflows and runaways
+        if {$addr < $iaddr || $addr > $iaddr + 256} { break }
+    }
+    if {$c == 0} {
+        # execute command
+        if {[catch {eval $s} msg]} {
```
m9710797: What if the MSX sends a command like `machine_info time`. This is a query, it doesn't do anything, it just returns a result. But currently you ignore the (successful) result from `eval $s`. Instead you should also return this result string to the MSX (just like you now already return the error string).

So instead of returning `OK` you should return the actual result string from the command.

And then of course you should also indicate whether the returned string is for success or for failure.
* That could be via the first byte in the result string (e.g. `0` is "success", `1` is "error during command execution",  `2` is "error during parsing the command", ...).
* Or maybe more logical is to store this byte at location `$base + 6`?

m9710797 (commented):
Several remarks. Most are smaller details. But also a few fundamental questions about the communication protocol itself.

share/scripts/tcl_bridge.tcl
pvmm: > I'm also not sure why you have `$osize - 2` was that to account for the `$` terminating character? But then why `-2` and not `-1`? But then you run into problems when the MSX passed `osize` equal to 0 (or 1).

Because `string range` expects an index as a second parameter, not a size (`[string range var 0 0]` for instance prints a single character). If osize is 0 or 1, nothing of the result is supposed to be written if the buffer is nul-terminated (or `$`-terminated in this case).

MSX-DOS' BDOS function that prints strings expect them to end with `$`, so I moved it here because it was only a MSX-DOS tool. I didn't intend to use it to send or receive binary data, just report an operation result.

m9710797:
> MSX-DOS' BDOS function that prints strings expect them to end with `$`, so I moved it here because it was only a MSX-DOS tool. I didn't intend to use it to send or receive binary data, just report an operation result.

_IF_ we add this Tcl script to openMSX than IMHO it should be fully generic. So not tied to a specific MSX-DOS tool, but usable by any MSX program that wants this functionality.

I think the requirements are the following (maybe this list is incomplete, please add if you see more):
* It must be disabled by default (because it is a security risk, you don't want to allow the MSX to send arbitrary commands without permission (e.g. commands like `exec rm -rf /`)). Or by default we don't want to give the MSX program the possibility to discover it's running in an emulator. This is sufficiently covered by the explicit `tcl_bridge start` command.
* (Once it has permission) the MSX should be able to execute arbitrary commands, so no limitations of how the command should be formatted (e.g. no "regular" symbol as a terminator character), and maybe no limit (of 256 chars) on the command length?
* The MSX should get feedback on whether the command was successful or whether it triggered an error.
* The MSX should have access to either the error message or the result returned by the command.
* There should be no restriction on the content of the returned result. E.g. it should be possible to execute a command like `debug read_block "VDP regs" 0 64` (to allow the MSX to read VDP registers, normally this isn't possible). This returns a block of 64 bytes, thus containing arbitrary characters (including 0x00 and '$'). Thus using any terminator character won't work.
* In principle the result length of a command can be arbitrary long, but the Tcl script should not overwrite arbitrary MSX memory. The MSX should have control over the maximum buffer size that will get written. The MSX should also be informed over the exact result size and whether or not the given buffer size was sufficient.

In the current version of the script, you start a Tcl command via writing a pointer (2 bytes) to I/O port 6. This points to a control block of 6 bytes (ptr-to-input, ptr-to-output, size-of-output). That's a nice idea, but I'd like to propose this change.

* We still start a command by writing a pointer to some I/O port. But I propose the control block is 11 bytes.
  * offset +0 (input-field, 2 bytes): Pointer to the input buffer.
  * offset +2 (input-field, 2 bytes): Length of the input.
  * offset +4 (input-field, 2 bytes): Pointer to the output buffer.
  * offset +6 (input-field, 2 bytes): Maximum length of the output.
  * offset +8 (output-field, 2 bytes): The actual length of the output.
  * offset +10 (output-field, 1 byte): 0 for success, 1 for error
* This means there's no restriction on the input, a command can be larger than 256 characters (rare but sometimes useful, e.g. to define some Tcl proc). There's also no restriction on the set of input characters (e.g .you can pass arbitrary binary data from MSX to tcl_bridge). It's no longer possible that parsing the input encounters a parse error.
* The MSX has full control over the output buffer (the Tcl script will not write outside this buffer). There are no terminator characters written, so a buffer of 64 bytes can contain 64 useful result bytes.
* The actual length of the result is written in the control block at offset `+8`. This length can be larger than the given maximum length (but that means only a partial result is written, we never write outside the output buffer).
* At offset +10 the script writes a single byte indicating the result buffer contains either the normal result, or the error message.
* Notice the symmetry between input and output buffer. Both use a pointer + size. Symmetry is always nice, makes the API easier to understand.
* Neither input nor output uses terminator characters. So no off-by-one-errors possible (e.g. buffer not large enough to hold the terminator character).

Alternatives:
* I had considered merging the max output length and actual output length fields. So the script would overwrite the output-length-field. I rejected that idea, because then it's more work for simple MSX programs that want to reissue the same command multiple times and/or reusing the same output buffer for multiple commands. It's nicer if there's a strict separation between input and output fields in the control block, then the MSX does not always have to re-setup the output-size-field.
* I had considered merging the success/error-field with the output length field. For example use bit 15 in the result-length field. But I think this will complicate using the data (better to make the control block one byte longer and have a simple MSX program that's many bytes shorter).

Is this API easy to use from the MSX side? Personally I think it is, but there are many different use cases, and I can't think of everything. So feedback welcome:
* If you want to execute a command, and are not interested in the result you can set the max-output-size field to 0 (the output-ptr field then doesn't matter). You can check success/error at offset `+10`, and ignore the result length at offset `+8` (or only look at it if there was an error).
* Input length instead of some termination character. If the MSX program want to issue fixed commands then this won't be an issue, it's easy enough to let the assembler "calculate" the length of the command (end_label - start_label). If the MSX program takes input from the user (e.g. a MSX-DOS command line tool) then the length must be calculated at run-time, but that's the same difficulty as making sure the input is zero-terminated.
* Output length instead of some kind of terminator. If the MSX program really wants a terminator then adding one is very simple. But usually having a length is equally efficient/easy (and more flexible, e.g. the `$` terminator used by MSX-DOS is a really dumb idea).

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,117 @@
...
+# https://github.com/pvmm/openmsx-tcl-bridge
+#
+# TODO: add connection structure description here
+#
+namespace eval tcl_bridge {
+
+variable wp {}
+variable latch 0
+variable address
+variable echo 0
```
m9710797: Do you think this echo feature is really useful for the end-user?

I can understand that now it's useful for you while still developing the script.
I users want this echo, then they can opt-in by doing an explicit `puts` in the command they execute.

So personally I'd drop this echo feature.

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,117 @@
...
+    if {$n2 > $n1} { return $n1 }
+    return $n2
+}
+
+proc execute_command {base} {
+    set input_addr  [peek16 $base]                      ;# input address
+    set input_size  [peek16 [expr {$base + 2}]]         ;# input size value
+    set output_addr [peek16 [expr {$base + 4}]]         ;# output address
+    set output_size_addr    [expr {$base + 6}]          ;# output size address
+    set output_max  [peek16 [expr {$base + 8}]]         ;# max output buffer size
```
m9710797: I would swap the order of `output_size_addr` and `output_max` (so keep the order as I proposed myself). The reasons are:
* Then you have two pairs of `ptr`+`size` next to each other, the first pair for the input buffer, the second pair for the output buffer.
* Then all input fields and all output fields in the control block are together (with your order input and output are interleaved).

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,117 @@
...
+            set status 1
+        } else {
+            set status 0
+        }
+    } else {
+        # overflow
+        set status 2
+        set output {}
+    }
+    debug write_block memory $output_size_addr [binary format s [string length $output]]
```
m9710797: Maybe I forgot to mention this, but in theory the length of `output` is not limited to 16-bit. In case it is larger we should report "overflow" to the MSX. And we can achieve that by using 0xffff as the required output size (because the MSX can anyway never specify such a large max output size, thus 0xffff will always be larger than this max size).

So we should write `[min 0xffff [string length $output]]`.

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,117 @@
...
+    # write output and fill remaining buffer with nul
+    set data [string range $output 0 [min [string length $output] [expr {$output_max - 1}]]]
```
m9710797: Why `$output_max - 1`? That's wrong.

One of the features of my proposal was that the full buffer can be used for useful bytes. And then you never have to deal with possible off-by-one mistakes because you forgot to reserve space for a terminator etc.

The MSX program should examine the output-size-field to detect how long the output is. It should not rely on having a zero-terminator. (But if a terminator character is useful, then the MSX program can trivially add it itself).

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,117 @@
...
+    debug write_block memory $output_addr [concat $data[string repeat \0 [expr {$output_max - [string length $data]}]]]
```
m9710797: We could indeed fill the unused part of the output buffer with zeros. And I think I even suggested that myself.
I'm not against this, but the MSX program cannot depend on this (because there might not be room for even a single zero character). And then maybe it's better to not even try to zero-fill. That way bugs(*1) in MSX programs will be detected sooner.

(*1) I mean the bug that the MSX program DOES depend on a zero-terminator.

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,117 @@
...
+            append s [format %c $c]
+        }
+    }
+    if {$c ne {}} {
+        # execute command
+        if {[catch {set output [eval $s]} output]} {
+            set status 1
+        } else {
+            set status 0
+        }
```
m9710797: Detail: this can be simplified to:
```
set status [catch {set output [eval $s]} output]
```

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,117 @@
...
+proc execute_command {base} {
+    set input_addr  [peek16 $base]                      ;# input address
+    set input_size  [peek16 [expr {$base + 2}]]         ;# input size value
+    set output_addr [peek16 [expr {$base + 4}]]         ;# output address
+    set output_size_addr    [expr {$base + 6}]          ;# output size address
+    set output_max  [peek16 [expr {$base + 8}]]         ;# max output buffer size
+    set status_addr         [expr {$base + 10}]         ;# status address
+    set c {}
+    set s ""
+    if {[expr {$input_addr + $input_size}] < 0xFFFF} {
```
m9710797: I don't think it's useful to check this condition (check that the input buffer ends before 0x10000 (you instead check for 0xffff, that might be a bug)). The reason why I think that's not useful is:
* This error would anyway be detected below in the `peek` command, it will produce an error that 0x10000 is outside the valid range.
* There can be a similar problem with the output buffer, and you don't check for that (and I indeed think you should not explicitly check that .. the `debug write` command below will already detect an out-of-range write).
* And then there are various extra conditions that the control block should not overlap with the input or output buffer, etc. We also don't check those (and I don't think we should).

Above I said that the `peek` and `debug write` commands will already detect out-of-bounds access. If that happens we get a Tcl error during execution of the `execute_command` proc. Then that proc then doesn't finish and then we don't produce proper output for the MSX. So if a MSX program _really_ wants to detect this, then it could set offset `+10` in the control block to `2` and detect that it was not overwritten with either `0` or `1`. But I would advice against that. Such a thing is always a bug in the MSX program itself (not in openmsx, not in this script, not in the command that was given, really in the MSX program itself), and in general (at least in MSX programs) we don't often add extra code to try to detect bugs in the program itself.

We should only take care that openMSX nor this Tcl script crashes with whatever (invalid) input the MSX sends. And I think we can do that with a single tcl `catch` statement (e.g. on line 107 around the call to `execute_command` ... but during development of this script it's easier to not yet add this).

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,117 @@
...
+        for {set i 0} {$i < $input_size} {incr i} {
+            set c [peek [expr {$input_addr + $i}]]
+            append s [format %c $c]
+        }
```
m9710797: I haven't tried myself yet, but can't this loop be replaced with a single `debug read_block` command?

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,117 @@
...
+        # execute command
+        if {[catch {set output [eval $s]} output]} {
+            set status 1
+        } else {
+            set status 0
+        }
+    } else {
+        # overflow
+        set status 2
+        set output {}
```
m9710797: As I (tried to) explain above, we shouldn't manually check for overflow (because the `debug read` or `peek`) command already does this. So this check for `$c ne {}` and this `else` block isn't needed.

m9710797 (changes requested):
Thanks for the update. Already looks a lot better, but still some important remarks.

m9710797:
I now also read the MSX code `tcl.c` on `https://github.com/pvmm/openmsx-tcl-bridge`.
And indeed an important case is "the tcl_bridge script isn't active". The MSX can reliably detect this by setting offset `+10` in the control block to e.g. `2` and then detect that it got changed to `0` or `1`. So the same detection mechanism as for detecting invalid control blocks (see above).

share/scripts/tcl_bridge.tcl
```diff
@@ -0,0 +1,92 @@
...
+    set input_size  [peek16 [expr {$base + 2}]]         ;# input size value
+    set output_addr [peek16 [expr {$base + 4}]]         ;# output address
+    set output_max  [peek16 [expr {$base + 6}]]         ;# max output buffer size
+    set output_size_addr    [expr {$base + 8}]          ;# output size address
+    set status_addr         [expr {$base + 10}]         ;# status address
+    set s [debug read_block memory $input_addr $input_size]
+    # execute command
+    set status [catch {set output [eval $s]} output]
+    # write command result and output data
+    debug write_block memory $output_size_addr [binary format s [expr {min([string length $output], $output_max)}]]
```
m9710797: Should this be
* `min([string length $output], $output_max)` or
* `min([string length $output], 0xffff)`

Only with the latter can the MSX detect that the given output buffer was too small (and know how big it should have been). So therefor I proposed the latter in my previous posts. But now I do see that:
* This requires an extra (16-bit) comparison on the MSX side.
* The information "how big the output buffer should have been" is maybe not that useful for the MSX?

In any case it's required that we somehow communicate to the MSX that the result was truncated (the output buffer was too small). So unless we find something better I'd still suggest the latter expression (the former is not OK IMHO).

m9710797 (changes requested):
Hi. Only a very quick review (I'm at work now).
I think it's close to ready. Only one main issue remaining.

The documentation on the protocol itself should be improved, but we can do that once the script is finished.

I'll try to make time this evening (or tomorrow) to do a full review and do some experiments myself.

share/scripts/tcl_bridge.tcl
pvmm: Isn't this part of the "adding extra code to detect errors" from MSX programs that we are not supposed to do?

m9710797: Not really, or at least not always.

Some commands return a string of unpredictable length. So the MSX can setup a fully correct control block, but nevertheless the output buffer may not be large enough to hold the result. In other words: no bug in the MSX program, the output-buffer-too-small condition is outside the control of the MSX program.

Maybe one example(*1) is: query the host filename of the rom that's located in the slot this program is executing from. That could be a "simple" path like `/home/wouter/msx/roms/myGame.rom`, but it could also be a _much_ longer path.
(*1) I know this is an artificial example, but I'm sure there are better more realistic examples as well.

Maybe an idea could be to extend the status code?
* `0`: OK
* `1`: command was successful, but output buffer too small (result truncated)
* `128`: command generated an error
* `129`: command generated an error, and result buffer too small to hold the full error message.

Another way to look at this:
* bit 7: 0 is command executed OK, 1 is command generated an error
* bit 0: 0 is buffer OK, 1 is output buffer too small

pvmm: I think comparing offset+08 with offset+06 should be enough?

pvmm: Namely, if offset+08 is larger than offset+06, it would overflow.

m9710797: Yes you're right. But let me clarify anyway to be very clear.

My first idea was to fill in the un-truncated length at offset `+8`. Then the MSX can indeed compare the values at offset `+6` and `+8` to detect overflow. And if needed it can place a terminator character at the smallest of these two. So with this version we also know how large the buffer should have been.

If I were to design the protocol for a _modern_ system I would go for this: 'comparison' and 'minimum' is trivial to calculate. Though, on Z80, comparing 16-bit values is a little bit more work, and it's easy to forget to take the minimum of both values when placing the terminator character (e.g. see your `tcl.com` application).

So, after reading your earlier remarks, I came up with this alternative version. Here the value at offset `+8` will never be larger than the one at offset `+6` (it is already the minimum of the values above). And then instead we return an extra bit via the status byte.

This alternative version is strictly less powerful: it does NOT communicate how large the buffer should have been. But it might be easier to use via Z80 code (no need for 16-bit comparison / minimum). And the info "how large should the buffer have been" is maybe not that useful. I mean for an MSX program, for a modern system is it often very useful to re-allocate a larger buffer and retry (but dynamically allocated buffers are not often used in MSX code).

So either solution is fine for me, pick which you like best:
* The more general solution where we return the un-truncated length, and the Z80 has to do more work.
* Or we return the truncated length plus an extra status bit, which might be easier to use for Z80 programs.

pvmm: This reminds me: since you want to use tcl_bridge not just with tcl.com and you also want it to be capable of sending binary data, do you see some other use for this that I don't?

I thought about this extension because `dirAsDisk` is too small to hold a complete compiler environment (Turbo Pascal or HITECH-C) with code and object files. So I could use a hard disk image instead and tcl_build would automatically move things to and from it without the hassle of doing it manually after you run a build.bat script from MSX-DOS. So it was conceived to be used with `diskmanipulator` (I still intend to create a dm.com after `diskmanipulator` that mimics scp and copy files to and from the virtual machine), but I saw no reason to limit it to that function and now tcl_bride accepts any Tcl command.

Since it was conceived as tooling, I don't see a reason for it to be extra efficient. The more general solution is good enough for me.

pvmm: If we go with the general solution, it's ready.

m9710797: I don't have a specific use case in mind. But I always try to make sure that whenever we add a new feature we try to do it in a general way, so not limited to the single use case that triggered the request.

I also think this is ready now. After dinner I'll do a final review and also actually test it. And if that doesn't find anything then I'll merge it. Thanks!

share/scripts/_tcl_bridge.tcl
```diff
@@ -0,0 +1,93 @@
...
+    set output_addr [peek16 [expr {$base + 4}]]         ;# output address
+    set output_max  [peek16 [expr {$base + 6}]]         ;# max output buffer size
+    set output_size_addr    [expr {$base + 8}]          ;# output size address
+    set status_addr         [expr {$base + 10}]         ;# status address
+    set s [debug read_block memory $input_addr $input_size]
+    # execute command
+    set status [catch {set output [eval $s]} output]
+    # write command result and output data
+    debug write_block memory $output_size_addr [binary format s [expr {min([string length $output], 0xFFFF)}]]
+    debug write_block memory $output_addr [string range $output 0 [expr {min([string length $output], $output_max) - 1}]]
```
m9710797: The script doesn't work for me. It produce this error:
```
invalid command name "tcl::mathfunc::min"
```

It's caused by the use of `min()` inside the above expressions. Actually this should have worked: `expr {min(a, b)}` is valid Tcl syntax. However this relies on a **full** Tcl installation. More in detail this internally calls the Tcl library proc `tcl::mathfunc::min`, but unfortunately openMSX doesn't ship with the full Tcl library (it only has the core builtin Tcl functionality). Strange that it did work for you??

But luckily it's easy to fix. I added the following `min` helper proc, and then changed these two lines into:
```
proc min {x y} {
	expr {$x < $y ? $x : $y}
}
...
    debug write_block memory $output_size_addr [binary format s [min [string length $output] 0xFFFF]]
    debug write_block memory $output_addr [string range $output 0 [min [string length $output] $output_max]-1]]
```
After this fix the script worked fine for me (including various corner cases that I tested).

share/scripts/_tcl_bridge.tcl
```diff
@@ -0,0 +1,93 @@
...
+#    char*    input;            // pointer to the received command: offset + 0
+#    uint16_t input_size;       // command length: offset + 2
+#    char*    output;           // pointer to store result of command: offset + 4
+#                               // tcl_bridge will not write more than output_max bytes.
+#    uint16_t output_max;       // max size of place to store result of command: offset + 6
+#    uint16_t output_size;      // actual length of result of command: offset + 8
+#                               // if output_size > output_max, output buffer was too small.
+#    int8_t   status;           // operation status of command when executed: offset + 10
+#                               // where 0 is OK and 1 is ERROR.
+# } tcl_data;
```
m9710797: I think the script is ready. Now it's time to improve the documentation. Here's a draft proposal (please review carefully, check for spelling/grammar errors, reformulate if needed, ...):
```
To execute a Tcl command from the MSX side, you need to perform the following steps:

1) Make sure the 'tcl_bridge' is active.
To do this execute the command 'tcl_bridge start' in the console.
The `tcl_bridge' is *intentionally* not active by default.

2) The MSX program needs to setup a control block.
That control block has 6 fields

  cmd_addr:      +0 (output, 2 bytes): pointer to the command string (plain string, no terminator needed)
  cmd_len:       +2 (output, 2 bytes): length of the command string
  res_addr:      +4 (output, 2 bytes): pointer to the result buffer
  res_max_len:   +6 (output, 2 bytes): length of the result buffer

  res_real_len:  +8 (input,  2 bytes): actual length of the result
  status:       +10 (input,  1 byte ): success (0) or failure (1)

Only the first 4 fields need to be filled in by the MSX program, the last 2 fields are filled in by the 'tcl_bridge' (see below).

3) Next the MSX program needs to send the address (2 bytes) of this control block to I/O port '6'.
This will trigger the 'tcl_bridge'.

4) The 'tcl_bridge' then executes the given Tcl command (located at 'cmd_addr').
If the command was executed successful, a '0' is written to the 'status' field in the control block'. If there was an error a '1' is written.
The result string (successful result or the error message) is written into the result buffer located at 'res_addr'.
The 'tcl_bridge' only writes at most 'res_max_len' bytes into that buffer. So if the result buffer is too small the result gets truncated. The result is written WITHOUT terminator characters.
The full (non-truncated) length is written to the 'res_real_len' field of the control block.

5) The MSX program should then
Check success / failure by examining the 'status' field.
Possibly check for overflow by comparing the actual length (field 'res_real_len') with the given buffer length (field 'res_max_len').
If the MSX want to detect whether the 'tcl_bridge' is active, it can e.g. write 0xff to the 'status' field and then check if it was changed into '0' or '1'.
```

share/scripts/_tcl_bridge.tcl
```diff
@@ -0,0 +1,93 @@
+# tcl_bridge v1.0 (c) 2025 Pedro de Medeiros
```
m9710797, line 1: To make this an _official_ openMSX Tcl script we should:
* Rename the script from `tcl_bridge.tcl` to `_tcl_bridge.tcl`.
* Register the script in `lazy.tcl`.

To improve openMSX startup time, we don't automatically execute tcl scripts in `.openMSX/share/scripts` that start with an `_` character. Instead we only do that on-demand (lazy) and to do that we need a little extra information that's provided via `lazy.tcl`.

m9710797 (commented):
I've made one fix (see review comments). After that it worked fine for me.
I'm happy with the functionality.
I think we can merge this after you've processed these last(?) review comments.

pvmm (commented):
I think it's ready now.

m9710797:
Merged. Thanks a lot for your effort.

PR #1987, opened 2025-10-04
Added the ability to specify variation indexes for .ttc font files.

Suzumizaki:
Added the ability to specify the variation index for TrueType Collection (.ttc) font files.

Four major improvements have been made:
+ Added fontPropIndex and fontMonoIndex to ImGuiManager
+ Added settings: gui_font_default_index, gui_font_mono_index
+ Updated the related GUI menu (Settings->GUI->Select Font)
+ Added an option for the osd create/info/configure commands: -fontfaceindex

### DIFF
--- a/src/console/OSDText.cc
+++ b/src/console/OSDText.cc
@@ -44,12 +44,19 @@ void OSDText::setProperty(
 		std::string val(value.getString());
 		if (fontFile != val) {
 			if (std::string file = systemFileContext().resolve(val);
-			    !FileOperations::isRegularFile(file)) {
+				!FileOperations::isRegularFile(file)) {
 				throw CommandException("Not a valid font file: ", val);
 			}
 			fontFile = val;
 			invalidateRecursive();
 		}
+	} else if (propName == "-fontfaceindex") {
+		int val = value.getInt(interp);
+		if (val < 0) {
+			throw CommandException("Not a valid value for -fontfaceindex, "
+					"should be 0 or greater integer: ", val);
+		}
+		fontFaceIndex = val;
 	} else if (propName == "-size") {
 		int size2 = value.getInt(interp);
 		if (size != size2) {
@@ -98,6 +105,8 @@ void OSDText::getProperty(std::string_view propName, TclObject& result) const
 		result = text;
 	} else if (propName == "-font") {
 		result = fontFile;
+	} else if (propName == "-fontfaceindex") {
+		result = fontFaceIndex;
 	} else if (propName == "-size") {
 		result = size;
 	} else if (propName == "-wrap") {
@@ -155,7 +164,7 @@ std::unique_ptr<GLImage> OSDText::create(OutputSurface& output)
 	if (font.empty()) {
 		try {
 			font = TTFFont(systemFileContext().resolve(fontFile),
-			               size * scale);
+			               size * scale, fontFaceIndex);
 		} catch (MSXException& e) {
 			throw MSXException("Couldn't open font: ", e.getMessage());
 		}
--- a/src/console/OSDText.hh
+++ b/src/console/OSDText.hh
@@ -20,7 +20,7 @@ private:
 		return concatArray(
 			imageBasedProperties,
 			std::array{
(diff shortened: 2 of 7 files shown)


m9710797:
Thanks a lot for your PR!
Technically it looks fine (I only have some minor remarks, but let's discuss those later).

But I don't like the way how the GUI is extended. Currently the user should specify a face index without any feedback on which indices are valid or which index means what.

Personally I was also surprised to learn that .ttf / .ttc file can contain multiple faces inside. Also because no GUI program (that I know of) allows to select the face index. And then my initial reaction was: "why should openMSX allow to select this". But then I did some research and I learned that most other GUI programs present this choice in a different way, they instead collapse the filename+faceindex into a single choice.

I've written a small test program (thanks CoPilot):
```
#include <iostream>
#include <ft2build.h>
#include FT_FREETYPE_H

int main(int argc, char** argv) {
    if (argc != 2) {
        std::cerr << "Usage: " << argv[0] << " <fontfile>\n";
        return 1;
    }

    const char* fontfile = argv[1];
    FT_Library library;
    FT_Error error = FT_Init_FreeType(&library);
    if (error) {
        std::cerr << "Could not initialize FreeType library\n";
        return 1;
    }

    FT_Face face;
    error = FT_New_Face(library, fontfile, 0, &face);
    if (error) {
        std::cerr << "Could not open font file: " << fontfile << "\n";
        FT_Done_FreeType(library);
        return 1;
    }

    std::cout << "Font file: " << fontfile << "\n";
    std::cout << "Number of faces: " << face->num_faces << "\n";

    for (FT_Long i = 0; i < face->num_faces; ++i) {
        FT_Face subface;
        error = FT_New_Face(library, fontfile, i, &subface);
        if (error) {
            std::cerr << "  [Face " << i << "] Error loading face\n";
            continue;
        }
        std::cout << "  [Face " << i << "] Name: ";
        if (subface->family_name)
            std::cout << subface->family_name;
        else
            std::cout << "(unknown)";
        if (subface->style_name)
            std::cout << " - " << subface->style_name;
        std::cout << "\n";
        FT_Done_Face(subface);
    }

    FT_Done_Face(face);
    FT_Done_FreeType(library);
    return 0;
}
```
When I compile and run this code, I get for example output like this:
```
./font-info /tmp/NotoSansCJK-Black.ttc 
Font file: /tmp/NotoSansCJK-Black.ttc
Number of faces: 5
  [Face 0] Name: Noto Sans CJK JP - Black
  [Face 1] Name: Noto Sans CJK KR - Black
  [Face 2] Name: Noto Sans CJK SC - Black
  [Face 3] Name: Noto Sans CJK TC - Black
  [Face 4] Name: Noto Sans CJK HK - Black
```
I propose that we do something similar in openMSX.
* Current the `ImGuiSettings::getAvailableFonts()` method only stores font filenames.
* But instead, using a mechanism like the above test program, we should:
  * (After removing duplicate files) open each font file.
  * Retrieve the number of faces.
  * Compute a "display string" for each face within the file.
  * Present these strings to the user.
  * Internally, when the user selects one of these, we still work with the filename + faceIndex.

In other words. Most of your PR will remain unchanged. We only need a better way for the GUI to select the filename + face index.
The `-fontfaceindex` property in `OSDText.cc` and the `fontPropIndex` and `fontMonoIndex` settings can remain, I think.

Suzumizaki:
Wow, that's too much work for me...

Thank you for your code review and feedback.

m9710797:
> Wow, that's too much work for me...
> 
> Thank you for your code review and feedback.

No problem.
It's an interesting idea. Hopefully next week(?) I'll have more time. Then I'll try to continue your work.
Thanks.

m9710797:
I've merged your PR with some edits (that's why this PR is not automatically closed, I'll do that manually).
The most important change I made is in the GUI, now you no longer need to guess the `faceIndex`.
Please help test and give feedback.

Thanks again for your PR, you did most of the work :-)

Suzumizaki:
I checked the updated master branch and it seems to work fine. Thank you!

PR #1988, opened 2025-10-04
Make the setup view editable

### LINKED ISSUE #1950: Allow quick set-up edit (to replace Catapult use case)
```
<Quibus> wouterv: one feedback I got is that people still want to use Catapult, because: "I can just setup what I want and power on"
<Quibus> (Just like the real machine.)
<Quibus> We of course always start powered up, so you have to set it up at power on
<wouterv> did they try the new openmsx setup-feature?
<Quibus> I don't think so
<Quibus> But how does that help?
<wouterv> does it need a better gui?
<Quibus> (We didn't implement editing the setup in it by the way)
<wouterv> ah right.  Allowing the change stuff in the "Current setup" menu (might then become a window) is still TODO
<wouterv> but if people don't even try to use it, then i don't think we can convince them to stop using catapult
<Quibus> Well, it's still not the same as what I described
<wouterv> the difference is power off?
<Quibus> Mostly, I guess
<Quibus> So I also asked: does the power off really make a difference?
<Quibus> But I guess it's also because Catapult offers the most important settings in one simple view
<wouterv> an extended version of "current setup" comes close, no?    (Maybe we can then also give it a better name).
<wouterv> We can make this window open by default (the first time you start openMSX .. next times it remembers the open state from previous run)
<Quibus> Yeah, I guess it does come close
<Quibus> This sounds pretty good actually
<Quibus> "Quick setup" menu!
```
As it is kind of a regression w.r.t. Catapult... it could have high prio?

MBilderbeek:
Experiments to make the setup view editable.

When we have a clue on what works and what not, we can merge this after properly rebasing.

TODO:

- [x] Make the setup view also a separate window, that is open at first run (similar to Catapult)
- [x] Make the machine view simpler: do not show the details, move the select new machine option in there (and the machine details are visible there).
- [x] Extensions section: investigate if we can highlight the full row on hover, or in other sections also only highlight the last column.
- [x] Connectors section: use sub menus instead of dropdown. The submenu then contains the list that's currently in the dropdown box (to be clear: not move the dropdown box itself to the menu, but the content of that box). 
- [x] Connectors section: highlight full row vs last column.
- [x] Media section: highlight full row vs last column.
- [x] Run time state section: remove
- [x] Fix flickering submenu/simpleToolTip stuff

Closes #1950

### DIFF
--- a/src/imgui/ImGuiMachine.cc
+++ b/src/imgui/ImGuiMachine.cc
@@ -449,6 +449,23 @@ void ImGuiMachine::showNonExistingPreview()
 
 void ImGuiMachine::showSetupOverview(MSXMotherBoard& motherBoard, ViewMode viewMode)
 {
+	showSetupOverviewMachine(motherBoard, viewMode);
+
+	const ImGuiTreeNodeFlags flags = (viewMode == ViewMode::VIEW || viewMode == ViewMode::EDIT) ? ImGuiTreeNodeFlags_DefaultOpen :
+					viewMode == ViewMode::NO_CONTROLS ? (ImGuiTreeNodeFlags_DefaultOpen | ImGuiTreeNodeFlags_Leaf | ImGuiTreeNodeFlags_Bullet) :
+					ImGuiTreeNodeFlags_None;
+
+	showSetupOverviewExtensions(motherBoard, viewMode, flags);
+
+	showSetupOverviewConnectors(motherBoard, viewMode, flags);
+	
+	showSetupOverviewMedia(motherBoard, viewMode, flags);
+	
+	showSetupOverviewState(motherBoard, viewMode, flags);
+}
+
+void ImGuiMachine::showSetupOverviewMachine(MSXMotherBoard& motherBoard, ViewMode viewMode)
+{
 	using enum SetupDepth;
 
 	auto configName = motherBoard.getMachineName();
@@ -486,10 +503,11 @@ void ImGuiMachine::showSetupOverview(MSXMotherBoard& motherBoard, ViewMode viewM
 		// machine config is gone... fallback: just show configName
 		showMachineWithoutInfo(configName);
 	}
+}
 
-	const ImGuiTreeNodeFlags flags = (viewMode == ViewMode::VIEW || viewMode == ViewMode::EDIT) ? ImGuiTreeNodeFlags_DefaultOpen :
-					viewMode == ViewMode::NO_CONTROLS ? (ImGuiTreeNodeFlags_DefaultOpen | ImGuiTreeNodeFlags_Leaf | ImGuiTreeNodeFlags_Bullet) :
-					ImGuiTreeNodeFlags_None;
+void ImGuiMachine::showSetupOverviewExtensions(MSXMotherBoard& motherBoard, ViewMode viewMode, ImGuiTreeNodeFlags flags)
+{
+	using enum SetupDepth;
 
 	im::StyleColor(viewMode == ViewMode::SAVE && saveSetupDepth < EXTENSIONS, ImGuiCol_Text, getColor(imColor::TEXT_DISABLED), [&]{
 		im::TreeNode(depthNodeNames[EXTENSIONS].c_str(), flags, [&]{
@@ -562,12 +580,24 @@ void ImGuiMachine::showSetupOverview(MSXMotherBoard& motherBoard, ViewMode viewM
 			});
 		});
 	});
+}
+
+void ImGuiMachine::showSetupOverviewConnectors(MSXMotherBoard& motherBoard, ViewMode viewMode, ImGuiTreeNodeFlags flags)
+{
+	using enum SetupDepth;
+
 	im::StyleColor(viewMode == ViewMode::SAVE && saveSetupDepth < CONNECTORS, ImGuiCol_Text, getColor(imColor::TEXT_DISABLED), [&]{
 		im::TreeNode(depthNodeNames[CONNECTORS].c_str(), flags, [&]{
 			using enum ImGuiConnector::Mode;
 			manager.connector->showPluggables(motherBoard.getPluggingController(), viewMode == ViewMode::EDIT ? SUBMENU : VIEW);
 		});
 	});
+}
+
+void ImGuiMachine::showSetupOverviewMedia(MSXMotherBoard& motherBoard, ViewMode viewMode, ImGuiTreeNodeFlags flags)
+{
+	using enum SetupDepth;
+
 	im::StyleColor(viewMode == ViewMode::SAVE && saveSetupDepth < MEDIA, ImGuiCol_Text, getColor(imColor::TEXT_DISABLED), [&]{
 		im::TreeNode(depthNodeNames[MEDIA].c_str(), flags, [&]{
 			im::Table("##MediaTable", 2, [&]{
@@ -676,6 +706,11 @@ void ImGuiMachine::showSetupOverview(MSXMotherBoard& motherBoard, ViewMode viewM
 			});
 		});
 	});
+}
+
+void ImGuiMachine::showSetupOverviewState(MSXMotherBoard& motherBoard, ViewMode viewMode, ImGuiTreeNodeFlags flags)
+{
+	using enum SetupDepth;
 
 	if (viewMode != ViewMode::EDIT) {
 		auto time = (motherBoard.getCurrentTime() - EmuTime::zero()).toDouble();
@@ -689,6 +724,7 @@ void ImGuiMachine::showSetupOverview(MSXMotherBoard& motherBoard, ViewMode viewM
 		}
 	}
 }
+
 
 void ImGuiMachine::paint(MSXMotherBoard* motherBoard)
 {
--- a/src/imgui/ImGuiMachine.cc
+++ b/src/imgui/ImGuiMachine.cc
@@ -458,9 +458,9 @@ void ImGuiMachine::showSetupOverview(MSXMotherBoard& motherBoard, ViewMode viewM
 	showSetupOverviewExtensions(motherBoard, viewMode, flags);
 
 	showSetupOverviewConnectors(motherBoard, viewMode, flags);
-	
+
 	showSetupOverviewMedia(motherBoard, viewMode, flags);
-	
+
 	showSetupOverviewState(motherBoard, viewMode, flags);
 }
 
@@ -472,8 +472,10 @@ void ImGuiMachine::showSetupOverviewMachine(MSXMotherBoard& motherBoard, ViewMod
 	if (auto* info = findMachineInfo(configName)) {
 		if (viewMode != ViewMode::SAVE) {
 			if (viewMode == ViewMode::EDIT) {
-				im::TreeNode("Machine", ImGuiTreeNodeFlags_DefaultOpen, [&]{
-					im::Table("##MachineTable", 2, [&]{
+				im::TreeNodeWithoutID("Machine", ImGuiTreeNodeFlags_DefaultOpen, [&] {
+					im::Table("##shared-table", 2, [&] {
+						ImGui::TableSetupColumn("Label", ImGuiTableColumnFlags_WidthFixed);
+						ImGui::TableSetupColumn("Value");
 						if (ImGui::TableNextColumn()) {
 							ImGui::TextUnformatted("Brand and model");
 						}
@@ -510,10 +512,12 @@ void ImGuiMachine::showSetupOverviewExtensions(MSXMotherBoard& motherBoard, View
 	using enum SetupDepth;
 
 	im::StyleColor(viewMode == ViewMode::SAVE && saveSetupDepth < EXTENSIONS, ImGuiCol_Text, getColor(imColor::TEXT_DISABLED), [&]{
-		im::TreeNode(depthNodeNames[EXTENSIONS].c_str(), flags, [&]{
+		im::TreeNodeWithoutID(depthNodeNames[EXTENSIONS].c_str(), flags, [&]{
 			const auto& slotManager = motherBoard.getSlotManager();
 			bool anySlot = false;
-			im::Table("##ExtTable", 2, [&]{
+			im::Table("##shared-table", 2, [&]{
+				ImGui::TableSetupColumn("Label", ImGuiTableColumnFlags_WidthFixed);
+				ImGui::TableSetupColumn("Value");
 				for (auto i : xrange(CartridgeSlotManager::MAX_SLOTS)) {
 					if (!slotManager.slotExists(i)) continue;
 					anySlot = true;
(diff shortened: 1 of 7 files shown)


MBilderbeek:
We aren't having good results with submenus, nor with pop-ups (file browser pop-up breaks on it). 
So, perhaps we should make things simple and just open the corresponding window to modify the setup.

src/imgui/ImGuiMachine.cc
```diff
@@ -254,8 +254,9 @@ void ImGuiMachine::showMenu(MSXMotherBoard* motherBoard)
 			});
 
 			im::Menu("Current setup", true, [&]{
-				showSetupOverview(*motherBoard);
+				showSetupOverview(*motherBoard, ViewMode::VIEW);
```
m9710797, line 259: This pattern wasn't started in this PR, but it does make it worse.

The `showSetupOverview()` method is way too complex. It's a method that takes an enum parameter and then internally contains very large chunks specific for one of those enum values.

I think it would be better to create separate methods for each of these enum values. And then factor out helper functions that are common to (some of) these methods.

Having some configurable behavior in functions is fine. But here we've gone too far.

src/imgui/ImGuiMachine.cc
```diff
@@ -610,6 +684,13 @@ void ImGuiMachine::paint(MSXMotherBoard* motherBoard)
 	if (showTestHardware) {
 		paintTestHardware();
 	}
+	if (showQuickSetupEditor) {
+		ImGui::PushItemFlag(ImGuiItemFlags_AutoClosePopups, false);
```
m9710797: This `ImGui::{Push,Pop}ItemFlag()` is a leftover from the (failed) popup experiment?

src/imgui/ImGuiMedia.cc
```diff
@@ -194,6 +194,28 @@ void ImGuiMedia::loadLine(std::string_view name, zstring_view value)
 	}
 }
 
+void ImGuiMedia::showMediaWindow(std::string_view mediaName)
+{
+		auto getIndex = [&] (std::string_view prefix) {
+			if ((mediaName.size() >= (prefix.size() + 1)) && mediaName.starts_with(prefix)) {
```
m9710797: `x >= (y + 1)` can be simplified to `x > y`.

src/imgui/ImGuiMedia.cc
```diff
@@ -444,6 +466,46 @@ bool ImGuiMedia::drawExtensionFilter()
 	return changed;
 }
 
+void ImGuiMedia::paintCurrent(const TclObject& current, std::string_view type)
+{
+	if (current.empty()) {
+		ImGui::StrCat("Current: no ", type, " inserted");
+	} else {
+		ImGui::StrCat("Current: ", current.getString());
```
m9710797, line 474: Detail: TclObject::empty() actually tests if the objects represents a list containing zero elements(*1). Usually that also works to test if the TclObject contains an empty string, but not always. For example if the string is not a valid list representation (e.g. `abc {`, so unbalanced braces) then `empty()` also returns `true`.

Also from an efficiency point of view, we should not first have to convert (the internal representation of) the TclObject to a list, and then check if it has zero elements. Instead we should get the string (getting the string from a TclObject is always fast), and then check if that string is empty.

So something like:
```
if (auto str = current.getString(); str.empty()) {
   ImGui::StrCat(... no ... inserted...);
} else {
   ImGui::StrCat("Current: ", str)
}
```


(*1) The original idea was to give TclObject the same API as the STL containers, so `begin()`, `end()`, `size()`, `empty()`, ... So you can use Tcl lists in much the same way as STL containers. But that indeed be confused with the similar API in `std::string`.

src/imgui/ImGuiMedia.cc
```diff
@@ -444,6 +466,46 @@ bool ImGuiMedia::drawExtensionFilter()
...
+	}
+	ImGui::Separator();
+}
+
+void ImGuiMedia::paintRecent(const std::string& mediaName, ItemGroup& group,
+		                      function_ref<std::string(const std::string&)> displayFunc,
+		                      const std::function<void(const std::string&)>& toolTip,
+				      std::function<void()>* actionToSet)
+{
+	if (!group.recent.empty()) {
```
m9710797: Detail. One problem with our ImGui code is that it very quickly indents (too) deeply.
Here we can reduce the indentation level by one by replacing
```
if (!group.recent.empty()) {
    ...
}
```
with:
```
if (group.recent.empty()) return;
...
```
That become possible because we split this part into a smaller helper function. That's another reason to try to split much more in smaller functions. (Not specifically to this PR, but in general much of our ImGui code uses way too large functions.)

m9710797 (commented):
I've only done a *very* quick review. I certainly did not read everything in detail, and I did not try to understand all new code. I think it's too soon for that. First we should get the functionality and the UI right.

I've already mentioned most of these points in IRC. But here's a repeats/summary:
* The UI to select a new machine is not good. I knew this functionality was present, I was searching for it under the `machine` section, but it wasn't there. If I couldn't find it, then users less familiar with openMSX might also have problems.
* Currently the `machine` section shows the machine details. But should that info be present in this `quick setup` window? The goal of this window is to allow the user to quickly make changes (like what we had in catapult), not to give a full detailed overview. E.g. we also don't show rom details (e.g. from the rom database) in this window.
* So I think we should remove what's currently in the `machine` section and instead move the select new machine option in there (and the machine details are visible there). Then it's consistent with what the other sections offer and behave.
* Extensions section. Looks good, but we should investigate if we can highlight the full row on hover, or in other sections also only highlight the last column. IOW try to make everything consistent.
* Connectors section. This is the only section that uses dropdown boxes. (and thus the rows take up more vertical space). We should at least try with using sub menus instead. The submenu then contains the list that's currently in the dropdown box (to be clear: not move the dropdown box itself to the menu, but the content of that box). That's the same functionality as we have now, but the look is then consistent with all the other sections in the `quick setup` window.
* Media section. OK, but same remark as above about highlight full row vs last column.
* Run time state section. Does not belong in `quick setup` window IMHO. For sure not with the current content. (You mentioned reversebar, but also don't doesn't belong here I think).

src/imgui/ImGuiMachine.cc
MBilderbeek: I guess so, no idea where that came from.

src/imgui/ImGuiMedia.cc
MBilderbeek: Of course.

src/imgui/ImGuiMedia.cc
MBilderbeek: Fixed.

src/imgui/ImGuiMedia.cc
MBilderbeek: Fixed.

src/imgui/ImGuiMachine.cc
MBilderbeek: Unfortunately, the code of this PR was mostly based on this change. So, to do this differently, will mean totally rewriting the code for this PR... I guess I'd better start another PR and use this one as an inspiration.... or do you have a better idea?

MBilderbeek:
@m9710797 I revised the connectors menu in the QSE. Please review. (I heavily modified the first commit of this PR for that.)
Now the connectors section has the same problem as some other sections: only the right side is highlighted. That is to be fixed later.

m9710797:
Personally I like the more consistent look of the connector section.
You're right that now only the 2nd column is highlighted on hover. But a good fix to make it globally consistent could be to have that behavior also everywhere in the media section.

MBilderbeek:
> Personally I like the more consistent look of the connector section. You're right that now only the 2nd column is highlighted on hover. But a good fix to make it globally consistent could be to have that behavior also everywhere in the media section.

Done. But there is some weird interaction with simpleToolTip and an open menu item: it will alternate to show every frame. Is that something wrong in my code @m9710797 ?

MBilderbeek:
@m9710797 what was your idea for the machine selector part exactly? I'm not sure I understood your remark fully: 
> move the select new machine option in there (and the machine details are visible there).

I modified the machine part of the QSE, please check if that is what you meant.

m9710797:
I'll check the code and the simpleToolTip stuff later this evening.

> @m9710797 what was your idea for the machine selector part exactly? I'm not sure I understood your remark fully:

I didn't have any concrete idea. But i did find the current approach non intuitive. My vague idea is to make it consistent with all the other sections in this window. So have a (single?) entry where if you click in the 2nd column it opens the machine selection window (similar to how most media entries behave). And I think it's fine to then remove the detailed machine info from this window (I don't think it belongs here, e.g. we also don't show the ROM cartridge details directly in this window),

m9710797:
> Done. But there is some weird interaction with simpleToolTip and an open menu item: it will alternate to show every frame. Is that something wrong in my code @m9710797 ?

Initially I couldn't reproduce this. But then I placed the `Quick Setup` window fully inside the main window, but close to the right border. That way, when the sub menu  opens, it falls (partially) outside the main window. And then I can reproduce this issue. I don't think that there's anything wrong with your code. It seems like a Dear ImGui bug???

But to be honest, I'm not sure the tooltip on the 2nd column in the connectors menu is useful. Especially because it too often overlaps with the items in the sub menu and then hides important info (also without this alternating frame issue). And the same info from this tooltip can easily be obtained by hovering over the items in the sub menu itself.

In other words: also from a usability point of view we should remove these 'middle' tooltips.

MBilderbeek:
> > Done. But there is some weird interaction with simpleToolTip and an open menu item: it will alternate to show every frame. Is that something wrong in my code @m9710797 ?
> 
> Initially I couldn't reproduce this. But then I placed the `Quick Setup` window fully inside the main window, but close to the right border. That way, when the sub menu opens, it falls (partially) outside the main window. And then I can reproduce this issue. I don't think that there's anything wrong with your code. It seems like a Dear ImGui bug???
> 
> But to be honest, I'm not sure the tooltip on the 2nd column in the connectors menu is useful. Especially because it too often overlaps with the items in the sub menu and then hides important info (also without this alternating frame issue). And the same info from this tooltip can easily be obtained by hovering over the items in the sub menu itself.
> 
> In other words: also from a usability point of view we should remove these 'middle' tooltips.

Removed them for all items that have a submenu and kept them for items that open a window.

MBilderbeek:
> I'll check the code and the simpleToolTip stuff later this evening.
> 
> > @m9710797 what was your idea for the machine selector part exactly? I'm not sure I understood your remark fully:
> 
> I didn't have any concrete idea. But i did find the current approach non intuitive. My vague idea is to make it consistent with all the other sections in this window. So have a (single?) entry where if you click in the 2nd column it opens the machine selection window (similar to how most media entries behave). And I think it's fine to then remove the detailed machine info from this window (I don't think it belongs here, e.g. we also don't show the ROM cartridge details directly in this window),

OK, please check how it looks now and please also look at the TODO and commented out code on TreeNode vs Indent. The TreeNode gave a lot of extra empty space at the top, which I didn't understand.

m9710797:
With the following patch I'm personally happy with the UI:
```
diff --git a/src/imgui/ImGuiMachine.cc b/src/imgui/ImGuiMachine.cc
index 3dc3fff36..9417038a6 100644
--- a/src/imgui/ImGuiMachine.cc
+++ b/src/imgui/ImGuiMachine.cc
@@ -454,11 +454,10 @@ void ImGuiMachine::showSetupOverview(MSXMotherBoard& motherBoard, ViewMode viewM
        if (auto* info = findMachineInfo(configName)) {
                if (viewMode != ViewMode::SAVE) {
                        if (viewMode == ViewMode::EDIT) {
-                               //im::TreeNode("##MachineTreeNode", ImGuiTreeNodeFlags_Leaf, [&]{
-                               im::Indent([&]{
+                               im::TreeNode("Machine", ImGuiTreeNodeFlags_DefaultOpen, [&]{
                                        im::Table("##MachineTable", 2, [&]{
                                                if (ImGui::TableNextColumn()) {
-                                                       ImGui::TextUnformatted("Machine");
+                                                       ImGui::TextUnformatted("Model");
                                                }
                                                if (ImGui::TableNextColumn()) {
                                                        if (ImGui::Selectable(info->displayName.c_str())) {
```
So, have also a collapsible section for the machine (and then to not have `Machine > Machine`, I renamed the latter to `Model`). Sure that's one extra line, and collapsing this section won't save much space, but see below. I also do like the tooltip on hovering the machine, then we don't even miss the info that was previously shown in this section.

Now everything is fully consistent:
* All sections are presented in the same way as collapsible tree nodes, with the same indentation, consistent row height, the same table structure, etc ...
* All use the same highlighting (hover over the 2nd column). And the same way of interaction (click or hover the 2nd column).

Of course this only my personal opinion. But I think, given the limitations of popups and file browser dialogs, and given the idea of trying to resemble the `Setup` window, this is the best(?) we can do.

MBilderbeek:
@m9710797 please check ImGuiMachine.cc:744 for what I did to try to auto resize the window when opening. But I guess this is wrong, as it only works when reopening, but not when opening from the start.

src/imgui/ImGuiMachine.cc
MBilderbeek: > Unfortunately, the code of this PR was mostly based on this change. So, to do this differently, will mean totally rewriting the code for this PR... I guess I'd better start another PR and use this one as an inspiration.... or do you have a better idea?

@m9710797 I changed the condition to default (always) and that seems to work exactly as I want it to work, also if the window is restored from a previous session.

Please check the commits again. If this is good enough for you to merge, please merge the PR.

m9710797:
> @m9710797 please check ImGuiMachine.cc:744 for what I did to try to auto resize the window when opening. But I guess this is wrong, as it only works when reopening, but not when opening from the start.

I have _not_ actually tested this yet. But after reading more documentation I found this:
```
ImGui::SetNextWindowSize(ImVec2(0, 0), ImGuiCond_Appearing);
```
However, I do have some concerns. Maybe invalid, but we should test well before adding this line:
* The QSE window will only change size (i mean the size that auto-fit would give) when something in the machine setup changes.
* The intended use case for this window are beginner users who will do most of the changes via this window.
* In other words it won't happen often that setup change is made while this window is closed.
* But that means the user will anyway have to manually resize. IOW the user doesn't really benefit for this auto-size-on-reopen.
* Always closing and reopening the QSE windows defeats the purpose. Then it's faster (less GUI clicks) to directly open/close the more specific window.

Also:
* Consider the scenario of a user having carefully arranged all his windows (including QSE).
* But then when he exits and restarts openMSX, all his windows will be in the expected position and have the expected size, except for QSE.
* I think it's fine, but we should check how this line interacts with a docked QSE window.

So in summary:
* This is a new behavior, we should test it well before starting to use it.
* For sure we don't want this behavior for most (all?) debugger and tool windows.
* Maybe it's fine for a window like QSE?
* But OTOH it also won't gain much (it does not resize at the moments it matters).

MBilderbeek:
Merged, after discussion with Wouter.

PR #1990, opened 2025-10-07
Add response modifier ("0": binary or "1": text)

pvmm:


share/scripts/_tcl_bridge.tcl
```diff
@@ -21,6 +21,7 @@
 #
 #   res_real_len:  +8 (input,  2 bytes): actual length of the result
 #   status:       +10 (input,  1 byte ): success (0) or failure (1)
+#   res_modifier: +11 (input,  1 byte ): binary (0) or text (1)
```
m9710797, line 24: Detail, the documentation is wrong, this new field is an `output` field, not `input` (seen from the MSX program point of view). And then this new field more logically belongs together with the 4 fields above rather than here.

Personally I think it's very unfortunate that a feature like this is needed. It's because of the different newline encoding used by Tcl and MSX. Actually is Tcl _always_ using this encoding? Or does Tcl just use the host OS convention? So is it different between Linux/Windows. Because in case of the latter this PR is wrong on Windows: then it will do too many substitutions.

Usually(*1) in openMSX we try to stick to emulating the MSX hardware and be agnostic about the operating system running on the MSX hardware. Usually this is the regular MSX-BIOS and/or MSX-DOS and that indeed uses CR+LF to go to the next line. But it's also possible to run an OS like UZIX on MSX hardware, and that one does follow the UNIX newline convention.
(*1) there are exceptions of course, for example in the `diskmanipulator` command we do make some assumptions about the filesystem and character encoding etc. But when possible we try to avoid this.

I see 3 possibilities to fix this newline encoding mismatch:
1) Do the conversion in the Tcl command itself. So in the MSX program instead of sending `my command` we send `string map "\n \n\r" [my command]`
2) Do the conversion in the MSX program that consumes the result. E.g. in the print loop of your `TCL.COM` application, print CR+LF for each LF character in the result.
3) Do the conversion in the `tcl_bridge` script (this PR).

I've listed them in order of my personal preference. In other words: personally I think the solution in this PR is not the right place to solve this problem. Personally I'd rather keep `tcl_bridge` encoding-agnostic (binary / text / which newline convention / ...).

Note that usually you know whether a command _might_ produce newlines (only the minority of commands) or will _definitely_ not (the majority). So the extra work in solution 1 is only needed for a few commands.

m9710797 (commented):
Thanks for the update. I understand the problem, but I think this is the wrong solution.

share/scripts/_tcl_bridge.tcl
pvmm: > Personally I think it's very unfortunate that a feature like this is needed. It's because of the different newline encoding used by Tcl and MSX. Actually is Tcl _always_ using this encoding? Or does Tcl just use the host OS convention?

Seems to me that Tcl always use `\n` only. This is on windows:
image

PR #1991, opened 2025-10-07
Add a manifest to force char type to utf-8 in non-ASCII Windows.

Suzumizaki:
On Japanese versions of Windows, I found that if a savestate file is named with non-ASCII characters, openMSX.exe throws an exception with the message "No mapping for the Unicode character exists in the target multi-byte code page." and terminate.

Specifically, this occurs when trying to get a list using Save state -> Load State in the GUI.

This Pull Request will fix this.

reference:
https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page

There are 2 Japanese related article / 日本語では下記を参照ください:
https://zenn.dev/tenka/articles/vc_utf8_prog2022_1
https://blog.techlab-xe.net/using-string-utf-8-winapi/


m9710797:
Thanks!
I have no windows system myself, so I can't test this PR. But you say it works so I'll trust you.

Suzumizaki:
This also works fine, thank you!

PR #1994, opened 2025-10-12
Tiny updates on completion

Suzumizaki:
I'd like to add new features to tab completion later, but I'm trying to clean up the code first.

I believe I've confirmed that it works on Windows and macOS, but please excuse me if you encounter any issues.

### DIFF
--- a/src/commands/Completer.cc
+++ b/src/commands/Completer.cc
@@ -96,7 +96,7 @@ bool Completer::completeImpl(std::string& str, std::vector<std::string_view> mat
 	matches.erase(u.begin(), u.end());
 
 	auto minsize_of_matches = std::ranges::min(
-		matches, [](auto& l, auto& r) { return l.size() < r.size(); }).size();
+		matches, {}, &std::string_view::size).size();
 
 	bool expanded = false;
 	while (str.size() < minsize_of_matches) {
--- a/src/commands/Completer.cc
+++ b/src/commands/Completer.cc
@@ -95,28 +95,23 @@ bool Completer::completeImpl(std::string& str, std::vector<std::string_view> mat
 	auto u = std::ranges::unique(matches);
 	matches.erase(u.begin(), u.end());
 
+	auto minsize_of_matches = std::ranges::min(
+		matches, [](auto& l, auto& r) { return l.size() < r.size(); }).size();
+
 	bool expanded = false;
-	while (true) {
+	while (str.size() < minsize_of_matches) {
 		auto it = begin(matches);
-		if (str.size() == it->size()) {
-			// match is as long as first word
-			goto out; // TODO rewrite this
-		}
-		// expand with one char and check all strings
 		auto b = begin(*it);
 		auto e = b + str.size();
-		utf8::unchecked::next(e); // one more utf8 char
-		std::string_view string2(std::to_address(b), e - b);
-		for (/**/; it != end(matches); ++it) {
-			if (!equalHead(string2, *it, caseSensitive)) {
-				goto out; // TODO rewrite this
-			}
-		}
-		// no conflict found
-		str = string2;
+		utf8::unchecked::next(e);
+		std::string_view test_str(b, e);
+		if (!std::ranges::all_of(matches,
+			[&](auto& val) {
+				return equalHead(test_str, val, caseSensitive);
+			})) { break; }
+		str = test_str;
 		expanded = true;
 	}
-	out:
 	if (!expanded && output) {
 		// print all possibilities
 		for (const auto& line : formatListInColumns(matches)) {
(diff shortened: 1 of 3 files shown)


src/commands/Completer.cc
```diff
@@ -95,28 +95,23 @@ bool Completer::completeImpl(std::string& str, std::vector<std::string_view> mat
 	auto u = std::ranges::unique(matches);
 	matches.erase(u.begin(), u.end());
 
+	auto minsize_of_matches = std::ranges::min(
+		matches, [](auto& l, auto& r) { return l.size() < r.size(); }).size();
```
m9710797: Detail: maybe using the 'projection' stuff is simpler:
```
auto min_match_size = std::ranges::min(matches, {}, &std::string_view::size).size();
```
or alternatively, using `transform`:
```
auto min_match_size = std::ranges::min(matches | std::views::transform(&std::string::size));
```

src/commands/GlobalCommandController.cc
```diff
@@ -403,22 +403,20 @@ void GlobalCommandController::tabCompletion(std::vector<std::string>& tokens)
 			TclObject command = makeTclList("openmsx::tabcompletion");
 			command.addListElements(tokens);
 			try {
-				TclObject list = command.executeCommand(interpreter);
+				TclObject org_list = command.executeCommand(interpreter);
+				std::vector<zstring_view> list;
+				std::ranges::copy(org_list, std::back_inserter(list));
```
m9710797: Would this be better? (More efficient? More readable?)
```
auto list = to_vector<zstring_view>(command.executeCommand(interpreter));
```

`to_vector()` is our version of `std::ranges::to()` but that one is c++23, and not yet available in all compilers we still want to support.

---

Though I don't really like that we make an unnecessary copy here. (I do understand that this is not performance critical code, so maybe this is not too bad at all).

I think you want to have the items available in a vector because then you can use the `Completer::completeString()` version that takes a 'range' instead of 'iterator+sentinel` (and then the latter is no longer used and can be removed).

I have not actually tried this, but what about using `std::subrange`, that can be constructed from two iterators, it does not make a copy, and it can be passed to the `completeString()` overload that takes a 'range'. Would that be a good compromise? Could you give it a try? But if it's too difficult, then we can of course use your version that does make a copy.

src/commands/Completer.hh
```diff
@@ -100,25 +94,15 @@ private:
...
 template<std::ranges::forward_range Range>
-inline std::vector<std::string_view> Completer::filter(
+std::vector<std::string_view> Completer::filter(
 	std::string_view str, Range&& range, bool caseSensitive)
 {
-	return filter(str, std::ranges::begin(range), std::ranges::end(range), caseSensitive);
+	std::vector<std::string_view> result;
+	std::ranges::copy_if(std::forward<Range>(range),
+	                     std::back_inserter(result),
+	                     [&](auto value) { return equalHead(str, value, caseSensitive); });
```
m9710797: Detail: maybe use `const auto& value` (add `const` and `&`).

Side question:
Instead of making a copy, would it be possible to return a `std::views::filter`?
But not needed for this PR. Maybe in a follow-up PR we could try this (the existing code also made a copy).

m9710797 (commented):
This code could indeed use some cleanups.
I'm looking forward to the new features you'd like to add.

I have some small remarks, but only about details. Or suggestions for small tweaks.

src/commands/Completer.cc
Suzumizaki: There seemed to be some issues with the second one so I used the first one. Thanks for suggestions.

src/commands/GlobalCommandController.cc
Suzumizaki: Thanks for suggestions, I tried to use `std::ranges::subrange`.

src/commands/Completer.hh
Suzumizaki: I added the `const` and `&`...oops, I just realized I forgot about the side question.

Suzumizaki:
I appreciate the code review and suggestions.

As I wrote above, I left one thing behind, so I'll try to update it. I'll let you know if it's too difficult for me.

m9710797:
Thanks!
I've merged this PR, with one small change (one extra commit).

Suzumizaki:
About "Instead of making a copy, would it be possible to return a `std::views::filter`?":

Well, it seems it's not that simple.

Because changing `filter()` requires changing `completeFileNameImpl()`, which has `append(matches, filenames)` which expects matches to be a `std::vector<std::string>`.

Sorry, this seems a bit out of my control.

m9710797:
No problem at all. The `append()` stuff is a good reason for why today this is not possible.
Thanks for looking into this.

PR #1997, opened 2025-10-14
Fix the force termenation when tab-completion is invoked with non-ASCII string.

Suzumizaki:
`utf8::unchecked::substr` is a code point unit process, so it should not be used here.

### DIFF
--- a/src/imgui/ImGuiConsole.cc
+++ b/src/imgui/ImGuiConsole.cc
@@ -305,8 +305,8 @@ int ImGuiConsole::textEditCallback(ImGuiInputTextCallbackData* data)
 	switch (data->EventFlag) {
 	case ImGuiInputTextFlags_CallbackCompletion: {
 		std::string_view oldLine{data->Buf, narrow<size_t>(data->BufTextLen)};
-		std::string_view front = utf8::unchecked::substr(oldLine, 0, data->CursorPos);
-		std::string_view back  = utf8::unchecked::substr(oldLine, data->CursorPos);
+		std::string_view front{ data->Buf, narrow<size_t>(data->CursorPos) };
+		std::string_view back{ data->Buf + data->CursorPos, data->Buf + data->BufTextLen };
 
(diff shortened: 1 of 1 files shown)


src/imgui/ImGuiConsole.cc
```diff
@@ -305,8 +305,8 @@ int ImGuiConsole::textEditCallback(ImGuiInputTextCallbackData* data)
 	switch (data->EventFlag) {
 	case ImGuiInputTextFlags_CallbackCompletion: {
 		std::string_view oldLine{data->Buf, narrow<size_t>(data->BufTextLen)};
-		std::string_view front = utf8::unchecked::substr(oldLine, 0, data->CursorPos);
-		std::string_view back  = utf8::unchecked::substr(oldLine, data->CursorPos);
+		std::string_view front{ data->Buf, narrow<size_t>(data->CursorPos) };
+		std::string_view back{ data->Buf + data->CursorPos, data->Buf + data->BufTextLen };
```
m9710797, line 309: Thanks! This is indeed a bug.

Historical context: before we switched the console code to 'Dear ImGui' we had our own implementation, and in our version the cursor position was in code-point-units. But it seems that in 'Dear ImGui' `data->CursorPos` is in byte-units. I overlooked this difference when changing the code.

Detail: in your version the `oldLine` variable is no longer used, right? (Can be removed.)
Or alternatively you could write this code as follows (I've NOT tested this):
```
std::string_view oldLine{data->Buf, narrow<size_t>(data->BufTextLen)};
auto front = oldLine.substr(0, data->CursorPos);
auto back  = oldLine.substr(data->CursorPos);
```

Thanks for your bug fix. I'll merge it later today when I'm back at home.

m9710797:
Thanks! I've pushed your fix. But with my edit applied.
Please help test.

Suzumizaki:
It seems to work correctly on my Windows 11 and macOS 15 Sequoia.
Thanks for accepting the pull request.

PR #2003, opened 2025-10-19
Enable file completion using tcl

Suzumizaki:
The description below may seem to be long, but my purpose is simple: I want to make better file-path tab completion usable from tcl script.

To see how this work instantly, try a script included in the zip file below.
[test_enhanced_tab_competion.zip](https://github.com/user-attachments/files/22992073/test_enhanced_tab_competion.zip)

Also, the patch code is not so difficult in C++ side. Tcl side is a bit complecated.

By this Pull Request, callback procedure, the 2nd argument of '`set_tabcompletion_proc`', will get the ability to control tab completion from Tcl code. I select triple-hyphen prefix and easy to understand English for me, but more better idea is acceptable. The callback procedure should prepend these options before candidates.

Compatibility is kept, only exception is when the candidates originally contain the same string options below. That's the reason I selected triple-hyphen which is similar but different from standard Tcl style.

All options you need should be written each callback. Otherwise they will be restored default behavior for each callback.

'`---rewritewith`', '`---donewith`' and '`---noappendable`' are used to select entry by number. In that case, candidates contain both number and result to show. I want this feature to select the path that contains non-ASCII string, because Input Method Editor (IME for Windows, IM for others) cannot work well. But ofcouse you can use any other purpose.

`---commonprefix <string>`
Set partially completed string. For example, parent directory path for a file. Given string should not be included in candidate tokens. Otherwise the string may be doubled.
Generally, even when the last component of the path is partially completed, just give the path of its parent directory to this option and candidates should have common beginning part of last component of the path. That manner makes usual behaivior of console.
Default is empty string.

`---goable <string for regular expression>`
When only one candidate matchs with last token and the token matches given regular expression, append new empty token to console input.
Default is empty string, means all the case appends new token.
For example,  if you want not to append new token only when the path points directory, set '`---goable .*\[^/\]$`' in tcl script.
Note that this regular expression is used with std::regex_match in C++ context, not std::regex_search.

`---sort`, `---nosort`
Sort candidates or not.
Use '---nosort' when your candiates are already sorted.
Default is `---sort`.

`---case`, `---nocase`
Set case-sensitive or case-insensitive for completion match.
It can overwrite previous or initial setting.
Default depends on the 3rd argument given to '`set_tabcompletion_proc`' procedure.
Currently (and originally), this opiton doesn't affect sorting.

`---rewritewith <string>`
Replace last token with given string. Even when only one candidate matchs the given string, no token added to console. Also, given string should be match one of candidate strings.
If you use `---commonprefix`, generally you should also prepend that string to the string for this option.
Default is empty string.

`---donewith <string>`
Complete the last token with given string and append new empty token. The candidates will be ignored.
If you use `---commonprefix`, generally also prepend that string to the string for this option.
Default is empty string.

`---appendable`, `---noappendable`
Allow append partially common matched string or not.
'`---noappendable`' especially works when only one candidate matches with current last token.
For exapmle, use '`---noappendable`' for the case the candidate strings contain the number or guide for each entry.
Default is `---appendable`.

`---`
Make clear the tab completion candidates start from next token.

### DIFF
--- a/share/init.tcl
+++ b/share/init.tcl
@@ -153,11 +153,11 @@ proc tabcompletion {args} {
 	lazy_handler $command
 	set result ""
 	if {[info exists tabcompletion_proc_sensitive($command)]} {
-		set result [namespace eval :: $tabcompletion_proc_sensitive($command) $args]
-		lappend result true
+		set result ---case
+		lappend result {*}[namespace eval :: $tabcompletion_proc_sensitive($command) $args]
 	} elseif {[info exists tabcompletion_proc_insensitive($command)]} {
-		set result [namespace eval :: $tabcompletion_proc_insensitive($command) $args]
-		lappend result false
+		set result ---nocase
+		lappend result {*}[namespace eval :: $tabcompletion_proc_insensitive($command) $args]
 	}
 	return $result
 }
--- a/share/scripts/_utils.tcl
+++ b/share/scripts/_utils.tcl
@@ -136,7 +136,8 @@ proc file_completion_by_number {args} {
 		}
 	}
 	if {[string length $last_arg]} {
-		puts "Give last component by number after: $last_arg"
+		set last_component [file tail $last_arg]
+		puts "Give last component by number instead of '$last_component'"
 	}
 	lassign [file_complete_common {*}$args] sub_params dirs files 
 	set entries ""
--- a/share/scripts/_utils.tcl
+++ b/share/scripts/_utils.tcl
@@ -111,16 +111,53 @@ proc clip {min max val} {
 
 # Provides file completion.
 proc file_completion {args} {
+	lassign [file_complete_common {*}$args] sub_params dirs files
+	return [list {*}$sub_params --- {*}$dirs {*}$files]
+}
+
+# Provides file completion, but last component should be typed by number.
+proc file_completion_by_number {args} {
+	set last_arg [lindex $args end]
+	set ends_with_separator [expr {[string index $last_arg end] eq "/"}]
+	if {!$ends_with_separator && [regexp ^\\d+$ [file tail $last_arg] num]} {
+		set dname [file dirname $last_arg]
+		if {[string index $dname end] ne "/"} { set dname $dname/ }
+		set modified_args [lreplace $args end end $dname]
+		lassign [file_complete_common {*}$modified_args] sub_params dirs files
+		set entries [list {*}$dirs {*}$files]
+		set subcom ""
+		if {$num < [llength $dirs]} {
+			set subcom "---rewritewith"
+		} elseif {$num < [llength $entries]} {
+			set subcom "---donewith"
+		}
+		if {$subcom ne ""} {
+			return [list {*}$sub_params $subcom "$dname[lindex $entries $num]"]
+		}
+	}
+	if {[string length $last_arg]} {
+		puts "Give last component by number after: $last_arg"
+	}
+	lassign [file_complete_common {*}$args] sub_params dirs files 
+	set entries ""
+	set cnt 0
+	foreach i [list {*}$dirs {*}$files] {
+		lappend entries ${cnt}:$i
+		incr cnt
+	}
+	return [list {*}$sub_params ---noappendable --- {*}$entries]
+}
+
+proc file_complete_common {args} {
 	set dirs [list]
 	set files [list]
 	set last_arg [lindex $args end]
 	set ends_with_separator [expr {[string index $last_arg end] eq "/"}]
 	set is_exists [file exists $last_arg]
-	set is_dir [file isdirectory $last_arg]
 	if {$ends_with_separator} {
 		set common_prefix $last_arg
 		if {![file isdirectory $last_arg]} {
-			puts "The path seems not be a directory."
+			puts "The path seems not be a directory: $last_arg"
 			set common_prefix [file dirname $last_arg]
 		}
 	} else {
@@ -144,14 +181,13 @@ proc file_completion {args} {
 			lappend files $fixed_i
 		}
 	}
-	set result [list ---goable .*\[^/\]$ ---nosort ---nocase]
+	set sub_params [list ---goable .*\[^/\]$ ---nosort ---nocase]
 	if {$common_prefix ne ""} {
-		lappend result ---commonprefix $common_prefix
+		lappend sub_params ---commonprefix $common_prefix
 	}
 	set dirs [lsort -dictionary $dirs]
 	set files [lsort -dictionary $files]
-	lappend result {*}$dirs {*}$files
-	return $result
+	return [list $sub_params $dirs $files]
 }
 
 # Replaces characters that are invalid in file names on the host OS or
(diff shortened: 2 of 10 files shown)


Suzumizaki:
...Hmm, is this a less elegant solution? For example, should we delegate more to the Tcl side?

Suzumizaki:
Ok, It seems I can make more smart version. Sorry, I cancel this PR.

m9710797:
Hi. Thanks for working on this.

I understand you've closed this PR because you want to create an improved version. But nevertheless I would already like to ask some questions. Not about the implementation, but about the motivation and about the usability from an openMSX user point of view.

Let me first describe the current behavior. Currently when in the console you type `<tab>`, openMSX will internally (I'm simplifying a bit):
* Collect all possible strings that are allowed at that position in that command (e.g. maybe we typed `plug jo`, so we're typing the 2nd token in the `plug` command, at that position we expect a `connector`, so the list of possible strings is `cassetteport`, `joyporta`, `joyportb` and `printerport`).
* Filter on the partial string that is already typed (in the above example we've already typed `jo`, so we filter the list for strings that start with that prefix, and then we only retain `joyporta` and `joyportb`).
* If after filtering the list only contains a single element anymore, then we complete with that element (but in our example there are 2 elements left)
* If there are multiple elements left in the list, then we only complete up-to the longest common prefix (thus in our example we complete up to `joyport`, so without the trailing `a` or `b`).
* If there are multiple elements in the list and the partially typed token is already the longest common prefix then we don't change the token in the command line, but instead we display all the (remaining) possibilities. (So in our example the first time the user presses `<tab>` we complete `jo` -> `joyport`, and the second time the user presses `<tab>` we show `joyporta` and `joyportb`).

My understanding is that this behavior is not good enough for you? And the reason is that if (e.g. in filenames) the next character is a non ASCII-character that currently(?) can't easily be typed in the openMSX console.
Is my understanding correct? Is that indeed the main motivation for your PR?

And as a solution you would print a numbered list of possible completions, and then the user can select from this list. It's not clear to me yet how exactly the user selects such a number.

I know you've closed your PR and are reworking it. Nevertheless I'd like to give some feedback:
* What I don't like in your current approach is that the command itself selects how completion should behave. The command itself decides whether or not to present the possibilities as a numbered list.
* Currently the command itself only selects whether completions are case-sensitive or not. But in your design you give the command a lot more options.
* My preference is to not do that. I prefer that tab-completion behaves, as much as possible, the same for all the commands.
* So I'd prefer to keep the decisions in the commands minimal, and instead improve the general completion feature.
* But I may have misunderstood your design. Please correct me if I'm wrong.

To fix the usability problem with non-ASCII filenames I'd propose the following:
* If the user presses `<tab>` and there's only a single match, then we use that and move to the next token.
* If there are multiple matches, we complete till the longest common prefix.
* If the previous step couldn't add any extra characters, then we show all possibilities.
* _Up to here it's exactly the same as we have now, the next step is new_.
* If the user then again presses `<tab>` we start cycling through all the remaining matches.
* So if we retake the above example, after we've shown `joyporta` and `joyportb` as possible matches, if the user again presses `<tab>` we will complete `joyporta`, and pressing `<tab>` once again completes `joyportb` (another `<tab>` and we wrap around back to `joyporta`).
* We could also support `<shift>+<tab>` to cycle in the other direction.

Would that we an good alternative for your problem?

I should also say that I didn't understand your `--goable` option. So maybe my solution doesn't address that??

Suzumizaki:
Thank you for the comment.
I think my understanding of current implementation is same as yours.

> If the user then again presses <tab> we start cycling through all the remaining matches.

That is good idea for a few candidates, but when the count exceeds about 10, I can't accept that.

> What I don't like in your current approach is that the command itself selects how completion should behave. The command itself decides whether or not to present the possibilities as a numbered list.

If you feel some difficulty about that, please consider my trial is as _last resort_. If openMSX has no problem with non-ASCII characters and IMEs (for Windows and IMs for other OSs), I don't have to make this approach.

_But anyway_, file completion is not work fine when tcl code want that, because it need pressing backspace for each path separator(resolved by `---goable`).  See the comment of `proc file_completion` in `share/scripts/_utils.tcl` on _master_ branch. Another one problem also exists, see description about `---commonprefix` below.

Here's the purpose of each options. Most of these are delegated to tcl code in #2004.
- `---commonprefix` is used to hide common parent path components. otherwise, all candidates should have whole common uplevel paths user entered.
- `---nosort` can be used when callbacked tcl code want to order candidates like directories first and followed by files.
- `---case` and `---nocase` can be used when some arguments are case-sesitive but the others are not. The NTFS or FAT on Windows are generally case insensitive, then the user can be use any case when only '---nocase` for the argument (not important but easy to implement and very effective).
- `---goable` (`---done` and `---rewrite` for new version) is used to change behavior whether the given path is the tail component or just parent or more upper path.
- `---appendable` is trickey hack to cope with existing code. No need this functionality for new version.

> Would that we an good alternative for your problem?

As described above, sorry to say it wouldn't.
Thank you again for considering my proposal.

m9710797:
OK, thanks for your clarifications. I now understand the problem you're trying to solve much better.

I see that you opened PR#2004 with an improved version. Let's use this thread for a general discussion on the topic and PR#2004 to talk about that specific implementation.

So the root problem is the lack of a proper IME for non-ASCII input. Should we instead try to address that? But I should immediately add that I'm not familiar with such things, so I don't really known what that would entail. So for sure I'll need help with that.
Dear ImGui is widely used, also outside of Europe, so most likely other projects have already solved the IME problem?? Can we find such example projects and borrow ideas from them? But again I'll need help, I cannot do this on my own, I simply lack the knowledge about IME.

What I don't like about PR#2003 and PR@2004 is that it is a specific solution only for tab-completion of filenames in Tcl. What about all calls to Completer::completeFilename() directly from c++. Those have the exact same problem, no? And these PRs don't solve that. Or what about completion of non-filenames. For example I can have a savestate with a filename containing non-ASCII characters (savestate files are located in `~/.openMSX/savestates`). Then tab-completion on the `loadstate` command will list all available save-states (the filename without the `.oms` extension). Or many other examples, e.g. in a machine description file (`~/.openMSX/share/machine/<some-machine>.xml`) I can give devices non-ASCII names, and then those names show up in e.g. the `debug` command, that command also has tab-completion, ....  All those are also problems, but they are not solved by these PRs.

So I do agree this is a problem and needs a solution. But I think we need a generic tab-completion solution. Not a solution that only works for Tcl. And preferably also a solution that doesn't need special annotations per command. Let me clarify this last point. Some annotations are fine. Like case-sensitive or not, or an indication that we're completing filenames (so that the completion code knows that on a directory we should stop at the trailing `/` character rather than moving to the next token). But let's keep these annotations very simple and minimal (I see that PR#2004 is already much better in this aspect).

Suzumizaki:
I wasn't sure how to respond, but I decided to do what I could :)

> What I don't like about PR#2003 and PR@2004 is that it is a specific solution only for tab-completion of filenames in Tcl. 

Resolved by #2017, includes #2004.

PR #2016, opened 2025-10-27
Unify tab completion for file path

Suzumizaki:
[Not working, cancelled] This PR has a bug. Until `scripts/_utils.tcl` are loaded, any tab completion does not work.

### Introduction
On the discussion #2003, I noticed the reviewer want to keep consistency between 2 versions of tab completion (C++ one and Tcl one). The issue would be done by this Pull Request.

### Main Enhancements
This PR replaces C++ version with the code calling Tcl one. The profits are:
- Ofcourse keep consistensy of tab completion for file path between commands implemented by C++ and by Tcl.
- End users can fix or replace tab completion for all command which they want to use it.
- All features only in C++ version previously are enabled to any tcl commands you want.

### Addtional features
Because of my initial motivation is this side (Also see #2003 discussion), new functionalities below are added. 
- Number Selection Mode
  - When type comma for the last component of the path and hit TAB key, the entries are shown with `,0`, `,1`, `,2` ... prefixed.
  - When type comma and number, tab completion replace it with associated path component. 
  - This is implemented as a _last resort_ to follow IMs/IMEs are not correctly working. Because of the reason, to make minimum changes, it does not support narrowing down like normal completion.

- Supporting Fullwidth characters to align possibility strings.
  - See https://www.unicode.org/reports/tr11/

### Implementation Details

In fact C++ version was not originally same as Tcl version at least 2 points below, even after #2004 applied. But this PR contains both.
- It has ability to merge contents of multiple parent directories (Try tab completion against `load_settings`).
- It has ability to add name of subcommands (Try one against `diska` or `carta`, etc.)

### DIFF
--- a/share/init.tcl
+++ b/share/init.tcl
@@ -154,10 +154,9 @@ proc tabcompletion {args} {
 	set result ""
 	if {[info exists tabcompletion_proc_sensitive($command)]} {
 		set result [namespace eval :: $tabcompletion_proc_sensitive($command) $args]
-		lappend result true
 	} elseif {[info exists tabcompletion_proc_insensitive($command)]} {
-		set result [namespace eval :: $tabcompletion_proc_insensitive($command) $args]
-		lappend result false
+		set result ---nocase
+		lappend result {*}[namespace eval :: $tabcompletion_proc_insensitive($command) $args]
 	}
 	return $result
 }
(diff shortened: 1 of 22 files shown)


Suzumizaki:
Sorry, I'm misunderstaning "Draft pull request". I close now.

PR #2017, opened 2025-10-27
Unify tab completion for file path (fixed)

Suzumizaki:
### Introduction
On the discussion #2003, I noticed the reviewer want to keep consistency between 2 versions of tab completion (C++ one and Tcl one). The issue would be done by this Pull Request.

### Main Enhancements
This PR replaces C++ version with the code calling Tcl one. The profits are:
- Ofcourse keep consistensy of tab completion for file path between commands implemented by C++ and by Tcl.
- End users can fix or replace tab completion for all commands even implemented by C++.
- All features only in C++ version previously are enabled to any tcl commands you want.

### Addtional features
Because of my initial motivation is this side (See #2003 discussion), new functionalities below are added. 
- Number Selection Mode -> **Totally Restructed**. See latest comment.
  - When type comma for the last component of the path and hit TAB key, the entries are shown with `,0`, `,1`, `,2` ... prefixed.
  - When type comma and number, tab completion replace it with associated path component. 
  - This is implemented as a _last resort_ to follow IMs/IMEs are not correctly working. Because of the reason, to make minimum changes, it does not support filter like normal completion.

- Supporting Fullwidth characters to align possibility strings.
  - See https://www.unicode.org/reports/tr11/
  - End users can switch or replace by `set_monospace_width_proc`.

- Possible strings (candidates) now only contain last component of the path.

  - Previously, all items have common path which typed in console.

### Implementation Details
This PR includes and built on #2004. About `---rewrite` and `---done`, see that PR. -> **Restructed**. `---cont` and `---done` for now.

New file `scripts/_file_completion.tcl` is added. It is large enough to split from `_utils.tcl`. -> **Renamed**. `_tab_completion.tcl` for now. 

In fact C++ version was not originally same as Tcl version at least 2 points below, even after #2004 applied. This PR keeps containing both.
- It has ability to merge contents of multiple parent directories (Try tab completion against `load_settings`).
- It has ability to add name of subcommands (Try one against `diska` or `carta`, etc.)

Fortunately to say, all related C++ classes has protected or private member to invoke tab completion Tcl version. I promoted them to protected.

Maybe this PR will address the remaining issues in #1994. -> **No need to do**. That part no longer exists.

Thanks.


m9710797:
Thanks a lot for continuing to work on this!
I'm at work right now. I'll try to review this evening when I'm back at home.

m9710797:
Again, thanks for working on this. I've now had some time to try out your code.

Let me first give some general feedback without looking at your actual implementation / code.
* Tab-completion has been available for a very long time (+20 years?) in openMSX. Many users use it, we should make sure that it continues to work well, and without breaking anything for existing users. Additionally we should make sure that your new use case (easy entry of non-ASCII input) works well.
* In an earlier review my feedback was that filename completion should work the same both for c++ and Tcl. You did address that in this new PR.
* But in that review I also mentioned that non-filenames can also have non-ASCII characters, and the solution should also address tab-completion for those. That's not solved in this new PR.
* I tried your PR. I have a directory containing all my ROM files (about ~900 files, other people have a much larger ROM collection). But with your PR when I press tab on `carta ~/msx/roms/` (so tab-completion for that whole directory), openMSX blocks for more than 1 second. That's not acceptable. It seems that the Tcl implementation is too slow. Without your PR, so with filename completion implemented in c++, there's no noticeable delay (less than 0.1 second). So I'm afraid we must implement tab-completion in c++ rather than in Tcl.
* Currently in your PR you can get a numbered list by starting the path component with `,`. I don't like that for these reasons:
  * It's not something "discoverable". New users will never think of trying this. The only thing we can do is document this feature in the manual, but we know that most users don't read the manual.
  * I cannot use this in the middle of a filename. Suppose I have some files that have a common ASCII prefix, I type the first few characters and then tab to complete that common prefix, I press tab again to see all possibilities, and then I see that the next character is non-ASCII (so I can't easily type it). The only possibility I have now is to erase the partial filename, type `,` and press tab again. But that then gives me the full list, maybe with 100s of files. So I loose the advantage of already having typed the common prefix (and then have a shorter list to choose from).
  * An ASCII prefix followed by some non-ASCII prefix is common for some European accented characters (like ë, è, é, ê), today that also doesn't work well in combination with tab-completion. So if we implement something we should make sure it covers both Japanese, European and any other input language.
  * I can no longer use tab-completion for filenames that start with `,`. That's unlikely a problem for filenames, but it can be a problem is we want to use this mechanism for generic (non-filename) tab-completion.

So I'm afraid I cannot accept your PR in it's current form :(
Please don't be discouraged by this. I will now try to help you address the above points.

Here are some requirements:
* For speed we must implement tab-completion in c++, not Tcl. (But if needed we can export this functionality from c++ to Tcl).
* The non-ASCII entry problem should be solved in a generic way. So we should not do something special that only works for filenames.
* We should solve this in a way that also helps users who have not read the manual.

I've only thought about this for maybe 10-15 minutes, so maybe I did overlook something. But here's an idea for a possible solution:
* From a user interface point of view:
  * Pressing tab once completes the longest common prefix (exactly what we have now).
  * Pressing tab again shows the full list of possibilities (also like what we have now).
  * But in addition the previous step show something like `press <tab> again for a selection popup`
  * If you then press tab a 3rd time you get a popup that shows a scrollable list (maybe 10 lines high?). It reacts to these key presses:
    * Cursor UP/DOWN move the selection up/down  (optionally PgUp/PgDown to move a full page, HOME/END to move to begin/end).
    * Enter and space pick the selection.
    * ESC closes the popup (as if you didn't press that 3rd tab).
    * Any other key: as if you didn't press the 3rd tab and typed this key instead.
    * You can also use the mouse to control this popup, but the intention is that everything in the console should be easy to control via the keyboard.
* From an implementation point of view:
  * Filename and non-filename completion should behave the same. So filename completion only generates the list of files and then calls the generic tab-completion function. The only(?) difference is that filename completion doesn't move to the next token when the completed token is a directory (ends with `/`).
  * For the popup we could probably use an ImGui::Popup(). This must be implemented in c++ (not possible from Tcl).
  * ... sorry I didn't have enough time yet to think about this and work out all the details.

I hope you can do something with this feedback. And if you're stuck with something ask for help. I am willing to help implement this feature. I do think it's an important problem that needs a solution.

Suzumizaki:
Thank you for new review!

> But with your PR when I press tab on carta ~/msx/roms/ (so tab-completion for that whole directory), openMSX blocks for more than 1 second. That's not acceptable. 

One reason is that I tried to trace the column ordering code in C++ (not optimized). I think the new commits have made it faster, but it possibly might not be enough if you have around 1000 entries in a directory. How it works in your environment?

Suzumizaki:
> An ASCII prefix followed by some non-ASCII prefix is common for some European accented characters (like ë, è, é, ê), today that also doesn't work well in combination with tab-completion. 

Ah... That's a problem. I have already used `lsort -dictionary`, but Tcl can't work well?

m9710797:
> ... I think the new commits have made it faster, but it possibly might not be enough if you have around 1000 entries in a directory. How it works in your environment?

Speed is not the most important reason to move away from Tcl.
IMHO the most important reason is that the UI isn't good (I mean the typing `,<number>` stuff). I proposed an alternative (with a ImGui::Popup()) but that's only available from C++, not Tcl.


> > An ASCII prefix followed by some non-ASCII prefix is common for some European accented characters (like ë, è, é, ê), today that also doesn't work well in combination with tab-completion.
> 
> Ah... That's a problem. I have already used `lsort -dictionary`, but Tcl can't work well?

I didn't mean that the sort-order isn't correct. I mean that typing such accented characters also isn't possible on a plain US-qwerty keyboard layout. So better tab-completion mechanism would also help for that.

Suzumizaki:
> > > An ASCII prefix followed by some non-ASCII prefix is common for some European accented characters (like ë, è, é, ê), today that also doesn't work well in combination with tab-completion.
> >
> > Ah... That's a problem. I have already used `lsort -dictionary`, but Tcl can't work well?
>
> I didn't mean that the sort-order isn't correct. I mean that typing such accented characters also isn't possible on a plain US-qwerty keyboard layout. So better tab-completion mechanism would also help for that.

Hm? You mean that is not brought by this PR? 

Anyway, I found Tcl CAN'T sort correctly...
https://wiki.tcl-lang.org/page/lsort#ae700b29c398fa05555ccec915e58adc9e497c914ecd2ca2d14a638446a0ff19

Suzumizaki:
> Anyway, I found Tcl CAN'T sort correctly...

That's a good reason to retry unifying into C++ not Tcl, even though I feel difficulty to accept some of other reasons you pointed.

Suzumizaki:
_I have succeeded to satify all the orders!_

You may be annoyed that it's not implemented in C++, but I hope you'll accept this.
Instructions on how to use the new method are at the second section of this post.

### Resolved issues

- > The non-ASCII entry problem should be solved in a generic way.
  - Done. New method can be used with all U+0080+ characters.

- > We should solve this in a way that also helps users who have not read the manual.
  - Done. Only end users have to know is 'tab completion exists' as usual.

- > Filename and non-filename completion should behave the same. 
  - Done. Only three features are different as same as master branch.

- > I cannot use this in the middle of a filename. 
  - Resolved by new method.

- > I can no longer use tab-completion for filenames that start with ,.
  - Resolved by new method, anyway I don't know why you could not before.

- > openMSX blocks for more than 1 second. That's not acceptable.
  - Resolved already by previous commit. I tested with 1000 files.

- Tcl cannot sort locale defined order
  - Resolved by delegating back to C++ code.

- Support Fullwidth characters when arranged in columns
  - Resolved. There's a way to alter the method to know length of any string.

### New method to enter 'not typable' characters

1. Type (or use tab completion) to fill string before such characters.
2. When you reached before the characters, press TAB key.
3. After the general possibilities list, you will also see a list of alternatives and their guides.
4. Append one of alternation and press TAB again. That's all.

#### The details of new method

The alternations are `,[0-9a-z]{1,4}$` (in regex), but automatically try to avoid conflict existing possibilities. When U+0080+ characters _doesn't_ exist at the same position of other possibilites, or given alternate expression is not binded to any possibilities, such expressions perform as normal string. As described in regex, the alternations work only at the end of console input.

There're problematic cases, but very rare.
1. When the _same position_ of the possibilities consume all `,[0-9a-z]`, and U+0080+ character exists in _same position_ of the other one, usual tab completion will not work for `,[0-9a-z]`. In this case, just directly type it (Alternation get the priority).
2. When over 46,656(36**3, minimum case when 1st id character is allowed only one) entries have the different `,[0-9a-z]{1,4}` each other at _same position_, there's no way to resolve until the trigger is changed by `set_tab_alternation_trigger`.
3. To be confortable in the case like above, `set_tab_alternation_trigger` procedure can be used to switch `,` to some better character like `@` or `#`.

Thank you.

m9710797:
Thanks a lot again!

Next two days, I'm very busy. After that I'll try to make time to review and hopefully merge this.

Suzumizaki:
There was a small problem so I fixed it.

And I've attached a set of simple test files. Please use it if necessary.
[pr2017.zip](https://github.com/user-attachments/files/23358043/pr2017.zip)

m9710797:
Hi. I found some time to examine your PR.

It works well, good job ... or at least it works well given the limitations of your current approach (keep it purely text based).

But what I (strongly) don't like is the user-interface (UI): typing `,<n>` to select from the list. It's weird. I don't know of any other application that uses this UI style. And we shouldn't invent our own (strange) style for such very common functionality. But as I said above I don't think it's possible to do much better than this if you're limited to text input/output.

So what if we remove that limitation? I started an experiment, see PR#2026. THIS IS NOT FINISHED. It's ONLY to investigate if a popup approach is feasible at all. What I'd like to do next is combine all the good parts from this PR with the UI from my experiment.

Here's a summary of my experiment:
* Typing TAB completes up-to the longest match (as before).
* If TAB was not able to (partially) complete anything, we now NO LONGER show the list of all possibilities as text.
* Instead we open a popup showing a (scrollable) table with all remaining possibilities.
* You can then use the CURSOR keys (or TAB) to select one of these.
* And with ENTER or SPACE you type the selection (TODO also allow to click with the mouse).
* ESC closes the popup.
* Typing any other key also closes the popup and types that key in the input field.

And again, my experiment does not yet include any of your other improvements. E.g. I do like your improvements for filename completion. But that's for later (I didn't have much time yet, and I already wanted you to show this experiment). Feedback is welcome.

Suzumizaki:
This review must have taken a long time. Thank you for taking the time. I'm glad that you recognized it as a nearly optimal solution without using a GUI, and that you appreciated aspects other than the selection method.

To be honest, I don't understand why you dislike my proposal so much, as I feel that some other features of openMSX are odd or wired. But if you dislike it that much, there's nothing I can do about it.

Of course, I'll cooperate with the work on PR#2026. Thank you.

m9710797:
Hi,

This is a combined reply to this thread and your last message in https://github.com/openMSX/openMSX/pull/2026.

> To be honest, I don't understand why you dislike my proposal so much, as I feel that some other features of openMSX are odd or wired. But if you dislike it that much, there's nothing I can do about it.

Maybe I used too strong wording. What I tied to say is: tab-completion (or completion is general) is a feature that many applications offer, and users have certain expectations for how it should look/behave. Your solution is (IMHO) too different from those existing solutions. I think it's better if openMSX stays closer to the existing solutions.

At the moment https://github.com/openMSX/openMSX/pull/2026 is still an experiment/prototype. I created it to test if using a popup-approach was even possible. But in its current state it still has several problems / bugs / usability issues.

Some of the things I want to improve are:
* Better sizing of the popup. Now the code tries to minimize the number of rows, but I think It may look nicer if we find a better balance between number of rows and columns.
* Placement of the popup. Currently it's always centered in the console window. But I want to change that so that it appears right above the text cursor. That will maybe address your concern about accidentally pressing TAB, and not noticing a small popup.
* It doesn't work well yet when the popup contains many items and requires a scrollbar.
* There are still bugs. For example in some cases when you select an item from the popup with SPACE or ENTER, that key is (wrongly) also typed in the text input field.

Once the behavior of the popup itself work well, I'll look at importing your improvements to filename completion.

m9710797:
@Suzumizaki  What should we do with this PR?

Most parts have been implemented on the `master` branch, but not everything. I think the most(?) important missing part is integration with Tcl, e.g. you cannot select the improved filename-completion from Tcl.

Do you see other parts of the PR that are still missing?

I think it would be a good idea to add Tcl integration. But personally I'm giving this lower priority. I mean there are other higher priority (IMHO) items to work on.

Because most of this PR is now no longer valid, should we close it? Should we open a new ticket as a reminder for the Tcl integration part?

Suzumizaki:
> Should we open a new ticket as a reminder for the Tcl integration part?

Please do so.

The reason is simple: the original purpose (see title of #2003, *Enable file completion using tcl*) has not been achieved.

Issue #2019, opened 2025-10-30
Instability when using "CIEL (Ademir Carchano) Expert Turbo MSX machine profile


djcaetano:
I am using openMSX v0.21 and I have noticed that when I select "**CIEL (Ademir Carchano) Expert Turbo**" as the current machine, openMSX crashes whenever I open **Machine > Current Setup** menu (the application is closed). 
If I try to save a setup based on this machine (CIEL), it seems to work... but when I try to load the setup... openMSX also crashes.

djcaetano:
Video of the problem: [https://www.amusementfactory.com.br/bugs/2025-10-30%2008-29-14.mp4](https://www.amusementfactory.com.br/bugs/2025-10-30%2008-29-14.mp4)

pvmm:
This crashes on me too.
```
openmsx: src/CartridgeSlotManager.cc:52: virtual void openmsx::CartridgeSlotManager::Slot::getMediaInfo(openmsx::TclObject&): Assertion `config->getType() == HardwareConfig::Type::ROM' failed.

Thread 1 "openmsx" received signal SIGABRT, Aborted.
0x00007ffff74803cc in __pthread_kill_implementation () from /lib64/libc.so.6
(gdb) bt
#0  0x00007ffff74803cc in __pthread_kill_implementation ()
   from /lib64/libc.so.6
#1  0x00007ffff742618e in raise () from /lib64/libc.so.6
#2  0x00007ffff740d6d0 in abort () from /lib64/libc.so.6
#3  0x00007ffff740d639 in __assert_fail_base.cold () from /lib64/libc.so.6
#4  0x0000000000411585 in openmsx::CartridgeSlotManager::Slot::getMediaInfo (
    this=0x38ef4c0, result=...) at src/CartridgeSlotManager.cc:52
#5  0x0000000000db8357 in operator() (__closure=0x7fffffffc110)
    at src/imgui/ImGuiMachine.cc:528
#6  0x0000000000dbf80a in im::Table<openmsx::ImGuiMachine::showSetupOverview(openmsx::MSXMotherBoard&, ViewMode)::<lambda()>::<lambda()>::<lambda()> >(const char *, int, ImGuiTableFlags, const ImVec2 &, float, struct {...}) (
    str_id=0x1ad805c "##MediaTable", column=2, flags=0, outer_size=..., 
    inner_width=0, next=...) at src/imgui/ImGuiCpp.hh:485
[...]

PR #2020, opened 2025-10-30
Fix CIEL_Expert-Turbo.xml

pvmm:
Problem described in #2019.


m9710797:
I've applied your patch in 2fe8f3dda5b7e693d4d03d46c9e1fe535219c534, though with a different commit message.

Thanks for this PR, it helped me located/fix the problem much quicker.

PR #2026, opened 2025-11-07
Experiment: pop-up for tab-completion

m9710797:
This is NOT yet finished:
* Code still looks ugly, not well structured, not well documented.
* Needs more fine tuning. E.g. pop-up placement, sizing, maybe scrollbar, ...
* Filename completion should be improved.
* ...

The current goal is an experiment, proof-of-concept. Is a pop-up for tab-completion feasible at all?

### DIFF
--- a/src/commands/Completer.cc
+++ b/src/commands/Completer.cc
@@ -59,31 +59,33 @@ std::vector<std::string> Completer::formatListInColumns(std::span<const std::str
 	return format(input, output->getOutputColumns() - 1);
 }
 
-bool Completer::equalHead(std::string_view s1, std::string_view s2, bool caseSensitive)
+bool Completer::equalHead(std::string_view s, const CompletionCandidate& c)
 {
-	if (s2.size() < s1.size()) return false;
-	if (caseSensitive) {
-		return std::ranges::equal(s1, subspan(s2, 0, s1.size()));
+	if (c.text.size() < s.size()) return false;
+	if (c.caseSensitive) {
+		return std::ranges::equal(s, subspan(c.text, 0, s.size()));
 	} else {
-		return strncasecmp(s1.data(), s2.data(), s1.size()) == 0;
+		return strncasecmp(s.data(), c.text.data(), s.size()) == 0;
 	}
 }
 
-bool Completer::completeImpl(std::string& str, std::vector<std::string_view> matches,
-                             bool caseSensitive)
+void Completer::completeImpl(std::vector<std::string>& tokens, std::vector<CompletionCandidate> matches)
 {
-	assert(std::ranges::all_of(matches, [&](auto& m) {
-		return equalHead(str, m, caseSensitive);
-	}));
+	auto& str = tokens.back();
+	assert(std::ranges::all_of(matches, [&](auto& m) { return equalHead(str, m); }));
 
 	if (matches.empty()) {
 		// no matching values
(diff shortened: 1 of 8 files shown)


Suzumizaki:
Hi, I built and tried this PR.
As a first attempt, it appears to work correctly.

The ability to use ImGui::CalcTextSize() is a great feature. It eliminates the need to count full-width characters.

Specific to this PR, it might be better to make it easier to notice when the user accidentally hits the TAB key too many times. It can be a little confusing, especially when the candidate popup window is small.

As for the remaining work, it seems like it would be a lot of work to add it carefully without breaking the surrounding implementation...

Suzumizaki:
I'm still checking this out. The new popup looks a lot better than the old one. Is it just not implemented yet that it crashes when I try to select it with Space or Enter?

m9710797:
> ... Is it just not implemented yet that it crashes when I try to select it with Space or Enter?

It's still experimental code of course, but for me it was working well (I didn't see crashes). So thanks for reporting this! I carefully reviewed the code, and I might have found the problem. Please try again. The latest version now also includes your fixes for proportional fonts in the console (I've rebased this PR on top of the master branch).

m9710797:
I implemented the filename tab-completion improvements. Inspired by https://github.com/openMSX/openMSX/pull/2017.

I believe this PR now implements the most important parts of #2017. One (the only??) part that is still missing is exposing the enhanced tab-completion possibilities to Tcl. But I'd like to split that part into a separate PR.
@Suzumizaki  if you have time, could you review, I mean the features and the usability (you're also welcome to review the code of course, but that's not what I'm asking here). Are there other parts of #2017 that I overlooked?

Suzumizaki:
Hi, I tested new version.
It looks fine, and no crashes I found. 

I think there's no critical difference and can go ahead, but compare with #2017:

1. Any plans to treat untypable characters? It's selectable but can't be filtered against next characters of that once encountered.
2. In my code, Directories first and files are following them. Would you plan to implement?
3. Also, would you keep merging 'extra's on sorting entries? Maybe I haven't talked about it in detail yet: In my code, when type `carta` and press TAB, `insert` and `eject` are placed on the end of the array. They are subcommand of `carta', not a file or a directory.

Thanks.

m9710797:
Hi, thanks for testing.

About 1, I'm not sure I fully understand your point, nevertheless I'll try to reply (but do correct me if I missed something). My preference is to not treat any characters special (in other words treat them all equal). The new popup is a replacement for the old list of all possibilities that appeared as text on double-tab (I do like that afterwards the popup disappears, so that this list no longer "pollutes" the text). So we also want this popup if the completions don't have any "untypable" characters. And of course it would still be best if we could fix these "untypable" characters so that they DO become typable (but that's not something I can fix myself, I don't have any experience with IME).

About 2 and 3. I didn't plan to do this. I would keep it simple and just sort the full list. You can see the difference between files and directories by the trailing '/' character. I would also not treat extra entries like `insert` and `eject` special. My reasoning is: it's _only_ tab-completion it's meant for more advanced users who are familiar with the command line. We do also have `Main menu bar > Media > Disk Drive A > Browse file` that opens a real file browser dialog. That one does better distinguish files and directories, and doesn't mix them with extra items like `insert` and `eject`.

Later today I'd like to do a final review of the code myself. And then merge it to the master branch.

PR #2027, opened 2025-11-09
Fixed an issue where full-width characters could not be displayed correctly in console input. 

Suzumizaki:
As a side effect, they now display properly even when proportional fonts are used (although other parts that require a monospaced font will be displayed incorrectly).

About FullWidth characters, see https://www.unicode.org/reports/tr11/

Note:
At Sep. 2023 (ImGuiConsole: fix rendering of wider than 256 pixel (colored) chunks, a0f43326e1f69ded8265f98f9b38f69fd98a9977), The comment says "avoid ImGui::CalcTextSize(): it's off-by-one for sizes >= 256 pixels". That's because overdrawing causes too bright or too bold text appearance at that time.

As another comment says "... we use ImGui::InputText() as-is, but then overdraw the text using the correct colors.", colored text is drawn on the text drawn by internally Imgui code.

But later, Mar. 2024 (ImGuiConsole: improve drawing of console input line, 723ba959f5be026298bd2e46dcf0272d04ff953a), The default color of console input text, given to Imgui system, is changed to transparent. After that, overdrawing problems are dissappered, like too bright or too bold.

This means, Sep. 2023 fix can be reverted at that time. Now we can safely call ImGui::CalcTextSize().

### DIFF
--- a/src/imgui/ImGuiConsole.cc
+++ b/src/imgui/ImGuiConsole.cc
@@ -263,7 +263,6 @@ void ImGuiConsole::paint(MSXMotherBoard* /*motherBoard*/)
 		/**/ gl::vec2 drawPos = topLeft + gl::vec2(style.FramePadding);
 		/**/ ImVec4 clipRect = gl::vec4(topLeft, bottomRight);
 		/**/ auto* drawList = ImGui::GetWindowDrawList();
-		/**/ auto charWidth = font->GetFontBaked(fontSize)->GetCharAdvance('A'); // assumes fixed-width font
 		/**/ if (ImGui::IsItemActive()) {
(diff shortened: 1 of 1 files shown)


m9710797:
Thanks!
Looks good at first sight.
Later today (or maybe tomorrow) I'll try to test it and then merge it.

BTW, I made some progress on https://github.com/openMSX/openMSX/pull/2026. The popup should behave a lot better now. Next step is to improve filename completion (inspired by your work, so thanks again).

m9710797:
Seems to work well. Merged. Thanks again!

Suzumizaki:
Thank you too!

PR #2046, opened 2025-12-02
Fixed an issue where bbc60d81 prevented Visual Studio 2022 from building.

Suzumizaki:


m9710797:
Thanks a lot for reporting and proposing a fix!
We didn't notice this bug because on github the code compiled fine (also with vs2022, possibly a newer version).

I did not merge your PR directly. Instead, based on your patch, I solved the problem in a different way (use `string_view` instead of `string`, the former can be `constexpr`).

Suzumizaki:
It works fine, thank you.

PR #2059, opened 2025-12-21
fix: Improve UI responsiveness when paused/breaked (#2024)

### LINKED ISSUE #2024: UI drawing delays when "breaked"
When openMSX is in "breaked" state, the UI drawing has significant drawing delays. One example is shown in the video below.

It was suggested that is is due to the lower rendering framerate when the emulation is paused, but the delays seem longer than that. In any case, now that there is a full UI in openmsx, I don't think any additional delays are justified.

https://github.com/user-attachments/assets/63238a81-0aee-4a14-bedd-d4371a618bc0

ej-sanmartin:
Fixes #2024, context on background and implementation justificaiton can be found in that issue's thread.

tldr; paused-state ImGui FPS drop in openMSX is intentional and caused by the combined `repaintDelayed()` throttle and a 20 ms sleep in `Reactor::run()`, my changes tunes the delays to be lower and ensure it only affects block state

### DIFF
--- a/src/Reactor.cc
+++ b/src/Reactor.cc
@@ -685,7 +685,7 @@ void Reactor::run()
 
 	while (running) {
 		auto isBlocked = [&] { return (blockedCounter > 0) || !activeBoard; };
-		if (isBlocked) {
+		if (isBlocked()) {
 			// Compute timeout, min of MAX_WAIT_MS and time until next
 			// RTScheduler event. This keeps UI responsive while avoiding
 			// busy-waiting when paused.
--- a/src/Reactor.cc
+++ b/src/Reactor.cc
@@ -682,10 +682,14 @@ void Reactor::powerOn()
 void Reactor::run()
 {
 	static constexpr int MAX_WAIT_MS = 8;
+	bool executionFailed = false;
 
 	while (running) {
-		auto isBlocked = [&] { return (blockedCounter > 0) || !activeBoard; };
+		auto isBlocked = [&] {
+			return executionFailed || (blockedCounter > 0) || !activeBoard;
+		};
 		if (isBlocked()) {
+			executionFailed = false;  // Reset when blocked by other conditions
 			// Compute timeout, min of MAX_WAIT_MS and time until next
 			// RTScheduler event. This keeps UI responsive while avoiding
 			// busy-waiting when paused.
@@ -712,7 +716,7 @@ void Reactor::run()
 			// copy shared_ptr to keep Board alive (e.g. in case of Tcl
 			// callbacks)
 			auto copy = activeBoard;
-			blocked = !copy->execute();
+			executionFailed = !copy->execute();
 		}
 	}
 }
--- a/src/Reactor.cc
+++ b/src/Reactor.cc
@@ -692,7 +692,7 @@ void Reactor::run()
 			// to also use a sleep/poll loop, with even shorter
 			// sleep periods as we use here. Maybe in future
 			// SDL implementations this will be improved.
-			eventDistributor->sleep(20 * 1000);
+			eventDistributor->sleep(8 * 1000); // 8ms
 		}
 	}
 }
--- a/src/Reactor.cc
+++ b/src/Reactor.cc
@@ -616,6 +616,11 @@ void Reactor::enterMainLoop()
 		if (activeBoard) {
 			activeBoard->exitCPULoopAsync();
 		}
+		// Wake SDL_WaitEventTimeout() in case main thread is blocked
+		// waiting for events. SDL_PushEvent() is thread-safe.
+		SDL_Event wakeEvent{};
+		wakeEvent.type = SDL_USEREVENT;
+		SDL_PushEvent(&wakeEvent);
 	}
 }
 
@@ -676,23 +681,38 @@ void Reactor::powerOn()
 
 void Reactor::run()
 {
+	static constexpr int MAX_WAIT_MS = 8;
+
 	while (running) {
-		eventDistributor->deliverEvents();
-		bool blocked = (blockedCounter > 0) || !activeBoard;
-		if (!blocked) {
+		auto isBlocked = [&] { return (blockedCounter > 0) || !activeBoard; };
+		if (isBlocked) {
+			// Compute timeout, min of MAX_WAIT_MS and time until next
+			// RTScheduler event. This keeps UI responsive while avoiding
+			// busy-waiting when paused.
+			auto& rtScheduler = getRTScheduler();
+			int timeoutMs = MAX_WAIT_MS;
+			if (auto nextTime = rtScheduler.getNextTime()) {
+				auto now = Timer::getTime();
+				if (*nextTime > now) {
+					auto deltaUs = *nextTime - now;
+					auto deltaMs = static_cast<int>(deltaUs / 1000);
+					timeoutMs = std::min(timeoutMs, std::max(1, deltaMs));
+				} else {
+					// Already past due, don't wait
+					timeoutMs = 0;
+				}
+			}
+			eventDistributor->deliverEvents(timeoutMs);
+		} else {
+			eventDistributor->deliverEvents();
+		}
+
+		// Re-check. State may have changed during event processing
+		if (!isBlocked()) {
 			// copy shared_ptr to keep Board alive (e.g. in case of Tcl
 			// callbacks)
 			auto copy = activeBoard;
 			blocked = !copy->execute();
-		}
-		if (blocked) {
-			// At first sight a better alternative is to use the
-			// SDL_WaitEvent() function. Though when inspecting
-			// the implementation of that function, it turns out
-			// to also use a sleep/poll loop, with even shorter
-			// sleep periods as we use here. Maybe in future
-			// SDL implementations this will be improved.
-			eventDistributor->sleep(8 * 1000); // 8ms
 		}
 	}
 }
--- a/src/events/EventDistributor.cc
+++ b/src/events/EventDistributor.cc
@@ -12,7 +12,6 @@
 
 #include <algorithm>
 #include <cassert>
-#include <chrono>
 
 namespace openmsx {
 
@@ -57,7 +56,6 @@ void EventDistributor::distributeEvent(Event&& event)
 		//             EventDistributor::unregisterEventListener()
 		//   thread 2: EventDistributor::distributeEvent()
 		//             Reactor::enterMainLoop()
-		condition.notify_all();
 		lock.unlock();
 		reactor.enterMainLoop();
 	}
@@ -68,14 +66,14 @@ bool EventDistributor::isRegistered(EventType type, EventListener* listener) con
 	return contains(listeners[size_t(type)], listener, &Entry::listener);
 }
 
-void EventDistributor::deliverEvents()
+void EventDistributor::deliverEvents(std::optional<int> timeoutMs)
 {
 	static PriorityMap priorityMapCopy; // static to preserve capacity
 	static EventQueue eventsCopy;       // static to preserve capacity
 
 	assert(Thread::isMainThread());
 
-	reactor.getInputEventGenerator().poll();
+	reactor.getInputEventGenerator().poll(timeoutMs);
 	reactor.getInterpreter().poll();
 	reactor.getRTScheduler().execute();
 
@@ -111,13 +109,6 @@ void EventDistributor::deliverEvents()
 		}
 		eventsCopy.clear();
 	}
-}
-
-bool EventDistributor::sleep(unsigned us)
-{
-	std::chrono::microseconds duration(us);
-	std::unique_lock lock(cvMutex);
-	return condition.wait_for(lock, duration) == std::cv_status::timeout;
 }
 
 } // namespace openmsx
--- a/src/events/InputEventGenerator.cc
+++ b/src/events/InputEventGenerator.cc
@@ -47,7 +47,7 @@ void InputEventGenerator::wait()
 	}
 }
 
-void InputEventGenerator::poll()
+void InputEventGenerator::poll(std::optional<int> timeoutMs)
 {
 	// Heuristic to emulate the old SDL1 behavior:
 	//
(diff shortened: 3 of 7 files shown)


src/Reactor.cc
```diff
@@ -692,7 +692,7 @@ void Reactor::run()
 			// to also use a sleep/poll loop, with even shorter
 			// sleep periods as we use here. Maybe in future
 			// SDL implementations this will be improved.
-			eventDistributor->sleep(20 * 1000);
+			eventDistributor->sleep(8 * 1000); // 8ms
```
m9710797: The purpose of this statement is to avoid consuming 100% CPU when openMSX is paused.

This change seems fine, this only affects the `blocked` code path.

I did check the comment right above this statement. It talks about why we didn't use `SDL_WaitEvent()`. And the reason was that `SDL_WaitEvent()` also basically just did a `sleep()` (*1) when no events are available. And a shorter sleep than we do (20ms), so consuming more CPU.
(*1) This was based on a very old SDL version, possibly things have changed.

 Though this patch lowers the sleep from 20ms to 8ms, so maybe now using `SDL_WaitEvent()` make more sense? I (very quicky) looked at the SDL source code, and I think the implementation of `SDL_WaitEvent()` and `SDL_WaitEventTimeout()` has improved a lot. So maybe now we should switch to that. And you already suggested the same in #2024.

I haven't worked out the details yet. But here's a start (I might look at this myself later, but I first want to finish some other stuff):
* Reactor.cc:680: we call `EventDistributor::deliverEvents()`. 
* EventDistributor.cc:78: that calls `InputEventGenerator::poll()`.
* InputEventGenerator.cc:86: and that in turn calls `SDL_PollEvent()` in a loop. So this handles all pending events, but doesn't wait for new events to arrive.

Maybe we could optionally (maybe via a `bool` parameter, when `blocked`) replace the first call to `SDL_PollEvent()` with `SDL_WaitEventTimeout()`. As the `timeout` parameter we could maybe pass the minimum of `8ms` and the earliest timeout of our `RTScheduler` (see EventDistributor.cc:80 and RTScheduler.hh:25).

I believe our screen drawing (via `repaintDelayed()`) is scheduled via `RTScheduler`, so if we don't sleep past this time it should also improve the responsiveness. Or maybe a better (simpler??) alternative is to integrate our `RTScheduler` with the SDL event loop?

src/video/Display.cc
```diff
@@ -357,7 +357,7 @@ void Display::repaintImpl()
 
 	// TODO maybe revisit this later (and/or simplify other calls to repaintDelayed())
 	// This ensures a minimum framerate for ImGui
-	repaintDelayed(40 * 1000); // 25fps
+	repaintDelayed(16 * 1000); // 60fps
```
m9710797: I think this is wrong.

`Display::repaint()` is called on every paint, including the normal paints when openMSX is not paused.

But then this patch, this last statement, requests to draw another frame 16ms from now (thus 62.5fps). So when openMSX is running and the emulated MSX only produces 50 (or even 60) frames per second. This MSX frame never arrives in time to cancel this additional request for a delayed repaint. So this effectively doubles the frame rate when openMSX is running normally.

m9710797 (changes requested):
Thanks a lot for this PR!

Though I think it needs changes:
* The change in `Reactor.cc` is fine. But I've also given some ideas for further improvements (could be done in a later PR, possibly by me).
* But the change in `Display.cc` is wrong I think.

src/Reactor.cc
ej-sanmartin: Okay thanks for all the background and code locations, helped a ton with me understanding what's going on!  I'll propose what would need done to do move from `SDL_PollEvent()` -> `SDL_WaitEventTimeout()` and why I see some issues if we go with integrating `RTScheduler` with the SDL event loop.

---

## Propose Implementation ##

With SDL_WaitEventTimeout with min of 8ms and computed RTScheduler timeout, code path is clear.
`Reactor::run()` -> `EventDistributor::deliverEvents()` -> `InputEventGenerator::poll()` -> `SDL_PollEvent()` loop. 
`RTScheduler` is already executed in deliverEvents() right after `poll()`, so we have easy access to the next scheduled time.

So te change would be fairly localized:
- Expose `RTScheduler::getNextTime()` to query `queue.front().time`
- Add an optional timeout parameter to `InputEventGenerator::poll()`
- Replace the first `SDL_PollEvent()` call with `SDL_WaitEventTimeout(min(8ms, rtTimeout))`
- Subsequent events in the same batch continue using `SDL_PollEvent()` as before

**However, we need one addition to account for threads**

I noticed this all needs to be thread-safe so I looked into this more (had issues with SDL and threads in the past cuz I didn't account for it RIP). `SDL_WaitEventTimeout()` only wakes for SDL events and it won't wake when `EventDistributor::distributeEvent()` is called from another thread. Currently this works because `sleep()` uses a `condition_variable` that gets notified.

To preserve this behavior, we'd have `distributeEvent()` also push an `SDL_USEREVENT`:

```cpp
void EventDistributor::distributeEvent(Event&& event) {
    // exisiting code
    condition.notify_all();
    
    // Wake SDL_WaitEventTimeout if main loop is blocked
    SDL_Event wakeEvent{};
    wakeEvent.type = SDL_USEREVENT;
    SDL_PushEvent(&wakeEvent);

    // existing code
}
```

This is a minimal, targeted change. `SDL_PushEvent()` is documented as thread-safe, and we just need to ignore/consume the `SDL_USEREVENT` in `InputEventGenerator::handle()`.

---

## Why we shouldn't integrate `RTScheduler` into the SDL event loop  ##

It did sound conceptually cleaner ("one event system"), but I had a few concerns when looking and thinking about this:

- **Intertwining two timing systems**: `RTScheduler` uses `Timer::getTime()` (microsecond precision, monotonic clock) while `SDL` uses its own internal timing. Bridging these introduces potential for drift or race conditions. The difference in how we use these two timing systems in the other proposal is we're just using `RTScheduler`'s data to make a smarter decision about how long to sleep. The timing systems remain decoupled in their execution.

- **Synchronization complexity**: `RTScheduler`'s queue would need synchronization with `SDL` event pushing. Currently `RTScheduler::add()` is only called from the main thread, but tying it to SDL events could introduce subtle ordering issues, especially if an `RTSchedulable` reschedules itself during `executeRT()`.

- All of which, results in **debugging difficulty**: When timing issues arise, having two intertwined systems makes it much harder to isolate whether the problem is in SDL event delivery or `RTScheduler` execution.

---

Let me know what you think and if you need clarification (tried explaining the best I could haha) and I'd love to update these changes depending on what you think would be best for openmsx.

src/video/Display.cc
ej-sanmartin: Oh yeah, you're right. I was assuming this only occurs in the `blocked` state but it would actually effect actual MSX frame rate. Reverting then and we'll focus on the other comment.

src/Reactor.cc
m9710797: I mostly agree with you. Below are only some small additions or tweaks.

* I do agree with your proposal to pass an optional timeout to `InputEventGenerator::poll()`. Notice: `RTScheduler::getNextTime()` will return an absolute time-point (in microseconds?), while `SDL_WaitEventTimeout()` expects a relative time-duration (in milliseconds). But that shouldn't be a problem. Not sure if it's useful, but I just read the documentation of `SDL_WaitEventTimeout()` and saw that it's allowed to pass `nullptr` for the `event` parameter (to not consume the event), that might fit better in the existing code.
* Account for threads. You're right (not for correctness, but to improve responsiveness) we should post a `SDL_USEREVENT` to (let the main thread) return early from `SDL_WaitEventTimeout()`. You propose to do that in `EventDistributor::distributeEvent()` and that would cover 99% of the cases (possibly 100% in todays code?). I would suggest to move this to `Reactor::enterMainLoop()`. That method is already called from `distributeEvent()`, it's purpose is specifically to request the main thread to perform another iteration of its main loop (e.g. it also requests to exit the call to `MSXMotherBoard::execute()` (the main emulation code) earlier (when possible). And the extra `SDL_USEREVENT` is only needed when called from the non-main thread, so maybe we can add this extra code after line `Reactor.cc:618`?
* About integrating the event loops. I think it might still be a good idea. But your concerns are certainly valid. I'm a big proponent of making small incremental changes. So let's consider this out-of-scope for this PR.

ej-sanmartin: Awesome, and I'll check about the `SDL_WaitEventTimeout()` and passing `nullptr` to the `event` param. From a quick glance that may work but need to double check while implementing.

I'll also update my plan and ensure to add appropriate changes to `Reactor::enterMainLoop()`. And actually it looks like `Reactor.cc:618` may make more sense in this case.

Give me a few days, happy holidays! 🎄

ej-sanmartin:
I also removed `EventDistributor::sleep` as there was no more usage of it elsewhere btw!

ej-sanmartin:
Successfully built and tested the new changes on this PR

Demo, on non-break and break modes:

https://github.com/user-attachments/assets/45ec6544-8173-4b19-ada0-458cd2b37fa9

Feels snappy with no noticeable lag from imgui now.

Btw on `Windows 11`.

src/events/InputEventGenerator.cc
```diff
@@ -83,7 +83,23 @@ void InputEventGenerator::poll()
...
+	// For the first event, optionally wait with timeout. This allows
+	// efficient blocking when the emulator is paused, waking immediately
+	// on any SDL event (mouse, keyboard, etc.) or on timeout.
+	auto getNextEvent = [&, firstEvent = true]() mutable -> bool {
+		if (firstEvent && timeoutMs) {
+			firstEvent = false;
+			return SDL_WaitEventTimeout(curr, *timeoutMs) == 1;
+		}
+		return SDL_PollEvent(curr) == 1;
+	};
```
m9710797, line 95: Can we simplify this to something like:
```
auto getNextEvent = [&]{
    if (timeoutMs) {
        auto t = *timeoutMs;
        timeoutMs = {};
        return SDL_WaitEventTimeout(curr, t);
    }
    return SDL_PollEvent(curr);
};
```
In other words: reuse the `timeoutMs` parameter instead of introducing a new variable `firstEvent`.
Just an idea, and untested.

Or even:
```
auto getNextEvent = [&]{
    return timeoutMs ? SDL_WaitEventTimeout(curr, *std::exchange(timeoutMs, {}))
                     : SDL_PollEvent(curr);
};
```

src/events/InputEventGenerator.hh
```diff
@@ -37,7 +38,13 @@ public:
...
 	void updateGrab(bool grab);
 
-	void poll();
+	/** Poll for SDL events and dispatch them.
+	  * @param timeoutMs If provided, wait up to this many milliseconds for
+	  *        the first event using SDL_WaitEventTimeout(). Subsequent
+	  *        events in the same batch use SDL_PollEvent(). If nullopt,
+	  *        uses SDL_PollEvent() for all events (non-blocking).
+	  */
+	void poll(std::optional<int> timeoutMs = std::nullopt);
```
m9710797, line 47: Very much a detail, no need to change.
Our typical style is to write this as:
```
void poll(std::optional<int> timeoutMs = {});
```

src/events/EventDistributor.cc
```diff
@@ -113,11 +111,4 @@ void EventDistributor::deliverEvents()
 	}
 }
 
-bool EventDistributor::sleep(unsigned us)
-{
-	std::chrono::microseconds duration(us);
-	std::unique_lock lock(cvMutex);
-	return condition.wait_for(lock, duration) == std::cv_status::timeout;
-}
```
m9710797, line 121: This is a nice simplification, thanks.

src/Reactor.cc
```diff
@@ -676,23 +681,42 @@ void Reactor::powerOn()
...
-			// SDL_WaitEvent() function. Though when inspecting
-			// the implementation of that function, it turns out
-			// to also use a sleep/poll loop, with even shorter
-			// sleep periods as we use here. Maybe in future
-			// SDL implementations this will be improved.
-			eventDistributor->sleep(20 * 1000);
+			executionFailed = !copy->execute();
 		}
 	}
 }
```
m9710797, line 722: This seems correct, but maybe more complex than needed?

Here's my attempt to simplify it:
```
void Reactor::run()
{
	bool blocked = (blockedCounter > 0) || !activeBoard;
	while (running) {
		auto timeoutMs = [&] -> std::optional<int> {
            static constexpr int MAX_WAIT_MS = 8;
			if (!blocked) return {};
            // sleep if blocked, but not past next RT-event
			auto nextTime = getRTScheduler().getNextTime();
			if (!nextTime) return MAX_WAIT_MS;
			auto deltaUs = int64_t(*nextTime - Timer::getTime()); // might be negative
			return std::clamp(narrow<int>(deltaUs / 1000), 0, MAX_WAIT_MS);
		}();
		eventDistributor->deliverEvents(timeoutMs);
		blocked = (blockedCounter > 0) || !activeBoard; // re-evaluate
		if (!blocked) {
			// copy shared_ptr to keep Board alive (e.g. in case of Tcl
			// callbacks)
			auto copy = activeBoard;
			blocked = !copy->execute();
		}
	}
}
```

I have NOT tested this! So very well possible I overlooked something.
One functional change compared to your version is that I changed `std::max(1, ...)` to `std::max(0, ...` (and then collapsed the combination of `min`+`max` into `clamp`).

m9710797 (commented):
Thanks a lot for your PR. I have already merged it because it seems correct and it is already an improvement.

Nevertheless I do have some review comment. If you want to continue working on this. (But do double check my comments, I can make mistakes).

src/events/InputEventGenerator.cc
ej-sanmartin: Good intuition, I do like the `std::exchange()` as it's more C++ idiomatic and fits better for this use case.

src/events/InputEventGenerator.hh
ej-sanmartin: I'll add so it's concistent with rest of repo.

src/Reactor.cc
ej-sanmartin: This does look a lot cleaner and I was also somewhat worried about the initially being a bit complex than necessary and was trying to simplify it.

ej-sanmartin:
Thanks for the quick review and merge!

I'll create another PR that adds these suggestions and test this out on my end to confirm this still works.

PR #2060, opened 2025-12-26
fix: enum serialization with std::initializer_list breaks toolchain builds.

### LINKED ISSUE #985: Enum cannot be properly serialized on Visual Studio build?
When Hydragon saves his replay on his Visual Studio build (starting at aeffa19), he gets a corrupted enum serialized there. It's the CassttePlayerPluggable:

```
            <pluggable id="24" type="CassettePlayer" version="2">
              <casImage>
                <original/>
                <resolved/>
              </casImage>
              <checksum/>
              <tapePos>
                <time>0</time>
              </tapePos>
              <prevSyncTime>
                <time>0</time>
              </prevSyncTime>
              <audioPos>0</audioPos>
              <state>internal-error-unknown-enum-value</state>
              <lastOutput>false</lastOutput>
              <motor>false</motor>
              <motorControl>true</motorControl>
            </pluggable>
```

When he then tries to load this replay, he doesn't get the usual error, but openMSX closes on him.

He can't reproduce this on a build I made with MinGW-w64, only on a Visual Studio build.

If this happens for more people, it means reverse is quite broken on it...

But note that "Mart" had a very similar issue on OS X, at least he also got an invalid enum in his replay with openMSX 0.12.0, see #980 .

So, I guess something is not 100% right with enum (de)serialization.

ej-sanmartin:
Fixes #985 

---

**tldr;** storing `std::initializer_list` as data members caused undefined behavior on all toolchains, update to use modern C++20  `std::span` and `std::array`

Since we also have to update all call sites to init with `std::array`, started by splitting this PR into two commits:

1. Change `SERIALIZE_ENUM` macro to take in `std::span`
2. Update all 35 usage of `SERIALIZE_ENUM` across 23 files to pass in `std::array` to make lifetime unambigious

### DIFF
--- a/src/serialize_core.hh
+++ b/src/serialize_core.hh
@@ -79,25 +79,23 @@ template<typename T1, typename T2> struct SerializeClassVersion<std::pair<T1, T2
  *
  * The serialize_as_enum class has the following members:
  *  - static bool value
- *      True iff this type must be serialized as an enum.
+ *      True if this type must be serialized as an enum.
  *      The fields below are only valid (even only present) if this variable
  *      is true.
- *  - std::string toString(T t)
- *      convert enum value to string
- *  - T fromString(const std::string& str)
- *      convert from string back to enum value
+ *  - static constexpr std::span<const enum_string<T>> info()
+ *      Returns a view over the enum-to-string mapping table.
  *
  * If the enum has all consecutive values, starting from zero (as is the case
  * if you don't explicitly mention the numeric values in the enum definition),
  * you can use the SERIALIZE_ENUM macro as a convenient way to define a
  * specialization of serialize_as_enum:
-
+ *
  * example:
  *    enum MyEnum { FOO, BAR };
- *    std::initializer_list<enum_string<MyEnum>> myEnumInfo = {
+ *    static constexpr auto myEnumInfo = std::to_array<enum_string<MyEnum>>({
  *          { "FOO", FOO },
  *          { "BAR", BAR },
- *    };
+ *    });
  *    SERIALIZE_ENUM(MyEnum, myEnumInfo);
  *
  * Note: when an enum definition evolves it has impact on the serialization,
--- a/src/serialize_core.hh
+++ b/src/serialize_core.hh
@@ -10,8 +10,8 @@
 
 #include <array>
 #include <cassert>
-#include <initializer_list>
 #include <memory>
+#include <span>
 #include <string>
 #include <type_traits>
 #include <variant>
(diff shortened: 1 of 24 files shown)


ej-sanmartin:
It's failing checks on code duplication, passing threshold (5% when limit is 3%).

On `src/ide/WD33C93.cc:443-457` and `src/ide/MB89352.cc:734-748`:

```cpp
static constexpr auto phaseInfo = std::to_array<enum_string<SCSI::Phase>>({
	{ "UNDEFINED",   SCSI::Phase::UNDEFINED   },
	{ "BUS_FREE",    SCSI::Phase::BUS_FREE    },
	{ "ARBITRATION", SCSI::Phase::ARBITRATION },
	{ "SELECTION",   SCSI::Phase::SELECTION   },
	{ "RESELECTION", SCSI::Phase::RESELECTION },
	{ "COMMAND",     SCSI::Phase::COMMAND     },
	{ "EXECUTE",     SCSI::Phase::EXECUTE     },
	{ "DATA_IN",     SCSI::Phase::DATA_IN     },
	{ "DATA_OUT",    SCSI::Phase::DATA_OUT    },
	{ "STATUS",      SCSI::Phase::STATUS      },
	{ "MSG_OUT",     SCSI::Phase::MSG_OUT     },
	{ "MSG_IN",      SCSI::Phase::MSG_IN      },
});
SERIALIZE_ENUM(SCSI::Phase, phaseInfo);
```

`TODO` in `src/ide/MB89352.cc:734-748` indicates these were *already* duplicated prior to this change.

The "best" thing to do would be to have this defined in a separate file that gets imported into thes files, but idk where the best place to put this in the `src/ide/`diriectory or if to name it to something like `PhaseInfo.h` there. We could also skip check which would retain status quo of this nit `TODO` but maybe not be 100% ideal.

lmk what you think

---
Note *haven't* built these changes yet in windows visual studio to validate, I'll have to find some time after work tomorrow to properly set up. Lmk any pointers there as well and if you manage to have time that's fine by me!

m9710797:
> It's failing checks on code duplication, passing threshold (5% when limit is 3%).

You mean the SonarQube checks, right? We don't strictly always follow those. We try to fix those that make sense, but not all of them.

This particular one is open for interpretation. There is indeed some code duplication, and it could be factored out. But will that make the code easier to understand, easier to maintain? Personally I doubt it.

But good that the tool mentions it anyway, in other situations a similarly sized duplication could be useful to factor out.

src/serialize_core.hh
```diff
@@ -79,25 +79,23 @@ template<typename T1, typename T2> struct SerializeClassVersion<std::pair<T1, T2
  *
  * The serialize_as_enum class has the following members:
  *  - static bool value
- *      True iff this type must be serialized as an enum.
+ *      True if this type must be serialized as an enum.
```
m9710797, line 82: This was not a typo, we use [iff](https://www.merriam-webster.com/dictionary/iff) as an abbreviation of "if-and-only-if".

m9710797:
Merged thanks!

PR #2061, opened 2025-12-27
enhancement: Simplify Reactor::run() and InputEventGenerator timeout handling.

### LINKED ISSUE #2024: UI drawing delays when "breaked"

ej-sanmartin:
Follow up on #2059, which fixed #2024

---

Reviewer mentioned the following improvements. I verified these do not add regressions to the improvements on `break` mode imgui framerate / responsiveness.

- Streamline `Reactor::run()` branching logic
- Simplify `getNextEvent()` with `std::exhange` usage
- `poll(std::optional<int> timeoutMs = {})` instead of `= std::nulptr` to align with rest of codebase

### DIFF
--- a/src/Reactor.cc
+++ b/src/Reactor.cc
@@ -681,42 +681,25 @@ void Reactor::powerOn()
 
 void Reactor::run()
 {
-	static constexpr int MAX_WAIT_MS = 8;
-	bool executionFailed = false;
-
+	bool blocked = (blockedCounter > 0) || !activeBoard;
 	while (running) {
-		auto isBlocked = [&] {
-			return executionFailed || (blockedCounter > 0) || !activeBoard;
-		};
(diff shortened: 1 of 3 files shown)


ej-sanmartin:
Demo with these changes:

https://github.com/user-attachments/assets/d3649587-25cd-4bb7-8871-a7f38dabb708

m9710797:
Thanks.

PR #2067, opened 2026-01-06
New profiler that uses the probe/tracer viewer

pvmm:
image

### DIFF
--- a/share/scripts/profiler.tcl
+++ b/share/scripts/profiler.tcl
@@ -5,8 +5,8 @@
 
 namespace eval symboltracer {
 
+variable names {}         ;# user traces
 variable counters {}      ;# count recursive calls
-variable functions {}     ;# function entry points
 variable symbolfiles {}   ;# symbols file name
 
 set_help_proc symboltracer [namespace code symboltracer_help]
@@ -71,10 +71,11 @@ proc _exit_function {name} {
 
 proc _add {name addr} {
 	# add user-defined name/addr pair and create function breakpoints and traces
-	variable functions
-	if {![dict exists $functions $addr]} {
-		dict set functions $addr [debug breakpoint create -address $addr -command [namespace code "_enter_function {$name}"]]
+	variable names
+	if {![dict exists $names $name]} {
+		dict set names $name {}
 		debug trace add $name
+		debug breakpoint create -address $addr -command [namespace code "_enter_function {$name}"]
 	}
 }
 
@@ -94,10 +95,10 @@ proc _start {{file ""}} {
 		set symbols [debug symbols lookup]
 	} else {
 		# load symbol file
-		variable symbolfiles
-		lappend symbolfiles $file
 		debug symbols load $file
 		set symbols [debug symbols lookup -filename $file]
+		variable symbolfiles
+		lappend symbolfiles $file
 	}
 	if {[llength $symbols] == 0} {
 		error "no symbols found"
@@ -107,15 +108,15 @@ proc _start {{file ""}} {
 
 proc _stop {} {
 	variable counters
-	set $counters {}
+	set counters {}
 
-	variable functions
-	foreach {symbol bp} $functions {
-		catch {debug trace drop $symbol}
+	variable names
+	dict for {name _} $names {
+		debug trace drop $name
 	}
-	set functions {}
+	set names {}
 
-	foreach {bp entry} [debug breakpoint list] {
+	dict for {bp entry} [debug breakpoint list] {
 		if {[string match "* [namespace current] *" [dict get $entry "-command"]]} {
 			catch {debug breakpoint remove $bp}
 		}
@@ -129,7 +130,7 @@ proc _stop {} {
 }
 
 proc _dispatcher {args} {
-	set params "[lrange $args 1 end]"
+	set params [lrange $args 1 end]
 	set cmd [lindex $args 0]
 	switch -- $cmd {
 		start   { return [_start {*}$params] }
--- a/share/scripts/profiler.tcl
+++ b/share/scripts/profiler.tcl
@@ -1,0 +1,146 @@
+# symboltracer: a profiler for the Probe/Trace viewer
+#
+# symboltracer uses the new Probe/Trace viewer as a profiler and symbol files
+# to register functions.
+
+namespace eval symboltracer {
+
+variable counters {}      ;# count recursive calls
+variable functions {}     ;# function entry points
+variable symbolfiles {}   ;# symbols file name
+
+set_help_proc symboltracer [namespace code symboltracer_help]
+proc symboltracer_help {args} {
+	if {[llength $args] == 1} {
+		return {symboltracer is a profiler that traces function calls in the Trace Viewer widget
+
+Recognized commands: start, stop, add
+
+Type 'help symboltracer <command>' for more information about each command.
+}
+    }
+	switch -- [lindex $args 1] {
+		"start" { return {Opens a symbol file and start tracing session
+
+'start' reads a symbol file (like a .noi or .sym file) and creates breakpoints and traces for each entry found.
+
+Syntax: symboltracer start [<sym-file>]
+}}
+		"add" { return {Creates breakpoint and trace for a user-defined name and address
+
+'add' inserts a user-defined breakpoint and trace. No previously defined symbol is necessary. This is most useful for quick tracing sessions.
+
+Syntax: symboltracer add <name> <address>
+}}
+		"stop" { return {Interrupts symboltracer session
+
+'stop' removes all breakpoints and traces associated with symboltracer and all files included with the 'start' command.
+
+Syntax: symboltracer stop
+}}
+	}
+}
+
+proc _enter_function {name} {
+	# find function return address
+	set retaddr [peek16 [reg SP]]
+	# check if we create caller breakpoint for CALL or RST
+	if {[peek [expr {$retaddr - 1}]] in "0 8 16 24 32 40 48 56" ||
+	    [peek [expr {$retaddr - 3}]] in "196 204 205 212 220 228 236 244 252"} {
+		variable counters
+		dict incr counters $name
+		set count [dict get $counters $name]
+		debug trace add $name $count
+		debug breakpoint create -once true -address $retaddr -command [namespace code "_exit_function {$name}"]
+	}
+}
+
+proc _exit_function {name} {
+	variable counters
+	# ignore a probable dangling breakpoint from a previous session
+	if {![dict exists $counters $name]} { return }
+	dict incr counters $name -1
+	set count [dict get $counters $name]
+	if {$count == 0} {
+		dict unset counters $name
+		debug trace add $name
+	} else {
+		debug trace add $name $count
+	}
+}
+
+proc _add {name addr} {
+	# add user-defined name/addr pair and create function breakpoints and traces
+	variable functions
+	if {![dict exists $functions $addr]} {
+		dict set functions $addr [debug breakpoint create -address $addr -command [namespace code "_enter_function {$name}"]]
+		debug trace add $name
+	}
+}
+
+proc _add_symbol_set {symbols} {
+	# run through collection of symbols
+	foreach entry $symbols {
+		dict with entry {
+			_add $name $value
+		}
+	}
+}
+
+proc _start {{file ""}} {
+	set symbols {}
+	if {$file eq ""} {
+		# use current symbols only
+		set symbols [debug symbols lookup]
+	} else {
+		# load symbol file
+		variable symbolfiles
+		lappend symbolfiles $file
+		debug symbols load $file
+		set symbols [debug symbols lookup -filename $file]
+	}
+	if {[llength $symbols] == 0} {
+		error "no symbols found"
+	}
+	_add_symbol_set $symbols
+}
+
+proc _stop {} {
+	variable counters
+	set $counters {}
+
+	variable functions
+	foreach {symbol bp} $functions {
+		catch {debug trace drop $symbol}
+	}
+	set functions {}
+
+	foreach {bp entry} [debug breakpoint list] {
+		if {[string match "* [namespace current] *" [dict get $entry "-command"]]} {
+			catch {debug breakpoint remove $bp}
+		}
+	}
+
+	variable symbolfiles
+	foreach file $symbolfiles {
+		debug symbols remove $file
+	}
+	set symbolfiles {}
+}
+
+proc _dispatcher {args} {
+	set params "[lrange $args 1 end]"
+	set cmd [lindex $args 0]
+	switch -- $cmd {
+		start   { return [_start {*}$params] }
+		add     { return [_add   {*}$params] }
+		stop    { return [_stop  {*}$params] }
+		default { error "Unknown command '[lindex $args 0]'." }
+    }
+}
+
+}
+
+proc symboltracer {args} {
+	symboltracer::_dispatcher {*}$args
+}


share/scripts/profiler.tcl
```diff
@@ -0,0 +1,98 @@
...
+			set symbols [debug symbols lookup]
+		}
+		1 {
+			lassign $args symbolfile
+			debug symbols load $symbolfile
+			set symbols [debug symbols lookup -filename $symbolfile]
+		}
+		default {
+			error "symbols file name expected"
+			return
```
m9710797: `error` already exits from the current proc (with an error status). So the following `return` statement isn't needed.
Here and in some other places.

share/scripts/profiler.tcl
```diff
@@ -0,0 +1,98 @@
...
+
+proc _enter_function {symbol} {
+	# find function return address
+	set retaddr [peek16 [reg SP]]
+	# check if we create caller breakpoint
+	set caller [lindex [debug disasm [expr {$retaddr - 3}]] 0]
+	if {[string match "call*" $caller]} {
+		debug trace add $symbol 1
+		variable returns
+		dict set returns $symbol [debug breakpoint create -once true -address 0x[h $retaddr] -command [list symboltracer::_exit_function $symbol]]
```
m9710797: It's not required to pass a value formatted as hex to the `-address` option, any integer (in any format) is fine.
So this can be simplified to just `-address $retaddr`. Here and in some other place. And then the `h` helper proc isn't needed anymore.

m9710797: `[list symboltracer::_exit_function $symbol]` is correct. Though it's more common to write this as:
`[namespace code "_exit_function $symbol"]`.
You pass a script to `namespace code` with the syntax you'd use in the current context (in this case in the `symboltracer` namespace) and it returns the transformed script that then can be invoked from any context (in this case breakpoint commands are invoked from the global namespace).

share/scripts/profiler.tcl
```diff
@@ -0,0 +1,98 @@
...
+	}
+}
+
+proc _exit_function {symbol} {
+	debug trace add $symbol 0
+	variable returns
+	dict unset returns $symbol
```
m9710797: The only reason to have this `returns` variable is to be able to cancel pending breakpoints in `stop`, right?

But I think it's wrong in case of recursive functions (calling a function directly or indirectly while another invocation of the same function is still active). Because the key in this dict is just `$symbol` you cannot have the same function present more than once.

I can't immediately think of a good way to fix this. Maybe the best way is to not have `returns` at all? And then in `stop` walk over all breakpoints and clear those with a `-command` that contains `_exit_function`??? `stop` is only run occasionally, so it's fine that it's a bit slower (walks over _all) breakpoints).

Note: I have not tried this. I can be wrong.

share/scripts/profiler.tcl
```diff
@@ -0,0 +1,98 @@
...
+}
+
+proc start {args} {
+	puts "Symbol tracer started"
+	# load symbol file
+	variable symbolfile {}
+	switch -- [llength $args] {
```
m9710797: This is correct, and future proof if you later want more complex argument parsing.
But for this specific (simple) case you could also write it like:
```
proc start {{file ""}} {
    if {$file eq ""} {
       # no filename passed
    } else {
       # here we do got a filename
    }
}
```
And you automatically get a nice error message when you pass more then 1 argument.

share/scripts/profiler.tcl
```diff
@@ -0,0 +1,98 @@
...
+}
+
+proc _exit_function {symbol} {
+	debug trace add $symbol 0
+	variable returns
+	dict unset returns $symbol
+}
+
+proc start {args} {
+	puts "Symbol tracer started"
```
m9710797: In general I don't like procs that print stuff. If you must show some message then use `return "my message"` instead of `puts`.

I'll explain via an example. Suppose that our `debug breakpoint remove` function would print `Succesfully removed breakpoint`. That could be nice when you interactively type that command. But when you use that function in a non-interactive context (e.g. like here in the `stop` command) then all those messages are useless and annoying. If instead the command returns a string, then on interactive use that string is still printed, but when called from another script that script can choose to either ignore that resulting string or return (or print) it.

So for sure replace `puts` with `return` and move it to the end. But is that even needed at all? When you interactively type `symboltracer::start` and there were no errors, do you even need this message?  (Ah and currently in case of error the message was already printed, that's another reason to move it to the end).

If you want to print/return something, maybe it's more informative to list which symbols will be traced? But personally I wouldn't print/return anything. Both here in `start` and in `stop`. (Only report errors, be silent on success).

share/scripts/profiler.tcl
```diff
@@ -0,0 +1,98 @@
...
+proc h {number} {
+	return [format %04x $number]
+}
+
+proc _enter_function {symbol} {
+	# find function return address
+	set retaddr [peek16 [reg SP]]
+	# check if we create caller breakpoint
+	set caller [lindex [debug disasm [expr {$retaddr - 3}]] 0]
+	if {[string match "call*" $caller]} {
```
m9710797: Nothing wrong with the Tcl implementation itself here. But I do want to discuss the reason for this check. E.g. if you insist on `call` then tail-calls (implemented via `jp` or `jr`) don't work. Ah and maybe `rst` should be supported as well? Though that's difficult because this is a mix of 1, 2 and 3 byte instructions. So can't correctly distinguish these in all cases.

But I don't see why you need to have this check at all. If `_enter_function` got called that must mean a breakpoint triggered, and that means a symbol a traced got called (in some way, call, jp, rst, whatever).

In case of tail-calls (`jp` instead of `call`) a single `ret` instruction will leave two (or more) functions at once. And this implementation will then register multiple `debug breakpoint -once ...` breakpoints, so it will correctly mark all with `debug trace add ... 0`.

share/scripts/profiler.tcl
```diff
@@ -0,0 +1,98 @@
...
+	}
+}
+
+proc _exit_function {symbol} {
+	debug trace add $symbol 0
+	variable returns
+	dict unset returns $symbol
+}
+
+proc start {args} {
```
m9710797: With the current script you must always have a symbol file to start from. In general that's a good idea. But sometimes I just want to trace one or a few functions and I may not have a symbol file ready.

Do you think it would be possible to factor out a helper function to setup such a single trace? And then the current `start` function makes use of that helper.

m9710797 (commented):
Very nice! Thanks a lot for working on this.

I assume this means you like the new trace functionality? It's very new and hasn't been tested yet very well. Did you find any bugs? Any suggestions for improvements? Usability or new (small) features?

I reviewed your PR. Overall it's very good. But I did make several small suggestions for improvements/tweaks. One bigger question about why check for `call` (because now tail-calls don't work). And one request to split this functionality is smaller reusable parts.

share/scripts/profiler.tcl
pvmm: In some situations, a symbol is not a function, but `_enter_function` will still be called (because it is a dynamically created block of code named `A$something$123`). Since it's not a function the stack doesn't contain a return address, but something else (probably a pointer to a variable or a literal value). So the `_exit_function` is never called and it keeps hanging there forever in the breakpoint list.

pvmm: So only CALLs are guaranteed to have a valid return address on the stack.

pvmm: I've added RST also since it adds a return address to the stack.

share/scripts/profiler.tcl
```diff
@@ -0,0 +1,90 @@
...
+
+variable functions {}     ;# function entry points
+variable symbolfiles {}   ;# symbols file name
+
+proc _enter_function {symbol} {
+	# find function return address
+	set retaddr [peek16 [reg SP]]
+	# check if we create caller breakpoint
+	set caller [lindex [debug disasm [expr {$retaddr - 3}]] 0]
+	if {[string match "call*" $caller] || [string match "rst*" $caller]} {
```
m9710797: Unfortunately this check on `rst` doesn't work. The reason is: `CALL` is a 3 byte instruction, so you indeed have to look for it at address `$retaddr - 3`. But `rst` is a single byte instruction. You could look for it at address `$retaddr - 1`, but then there is a chance (8 out of 256) that this byte is the 2nd, 3rd, 4th byte from a multi-byte instruction that happens to be one of the 8 possible `rst` instructions. That's what I mean before with "you cannot unambiguously decode the previous instruction".

Actually also your check for CALL can go wrong. The previous 4 bytes could have been:
```
 3E CD     ld a,#cd
 06 23     ld b,#23
```
When you decode that starting from the 3rd previous byte you get `CALL #2306`, but that was not the actual instruction sequence. 

But this is anyway only an heuristic, maybe it's fine if it doesn't always work???

---

BTW instead of doing `debug disasm` it's a lot faster to check this via something like:
```
if {[peek [expr {$retaddr - 3}]] in "196 204 205 212 220 228 236 244 252"} { ...
```
So check the 9 possible starting bytes for `call` or `call <cc>`.

share/scripts/profiler.tcl
```diff
@@ -0,0 +1,90 @@
...
+		debug trace add $symbol 1
+		debug breakpoint create -once true -address $retaddr -command [namespace code "_exit_function {$symbol}"]
+	}
+}
+
+proc _exit_function {symbol} {
+	debug trace add $symbol 0
```
m9710797: Maybe this solution is good enough. But it doesn't correctly visualize recursive function calls. Suppose we have something like:
* myFunction, calls (possibly indirectly) myFunction again.
* Only then we leave the inner and later the outer myFunction call.

Then:
* We trace: `debug trace add myFunction 1`
* Followed by again: `debug trace add myFunction 1`, but this doesn't show up in the graph because the value was already `1`.
* Then we leave the inner function and do `debug trace add myFunction 0`. Now the traced signal already goes low, but the outer call is still active.
* Finally we again do `debug trace add myFunction 0`, but visually this also doesn't show up.

I don't know how common recursive function calls are on MSX (in general it's relatively common). So maybe in the first version of this script we don't need to solve it?

---

A possible solution could be to not trace a binary signal 0/1, but a counter. So counting the number of recursive invocations. And then specifically for the openMSX visualization it will look nicer if instead of `0` you trace the `none` value (with `debug trace myFunction`, so _without_ a value argument).

share/scripts/profiler.tcl
pvmm: > When you decode that starting from the 3rd previous byte you get `CALL #2306`, but that was not the actual instruction sequence.
> 
> But this is anyway only an heuristic, maybe it's fine if it doesn't always work???

I don't think that would impact this script very much, because:
* a false positive will only result in a dangling breakpoint that the user will probably ignore anyway if they don't recognise it, since it's not a function but a block of code with some weird name like `A$something$number` generated by SDCC;
* it should be somewhat rare because two simultaneous conditions are required: SP has to point to a position that resembles a CALL instruction prefix (or byte containing 00, 08, 10, 18, 20, 28, 30 or 38 for RST) and the sym file has to point to an instruction where the former happens;
* The viewer has names disabled by default, so the user must select which names they are interested in. They won't select a name that they don't know;

share/scripts/profiler.tcl
pvmm: > A possible solution could be to not trace a binary signal 0/1, but a counter. So counting the number of recursive invocations. And then specifically for the openMSX visualization it will look nicer if instead of `0` you trace the `none` value (with `debug trace myFunction`, so _without_ a value argument).

But I still have to store the previous value of trace, probably in a dictionary. So I use  `debug trace add [expr {[dict get $counter $symbol] + 1}]` instead of `debug trace add 1` in `_enter_function`, right?

m9710797: Yes, somewhere you'll need to keep track of the recursion count.

BTW for me it's fine to already merge the current version. And then leave enhancements like recursive functions and tail-calls for a later version.

share/scripts/profiler.tcl
```diff
@@ -0,0 +1,92 @@
...
+}
+
+proc add {symbols} {
```
m9710797: Thanks for splitting the original `start` proc into `start` + `add`. Though it's not yet fully what I had in mind. I'll try to explain better.

My goal was to make this feature also useful for small one-off traces when you don't have a symbol file at hand. What I can do with this version is:
```
symboltracer::add {{name chget value 0x009f} {name chput value 0x00a2}}
```
But that's not easy to use (or easy to remember, or figure out and it's not documented). For me it would be nicer if I could (also) do something like:
```
symboltracer::add chget 0x009f
symboltracer::add chput 0x00a2
```
And then maybe rename the current `add` proc to `add_symbol_set` (and OK for me to then make it an internal non-documented) proc).
And if I don't know/remember the syntax, I can execute just `symboltracer::add`, and then I'll get a helpful error message like: `wrong # args: should be "symboltracer::add name value".

Ah, maybe it would also be nice if you could implement help for this script (see `set_help_text` in our other scripts). Maybe just turn the comments at the top of this script into the help text (fine if all symboltraces::* procs share this single help text, or feel free to write specific text for each).

share/scripts/profiler.tcl
```diff
@@ -16,36 +16,43 @@
...
-		debug trace add $symbol 1
-		debug breakpoint create -once true -address $retaddr -command [namespace code "_exit_function {$symbol}"]
-	} elseif {[peek [expr {$retaddr - 3}]] in "196 204 205 212 220 228 236 244 252"} {
-		debug trace add $symbol 1
-		debug breakpoint create -once true -address $retaddr -command [namespace code "_exit_function {$symbol}"]
+	# check if we create caller breakpoint for CALL or RST
+	if {[peek [expr {$retaddr - 1}]] in "0 8 16 24 32 40 48 56" ||
+	    [peek [expr {$retaddr - 3}]] in "196 204 205 212 220 228 236 244 252"} {
+		variable counter
+		dict set counter $name [set count [expr {[dict getdef $counter $name 0] + 1}]]
```
m9710797: `dict getdef` is a Tcl-9 feature. ATM openMSX should still work with Tcl-8.6 as well.
But you can rewrite this as:
```
dict incr counters $name
set count [dict get $counters $name]
```
If the key did not yet exist in the dict, then `dict incr` acts as-if it did exist with value 0.

share/scripts/profiler.tcl
```diff
@@ -16,36 +16,43 @@
...
+		if {$count > 1} {
+			debug trace add $name $count -type int
+		} else {
+			debug trace add $name $count
+		}
```
m9710797: It is allowed to repeat the `-type int` part in `debug trace add` (and because it's handled in c++, it's probably faster than doing this check in Tcl).

share/scripts/profiler.tcl
```diff
@@ -16,36 +16,43 @@
...
 	}
 }
 
-proc _exit_function {symbol} {
-	debug trace add $symbol 0
+proc _exit_function {name} {
+	variable counter
+	# ignore it: probably a dangling breakpoint from a previous session
+	if {![dict exists $counter $name]} { return }
+	dict set counter $name [set count [expr {[dict get $counter $name] - 1}]]
```
m9710797: Here we can also use `dict incr`:
```
dict incr counters $name -1
set count [dict get $counters $name]
```

share/scripts/profiler.tcl
```diff
@@ -16,36 +16,43 @@
...
+	debug trace add $name $count
```
m9710797: Above I suggested to not have separate `debug trace add` calls (for speed). How I do recommend to have separate calls for count=0 and !=0:
```
if {$count == 0} {
    debug trace add $name
} else {
    debug trace add $name $count
}
```
 The reason is that no-value is displayed differently (nicer?) than the value 0.

In the `$count == 0` branch you might want to do a `dict unset counters $name`. Normally this shouldn't be needed. But it helps with the "dangling breakpoint protection stuff" you do on line 43.

share/scripts/profiler.tcl
```diff
@@ -16,36 +16,43 @@
...
 }
 
-proc add {symbols} {
-	# scan symbols and create function breakpoints
+proc add_symbol_set {name addr} {
```
m9710797: I think my earlier review comment might have been unclear:
* `add_symbol_set` would be a good name for a proc that takes a set of symbols (the old implementation of `add`).
* Just `add` would be a good (short, convenient) name for this implementation (which takes a single name/value pair, not a whole set).

share/scripts/profiler.tcl
```diff
@@ -16,36 +16,43 @@
...
-		dict with entry {
-			debug trace add $name 0
-			if {![dict exists $functions $value]} {
-				dict set functions $value [debug breakpoint create -address $value -command [namespace code "_enter_function {$name}"]]
-			}
-		}
+	if {![dict exists $functions $addr]} {
+		variable counter
+		dict set counter $name 0
+		debug trace add $name 0
```
m9710797: See earlier comments:
* If you use `dict incr` (or also `dict getdef` in Tcl-9) there's no need to initialize the counter to 0.
* I would suggest to not use `0` but no-value for "function-is-not-active".

m9710797 (commented):
Thanks for the improvements. Good progress (but needs some fixes now).
After those fixes we could already merge this as a first version.

Or you could immediately continue with implementing tail calls (that is when the function is entered via a `JP` or `JR` instruction instead of `CALL`). I haven't worked out all the details, but _I think_ the only thing this needs is extending the check for `CALL` or `RST`  to also allow `JP` or `JR` (the other logic to increment/decrement the counter should be fine).

share/scripts/profiler.tcl
```diff
@@ -0,0 +1,160 @@
...
+    }
+}
+
+}
+
+proc symboltracer {args} {
+	if {[catch {set result [symboltracer::dispatcher {*}$args]} msg]} {
+		error $msg
+	}
+	return $result
```
m9710797: Did I miss something? You catch a possible error, only to re-issue that same error. And you capture the result, only to return the same result. Isn't this fully the same as just: `symboltracer::dispatcher {*}args`?
In other words, why not inline the implementation of `dispatcher` here?

share/scripts/profiler.tcl
```diff
@@ -0,0 +1,160 @@
...
+# Commands:
+# symboltracer start
+# 	* Create breakpoints and traces from each of the already loaded symbols
+# symboltracer start ?<sym-file>?
+# 	* Load symbol file and create breakpoints and traces for each function included in the symbol file
+# symboltracer add <name> <value>
+# 	* Insert an user-defined name/value pair and create breakpoint and trace for each pair
+# symboltracer stop
+# 	* Delete all breakpoints and traces and remove all symbol files included with 'start'
+#
```
m9710797, line 2: This comment block fully overlaps with the help text, no?
Maybe remove it? It's redundant (will only get out of sync with the help text when the script evolves)

m9710797 (commented):
Thanks again. I think it's close to ready.
Let's leave tail-calls for a follow-up PR.

pvmm:
> I haven't worked out all the details, but _I think_ the only thing this needs is extending the check for `CALL` or `RST` to also allow `JP` or `JR` (the other logic to increment/decrement the counter should be fine).

I don't think it's necessary to worry about tail calls. Consider this for instance:

If PC is at 0x8000 and there it calls function `f9000` at 0x9000. This is the result in the stack:
8003 <- top of the stack
xxxx
(...)

If somewhere above 9000, a tail call (JP) is called to function `fA000` at A000, the top of the stack still has to be 8003 if `fA000`'s RET instruction is supposed to return to 8003.

As it is, before returning from `fA000`, `_enter_function` is called twice (for both `f9000` and `fA000`, because top of the stack is still at the same CALL-address+3) and so `_exit_function` breakpoint is also registered twice, so their counters will be properly updated.

share/scripts/profiler.tcl
```diff
@@ -0,0 +1,146 @@
...
+
+	variable symbolfiles
+	foreach file $symbolfiles {
+		debug symbols remove $file
+	}
+	set symbolfiles {}
+}
+
+proc _dispatcher {args} {
+	set params "[lrange $args 1 end]"
```
m9710797: I overlooked this in my previous review(s), sorry about that.
The double quotes here aren't needed. I think from a functional point of view it's not wrong, but it's unnecessary (and slower):
* The `lrange` command already returns a proper list.
* Putting quotes around it, in principle, doesn't change anything. Though internally the list is now converted to string representation.
* Later on `{*}$params` then re-parses that string then back into a list.

share/scripts/profiler.tcl
```diff
@@ -0,0 +1,146 @@
...
+	}
+	if {[llength $symbols] == 0} {
+		error "no symbols found"
+	}
+	_add_symbol_set $symbols
+}
+
+proc _stop {} {
+	variable counters
+	set $counters {}
```
m9710797: Typo: should be `set counter {}` (remove the `$`).

share/scripts/profiler.tcl
```diff
@@ -0,0 +1,146 @@
...
+
+	variable functions
+	foreach {symbol bp} $functions {
```
m9710797: No need to change. This line works perfectly fine.

A good alternative is: `dict for {symbol bp} $functions {`:
* This shows we're iterating over a dict rather than a list of an even number of elements.
* Internally this doesn't require building a list-representation of the dict.

share/scripts/profiler.tcl
```diff
@@ -0,0 +1,146 @@
...
+		catch {debug trace drop $symbol}
```
m9710797: Not wrong, no need to change.

The `catch` isn't needed. I intentionally did NOT make it an error to drop a trace that doesn't exist. But then later for `debug breakpoint remove` (line 120) the `catch` is needed. That's not very consistent, I mean the openMSX implementation is not consistent in the behavior of different debug sub commands, I should probably fix that: remove the error from `debug breakpoint remove`.

So for now maybe keep the `catch`, both here and on line 120.

share/scripts/profiler.tcl
```diff
@@ -0,0 +1,146 @@
...
+proc _start {{file ""}} {
+	set symbols {}
+	if {$file eq ""} {
+		# use current symbols only
+		set symbols [debug symbols lookup]
+	} else {
+		# load symbol file
+		variable symbolfiles
+		lappend symbolfiles $file
+		debug symbols load $file
```
m9710797, line 98: Minor tweak. Suppose `debug symbols load` goes wrong, then we nevertheless already added `$file` to `symbolfiles`. Better swap the order of `debug symbols load` and `lappend`.

m9710797 (commented):
I did a (hopefully last) detailed review. I caught some details that I missed in my previous reviews (sorry about that). None of my remarks are really critical.

I do agree with your analysis of tail-calls. It should already work fine.

pvmm:
A `functions` dictionary is not really necessary, since breakpoints are deleted by string matching with `-command`. Besides that, stopping breakpoint creation because of a repeated address stops names from being properly registered and counted.

m9710797:
Thanks, merged!

If you want to continue working on this, here are some more ideas you could try:
* Currently you have a `names` dict (for which only the keys are relevant, not the values). I think it should be possible to merge this with the `counters` dict (and then initialize the counters dict to 0, and don't unset the key when the counter value drops down to zero).
* You could implement the current `_dispatcher` stuff with a [namespace ensemble](https://wiki.tcl-lang.org/page/namespace+ensemble). I only relatively recently learned about this myself, so many other of our scripts also don't (yet) use this.

Issue #1579, opened 2024-02-03
Support emulation of plotter (like Sony PRN-C41)


MBilderbeek:
There are several (very similar) MSX plotters out there and there has been made some software for them as well. E.g. by Sony but also a few programs released by "Lunatic" (like Plotterdisk I and II). 

Would be fun to emulate such plotters and let the output be an SVG file :)

The command codes are available online in the Service Manual: https://archive.org/details/sonyprnc41sm/

sndpl:
@MBilderbeek I think this issue can be closed?

MBilderbeek:
Yes, this has been implemented via #2062 - THANKS AGAIN! Next time, please include a Closes #1579 thing, so it is cleaned up automatically :)

PR #2069, opened 2026-01-11
feat: add plotter support

sndpl:
Add support for a MSX plotter based on the Sony - PRN-C41 (4 color pens). It supports the following:

- all commands that are mentioned in the service manual are supported
- all character sets are supported (International, Din, Japanese)
- supports 2 pen sizes (PRK-C41  small and PRK-C42 big).
- supports the dip switch 4 that you can turn off (default on, unless the software says that you need to turn it off)
- added a plotter preview screen, so you can see what's its plotting
- supports the change pen, up and down buttons in the plotter preview screen.
- added an additional eject page button for when you drawing is completed (else you had to unplug the plotter before it writes the image) on the plotter preview screen.
- added a plotter settings menu under settings -> output

What is not supported is the additional Kanji ROM  (level 1) that you can add in the Plotter. Couldn't find any info about his and how to print those Kanji chars. Also it only supports A4 paper, you can plot to A5 for example, but then it will just print an smaller image on the A4 paper.

Plotter viewer:
image
The red cross is the plotter pen.

Plotter settings:
image

Test software:
https://download.file-hunter.com/Programs/Plotterdisk%20I%20(Lunatic).zip
https://download.file-hunter.com/Programs/Plotterdisk%20II%20(Lunatic).zip
https://hansotten.file-hunter.com/uploads/files/sonyprnc41sm.pdf (Basic programs in the manual)

Some extra plotter info:
The Sony Plotter is also sold as the General PCP-5501 and Toshiba HX-P570. 
There is also the Daewoo DPL-400 and Talent DPL-400, almost the the same commands but not 100% compatible. Could be emulated, but couldn't find any software for it.
And there is the National CF-2311/Panasonic KX-08P, this plotter also has an other command set. There is commercial software for it from National, but not manuals of these plotters are available at the moment.

[test.zip](https://github.com/user-attachments/files/25325609/test.zip)
Test disk with 4 test programs on it, for each program also a png is added (the snatch is drawn with a magnification of 5).

[auto-demo.zip](https://github.com/user-attachments/files/25349469/auto-demo.zip)
Disk images that runs all tests scripts at once, also includes a tcl script, run with: `openmsx -script openmsx.tcl -diska ./plotter-demo.dsk`


m9710797:
Nice!

I understand this PR is still in draft. But I do already have questions, and some very high level review feedback.
* Can you point us to example(s) of MSX software that uses this. So that we can (help) test this.
* May I ask that you separate the commits into: 1) purely extract the charset and paper stuff from the printer files, 2) only then make changes.  Because now the diff is hard to read: it's not immediately clear which parts are only extracted and which parts are new or modified.
* I was surprised to see an internal method called `drawLineBresenham`. "Bresenham" is a (integer-only) line rasterization algorithm (split a line in staircase segments). While the key property of a plotter is that it can draw "real" lines (no staircase effect), no? But then when I look at the implementation of this function, I see it's not using Bresenham. The implementation is perfectly fine, only the name of the function was a bit misleading (to me).
* I see that for drawing lines you approximate that by a dense series of dots. Probably that's fine, I didn't actually test it yet. But then without having seen the result, my concern is: doesn't that remove the anti-aliasing effect of the dots? I mean currently the dots are drawn in various shades of grey, if you overdraw multiple dots close to each other then the colors get darker. So depending on how dense you make the dots for your line drawing you get more overdraw and thus darker colors. I'm just wondering if you took that into account? If the current result looks OK, then perfectly fine. Or else we should extend the Paper class itself with a line-drawing function? To better see the effect of this, you could try increasing the `print-resolution` setting.

MBilderbeek:
See also #1579.

m9710797:
Just for fun I started playing with code to rasterize wide lines to emulate the behavior of ink on paper:
* Lines have a semi-circle start and end cap.
* Overdrawing gives a darker color.
* So if you draw a closed loop (e.g. a square or a hexagon), the start/stop corner is visited twice and slightly darker than the other corners.
* If 3 lines meet in a point (e.g. a square with diagonals) then that point is also slightly darker because of overdraw.
* If you draw 2 line segments that meet at a very sharp angle, then the lines close to the corner overlap and are drawn darker.

I've written some proof of concept code (just to play around with the idea, not yet in a form that's 100% suited for openMSX):
* You can find the code here: https://godbolt.org/z/v8v4eq1sz
* Copy paste that code and compile locally, it needs zlib and libpng. So compile with something like: `g++ lines.cpp -Wall -Wextra -O3 -lpng -lz -std=c++20`
* If you execute the resulting program, it will generate a file `output.png`.
* Open that PNG file and zoom in on interesting features.

Of course this is only a test program. It intentionally draws wide lines and starting with not fully black color, so that overdrawn is more clearly visible.
Drawing dots is also possible, so it could also be used for the dot-matrix printer emulation. It uses a very different (simpler, more accurate, but slower(?)) approach than the current openMSX code in the `Paper` class. Though that code was written 18(?) years ago, and at that time the quality/speed tradeoff was still different.

Maybe (later) this code could be reworked and be used for the plotter rendering code. Of course MBilderbeek's idea of rendering to SVG is also nice.

sndpl:
hi @m9710797 , thanks for the high level review :) Here some answers on your questions:

1) I updated the description of the PR with some links to software (will add more when it's available)
2) Sure, I only extracted the Charset and Paper from the printer class because the plotter also needed them. And in the Charset I added the Japanese front and Din (german) font.
3) Ok, have to admin this was an idea of Gemini. I asked for a routine that could draw those lines and it came with that function. But it didn't do what I wanted and after some testing and tweaking it because an other implementation :) I have renamed the function in my latest commits.
4) I haven't thought about the fact that if you print multiple lines over the same spot that it should become darker.  Can have a look at it, but maybe better to this in a separate PR.

Currently this PR supports all functionality that I wanted to build to have some (basic) Plotter support. It started as a fun challenge because I got a cartridge that need the plotter and I was just wondering how this would work and how it looks. 
And thanks to the autocomplete it became a lot easier to implement this and luckily also the docs where available on how the Sony plotter works.

Will some some final checks this weekend and check all the committed code to see if I removed all the debug stuff. I made a small debug function that can be enabled with a const in the code, for if something goes wrong which makes it easier to debug. Not sure what is the best practice for this.

Also it's now named "MSX Plotter", but since there a multiple Plotters available with different commands, it's maybe better to give it a proper name like "Sony PRN-C41(D) Plotter" ?

m9710797:
>     4. I haven't thought about the fact that if you print multiple lines over the same spot that it should become darker.  Can have a look at it, but maybe better to this in a separate PR.

Fully agree, let's do this in a separate commit. And I might even volunteer to do this myself, or help you do it (*1).


> Currently this PR supports all functionality that I wanted to build to have some (basic) Plotter support. It started as a fun challenge because I got a cartridge that need the plotter and I was just wondering how this would work and how it looks. And thanks to the autocomplete it became a lot easier to implement this and luckily also the docs where available on how the Sony plotter works.

It's a nice project, and I'm glad you're now contributing it to openMSX. Thanks!

 
> Will some some final checks this weekend and check all the committed code to see if I removed all the debug stuff. I made a small debug function that can be enabled with a const in the code, for if something goes wrong which makes it easier to debug. Not sure what is the best practice for this.

Leaving the debug code in is fine. (Optionally) we can do some small tweaks, but I'll discuss those later in the review.

Speaking of review, I understand you're still working on this, nevertheless are you already interested in review comments? Or do you prefer to wait till you're finished? (Actually I don't think I'll have much time this weekend, so maybe it won't make any difference).

 
> Also it's now named "MSX Plotter", but since there a multiple Plotters available with different commands, it's maybe better to give it a proper name like "Sony PRN-C41(D) Plotter" ?

I don't have a strong opinion on this. How likely/useful is it that we''ll add multiple plotters? @MBilderbeek what do you think?

----

(*1) In an earlier comment I posted my experiment with nicer line rasterization. I've since made an improved version. The first version didn't look right in corners.

build/msvc/openmsx.vcxproj
```diff
@@ -1,1694 +1,1698 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
m9710797, line 1: The whole file changed. Possibly you changed the line-ending? Just a guess maybe the original file use CR/LF line endings (does msvc maybe require that), and you changed that to LF?

Can you amend that?

doc/plotter.md
```diff
@@ -0,0 +1,123 @@
+# Plotter Command Reference
```
m9710797: Very nice to add this documentation. Maybe also add the plotter type for which this documentation is valid.

src/MSXCharacterSets.hh
```diff
@@ -0,0 +1,28 @@
...
+#ifndef MSX_CHARACTER_SETS_HH
+#define MSX_CHARACTER_SETS_HH
+
+#include <cstdint>
+#include <span>
+
+namespace openmsx {
+
+// International MSX font (from NMS8250 BIOS ROM)
+[[nodiscard]] std::span<const uint8_t> getMSXFontRaw();
```
m9710797: Very much a detail. I wonder if we should change the return type from `std::span<const uint8_t>` to `std::span<const uint8_t, 8 * 256>`. 

Functionally it won't make any difference, and the speed difference will be insignificant. But it could be useful as documentation: when only reading this header file it gives a slightly better idea about what the format of the data could be.

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+{
+    // Initialize default font (8x8) if not loaded
+    // This prevents crashes in printVisibleCharacter if rom is empty
+    if (fontInfo.charWidth == 0) {
+	fontInfo.charWidth = 8;
+	// FontInfo struct does not have charHeight, and rom is std::array
+
+	// Populate with a very simple pattern to distinguish chars?
+	// Let's make it stripes: 0x55
+	std::fill(fontInfo.rom.begin(), fontInfo.rom.end(), 0x55);
```
m9710797: Detail: `std::ranges::fill(fontInfo.rom, 0x55);`

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+    lineType = savedLineType;
+    dashDistance = savedDashDistance;
+}
+
+void MSXPlotter::printDebug(std::string_view message) const
+{
+	if (PLOTTER_DEBUG) {
+		motherBoard.getMSXCliComm().printInfo(message);
+	}
+}
```
m9710797, line 924: Since you mentioned this debug print function before, here are some suggestions:

* You often call it like: `printDebug(strCat(...))`, we could postpone that `strCat` part till here. With something like:
```
template<typename... Args>
static void printDebug(Args&&... args)
{
    std::cerr << strCat(std::forward<Args>(args)...) << '\n';
}
```
In this example I also printed to `std::cerr` instead of using `printInfo()`. For pure debug prints this might be fine. But also no problem if you prefer the latter. (`std::cerr` has the advantage that `printDebug()` doesn't need to be a member function anymore).

* We could also implement it like:
```
#if 1
template<typename... Args> static void printDebug(Args&&...)
{
    // nothing
}
#else
template<typename... Args> static void printDebug(Args&&... args)
{
    std::cerr << strCat(std::forward<Args>(args)...) << '\n';
}
#endif 
```

Most likely the compiler will inline the first call and then, in case debug prints are disabled, they also have zero-cost. In the old implementation we still did the `strCat()`, even if that string won't get printed.

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+
+// Serialization and registration macros
+#include "serialize.hh"
```
m9710797: Technically correct, and indeed only need here. Nevertheless, we prefer to group all `#include` statements at the top of the file.

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+        "plotter-charset",
+        motherBoard.getCommandController(), "plotter-charset",
+        "character set for the MSX plotter",
+        PlotterCharacterSet::International,
+        EnumSetting<PlotterCharacterSet>::Map{
+            {"international", PlotterCharacterSet::International},
+            {"japanese", PlotterCharacterSet::Japanese},
+            {"din", PlotterCharacterSet::DIN}}))
+    , dipSwitch4Setting(motherBoard.getSharedStuff<BooleanSetting>(
+        "plotter-dipswitch4",
```
m9710797: Without reading the documentation, I have no idea what "dipswitch4" means. And the setting description "dipswitch 4 setting for the MSX plotter" doesn't give any more info.

Can we give this a more descriptive name?

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+    resetSettings(); // Resets pen pos and other states
+    ensurePrintPage();
+}
+
+void MSXPlotter::processCharacter(uint8_t data) {
+    // Debug: log every byte received
+    printDebug(
+	strCat("Plotter: received 0x", hex_string<2>(data),
+	       " mode=", (mode == Mode::TEXT ? "TEXT" : "GRAPHIC"),
+	       " escState=", static_cast<int>(escState)));
```
m9710797: Nothing wrong with `static_cast<int>(...)`. In some places in openMSX we use these alternatives:
* `narrow<int>(...)` to indicate this is a narrowing conversion and it's an error if the numerical value changes (if it doesn't fit in the more narrow type). In non-release builds we check for this error.
* `narrow_cast<int>(...)` a narrowing conversion, but now not an error if the value doesn't fit. Internally this is exactly the same as `static_cast<int>(...)`, but this better documents the intention.
* `int(...)` as a shorthand notation for `static_cast<int>(...)`, typically we use this for all to-int conversions that are not necessarily narrowing conversions.

I think in this case you want to ensure `escState` is printed as a numerical value (and not as a character). Then the 3rd form would be appropriate. Or leave it as `static_cast<int>(...)` that's also fine.

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+    switch (escState) {
+	case EscState::ESC:
+	    // Got ESC, waiting for next char
+	    if (data == '#') {
+		// ESC # = enter graphic mode
+		printDebug("Plotter: entering GRAPHIC mode");
+		mode = Mode::GRAPHIC;
+		graphicCmdBuffer.clear();
+		escState = EscState::NONE;
+		return;
```
m9710797: Indentation looks wrong sometimes. Did you use a mix of indentation with tabs and spaces?

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+	    }
+	    break;
+    }
+}
+
+void MSXPlotter::processGraphicMode(uint8_t data) {
+    // In graphic mode, accumulate command chars until delimiter
+    // Commands are ASCII text like "J100,-100,50,-100" terminated by CR/LF or next command
+
+    if (data == 0x0D || data == 0x0A) {
```
m9710797: Nothing wrong with this. I just wanted to point out a utility function we have in openMSX:
```
if (data == one_of(0x0d, 0x0a)) {
```
This is more useful if:
* The left-hand-side is an expression you don't want to re-evaluate.
* The right-hand-side has more than 2 alternatives.

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+	    bool isPrintCmd = (graphicCmdBuffer[0] == 'P');
+
+	    char lastChar = graphicCmdBuffer.back();
+	    if (!isPrintCmd && std::isalpha(static_cast<unsigned char>(lastChar)) &&
+	        std::toupper(lastChar) != 'E' &&
+	        lastChar != graphicCmdBuffer[0]) {
+		// New command starting, execute previous
+		std::string cmd = graphicCmdBuffer.substr(0, graphicCmdBuffer.size() - 1);
+		graphicCmdBuffer = graphicCmdBuffer.substr(graphicCmdBuffer.size() - 1);
+		std::swap(cmd, graphicCmdBuffer);
```
m9710797: Are these last 3 statements equivalent with the following?
```
auto cmd = std::string(1, graphicCmdBuffer.back());
graphicCmdBuffer.pop_back();
```
If correct (please verify), this should be a bit more efficient.

But also, I don't see you use `cmd` anywhere, and then we can simplify further.

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+
+void MSXPlotter::executeGraphicCommand() {
+    if (graphicCmdBuffer.empty()) return;
+
+    // Debug: log the command being executed
+    printDebug(
+	strCat("Plotter: executing graphic command '", graphicCmdBuffer, "'"));
+
+    char cmd = graphicCmdBuffer[0];
+    std::string args = graphicCmdBuffer.substr(1);
```
m9710797: Do we need to make a copy? Two alternatives are:
* `auto args = std::string_view(graphicCmdBuffer).substr(1);`.    A view into the original string, only valid of the original doesn't change, I didn't verify that.
* Start with `pos = 1` instead of `0`.  Also only valid when `graphicCmdBuffer` doesn't change.

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+
+	// Try to parse a number
+	size_t nextPos;
+	try {
+	    // Using stod is easiest for double parsing, but we need to check if it actually parses
+	    // std::stod throws if no conversion.
+	    // Manually check if valid start of number?
+	    // Valid: digit, +, -, or .
+	    char c = args[pos];
+	    if (std::isdigit(static_cast<unsigned char>(c)) || c == '-' || c == '+' || c == '.') {
```
m9710797: Here our `one_of` utility is already more useful:
```
if (std::is_digit(...) || c == one_of('-', '+', '.'))
```

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+		 // Check if it's just a lonely '+' or '-' or '.' which stod might handle or reject
+		 // Standard library stod parses.
+		 double val = std::stod(args.substr(pos), &nextPos);
```
m9710797: Using `std::stod()` is good enough for the initial version. But I think later we should replace this with a more robust version. E.g. I think it's unlikely that the parsing in `std::stod()` and in the plotter itself behaves the same in all corner cases. E.g. what with invalid input like `-8+3` or `4...5` or overflow `9999999999999999999999999999`? 

So later we should replace this with our parser that can only(?) handle `([+-]?[0-9]+(.[0-9]*)?` Note: I'm using regexp notation, but I'm NOT suggesting we should use a regexp in our parser. It shouldn't be too hard to write a simple parser that accepts `+-` only in front, followed by some digits and optionally a decimal point and again some digits.

Ideally we'd also investigate what format exactly the plotter accepts. Is it OK to have leading zeros e.g. `0012.3`? Can there be no digit before the decimal point e.g. `-.123`? Probably not, but is scientific notation accepted e.g. `1e+02`?

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+	    // parsing failed, skip char
+	    ++pos;
+	}
+    }
+
+    switch (cmd) {
+	case 'H': // Home - move to origin
+	    printDebug("Plotter: H - Home");
+	    plotterX = 0.0;
+	    plotterY = 0.0;
```
m9710797: I haven't studied the code well enough yet to know if this suggestion is a good idea. So feel free to ignore if you think it's not.

In openMSX we have a 2D math vector type `gl::vec2`. This is sort of like a struct with two members `float x; float y` (I know you use `double`, but I think `float` should be sufficient?).

It can be used everywhere where you now have a pair of floating point values. Either as 2 member variables, or as 2 function parameters.

If offers various operations like `gl::vec2 delta = pos2 - pos1;`  (shorthand for delta.x = pos2.x - pos1.x; delta.y = pos2.y - pos1.y;).

In this case setting both `x` and `y` to `0.0` can be done via:
* `plotter = {0.0f};` The single constant is "broadcast" to all vector-components.

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+	    break;
+
+	case 'M': // Move (absolute) - M x,y
+	    if (coords.size() >= 2) {
+		printDebug(
+		    strCat("Plotter: M - Move to (", coords[0], ",", coords[1], ")"));
+		moveTo(coords[0], coords[1]);
```
m9710797: IF you decide using `gl::vec2` is a good idea (see previous comment). Then this could become (notice the `{}`):
```
moveTo({coords[0], coords[1]});
```
Assuming `moveTo()` is changed to take a single `gl::vec2` parameter.

Or alternatively if `coords` _always_ contains an even number of elements (I don't know if that is true, maybe not), then we could change it to a `vector<gl::vec2>`.

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+		lineTo(newX, newY);
+	    }
+	    break;
+
+	case 'J': // Draw relative - J dx,dy[,dx,dy,...]
+	    // The J command draws a series of relative line segments
+	    printDebug(
+		strCat("Plotter: J - Draw relative, ", coords.size() / 2, " segments, penDown=", penDown));
+	    for (size_t i = 0; i + 1 < coords.size(); i += 2) {
+		lineTo(plotterX + coords[i], plotterY + coords[i + 1]);
```
m9710797: So here's an example where vector math could be useful:
```
lineTo(plotter + gl::vec2{coords[i], coords[i + 1]});
```

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+	plotWithPen(x, y, (i == 0) ? 0.0 : dDist);
+	x += xInc;
+	y += yInc;
+    }
+}
+
+void MSXPlotter::plotWithPen(double x, double y, double distMoved) {
+    ensurePrintPage();
+    auto* p = getPaper();
+    if (p) {
```
m9710797: Only a suggestion. Currently you have:
```
if (p) {
    ... ~30 lines of "happy" code, indented
} else {
   ... 1 line of error handling, far removed from the error condition
}
```

An alternative is:
```
if (!p) {
   ... handle error, close to the check
   return;
}
... the "happy" code, NOT indented.
```

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+}
+
+void MSXPlotter::ensurePrintPage() {
+    bool alreadyHadPaper = (getPaper() != nullptr);
+    ImagePrinter::ensurePrintPage();
+    if (!alreadyHadPaper) {
+	// Sony PRN-C41 uses pens that are likely ~0.4-0.5mm wide.
+	// Standard step is 0.2mm. Set dot size to 200% of step size for solid lines.
+	auto* p = getPaper();
+	if (p) {
```
m9710797: In openMSX we typically combine these 2:
```
if (auto* p = getPaper()) { ...
```
For a small function like this it doesn't really matter. But in general:
* It limits the scope of the `p` variable.
* It's one line shorter ;-)

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+	    double sizeMultiplier = (getPenThicknessSetting().getEnum() == PlotterPenThickness::Thick) ? 1.5 : 1.0;
+	    p->setDotSize(pixelSizeX * sizeMultiplier, pixelSizeY * sizeMultiplier);
```
m9710797: IF you would go for vector math, our `gl::vec2` type also supports vector-times-scalar. Then this becomes:
```
p->setDotSize(pixeSize * sizeMultiplier);
```

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+            font = getMSXJPFontRaw();
+            break;
+        case PlotterCharacterSet::DIN:
+            font = getMSXDINFontRaw();
+            break;
+        case PlotterCharacterSet::International:
+        default:
+            font = getMSXFontRaw();
+            break;
+    }
```
m9710797: Coding style. No need to change if you don't like it.

Personally I try to avoid uninitialized variables (like on line 670), even if soon after they do get initialized (lines 673, 676, 680). Here that could be achieved with something like this:
```
auto font = [&] {
    switch (getCharacterSet()) {
        using enum PlotterCharacterSet;
        default:
        case: International: return getMSXFontRaw();
        case Japanese:       return getMSXJPFontRaw();
        case DIN:            return getMSXDINFontRaw();
    }
}();
```
Notice the `()` at the end. We create an (unnamed) lambda function (which doesn't take any parameters), and then immediately invoke it (via passing zero parameters).


EDIT: OK, technically `std::span<const uint8_t>` does initialize the value (to the empty span), but then I meant "dummy initialize, only to overwrite soon after".

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+    const uint8_t* fontPtr = font.data() + c * 8;
```
m9710797: No need to change if you don't like it.

For extra safety we could do something like this (requires our `#include "ranges.hh"`):
```
auto glyph = subspan<8>(font, 8 * c);
```
Now the type is `std::span<const uint8_t, 8>` instead of just `const uint8_t*`. So we retain the knowledge that `glyph` points to (only) 8 bytes. (So in a debug build this can do better bound-checks).

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+    //double gridSpacingX = 0.54 * scaleFactor;
+    //double gridSpacingY = 0.95 * scaleFactor;
+
+    maxLineHeight = std::max(maxLineHeight, 8.0 * gridSpacingY);
+
+    // Transform local (u, v) [0..width, 0..height] to global (px, py)
+    // Cursor is at the START of the character baseline.
+    // u=0 is at cursor, character extends in reading direction (+u).
+    // v=0 is at baseline, character extends upward (+v).
+    auto transform = [&](double u, double v) -> std::pair<double, double> {
```
m9710797: Returning a `pair<double, double>`, that's another hint that `gl::vec2` might be a good fit.

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+    double charWidthOnly = 4.12 * gridSpacingX;   // just the character
+    double charGap = 2.3 * gridSpacingX;          // gap between characters (11.12 - 4.12 = 7.0)
+    double charAdvance = charWidthOnly + charGap; // total advance
+
+    switch (rotation) {
+	case 0: plotterX += charAdvance; break;
+	case 1: plotterY -= charAdvance; break;
+	case 2: plotterX -= charAdvance; break;
+	case 3: plotterY += charAdvance; break;
+    }
```
m9710797: With vector math this could become:
```
plotter += transform({charAdvance, 0.0f});
```

src/Plotter.hh
```diff
@@ -0,0 +1,157 @@
...
+#include <memory>
+#include <string>
+#include <utility> // for std::pair
+#include <cstdint> // for uint8_t
+
+namespace openmsx {
+
+// Character set options for the plotter
+enum class PlotterCharacterSet { International, Japanese, DIN };
+enum class PlotterPenThickness { Standard, Thick };
```
m9710797: Maybe these two belong inside the `MSXPlotter` class. I haven't checked if that's indeed possible?

And if so I'd also drop the, then redundant, "Plotter" prefix. So externally the name changes from `PlotterCharacterSet` to `MSXPlotter::CharacterSet`.

src/Plotter.hh
```diff
@@ -0,0 +1,157 @@
...
+	// Graphic mode command buffer
+	std::string graphicCmdBuffer;
+
+	// Pen colors: RGB tuples for each pen (0=black, 1=blue, 2=green, 3=red)
+	static constexpr std::array<std::array<uint8_t, 3>, 4> penColors = {{
+		{{0, 0, 0}},       // black
+		{{0, 0, 255}},     // blue
+		{{49, 153, 90}},     // green
+		{{255, 0, 0}}      // red
+	}};
```
m9710797: (Applies here and also to the constants below.)

Constants in the `private` section can often be moved outside the `.hh` file, inside the `.cc` file.
Either place them near the top of the `.cc` file (closely after the `#include` section). Or, when possible, inside the (only) function that uses them.

src/Plotter.hh
```diff
@@ -0,0 +1,157 @@
...
+
+	// A4 Plotting Area: 192mm x 276.8mm => 960 x 1384 steps.
+	static constexpr unsigned PLOT_AREA_WIDTH = 960;
+	static constexpr unsigned PLOT_AREA_HEIGHT = 1384;
+
+	// Center the plotting area on the paper.
+	static constexpr unsigned MARGIN_X = (PAPER_WIDTH_STEPS - PLOT_AREA_WIDTH) / 2; // 45 steps (9mm)
+	static constexpr unsigned MARGIN_Y = (PAPER_HEIGHT_STEPS - PLOT_AREA_HEIGHT) / 2; // 48 steps (9.6mm)
+
+	// Private methods
```
m9710797: Coding style. Good that you grouped private member variables and function. In openMSX we usually order them like this:
```
private:
    ... all private member functions

private:
    ... all private member variables
};
```

So functions before variable, and with a (redundant) `private:` keyword between the two sections.

src/imgui/ImGuiMachine.cc
```diff
@@ -1101,16 +1101,26 @@ const std::string& ImGuiMachine::getTestResult(MachineInfo& info)
...
+				//assert(currentInfo->testResult->empty());
+				amendConfigInfo(mb, *currentInfo);
+			} catch (const MSXException& e) {
+				if (auto* currentInfo = findMachineInfo(configName)) {
+					currentInfo->testResult = e.getMessage(); // error
+				}
+			} catch (const std::exception& e) {
+				if (auto* currentInfo = findMachineInfo(configName)) {
+					currentInfo->testResult = strCat("std::exception: ", e.what()); // error
+				}
```
m9710797, line 1123: It's not immediately clear why this change is needed. I see you added an extra catch-block for `std::exception`. I don't know why yet (I also didn't search for it yet). Maybe there's a good reason, but my first idea is that instead we should catch that exception closer to where it's thrown and, if needed, re-throw it as an `MSXException`.

src/imgui/ImGuiPlotterViewer.cc
```diff
@@ -0,0 +1,141 @@
...
+			if (auto* pluggable = &connector->getPlugged()) {
+				if (auto* p = dynamic_cast<MSXPlotter*>(pluggable)) {
+					plotter = p;
+					break;
+				}
+			}
+		}
+
+		if (!plotter) {
+			ImGui::TextUnformatted("No plotter present in the system.");
```
m9710797: Detail. Above we test that the plotter is actually plugged into "the" printer port, right? Maybe then formulate the message like that e.g. "Plotter not plugged in" (because otherwise it might be confusing that the user _does_ see the plotter in the list of pluggable, while this message says it's not present in the whole system).

In the previous paragraph I said `"the" printer port` with "the" in quotes, because in theory the emulated MSX machine could have more than one printer port (e.g. the standard builtin printer port, plus the `Philips_VU_0040` cartridge with an extra printer port (or better a modified version that uses different IO ports). But let's ignore that possibility, most likely nobody will ever use more than one printer port. IMHO it's good that the openMSX core supports it, but the GUI should not support these very exotic scenarios.

src/imgui/ImGuiPlotterViewer.cc
```diff
@@ -0,0 +1,141 @@
...
+			if (texture) texture.reset();
+			return;
+		}
+
+		auto* paper = plotter->getPaper();
+		if (!paper) {
+			ImGui::TextUnformatted("Plotter is present, but has no paper.");
```
m9710797: Maybe simplify this message to just `"Plotter has no paper."`?
No need to say everything else that went well, e.g. the MSX machine is powered on etc ;-)

src/imgui/ImGuiPlotterViewer.cc
```diff
@@ -0,0 +1,141 @@
...
+			if (texture) texture.reset();
```
m9710797: Very much a detail. It's ok to call `reset()` on an already empty `std::optional`. So remove the `if (texture)` part.

(Internally the first thing that `reset()` does is check whether it's already empty, so no need to repeat that check externally).

src/imgui/ImGuiPlotterViewer.cc
```diff
@@ -0,0 +1,141 @@
...
+		if (!texture || texture->getWidth() != (int)width || texture->getHeight() != (int)height) {
+			texture.emplace(width, height);
+			texture->setInterpolation(false);
+			texture->setWrapMode(false);
+			texture->bind();
+			glTexImage2D(GL_TEXTURE_2D, 0, GL_RGB, width, height, 0, GL_RGB, GL_UNSIGNED_BYTE, rgbData.data());
+		} else {
+			texture->bind();
+			glTexSubImage2D(GL_TEXTURE_2D, 0, 0, 0, width, height, GL_RGB, GL_UNSIGNED_BYTE, rgbData.data());
+		}
```
m9710797: Functionally this is 100% correct. Nevertheless I'm a bit concerned about this for two reasons:
* The image can be very large. For A4 paper at 300 dpi it's 2480 x 3508 pixels. But the user can also choose a higher dpi setting (e.g. 600 or even 1200). Such large images might run into the maximum texture-size limitation of some openGL implementations. Maybe we need to down-sample the image before uploading it as a texture? You've chosen `setInterpolation(false)` that's fine. But that means down-sampling could be very easy, just drop pixels. E.g. maybe drop `n-1` out of every `n` pixels until (each dimension) of the image is <= 2048 pixels?? That should not result in any image quality loss in this viewer (and still retain the higher resolution in the `.png` output).
* Currently we upload a new version of this texture on _every_ frame (e.g. 60 times per second). That's OK if the plotter is actually working, but maybe we should add a `dirty` flag in the `Paper` class to indicate something changed (a flag with a reset method, or a generation counter). That way we can skip the (semi-expensive) texture upload (and down-sampling) when nothing changed.

src/imgui/ImGuiPlotterViewer.cc
```diff
@@ -0,0 +1,141 @@
...
+
+		// Calculate available space and aspect ratio
+		float availableWidth = ImGui::GetContentRegionAvail().x;
+		float availableHeight = ImGui::GetContentRegionAvail().y - ImGui::GetTextLineHeight() * 2.5f;
+		float scale = std::min(availableWidth / (float)width, availableHeight / (float)height);
+		ImVec2 displaySize = ImVec2(width * scale, height * scale);
```
m9710797: Initially I thought this computed `displaySize` to be equal as `availableSize`. But that's not true because of the `min()` operation. You instead scale to fit `availableSize` while retaining the aspect ratio.

BTW with vector-math this can be simplified to:
```
auto paperSize = gl::vec2{width, heigh}; // or better add a method: "gl::vec2 Paper::getSize()"
auto availableSize = ImGui::GetContentRegionAvail() - gl::vec2{0.0f, 2.5f * ImGui::GetTextLineHeight()};
auto scale = paperSize / availableSize;
auto displaySize = paperSize * min_component(scale);
```
And then if you also want to center the scaled image:
```
auto offset = (availableSize - displaySize) * 0.5f;
```

src/imgui/ImGuiSettings.cc
```diff
@@ -384,6 +386,50 @@ void ImGuiSettings::showMenu(MSXMotherBoard* motherBoard)
 			});
 			ImGui::MenuItem("Configure messages", nullptr, &manager.messages->configureWindow.open);
 		});
+		im::Menu("Output", motherBoard != nullptr, [&]{
+			im::TreeNode("MSX Plotter", ImGuiTreeNodeFlags_DefaultOpen, [&]{
```
m9710797: Currently `Output` only has a single item inside. That's unusual for a GUI, no?
Until we have more stuff to group below `Output`, can we move it someplace else? Or possibly instead make `MSX Plotter` the sub-menu, and omit the `TreeNode`?

src/imgui/ImGuiSettings.cc
```diff
@@ -384,6 +386,50 @@ void ImGuiSettings::showMenu(MSXMotherBoard* motherBoard)
...
+				// Find the MSXPlotter in the pluggables
+				MSXPlotter* plotter = nullptr;
+				for (const auto& plug : motherBoard->getPluggingController().getPluggables()) {
+					if (auto* p = dynamic_cast<MSXPlotter*>(plug.get())) {
+						plotter = p;
+						break;
+					}
+				}
+
+				if (!plotter) {
```
m9710797: Can this ever be `nullptr`? Why would there be no `MSXPlotter`? Note we just want the plotter, independent of whether it's plugged in or not (good).

src/imgui/ImGuiTools.cc
```diff
@@ -143,6 +144,7 @@ void ImGuiTools::showMenu(MSXMotherBoard* motherBoard)
 		ImGui::MenuItem("SCC viewer", nullptr, &manager.sccViewer->show);
 		ImGui::MenuItem("MSX-Music viewer", nullptr, &manager.msxMusicViewer->show);
 		ImGui::MenuItem("Audio channel viewer", nullptr, &manager.waveViewer->show);
+		ImGui::MenuItem("Plotter viewer", nullptr, &manager.plotterViewer->show);
```
m9710797, line 147: Fine tuning. Maybe we can add a tooltip here, mentioning that the plotter should be plugged into the printer port.
```
ImGui::MenuItem("Plotter viewer", ...);
simpleToolTip("...");
```

m9710797 (commented):
Hi,

I did a quick review. I did not yet try to understand what the code is actually doing. I only looked at small code snippets without the larger context. So possibly some of my comments don't make sense (in the larger context). Many of comments are also merely suggestions, feel free to ignore.

src/Plotter.cc
```diff
@@ -0,0 +1,774 @@
...
+    switch (data) {
+	case 0x01: // Header / Literal prefix
+	    printNext = true;
+	    break;
+	case 0x08: // BS - Backspace
+	    plotter.x = std::max(plotter.x - (8 + 2 * charScale), 0.0f);
+	    break;
+	case 0x0a: // LF - Line feed
+	    plotter.y -= float(lineFeed);
+// ...
```
m9710797: I missed this in my previous review (this `// ...`). Now I noticed because it gives a compiler warning:
```
src/Plotter.cc: In member function ‘void openmsx::MSXPlotter::processTextMode(uint8_t)’:
src/Plotter.cc:202:23: warning: this statement may fall through [-Wimplicit-fallthrough=]
  202 |             plotter.y -= float(lineFeed);
      |             ~~~~~~~~~~^~~~~~~~~~~~~~~~~~
src/Plotter.cc:204:9: note: here
  204 |         case 0x0b: // VT - Line up (feed paper to previous line)
      |         ^~~~
```
What is the intention of this code? Is there a missing break? Or is the fall-through intended? In case of the latter there's a _standard_ way to suppress the warning (= to document the intention), like this:
```
case 0x0a: // LF - Line feed
    plotter.y -= float(lineFeed);
    [[fallthrough]];
case 0x0b: ...
```

src/Plotter.cc
```diff
@@ -0,0 +1,774 @@
...
+	case 0x1b: // ESC - Start escape sequence
+	    escState = EscState::ESC;
+	    break;
+	default:
+	    // Printable character in text mode (Allow all >= 32)
+	    if (data >= 32) {
+		// Check for line wrap before drawing
+		// Estimate width: 6 steps base * scaleFactor
+		// Or simply check current position
+		if (plotter.x >= float(rightBorder)) {
```
m9710797: Thanks for trying out the `gl::vec2` stuff. I now see this requires a lot of casts from `double` to `float`. That's because `ImagePrinter` currently also uses `double` everywhere (when `float` should be more than sufficient (*1)). Maybe I should clean that up (= use `float' instead of `double`). But let's not do that also in this PR.

I do wonder if `ImagePrinter` is the right base class for `MSXPlotter`, because it seems to offer a lot of stuff that's irrelevant for the plotter. Maybe we should introduce a new class that sits between `PrinterCore` and `ImagePrinter`, and then let `MSXPlotter` and `PrinterCore` inherit from that new class? Or alternatively (depending on how much can still be shared between `ImagePrinter` and `MSXPlotter`) we could let `MSXPlotter` inherit directly from `PrinterCore`? But it's fine to postpone this refactoring till later (I mean not needed for the first version of this PR).



---

(*1) The `ImagePrinter` code was written more than 20(?) years ago. At the time x86 floating point was still mainly done on the old x87 FPU, on that unit `float` and `double` computations execute equally fast (read `float` is as slow as `double`). Since then x86 CPUs introduced the `SSE` (and newer) instruction sets. And now `float` does have an advantage over `double` (also memory bandwidth which has become more important). All this to say: nowadays, if I were to re-implement the `ImagePrinter` class, I would for sure use `float` instead of `double`.

src/Plotter.cc
```diff
@@ -0,0 +1,774 @@
...
+    switch (cmd) {
+	case 'H': // Home - move to origin
+	    printDebug("Plotter: H - Home");
+	    plotter = gl::vec2{0.0f};
+	    break;
+
+	case 'M': // Move (absolute) - M x,y
+	    if (coords.size() >= 2) {
+		printDebug("Plotter: M - Move to (", coords[0], ",", coords[1], ")");
+		moveTo({float(coords[0]), float(coords[1])});
```
m9710797: You changed `coords` to `vector<float>`, so these casts are no longer needed. (And in various places below.)

src/Plotter.cc
```diff
@@ -0,0 +1,774 @@
...
+void MSXPlotter::processGraphicMode(uint8_t data) {
+    // In graphic mode, accumulate command chars until delimiter
+    // Commands are ASCII text like "J100,-100,50,-100" terminated by CR/LF or next command
+
+    if (data == one_of(0x0d, 0x0a)) {
+	// CR or LF terminates command
+	if (!graphicCmdBuffer.empty()) {
+	    executeGraphicCommand();
+	    graphicCmdBuffer.clear();
+	}
```
m9710797: Suggestion: `executeGraphicsCommand()` already contains a check for `graphicCmdBuffer.empty()`. We could also make `executeGraphicCommand()` responsible for `clear()`ing `graphicCmdBuffer` before it returns. That would simplify all places where we currently call `executeGraphicsCommand()`.

src/Paper.hh
```diff
@@ -0,0 +1,52 @@
...
+
+	[[nodiscard]] unsigned getWidth() const { return sizeX; }
+	[[nodiscard]] unsigned getHeight() const { return sizeY; }
+	[[nodiscard]] std::span<const uint8_t> getRGBData() const { return colorBuf; }
+
+private:
+	uint8_t& dot(unsigned x, unsigned y);
+
+	std::vector<uint8_t> buf;
+	std::vector<uint8_t> colorBuf; // RGB triples for color output
```
m9710797: This is fine for the first version, but eventually we should clean this up (i can help/volunteer to do that).

Currently `Paper` contains both `buf` and `colorBuf`. And then `plot()` vs `plotColor()` renders in 1 of these 2 buffers, and in the end you call `save()` with a bool parameter indicating which buffer to save. So a single `Paper` object represents in fact 2 independent papers, 1 grayscale and 1 colored. We should split that up.

src/imgui/ImGuiMachine.cc
sndpl: Will remove it since I can't reproduce it anymore, but during developing openMSX crashed when I opened the recently used machines dropdown. It didn't give any real error message, so that's why i added the try/catch. But this didn't give any better results. At the end I remove the machine recent entries from the imgui.ini which solved the problem. When I opened the same machines again like I had before I removed them, there was no issue. So not sure what was the issue.

build/msvc/openmsx.vcxproj
sndpl: should be fixed, my git setting did convert the CR/LF to LF.

doc/plotter.md
sndpl: Updated the doc for the Sony, also added some info about the other Plotters. Who knows, maybe someday the information will be available :)

src/MSXCharacterSets.hh
sndpl: I updated the types

src/Plotter.cc
sndpl: looks cleaner, so updated

src/Plotter.cc
sndpl: I implemented the template for debugging.

src/Plotter.cc
sndpl: Moved it to the top

src/Plotter.cc
sndpl: updated it to int.

src/Plotter.cc
sndpl: editor didn't have a formatter for C++, I installed one and did a reformat of the code.

src/Plotter.cc
sndpl: used the one_of function,

src/Plotter.cc
sndpl: yes, I updated the code

src/Plotter.cc
sndpl: Updated the code.

src/Plotter.cc
sndpl: Used the one_of

src/Plotter.cc
sndpl: I need to do some more testing to see what it does accept and what not. It does send the scientific notation values like "D 4.9235E-08", I had to fix this because the parser did see this as a new command :)

src/Plotter.cc
sndpl: as you already saw, I used the gl::vec2, still needed some double -> float casts because of other classes that uses doubles. But you also saw it but it's out of this scope.

src/Plotter.cc
sndpl: updated the moveTo

src/Plotter.cc
sndpl: also updated to use the gl::vec2

src/Plotter.cc
sndpl: ah yes, updated the code

src/Plotter.cc
sndpl: updated

src/Plotter.cc
sndpl: it's using the gl::vec2 now

src/Plotter.cc
sndpl: updated the code

src/Plotter.cc
sndpl: Used the subspan and it still works :)

src/Plotter.cc
sndpl: it's using gl::vec2 now

src/Plotter.cc
sndpl: it's using the vector now

src/Plotter.hh
sndpl: Not complete sure what to do with this, since I need the type also for the function declarations in the Plotter.hh. 
I thought when I use it in both files the .hh file would be better to place these kind of constants? or is this not way to go in C++ ?

src/Plotter.hh
sndpl: I left them here, because I also need them in imGui preview screen. Then this file looks the best place to me?

src/Plotter.hh
sndpl: updated the coding style

src/imgui/ImGuiPlotterViewer.cc
sndpl: updated the text.

src/imgui/ImGuiPlotterViewer.cc
sndpl: updated the text

src/imgui/ImGuiPlotterViewer.cc
sndpl: Removed the if statements

src/imgui/ImGuiPlotterViewer.cc
sndpl: Ok, i tried to add a generation counter which seems to work, the only downside of the change I made is that the preview is not that sharp any more as the image it's saving to the png file. But maybe I made a mistake somewhere.

src/imgui/ImGuiPlotterViewer.cc
sndpl: updated it to vectors.

src/imgui/ImGuiSettings.cc
sndpl: removed the threenode

src/imgui/ImGuiSettings.cc
sndpl: updated the code

src/imgui/ImGuiTools.cc
sndpl: Added a tooltip

src/Plotter.cc
sndpl: Good question, the Plotter has four dip switches, but the service manual and user manual only write about dip switch 4.  In case of sending a carriage return code (0DH) in the text mode it does the following:
On: the pen position is set to the beginning of the next line
Off: the pen position is set to the beginning of the line

The software should describe how to set these switches. In the manual of Creative Greetings is shows a picture where switch 4 is on, the rest of off. I could implement all 4 switches in the settings screen, but no idea what they do, that can only be tested on a real plotter.

MBilderbeek:
> > Also it's now named "MSX Plotter", but since there a multiple Plotters available with different commands, it's maybe better to give it a proper name like "Sony PRN-C41(D) Plotter" ?
> 
> I don't have a strong opinion on this. How likely/useful is it that we''ll add multiple plotters? @MBilderbeek what do you think?

As we know there are multiple plotter types available, I think it indeed would be best to use "SonyPlotter" or something for the class name which implements the command set and for the user, you could indeed call it specifically the Sony PRN-C41. Because that's what it is.

src/Plotter.cc
MBilderbeek: I checked the service manual on page 5-1 and 6-1:
- DIP switch 1 and 2 are apparently not connected at all.
- DIP switch 3 is only connected on the Japanese model and selects between JIS and SHIFT-JIS. It is short circuited on non-Japanese models (so GND is on the pin to the chip that makes the selection).
- DIP switch 4 is called CR/CR,LF in that manual, but I think it's fine to just keep calling it "DIP switch 4".

src/Plotter.hh
m9710797: What I meant was move them inside the `MSXPlotter` class, but still in the `.hh` file:
```
class MSXPlotter ... {
    ... all private types, enums, constants
public:
     ... all public types, enums, constants
public:
    ... all public functions
private:
   ... all private functions
private:
   ... all private variables
};
```
In general we prefer to move stuff to the smallest possible scope:
* `PlotterCharacterSet` to prefix `Plotter` suggests that this is something specific for the `MSXPlotter` class. So, if possible, we should move it inside that class.
* And indeed this `enum` is used in the declaration of other functions (e.g. `getCharacterSet()`). So this `enum` must already be declared. That means the smallest scope we can move it is inside `MSXPlotter`, but before the sections with public functions.

Another example is the constant `MARGIN_X`. Currently it's defined in the `.hh` file, but nothing in the file already uses it. Then it becomes possible to move it to the `.cc` file. We could even go further: currently it's only used in the `plotWithPen()` function. So we can consider moving it there (thus to a even smaller scope).

Moving from the `.hh` to the `.cc` file has a small technical advantage: when we ever change that value (unlikely) or change the type (currently it's `unsigned`, but we only ever use it as a `float`), then the build system needs to recompile fewer files.

The next step, moving from the global scope in the `.cc` file to inside the `plotWithPen()` function is more of a judgement call. If you think that constant is strongly tied to (only) that specific function then inside that function is a good place. OTOH if you think in the future it's likely other functions will also need that constant, then it's better to make it globally visible (in the `.cc` file).

There's no single right or wrong answer here.

src/imgui/ImGuiPlotterViewer.cc
```diff
@@ -0,0 +1,180 @@
...
+
+    if (!texture || texture->getWidth() != (int)width ||
+        texture->getHeight() != (int)height ||
+        currentGeneration != lastGeneration) {
+
+      auto rgbData = paper->getRGBData();
+      int n = 1;
+      while ((width / n) > 2048 || (height / n) > 2048) {
+        n *= 2;
+      }
```
m9710797: Thanks for implementing down sampling. Though you say it's now looking less sharp. My assumption was that it wouldn't make a noticeable difference (for viewing on screen with probably fewer than 2048x2048 pixels). But if my assumption is wrong then we may need to look for other solutions. What is the original resolution? What is the down sample factor? What are the viewing dimensions you use?

BTW, you calculate the down sample factor as a power-of-2. That's often used in practice in combination with filters (e.g. recursively average blocks of 2x2 pixels). Here we're just dropping pixels (maybe averaging pixels will give a nicer visual result). But we could also down sample by a non-power-of-2. E.g. reduce 6000 to 2000 instead of 6000 to 1500. We could even choose a different factor for X and Y. (BTW we can then compute the factor directly via int(ceil(width / 2048)), no need for a loop).

BTW may I suggest to put the whole computation of `texture` in a separate function. Just to avoid having too large functions (many of the other openMSX GUI code has the same problem).

src/imgui/ImGuiPlotterViewer.cc
```diff
@@ -0,0 +1,180 @@
...
+
+      int texW = width / n;
+      int texH = height / n;
+
+      const uint8_t *dataToUpload = rgbData.data();
+      if (n > 1) {
+        downsampledBuf.assign(size_t(texW) * texH * 3, 255);
```
m9710797: You will anyway overwrite all pixels in this buffer, right? Then there's no need to first set them all to `255`. You can replace `assign()` with `resize()`.

src/imgui/ImGuiPlotterViewer.cc
```diff
@@ -0,0 +1,180 @@
...
+
+    // Draw crosshair at pen position
+    auto [plotX, plotY] = plotter->getPlotterPos();
+    // MSXPlotter: X-axis is short axis (carriage), Y-axis is long axis (paper
+    // feed). Origin (0,0) is bottom-left of plotting area. Paper location in
+    // plotWithPen: double paperX = MARGIN_X + (x + originX); double paperY =
+    // (PAPER_HEIGHT_STEPS - MARGIN_Y) - (y + originY); NOTE: Plotter steps
+    // might be slightly different than paper size, but MARGIN_X handles it.
+    // Let's use the same logic as plotWithPen:
+    double paperX = (double)plotX + 45.0; // MARGIN_X is 45
```
m9710797: Aha, here I see that you _do_ use (or want to use) `MARGIN_X`. In one of my earlier comments I used it as an example of a private constant. Maybe that was then a bad example.

OTOH I don't immediately see why this viewer needs to know what the margins. Don't you want to view the full paper, including margins?

src/imgui/ImGuiPlotterViewer.hh
```diff
@@ -0,0 +1,35 @@
...
+  }
+  void save(ImGuiTextBuffer &buf) override;
+  void loadLine(std::string_view name, zstring_view value) override;
+  void paint(MSXMotherBoard *motherBoard) override;
+
+  bool show = false;
+
+private:
+  std::optional<gl::ColorTexture> texture;
+  uint64_t lastGeneration = uint64_t(-1);
```
m9710797, line 27: Hehe, 64-bit generation counter. You expect to see A LOT of generations ;-)

src/Plotter.cc
```diff
@@ -0,0 +1,788 @@
...
+  }
+
+  resetSettings();
+  ensurePrintPage();
+}
+
+zstring_view MSXPlotter::getName() const { return "MSXPlotter"; }
+
+zstring_view MSXPlotter::getDescription() const {
+  return "Sony PRN-C41 Color Plotter (4 pens)";
```
m9710797: Detail, coding style (we can fix this later).

You formatted your code with indentation of 2 spaces. All other openMSX code indents with 1 tab character (and tab stops set to 8 characters).

That's mostly a historical decision: 8 characters per indentation level forces to split large functions (with deep indentation levels) into smaller parts. Maybe nowadays for a new project I'd choose 4 characters indentation (and spaces instead of tabs). But not worth reformatting all openMSX code, so let's stick with the current convention of 1 tab character.

src/Plotter.cc
```diff
@@ -0,0 +1,788 @@
...
+    if (data == ' ')
+      return; // Ignore spaces
+    if (data >= '0' && data <= '3') {
+      selectedPen = data - '0';
+      printDebug("Plotter: selected pen ", selectedPen);
+    }
+    escState = EscState::NONE;
+    return;
+  case EscState::NONE:
+    break;
```
m9710797: Did I see correctly that all the above blocks except for this one end with `return`? Then maybe the code becomes simpler if we move the `if (mode == Mode::TEXT) { ... } else { ... }` block here. And remove all the above `return` statements?

Just a suggestion, pleas check if it's correct, and if it indeed makes the code easier to understand.

src/Plotter.cc
```diff
@@ -0,0 +1,788 @@
...
+
+    // Execute immediately if buffer starts with known command and we hit a
+    // letter (next command) Commands start with a letter, followed by numbers
+    // and commas
+    if (graphicCmdBuffer.size() > 1) {
+      // Special handling for P command which takes arbitrary text
+      bool isPrintCmd = (graphicCmdBuffer[0] == 'P');
+
+      char lastChar = graphicCmdBuffer.back();
+      if (!isPrintCmd && std::isalpha(static_cast<unsigned char>(lastChar)) &&
```
m9710797: I suggest to not use the `isalpha()` function here. Because according to [the documentation](https://en.cppreference.com/w/cpp/string/byte/isalpha) the result is locale dependent. E.g. it gives this example:
```
isalpha('\xdf', default C locale) returned false
isalpha('\xdf', ISO-8859-1 locale) returned true
```
The behavior of the plotter emulation should not depend on users host locale settings.

Ah, and similar for `toupper()` on the next line, that result is also locale dependent. In this case there's an easy replacement: `lastChar != one_of('E', 'e', grahichCmdBuffer.front())`.

src/Plotter.cc
```diff
@@ -0,0 +1,788 @@
...
+                 ")");
+      moveTo(plotter + gl::vec2{coords[0], coords[1]});
+    }
+    break;
+
+  case 'D': // Draw (absolute) - D x,y
+    if (coords.size() >= 2) {
+      gl::vec2 target{coords[0], coords[1]};
+      printDebug("Plotter: D - Draw to (", target.x, ",", target.y, ") from (",
+                 plotter.x, ",", plotter.y, ") penDown=", penDown);
```
m9710797: Detail, I'm also not 100% sure, just try it.

If you don't care about the exact formatting in your debug prints, then you can simplify this to:
```
printDebug("Plotter: D - draw to ", target, " from ", plotter, " penDown=", penDown);
```
In other words: the gl::vec2 type can be printed directly. I believe it gets formatted like: `[123.0, 45.7]`.

m9710797 (commented):
Thanks for your updates. Here are some more suggestions.

src/Plotter.cc
```diff
@@ -0,0 +1,788 @@
...
+  // Transform local (u, v) [0..width, 0..height] to global (px, py)
+  // Cursor is at the START of the character baseline.
+  // u=0 is at cursor, character extends in reading direction (+u).
+  // v=0 is at baseline, character extends upward (+v).
+  auto transform = [&](gl::vec2 p) -> gl::vec2 {
+    static constexpr std::array<gl::vec2, 4> uDir = {
+        {{1.0f, 0.0f}, {0.0f, -1.0f}, {-1.0f, 0.0f}, {0.0f, 1.0f}}};
+    static constexpr std::array<gl::vec2, 4> vDir = {
+        {{0.0f, 1.0f}, {1.0f, 0.0f}, {0.0f, -1.0f}, {-1.0f, 0.0f}}};
+    return plotter + p.x * uDir[rotation] + p.y * vDir[rotation];
```
m9710797: Depending on how familiar you are with vector/matrix math you could _simplify_ this to:
```
static constexpr std::array rotations = {
    gl::mat2{gl::vec2{1.0f, 0.0f}, gl::vec2{0.0f, 1.0f}},
    gl::mat2{...},
    ...
};
return plotter + rotations[rotation] * p;
```
In other words: we also have a `gl::mat2` type representing a 2x2 matrix, and it supports matrix-times-vector operation. Rotations are a natural fit for matrices.

src/Plotter.hh
m9710797: Agree. At the time I wrote my comment, those constants were still `private`.
If you use those constants are externally (and thus make them `public`) then absolutely this is the correct place.

src/Plotter.cc
sndpl: Ok, used the gl:mat2 and it still works :)

src/Plotter.cc
sndpl: Updated the debug command, it's easier to read so and don't care that much about the format, it's debug

src/Plotter.cc
sndpl: Updated the toupper and isalpha functions

src/Plotter.cc
sndpl: I installed a clangd plugin in vscode that did the formatting, will check if I can set it to 1 tab.

src/imgui/ImGuiPlotterViewer.cc
sndpl: I updated the code a bit, the result is better but I removed the power-of-2 downsampling.... the texture computation has been moved to it's own function.

src/imgui/ImGuiPlotterViewer.cc
sndpl: Updated it with resize.

src/Plotter.cc
sndpl: I tried to update this, but for some reason when I removed the returns it didn't work any more. Will have a look at it later when I have more time to see if we can improve this.

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+              EnumSetting<MSXPlotter::CharacterSet>::Map{
+                  {"international", MSXPlotter::CharacterSet::International},
+                  {"japanese", MSXPlotter::CharacterSet::Japanese},
+                  {"din", MSXPlotter::CharacterSet::DIN}})),
+      dipSwitch4Setting(motherBoard.getSharedStuff<BooleanSetting>(
+          "plotter-dipswitch4", motherBoard.getCommandController(),
+          "plotter-dipswitch4", "dipswitch 4 setting for the MSX plotter",
+          false)),
+      penThicknessSetting(
+          motherBoard.getSharedStuff<EnumSetting<MSXPlotter::PenThickness>>(
```
m9710797: I assume you copied that `getSharedStuff<...>(...)` from the code in `Printer.cc`?

In the printer code we had that to create a setting `print-resolution` which was _shared_ between all (2 in this case) `MSXPrinter` and `EpsonPrinter` instances. The reasoning was that `print-resolution` is more of a global emulator setting, not a per-printer setting.

In this case for the `plotter-charset`, `plotter-dipswitch4` and `plotter-pen-thickness` settings, I do think they should be per instance of the plotter, not shared global settings. In practice it won't make a difference of course, because there is only one instance(*1) of the `MSXPlotter` class. But it is conceptually more correct, and it also simplifies the code if we create these settings simply as:

```
In plotter.hh:
    EnumSetting<CharacterSet> charSetSetting;
In Plotter.cc:
    charSetSetting(
        motherBoard.getCommandController(),
        "plotter-charset", "character set for the MSX plotter",
        CharacterSet::International,
        EnumSetting<CharacterSet>::Map{
            {"international", CharacterSet::International},
            {"japanese", CharacterSet::Japanese},
            {"din", CharacterSet::DIN}})
```

(*1) We should _not_ do this IMHO, but if we would want to create multiple instances of `MSXPlotter`, each with their own settings, we should make the setting name include the instance name. But again, let's not do that, let's just use a single instance with the current setting names.

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+
+bool MSXPlotter::getStatus(EmuTime /*time*/) { return false; }
+
+void MSXPlotter::setStrobe(bool strobe, EmuTime time) {
+  ImagePrinter::setStrobe(strobe, time);
+}
+
+void MSXPlotter::writeData(uint8_t data, EmuTime time) {
+  ImagePrinter::writeData(data, time);
+}
```
m9710797, line 106: `setStrobe()` and `writeData()` don't do anything more than delegate to the base class implementation. Doesn't that mean we can leave out our implementation?

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+  dipSwitch4Setting->setBoolean(enabled);
+}
+
+void MSXPlotter::cyclePen() { selectedPen = (selectedPen + 1) % 4; }
+
+void MSXPlotter::moveStep(gl::vec2 delta) {
+  gl::vec2 oldAbsPos = plotter + origin;
+  gl::vec2 newAbsPos =
+      gl::clamp(oldAbsPos + delta, gl::vec2{0.0f},
+                gl::vec2{float(PLOT_AREA_WIDTH), float(PLOT_AREA_HEIGHT)});
```
m9710797: Detail: can we join `PLOT_AREA_WIDTH` and `PLOT_AREA_HEIGHT` into a single `gl::vec2` constant?

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+  plotter = newAbsPos - origin;
```
m9710797: Detail: `plotter` is not a very descriptive name. Maybe `position` or `penPosition` is clearer?

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+      printDebug("Plotter: L - Line type set to ", lineType);
+    }
+    break;
+
+  case 'Q': // Alpha rotate - Q n (0-3)
+    if (!coords.empty()) {
+      // If a character was just printed, undo the gap
+      // (the last char before Q should not have the trailing gap)
+      if (pendingCharGap > 0.0f) {
+        switch (pendingGapRotation) {
```
m9710797: I wonder if we can reuse the rotation code here? So factor out the transform function from `drawCharacter` and then do something like:
```
plotter += transform({pendingCharGap, 0.0f});
```
First I thought we could do that same in `case 'F'`, but probably not.

src/Plotter.cc
```diff
@@ -0,0 +1,813 @@
...
+  // Handle Line Type (Dashing)
+  // 0 and 15 are solid. 1-14 are broken lines.
+  bool draw = true;
+  if (lineType > 0 && lineType < 15) {
+    // HP-GL/2 style or similar: start with a mark.
+    // halfPeriod is defined by lineType.
+    unsigned halfPeriod = lineType + 2;
+    unsigned phase = static_cast<unsigned>(dashDistance / float(halfPeriod));
+    if (phase % 2 == 0) { // Skip if 0 (even), Mark if 1 (odd)
+      draw = false;
```
m9710797: I admit I don't fully understand this code. But at least this code is surprising to me. My expectation was that to draw a dashed line, you need to split up a long line into shorter segments. But if i read this code correctly, you only make a single `draw = true/false` decision for the full line (also for a very long line).

m9710797 (commented):
Some more review comments. Comments are more and more about details. Meaning the code is getting closer to being ready for merge.

How do you see this this? Do you still have many changes you like to make before you consider this ready for merge.

And it doesn't have to be perfect before the initial merge. We can still improve further later. And then it's also easier to let others help. Or what do you prefer? Would you like to continue reviewing/changing like we do now, or do you prefer that I do small changes myself directly on the code? Either is fine for me.

src/Plotter.cc
sndpl: yes, I removed them, it where kind of stubs but didn't need an own implementation.

src/Plotter.cc
sndpl: I made a gl::vec2 const of it

src/Plotter.cc
sndpl: you are right, i renamed it to penPosition, thats the most descriptive name for this var.

src/Plotter.cc
sndpl: real junior code :) I wrote it like this because drawing a dotted line is just a strait line in basic, see for example this example program:
```
7000 'Doted Line
7005 LPRINT
7020 LPRINT CHR$(27);"#"
7030 LPRINT "C0,L1"
7040 LPRINT "J160,0":LPRINT "P Black"
7050 LPRINT "R-232,-20"
7060 LPRINT "C1,L5"
7070 LPRINT "J160,0":LPRINT "P Blue "
7080 LPRINT "R-232,-20"
7090 LPRINT "C2,L9"
7100 LPRINT "J160,0":LPRINT "P Green"
7110 LPRINT "R-232,-20"
7120 LPRINT "C3,L13"
7130 LPRINT "J160,0":LPRINT "P Red  "
7140 LPRINT "R-232,-20"
7150 LPRINT "C0,L0"
7160 LPRINT "A"
```
So with L<number> you choose the size and with J it does a relative draw. So that's why choose to draw a line with the if in it to make the spaces between them.

src/Plotter.cc
sndpl: When I started the Plotter, I thought it was a good idea to kind of extend from the MSXPrinter, so I also copied that constructor from the printer and added some new settings to it. At the end maybe not the best choice, but as you said in one of your earlier comments it would maybe better to have a kind of code class above the printer/plotter.

src/Plotter.cc
m9710797: Sorry, I don't understand. Is the current emulation code correct or wrong? In other words: when you select a different line type (via the L<n> command), does the plotter then by itself raise/lower the pen while drawing a (long) line?

m9710797:
As already discussed on IRC: I think the code is in reasonably good shape. I'd like to help with getting the last details right, and maybe it's more efficient if I can make the required changes directly in the code myself?

Though before making changes, I would like to have a reasonable set of test cases (to make sure I don't accidentally break anything). Could you maybe prepare a set of BASIC programs together with the expected output (a .png file)?

src/Plotter.cc
sndpl: yes, this works now correctly.

sndpl:
Ok, i will create a disk with some demo programs and the png's. Already started with creating a demo and found a bug with printing all chars of the font. Will have a look at it and let attach the disk image and png when ready.

src/imgui/ImGuiSettings.cc
```diff
@@ -57,568 +59,683 @@ using namespace std::literals;
 
 namespace openmsx {
 
-ImGuiSettings::ImGuiSettings(ImGuiManager& manager_)
```
m9710797, line 60: The whole file is changed: reformatted in a way different from the usual openMSX formatting.
Please revert that (keep only your changes).

Reason I'm asking: I was trying to merge your work in master (I mean in a local branch to do some experiments), but this is giving non-trivial merge conflicts.

sndpl: @m9710797 fixed it, also added a test.zip file to this PR and commited some fixes of bugs I found during creating those tests.

m9710797:
Hi. I hope you've seen it. I've pushed a new branch 'plotter' that takes this PR as a starting point and makes several improvements / cleanups. See the commit messages for details (or ask for more info if anything is unclear).

Apart from some minor stuff, I'm finished with that work. For the next steps I'll need your help again.

Here's a list of still open items:

* The documentation mentions DIP switch 4 controls CR vs CR+LF. It can be controlled via the GUI, but the actual plotter code doesn't use this value yet?
* Similar for the `plotter-pen-thickness` setting. It can be controlled via the GUI, but the plotter code doesn't care about the value. The refactored code has a `penRadius()` method. So maybe now it's easy to implement this?
* I've added the new line rasterization code (before we would draw a line as a dense sequence of dots). At least in my opinion the result looks nice. Though one thing that's still missing is the "dwell time" when the pen is stationary at the start/end of a line. So currently the start/end points of a line don't look dark enough. For example if you draw a square, the start /end corner should look a little darker than the other 3 corners. Because most likely when the pen is lowered, it doesn't _immediately_ start moving, and when the 4 sides are drawn, the pen also isn't _immediately_ lifted. See also next point.
* There is a `penDown` variable. It's initialized to `true`, and tested in `drawLine()`, but nowhere in the current code is this variable ever changed. I think we need this to properly implement the previous point (dwell time).
* The `drawCharacter()` method has a `hasNextChar` parameter, but it isn't used currently. Should we use it? Or should we remove that parameter (and don't compute this in `executeGraphicCommand(), case 'P'`?
* In `executeGraphicCommand(), case 'F'`, are the bound checks missing for rotation 1-3?

sndpl:
Will try to look and compile the plotter branch tomorrow, and do some checks. Here are already some answers:

1) hmm, I thought I implemented this, but you are right, the processTextMode function doesn't use it for the 0x0D character. It should contain an IF that if it's on, do a CR/LF and else just go back the beginning of a line.
2) The plotter has support for 2 types of pens, I was sure I implemented it, but maybe the code has moved during refactoring. It's currently located in: https://github.com/sndpl/openMSX/blob/0e73986285252f67c5f3982d40440004d0200947/src/Plotter.cc#L698 There I changed the dotsize based on the sizeMultiplier that uses the setting. And the thickness of 1.5 is an estimate, since I cant find the specs about the real size of those ticker pens. I think this needs to move to an other function :)
3) Nice, will have a look at it when I compiled that branch
4) I think this is a left over of the dashed lines, in beginning I did this with the penDown/Up, but I changed that in the plotWithPen. But maybe it's indeed better to implement it properly
5) the hasNextChar isn't needed anymore in that function. 
6) hmm, that could be a bug indeed. Never checked that with printing a lot of text in other directions.

m9710797:
I fixed point 2 and 5. For the others I still need your help.

m9710797:
Progress on this PR got stuck. What can we do to unblock it?

m9710797:
I'm doing some final(?) test before merging to master.

One thing I noticed is that now the circle test program (`plotter.md` BASIC lines 5000-5140) draws line segments with _very_ clear dots at the start/stop of each segment. Maybe there are also dots on real hardware with real pen/paper, but possibly not as clear as these?

My idea with begin/end dots at line segments was to only draw them when the pen gets lowered/lifted (or when there's a significant pause when the pen is down .. but that requires more detailed time emulation, let's leave that out for now).

Can you take a look?

m9710797:
Replying to my own comment (from yesterday). The current implementation of the (gfx) `D` command does lower and raise the pen, resulting in those very visible dots on the line segments (in the circle example). But is that also how the real plotter behaves? Maybe a more logical(?) behavior would be to only lower the pen when not already lowered at the start of a `D` and `J` command, and leave it lowered at the end of the command. And then only lift the pen at the start of a `M` and `R` command (and maybe various other commands?).

Or is the only way to draw polylines (connected segments) without lifting the pen via a single `D/J` command with a whole list of coordinates?

So maybe the code is fine as is. But this is something I'd like to have checked on real hardware.

m9710797:
Merged into master. Thanks a lot for all the work and revisions!

PR #2070, opened 2026-01-13
profiler: use namespace ensemble

pvmm:


share/scripts/profiler.tcl
```diff
@@ -41,7 +43,7 @@ Syntax: symboltracer stop
 	}
 }
 
-proc _enter_function {name} {
+proc _enter_function name {
```
m9710797, line 46: Removing the `{}` around the single parameter is technically correct, the meaning is exactly the same.
Though the common style is to always use `{}` around the parameter list.

Issue #2075, opened 2026-01-20
Improve Tcl code completion of namespace functions


pvmm:
Change Tcl code completion so that the console doesn't expose inner functions and variables hidden in the namespace (through the `::` operator), providing a more user-friendly experience to the user.

<img width="644" height="543" alt="Image" src="https://github.com/user-attachments/assets/27fe4813-b07e-48a8-b527-124861bdc102" />

m9710797:
I think this needs more discussion/investigation. Because occasionally I do use tab-completion also on namespaced functions. A typical example is functions like "machine1::reset" so to give command to specific MSX machines (when you have multiple MSX machines active in multiple 'tabs').

Of course one outcome of the discussion could be that this use cases (multiple machines) is rare enough that it doesn't matter?

pvmm:
Isn't there a way to distinguish ensemble namespaces from conventional namespaces and hide the nitty-gritty details just for the ensemble namespaces to provide a more friendly interface to the user?

m9710797:
To be honest I have no idea (I'd have to investigate just like you). But it may indeed be related to the `namespace ensembles`, because e.g. the `record_channels` proc located in the `record_channels` namespace doesn't seem to have this problem.

pvmm:
> To be honest I have no idea (I'd have to investigate just like you). But it may indeed be related to the `namespace ensembles`, because e.g. the `record_channels` proc located in the `record_channels` namespace doesn't seem to have this problem.

What `record_channels` does is overwrite the namespace reference with a function of the same name, turning it effectively invisible to the global scope. This is actually a reason not to use ensembles and convert it to the old form, because a function called `symboltracer` that receives all commands would hide the symboltracer namespace.

pvmm:
Actually, a more thorough investigation suggests that **lazy.tcl** specifically registers all exported functions from `record_channels`. Since typing `record_channels`, partially or completely, triggers the autocomplete, it shadows the `record_channels::` namespace which is not scanned. Only when you type `record_channels::` the hidden namespace functions become visible to the tab-completion mechanism. So this is the solution: rename **script.tcl** to **_script.tcl** and declare all its exported functions in **lazy.tcl** and create a function that carries the same name of the namespace.

Since `symboltracer` is not shadowed by a function, it completes to the full `symboltracer::` prefix if you start typing its name.

m9710797:
The lazy-loading mechanism may appear to solve this, but only temporarily. Let me give some background.

[...]

PR #2077, opened 2026-01-23
Rename profiler.tcl to force lazy evaluation

pvmm:
* Lazy evaluation makes the interpreter autocomplete only what we want;
* Add `set_tabcompletion_proc` to evaluate subcommands;
* And now it works like `record_channels` (referencing #2075):
image

### DIFF
--- a/share/scripts/_profiler.tcl
+++ b/share/scripts/_profiler.tcl
@@ -43,7 +43,7 @@ Syntax: symboltracer stop
 	}
 }
 
-proc _enter_function name {
+proc _enter_function {name} {
 	# find function return address
 	set retaddr [peek16 [reg SP]]
 	# check if we create caller breakpoint for CALL or RST
@@ -57,7 +57,7 @@ proc _enter_function name {
 	}
 }
 
-proc _exit_function name {
+proc _exit_function {name} {
 	variable counters
 	# ignore a probable dangling breakpoint from a previous session
 	if {![dict exists $counters $name]} { return }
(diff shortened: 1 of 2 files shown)


share/scripts/_profiler.tcl
```diff
@@ -131,4 +131,24 @@ proc stop {} {
...
+		start -
+                stop { concat }
+                a* { concat add }
+                s* {
+                        concat start stop
+                }
+                default {
+                        concat start add stop
+                }
+        }
```
m9710797: This seems wrong, or at least more complex than needed.
* Usage of `concat` is not technically wrong, but weird. Is the intention to return a list? In that case better write it like `list add start stop`, this creates a list from the 3 given elements. The `concat` command OTOH interprets all given elements as lists and then joins these lists together into a new list. In this case `add`, `start` and `stop` can be parsed as a list of one element, so technically you'll get the same result in the end, but via a detour.
* It seems you are matching the start of the sub-command and then returning the possible remaining sub-commands. That's not neded. You can just always return `list add start stop`, and then (in the builtin tab-completion handling) we'll retain only those possibilities that still match.

EDIT:
Detail: makes no functional difference. There are two common styles to write this:
* `list add start stop`
* `list "add" "start" "stop"`

Both mean exactly the same, but the latter has nicer syntax highlighting, shows more clearly that the last 3 are meant as strings (and not e.g. sub-commands of the `list` command).
Pick which you like best.

m9710797 (changes requested):
Good idea. We should always try to use the lazy-loading-mechanism for Tcl scripts (to improve startup time). This is possible for most script (exceptions are scripts that must already do something at startup). I should have caught this during my review, but I didn't, sorry about that.

The lazy registration part of the PR is fine. Though this PR also includes a change related to `_tab_symboltracer`, and that part should be simplified.

m9710797:
Thanks.

PR #2079, opened 2026-01-30
colorblind theme for GUI (based on StyleColorsClassic)

jeangit:
It's for people with have difficulties with red , in the GUI.


m9710797:
Merged (with a typo fix). Thanks!

Issue #2085, opened 2026-02-10
Debugger: show sprite data in hexadecimal


Colpocorto:
When using the debugger for reverse engineering, showing coords or sprite numbers in decimal format is not very useful, since you normally compare values with those shown on the memory inspector or the disassembly window (where they are shown in hexadecimal format).

This forces the developer to keep converting between dec and hex all the time to check or compare values between different debugging windows.  Showing the values in hex makes much more sense.

A good compromise would be showing both simultaneously, for instance, using parentheses:

x=226 (0xe2) y=80 (0x50)
sprite=12 (0x0c) x=222 (0xde) y=80 (0x50) pat=12 (0x0c)

If there already is a way to achieve this, please close this ticket.

Thank you in advance

m9710797:
Implemented in f64a3aa59c453103656886fa2e2407554c6776a1.
Also thanks to pvmm for creating a few prototypes from which I could pick good ideas.

@Colpocorto please test, and close this issue if it's ok for you.

Colpocorto:
It does the trick, thank you.

PR #2086, opened 2026-02-11
Show sprite data as hexadecimal

pvmm:
https://github.com/openMSX/openMSX/issues/2085

### DIFF
--- a/src/imgui/ImGuiSpriteViewer.cc
+++ b/src/imgui/ImGuiSpriteViewer.cc
@@ -793,7 +793,7 @@ void ImGuiSpriteViewer::paint(MSXMotherBoard* motherBoard)
 					if (hovered) {
 						ImGui::Separator();
 						auto [hx, hy] = hoverPos;
-						ImGui::Text("x=%d y=%d", hx, hy);
+						ImGui::Text("x=%d, y=%d", hx, hy);
 						ImGui::Spacing();
 
 						for (int i : xrange(nrClippedBoxes)) {
@@ -803,7 +803,8 @@ void ImGuiSpriteViewer::paint(MSXMotherBoard* motherBoard)
 								if (drawBoundingBox && (boundingBoxOnAll == 0)) {
 									drawBox(b.x, b.y, b.w, b.h);
 								}
-								ImGui::Text("sprite=%d x=%d y=%d pat=%d", b.sprite, b.vramX, b.vramY, b.pattern);
+								ImGui::Bullet();
+								ImGui::StrCat("sprite n#: 0x", hex_string<2>(b.sprite), "\nx: 0x", hex_string<2>(hx), ", y: 0x", hex_string<2>(hy), ", pat: 0x", hex_string<2>(b.pattern));
 							}
 						}
 					}


src/imgui/ImGuiSpriteViewer.cc
```diff
@@ -793,7 +793,7 @@ void ImGuiSpriteViewer::paint(MSXMotherBoard* motherBoard)
 					if (hovered) {
 						ImGui::Separator();
 						auto [hx, hy] = hoverPos;
-						ImGui::Text("x=%d y=%d", hx, hy);
+						ImGui::Text("x=%d (%x) y=%d (%x)", hx, hx, hy, hy);
```
m9710797: Please do include the `0x` prefix (to make it clear that this is a repetition of the previous value, but now in hex).
It may also be a good idea to print leading zeros, so all in all replace `%x` with `0x%02x` (if indeed both are 8 bit values, i think that's the case here (but please check) but probably not for all values on line 806 below).

An alternative could be to use something like this (not tested):
```
ImGui::StrCat("x=", hx, " (0x", hex_string<2>(hx), ") y=", hy, " (0x", hex_string<2>(hy), ')');'
```

m9710797 (commented):
Later today when I'm back at home I'll test and merge this.

My only concern is that the string became longer and may no longer fit in the available space.

src/imgui/ImGuiSpriteViewer.cc
pvmm: hx and hy are integers, but sprite coordinates are always in their own 8-bit space coordinate, even when screen width is 512.

On line 806, b.vramX, b.vramY, b.pattern and b.sprite are all 8-bit variables.

m9710797:
I've tested it. Thanks for your effort, but I'm rejecting the patch in it's current form:
* Before this patch all info `sprite=12 x=123 y=123 pat=123` fits in the default window size(*1).
* After this patch only `sprite` and `x` fit. There appears a scrollbar, but there's no way to access it: when you move the mouse towards the scrollbar, the text disappears together with the scrollbar. So currently it's a clear regression in functionality.
* Only if you make the whole window wider does the new text fit, but then it's also wasting a lot of horizontal space in all other regions of this window.

If we want something like this we need to rethink the layout. Maybe just spreading this info over two lines could be a solution?


(*1) To restore the default window size, you can delete (or temporarily rename) your `imgui.ini` file.

BTW: Thanks for trying the solution with `ImGui::StrCat()`, but then for consistency we should also use that for the 2nd line. (`ImGui::StrCat()` runs a bit faster than `ImGui::Text()`, it doesn't need to parse the format string at run-time, it's already handled at compile-time).

m9710797 (changes requested):
See previous comment.

pvmm:
Added a bullet point to order sprite data in two lines, and if I have to chose between hexadecimal or decimal, I chose hexadecimal. So now it looks like this:
image

m9710797:
I've re-implemented this in f64a3aa59c453103656886fa2e2407554c6776a1.
I took your idea of using `ImGui::Bullet()`, but apart from it's totally different (e.g. bug showed wrong x,y coordinates).
I also show both decimal and hexadecimal.
Comments welcome of course.

PR #2091, opened 2026-02-28
fix: implement the dipswitch 4 logic for carriage return

sndpl:
This should fix the behaviour of the dip switch with the carriage return.

10 LPRINT "TEST";CHR$(&H0D);
20 LPRINT "JE"

When the switch is on, it prints:
TEST
JE

When the switch is off, it prints:
JEST

Because it only goes to the beginning of the line without the newline and then prints JE over the TEST.

### DIFF
--- a/src/Plotter.cc
+++ b/src/Plotter.cc
@@ -300,8 +300,16 @@ void MSXPlotter::processTextMode(uint8_t data)
 		ensurePrintPage();
 		penPosition = {0.0f, 30.0f};
 		break;
-	case 0x0d: // CR - Carriage return
+	case 0x0d: // CR - Carriage return, if dipswitch 4 is set, also do a line feed
 		penPosition.x = 0.0f;
+		if (getDipSwitch4()) {
+			penPosition.y -= lineFeed;
+			if (penPosition.y < -1354.0f) { // Page end reached
+				flushEmulatedPrinter();
+				ensurePrintPage();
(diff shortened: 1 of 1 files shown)


src/Plotter.cc
```diff
@@ -300,8 +300,16 @@ void MSXPlotter::processTextMode(uint8_t data)
...
+	case 0x0d: // CR - Carriage return, if dipswitch 4 is set, also do a line feed
 		penPosition.x = 0.0f;
+		if (getDipSwitch4()) {
+			penPosition.y -= lineFeed;
+			if (penPosition.y < -1354.0f) { // Page end reached
+				flushEmulatedPrinter();
+				ensurePrintPage();
+				penPosition.y = 30.0f; // Reset to top of new page
+			}
+		}
```
m9710797, line 312: I see you copy/pasted the behavior of `case 0x0a` here. That's indeed as expected.

But that behavior itself is unexpected (at least to me). If I understand the code correctly, when printing multiple lines of text, `penPosition.y` starts at `30`, then decreases (in steps of 9 or 18) until `-1354`. So according to this the "working area" is [-1354, +30]. Is that correct?

m9710797 (commented):
Thanks. I've merged your patch, though in a slightly different way (without duplicating the `LF` code).
But please also review my question about the `LF` behavior itself.

Issue #2092, opened 2026-03-07
Crash with SDL2-compat instead of normal SDL2


pvmm:
When OpenMSX is compiled in a system with both SDL2 and SDL3, it adds both as dependencies and crashes during initialization.

```
(gdb) r
Starting program: /usr/bin/openmsx 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
[New Thread 0x7ffff68d86c0 (LWP 23916)]

Thread 1 "openmsx" received signal SIGSEGV, Segmentation fault.
0x00007ffff6b37aeb in SDL_GetGrabbedWindow_REAL () from /lib64/libSDL3.so.0
[...]

PR #2093, opened 2026-03-11
Fix OpenMSX crash when compiled against SDL2-compat

pvmm:
OpenMSX calls `SDL_SetWindowGrab` before the video subsystem is initialised, which when compiled against the sdl2-compat compatibility layer causes a segfault. That didn't happen before because SDL2 has a macro called `CHECK_WINDOW_MAGIC` that initialised video subsystem before any attempt to reference it. But SDL2-compat doesn't use that macro and calls SDL3 functions directly, so `_this` in SDL3 is not initialised before `_this->grabbed_window` is referenced in `SDL_GetGrabbedWindow`, causing OpenMSX to abort.

This fixes issue #2092

### DIFF
--- a/src/events/InputEventGenerator.cc
+++ b/src/events/InputEventGenerator.cc
@@ -26,7 +26,6 @@ InputEventGenerator::InputEventGenerator(CommandController& commandController,
 		false, Setting::Save::NO)
 	, escapeGrabCmd(commandController)
 {
-	setGrabInput(grabInput.getBoolean());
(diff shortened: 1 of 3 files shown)


pvmm:
`grabinput `is not saved in settings.xml, so should not be restored when OpenMSX is starting.

m9710797:
Thanks!
I've merged it with some small edits (that's why github doesn't automatically detect it as merged).

Issue #2098, opened 2026-04-03
Proposal: automatic user trace creation from symbol manager


pvmm:
I propose a new checkbox in a column in the Symbol Manager called (tentatively) `function trace` that when checked creates an user trace automatically. It will also set an equivalent to `symboltracer add <name> $::sym(<name>)` for the specified symbol, basically closing the functional gap between symbol manager and trace viewer. A second parameter can be provided to the trace as the function parameter. For instance, writing `[reg A]` is equivalent to calling the command `debug trace add myfunction [reg A]`

<img width="1330" height="1047" alt="Image" src="https://github.com/user-attachments/assets/a1168537-352f-49b8-b14a-cb38433fb0a6" />

I believe this will encourage more developers to profile their own code using the new trace viewer widget. It also creates user traces for a selected number of symbols only instead of the whole symbol file, which is a lot heavier.

m9710797:
Thanks. I agree we should make this useful functionality more visible in the GUI.

I'm not yet sure if your proposal (an extra column in the symbol manager) is the best approach. Maybe it is, but let's also examine other options.

I didn't have time yet to give this much thought (and I probably also don't have much time this weekend). But my first idea was to insert a new section in the "Select probes and traces" window. Maybe a "function traces" section between "Probes" and "User traces". I don't know yet how that section should look .. maybe an "Add" button that opens a dialog where you can either select a symbol or manually type an address? Because of my concerns with the column in the "Symbol manager" approach was that this requires to have a symbol file (in most of my ad-hoc debug sessions I don't have a symbol file).

And in the symbol manager we could still have an entry like "Add function trace" in the right-click context menu.

Anyway, just an idea ... needs to be worked out further.

pvmm:
Here is something more akin to the context menu from the disassembly widget, but it applies to the symbol manager only. This looks better and doesn't clutter the interface. I think this and a new section in the "select probes and traces" window to add addresses directly that you mentioned could both exist. And this is the alternative shortcut version.
<img width="1330" height="1047" alt="Image" src="https://github.com/user-attachments/assets/2e3d77e0-3df1-4bc5-af23-4323a90450f4" />

pvmm:
New look
<img width="671" height="448" alt="Image" src="https://github.com/user-attachments/assets/6655ba5c-5f7a-4560-8867-f3a3cabcb791" />

m9710797:
To be honest, I don't like this context menu.
[...]

PR #2099, opened 2026-04-04
Implement proposal #2098 

pvmm:
Useful for future updates stemming from #2098

### DIFF
--- a/share/scripts/_profiler.tcl
+++ b/share/scripts/_profiler.tcl
@@ -4,6 +4,7 @@
 # to register functions.
 
 namespace eval symboltracer {
+namespace export add list remove start stop
 namespace ensemble create -prefixes 0
 
 variable names {}         ;# user traces
@@ -15,7 +16,7 @@ proc symboltracer_help {args} {
 	if {[llength $args] == 1} {
 		return {symboltracer is a profiler that traces function calls in the Trace Viewer widget
 
-Recognized commands: start, stop, add
+Recognized commands: start, stop, add, list, remove
 
 Type 'help symboltracer <command>' for more information about each command.
 }
@@ -39,6 +40,20 @@ Syntax: symboltracer add <name> <address>
 
 Syntax: symboltracer stop
 }}
+		"list" { return {Verifies if a user trace exists
+
+'list' returns all user traces created by symboltracer when executed without parameters or the address of the user trace if a user trace name is specified.
+
+Syntax: symboltracer list my_function
+        symboltracer list
+}}
+		"remove" { return {Removes user trace from the trace viewer
+
+'remove' deletes the user specified user trace and associated breakpoint.
+
+Syntax: symboltracer remove my_function
+}}
+		default { error "Unknown subcommand. Must be one of 'add', 'list', 'remove', 'start' or 'stop'." }
 	}
 }
 
@@ -73,10 +88,35 @@ proc _exit_function {name} {
 proc add {name addr} {
 	# add user-defined name/addr pair and create function breakpoints and traces
 	variable names
-	if {![dict exists $names $name]} {
-		dict set names $name {}
-		debug trace add $name
-		debug breakpoint create -address $addr -command [namespace code "_enter_function {$name}"]
+	# replace old entry
+	if {[dict exists $names $name]} {
+		catch {debug breakpoint remove [dict get $names $name]}
+		debug trace drop $name
+	}
+	set bp [debug breakpoint create -address $addr -command [namespace code "_enter_function {$name}"]]
+	dict set names $name $bp
+	debug trace add $name
+}
+
+proc remove {name} {
+	variable names
+	if {[dict exists $names $name]} {
+		debug trace drop $name
+		catch {debug breakpoint remove [dict get $names $name]}
+		dict unset names $name
+		variable counters
+		dict unset counters $name
+	}
+}
+
+proc list {{name {}}} {
+	variable names
+	if {$name ne {}} {
+		if {[dict exists $names $name]} {
+			return [dict get $names $name]
+		}
+	} else {
+		return [dict keys $names]
 	}
 }
 
@@ -133,9 +173,7 @@ proc stop {} {
 set_tabcompletion_proc symboltracer [namespace code _tab_symboltracer]
 
 proc _tab_symboltracer {args} {
-	list "start" "add" "stop"
+	::list "add" "list" "remove" "start" "stop"
 }
 
 } ;# namespace symboltracer
-
-namespace import symboltracer::*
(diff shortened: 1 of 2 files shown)


share/scripts/_profiler.tcl
```diff
@@ -74,12 +83,27 @@ proc add {name addr} {
...
+	if {$name ne {}} {
+		if {[dict exists $names $name]} {
+			return [dict get $names $name]
+		}
+	} else {
+		set result {}
+		dict for {name _} $names {
+			lappend result $name
+		}
+		return $result
```
m9710797: I haven't tried, but can this be simplified to:
`return [dict keys $names]`
But see also my other remark.

share/scripts/_profiler.tcl
```diff
@@ -39,6 +40,14 @@ Syntax: symboltracer add <name> <address>
 
 Syntax: symboltracer stop
 }}
+		"check" { return {Verifies if a user trace exists
+
+'check' returns all user traces created by symboltracer when executed without parameters or the address of the user trace if a user trace name is specified.
+
+Syntax: symboltracer check my_function
+        symboltracer check
+}}
```
m9710797, line 49: For consistency, maybe we should follow the design from `debug breakpoints`. There we have a sub command `debug breakpoints list`. That command returns all available info at once. For example it can return stuff like:
```
bp#1 {-address 1 -condition {} -command {debug break} -enabled 1 -once 0} bp#2 {-address 2 -condition {} -command {debug break} -enabled 1 -once 0}
```
In other words:
* The keys in the (top-level) dict are what your `check` command (without extra arguments) would return.
* The corresponding values are again a dict. It are key-value pairs. In your case This would be a dict with only one key-value pair `address 0x1234`. But I think soon(?) you want to extend this with extra parameters that should be traced?

m9710797 (commented):
I still have some questions.

In issue #2098 you talk about including parameters in the trace. Do you already know how to technically implement that? Because currently the tracing already logs an integer (the recursion depth). How will you combine that with extra parameter(s)?

If you indeed want to include an extra expression(?) to be logged, then it might be required to query (and maybe also edit) this expression. In that case I prefer that the `list <name>` sub command returns a dict (instead of a list). That way we can later easily add another key-value pair without breaking compatibility.

If I understood your patch correctly, then `list <name>` currently returns a list of two elements: `address breakpoint-id` (in the previous paragraph I proposed to change that to a dict (at least if you plan to extend this with an expression)). But in this paragraph I'd like to argue that the `address` part is redundant (can be queried from `breakpoint-id`). Often it's better to not store redundant info (then it cannot get in an inconsistent state).

pvmm:
I will probably need to add `-type` and `-format` to the optional expression  too.

share/scripts/_profiler.tcl
```diff
@@ -48,35 +63,100 @@ proc _enter_function {name} {
...
+			# evaluate expression and update user trace
+			set result [uplevel #0 [dict get $names $name expression]]
+			dict with names $name {
+				debug trace add $name $result -type $type -format $format
+			}
+		} else {
+			# update recursion count and update user trace
+			set result [expr [lindex [dict get $context $name] end] + 1]
+			debug trace add $name $result
+		}
```
m9710797: I'm surprised by this change. i was thinking of this expression stuff as purely an additional functionality. But here I see that when you have an expression it replaces the recursion counter. Wasn't the counter needed for correctness? E.g. for when to trace an empty value again (no recursion active) and cleanup some data structures. Ah, later I see that you replaced the counter with a stack (also for the empty expression case, ok).

Could we maybe share more code between the empty/non-empty expression case? Not sure if that's possible.

Technical detail, the `names` and `context` dict seem to always have the same key (is that true?). Could both be merged into a single dict? (But I don't know if that's better) (In `remove` you do clear `names` but not `context`, is that a bug?)

Did you think about functions with more than one parameter. E.g. where you want to trace both the A and the HL register. Setting an expression like `"[reg A] [reg HL]"` might work (notice the `""`). But then the `-type` and `-format` parameters you added don't make sense anymore (`-type` will be a string, but a general string can't be formatted with the current `-format`). A solution could be to make the user responsible for this, so drop the `-type` and `-format` stuff, and then if the user wants a specific format he should register an expression like `"[format %08b [reg A] [format %x [reg HL]"`.

share/scripts/_profiler.tcl
```diff
@@ -48,35 +63,100 @@ proc _enter_function {name} {
...
-	if {![dict exists $names $name]} {
-		dict set names $name {}
-		debug trace add $name
-		debug breakpoint create -address $addr -command [namespace code "_enter_function {$name}"]
+	if {![dict exists $names $name]} { return }
+	return [dict get $names $name $key]
+}
+
+# used internally (ImGUI)
+proc _replace {name key value} {
```
m9710797: Ideally, everything that's possible via the GUI should also be possible via the console. E.g. breakpoint properties can be changed via the `configure sub command.

It's nice that a function profiler can be written as a Tcl script. But if it becomes too extensive (e.g. if ImGui integration is too difficult) than we can consider moving it to c++. But i think for now, Tcl is still fine.

src/imgui/ImGuiSymbols.cc
```diff
@@ -207,6 +207,19 @@ void ImGuiSymbols::drawContext(MSXMotherBoard* motherBoard, const SymbolRef& sym
...
+	auto cmdResult = manager.execute(makeTclList("symboltracer", "list", sym.name(symbolManager)));
+	bool traceAsFunction = cmdResult && cmdResult->getString().size() > 0;
+
+	if (ImGui::Checkbox("Trace as a function", &traceAsFunction)) {
+		TclObject cmd = makeTclList("symboltracer", traceAsFunction ? "add" : "remove", sym.name(symbolManager));
+		if (traceAsFunction) {
+			cmd.addListElement(sym.value(symbolManager));
+		}
+		manager.execute(cmd);
+	}
```
m9710797, line 222: See my comment on #2098. I would propose to simplify this to just a single fixed line "Add function trace". So not even a possibility to remove the trace, same as the current "Set breakpoint".
And then the real ImGui function tracing integration is done in the "Select probes and traces" window.

share/scripts/_profiler.tcl
pvmm: This is getting out of scope. I added expressions because user traces defined them in the first place and also their format, not the other way. If we make format user-defined, I think the data should be saved somewhere because it would be a pain to add them all over again when restarting OpenMSX.

src/imgui/ImGuiSymbols.cc
pvmm: Here is an idea for starters.
image

pvmm: > See my comment on #2098. I would propose to simplify this to just a single fixed line "Add function trace". So not even a possibility to remove the trace, same as the current "Set breakpoint".

Breakpoints and traces are different beasts. You can have multiple breakpoints with same conditions, addresses and commands. But a "function trace" is unique because it has a name. An "add function trace" menu item does nothing if a function trace already exists. I think a checkbox would make more sense.

PR #2103, opened 2026-04-12
Proposal: a second debuggable for ROMs with 16-bit segment indexes

pvmm:
This proof of concept changes RomAscii16X ROM device to add a second debuggable to allow Tcl scripts to access the remaining bits of a 16 bit segment index in a compatible manner.

The low 8-bit segment index is still accessible through:
```debug read {MSXdev25_ASTEROIDS_v1.0.rom romblocks} <address>```

but the remaining bits of the segment index are accessible through: 
```debug read {MSXdev25_ASTEROIDS_v1.0.rom romblocks16} <address>```

Since this is a proposal, I have set to draft, but if this is accepted as the right approach, I intend to do the same to other ROM like Yamanooto in the same PR.

### DIFF
--- a/src/imgui/ImGuiDebugger.cc
+++ b/src/imgui/ImGuiDebugger.cc
@@ -461,12 +461,12 @@ ImGuiDebugger::SlotInfo ImGuiDebugger::getSlotInfo(MSXCPUInterface& cpuInterface
 		const auto* device = cpuInterface.getVisibleMSXDevice(page);
 		if (const auto* mapper = dynamic_cast<const MSXMemoryMapperBase*>(device)) {
 			r.segment = strCat(mapper->getSelectedSegment(page));
-		} else if (auto [rom, romBlocks] = ImGuiDisassembly::getRomBlocks(debugger, device); romBlocks) {
+		} else if (auto [rom, debuggable] = ImGuiDisassembly::getDebuggable(debugger, device); debuggable) {
 			if (unsigned blockSize = RomInfo::getBlockSize(rom->getRomType())) {
 				auto addr = 0x4000 * page;
 				char separator = 'R';
 				for (int offset = 0; offset < 0x4000; offset += blockSize) {
-					auto seg = romBlocks->readExt(addr + offset);
+					auto seg = debuggable->readExt(addr + offset);
 					strAppend(r.segment, separator, strCat_if(seg != unsigned(-1), seg).else_('-'));
 					separator = '/';
 				}
--- a/src/imgui/ImGuiDisassembly.cc
+++ b/src/imgui/ImGuiDisassembly.cc
@@ -66,14 +66,14 @@ void ImGuiDisassembly::setGotoTarget(uint16_t target)
 	setDisassemblyScrollY.reset(); // don't restore initial scroll position
 }
 
-std::pair<const MSXRom*, RomBlockDebuggableBase*>
-	ImGuiDisassembly::getRomBlocks(Debugger& debugger, const MSXDevice* device)
+std::pair<const MSXRom*, RomBlockDebuggableBase::Debuggable16*>
+	ImGuiDisassembly::getDebuggable(Debugger& debugger, const MSXDevice* device)
 {
-	RomBlockDebuggableBase* debuggable = nullptr;
+	RomBlockDebuggableBase::Debuggable16* debuggable = nullptr;
 	const auto* rom = dynamic_cast<const MSXRom*>(device);
 	if (rom && !dynamic_cast<const RomPlain*>(rom)) {
-		debuggable = dynamic_cast<RomBlockDebuggableBase*>(
-			debugger.findDebuggable(tmpStrCat(rom->getName(), " romblocks")));
+		debuggable = dynamic_cast<RomBlockDebuggableBase::Debuggable16*>(
+			debugger.findDebuggable(tmpStrCat(rom->getName(), " romblocks16")));
 	}
 	return {rom, debuggable};
 }
@@ -98,8 +98,8 @@ struct CurrentSlot {
 		const auto* device = cpuInterface.getVisibleMSXDevice(page);
 		if (const auto* mapper = dynamic_cast<const MSXMemoryMapperBase*>(device)) {
 			result.seg = mapper->getSelectedSegment(narrow<uint8_t>(page));
-		} else if (auto [_, romBlocks] = ImGuiDisassembly::getRomBlocks(debugger, device); romBlocks) {
-			result.seg = romBlocks->readExt(addr);
+		} else if (auto [_, debuggable] = ImGuiDisassembly::getDebuggable(debugger, device); debuggable) {
+			result.seg = debuggable->readExt(addr);
 		}
 	}
 	return result;
--- a/src/memory/RomAscii16X.cc
+++ b/src/memory/RomAscii16X.cc
@@ -22,6 +22,7 @@ namespace openmsx {
 RomAscii16X::RomAscii16X(DeviceConfig& config, Rom&& rom_)
 	: MSXRom(config, std::move(rom_))
 	, debuggable(*this)
+	, debuggableExt(*this, getName() + " romblocks ext")
 	, flash(rom, AmdFlashChip::S29GL064S70TFI040, {}, config)
 {
 	reset(EmuTime::dummy());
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

**Comment by MBilderbeek** [comment]
My initial thoughts was to make the 2nd debuggable named "<device> romblocks16" to explicitly indicate the 16-bit values. For even addresses it would be backwards compatible with the original 8-bit variant and for odd addresses (+1) the upper 8-bits would be accessible.

**Comment by MBilderbeek** [comment]
This is related to #2101

**Comment by m9710797** [comment]
Thanks for working on this.

I had two alternatives in mind:
* Adding a 2nd debuggable which exposes bits 8-15.
* Adding a 2nd debuggable which has bits 0-7 on even and bits 8-15 on odd addresses (IOW 16-bit little endian).

I think I like the 2nd version better, because it's a little bit more efficient to use in a Tcl script: then you can read both bytes at once with "debug read_block", and "binary scan" to extract the 16 bit value. Though the disadvantage is that you have to ensure you read from an even address.

I think we should not only implement this for RomAscii16X and other large mappers, but for _all_ mappers, even if the high byte will always be zero. Reasoning: generic code should not have to know about the details of a mapper (is it large or not).

I've only thought about this for a few minutes, maybe the following doesn't work. But my implementation idea would be:
* Currently `RomBlockDebuggableBase` inherits from `SimpleDebuggable`. Remove that.
* Instead `RomBlockDebuggableBase` will contain 2 `SimpleDebuggable` member variables. One for the existing debuggable, one for the new 16-bit version. E.g. look at `VDP.hh`, `RegDebug` and `StatusRegDebug` for an example of how to implement multiple debuggables in one class.
* Both these embedded `SimpleDebuggable` members can access the outer object `RomBlockDebuggableBase` (search for examples using `OUTER` in our code, this saves having to store a pointer in the 2 objects).
* Via this outer object, you can call the existing `virtual unsigned readExt()` member function from `RomBlockDebuggableBase`, this does already gives you the full segment number (e.g. the ImGui debugger uses this).
* One of the two `SimpleDebuggable` objects only exposes the low 8 bit, the other low and high at even/odd addresses.
* I think this way we _only_ need to change the `RomBlockDebuggable` file, and none of the mapper code. And it immediately works for all mappers, even small ones.

But again: just an idea, not tested at all.

File: src/imgui/ImGuiDebugger.cc [hunk 1]
```diff
@@ -461,12 +461,12 @@ ImGuiDebugger::SlotInfo ImGuiDebugger::getSlotInfo(MSXCPUInterface& cpuInterface
...
 		if (const auto* mapper = dynamic_cast<const MSXMemoryMapperBase*>(device)) {
 			r.segment = strCat(mapper->getSelectedSegment(page));
-		} else if (auto [rom, romBlocks] = ImGuiDisassembly::getRomBlocks(debugger, device); romBlocks) {
+		} else if (auto [rom, debuggable] = ImGuiDisassembly::getDebuggable(debugger, device); debuggable) {
 			if (unsigned blockSize = RomInfo::getBlockSize(rom->getRomType())) {
 				auto addr = 0x4000 * page;
 				char separator = 'R';
 				for (int offset = 0; offset < 0x4000; offset += blockSize) {
-					auto seg = romBlocks->readExt(addr + offset);
+					unsigned seg = (debuggable->read(addr + offset + 1) << 8) | debuggable->read(addr + offset);
```
m9710797: Why this change?
* Before we directly used `readExt()`.
* Now we do 2 read() calls, each of which internally uses `readExt()` and return part of it, and then recombine it.

That seems like an unneeded detour. Internally in openMSX we're not limited to using the "debuggable API". if there's a better API then we can use that.

src/imgui/ImGuiDisassembly.cc
```diff
@@ -98,8 +98,9 @@ struct CurrentSlot {
 		const auto* device = cpuInterface.getVisibleMSXDevice(page);
 		if (const auto* mapper = dynamic_cast<const MSXMemoryMapperBase*>(device)) {
 			result.seg = mapper->getSelectedSegment(narrow<uint8_t>(page));
-		} else if (auto [_, romBlocks] = ImGuiDisassembly::getRomBlocks(debugger, device); romBlocks) {
-			result.seg = romBlocks->readExt(addr);
+		} else if (auto [_, debuggable] = ImGuiDisassembly::getDebuggable(debugger, device); debuggable) {
+			addr = (addr & ~1);
+			result.seg = (debuggable->read(addr + 1) << 8) | debuggable->read(addr);
```
m9710797: Same for all these changes.

src/memory/RomAscii16X.cc
```diff
@@ -74,10 +77,10 @@ byte* RomAscii16X::getWriteCacheLine(uint16_t /* addr */)
 	return nullptr; // not cacheable
 }
 
-unsigned RomAscii16X::Debuggable::readExt(unsigned address)
+unsigned RomAscii16X::Blocks::readExt(unsigned address)
 {
-	auto& outer = OUTER(RomAscii16X, debuggable);
-	return outer.bankRegs[((address >> 14) & 1) ^ 1];
+	uint8_t index = ((address >> 14) & 1) ^ 1;
+	return (mappedHigh[index] << 8) | mappedLow[index];
```
m9710797: Why these changes? The existing `readExt()` implementation was already working fine, no? And more efficient than this new version.

src/memory/RomAscii16X.cc
```diff
@@ -87,7 +90,8 @@ void RomAscii16X::serialize(Archive& ar, unsigned /*version*/)
 	ar.template serializeBase<MSXDevice>(*this);
 
 	ar.serialize("flash",       flash,
-	             "bankRegs",    bankRegs);
+	             "mappedLow",   romBlockDebug.mappedLow,
+	             "mappedHigh",  romBlockDebug.mappedHigh);
```
m9710797: This breaks backwards compatibility with old savestates. Please revert.

src/memory/RomBlockDebuggable.hh
```diff
@@ -2,34 +2,65 @@
...
+		{
+		}
+		[[nodiscard]] byte read(unsigned address) override
+		{
+			auto& outer = OUTER(RomBlockDebuggableBase, debuggable);
+			return narrow_cast<byte>(outer.readExt(address));
+		}
+        } debuggable;
+
+        struct DebuggableExt final : Debuggable {
```
m9710797: Detail. I'd suggest to directly inherit from `SimpleDebuggable` (instead of `Debuggable`). Because apart from the size `0x10000` you don't reuse anything from the intermediate `Debuggable` class. And then both classes can be made `final`.

May I suggest the names `Debuggable8 debuggable8` and `Debuggable16 debuggable16` (so rename both the class and the variable, and include 8 or 16 in both).

src/memory/RomBlockDebuggable.hh
```diff
@@ -2,34 +2,65 @@
...
+		}
+		[[nodiscard]] byte read(unsigned address) override
+		{
+			auto& outer = OUTER(RomBlockDebuggableBase, debuggable16);
+			return narrow_cast<byte>(address & 1 ? outer.readExt(address) >> 8 : outer.readExt(address));
+		}
+        } debuggable16;
+
+	std::array<uint8_t, 2> mappedLow = {0, 0};
+	std::array<uint8_t, 2> mappedHigh = {0, 0};
```
m9710797: These 2 variables are leftovers from a previous version?

m9710797 (commented):

I only did a very quick review, without doing any tests. It's possible I'm overlooking something of course.

However I think you should revert most of your changes. I think you **only** need to change the `RomBlockDebuggable.hh` file, nothing else. Your current changes in that file look mostly OK, only some minor suggestions.

src/imgui/ImGuiDebugger.cc
pvmm: Because the returned debuggable doesn't have `readExt` any longer since RomBlockDebuggableBase is no longer a debuggable, so `ImGuiDisassembly::getRomBlocks` doesn't return a RomBlockDebuggableBase (I renamed it to `getDebuggable`). But I fixed it by extending Debuggable16 to include a `readExt` function.

src/memory/RomAscii16X.cc
pvmm: It was remnant of an older version that separated bank registers differently because a debuggable's read() returned only 8bits.

m9710797:
I took this PR as a starting point, heavily simplified it, and committed it as faa9549b4273d8acbb2c8a8d4e9bed4065c6146e.
Thanks for the starting point.

PR #2105, opened 2026-04-18
fix(plotter): enforce bounds on pen position during rotation

sndpl:
- Update pen position calculations to ensure it does not exceed defined borders.
- Implement clamping for X and Y coordinates based on rotation direction.
- Add test cases in documentation to verify bounds handling for each rotation.


m9710797:
Thanks!

PR #2107, opened 2026-04-18
fix(plotter): set default size for plotter viewer window

sndpl:
- Add initial window size of 480x720 for the plotter viewer
- Ensure size is applied on first use to enhance user experience

### DIFF
--- a/src/imgui/ImGuiPlotterViewer.cc
+++ b/src/imgui/ImGuiPlotterViewer.cc
@@ -33,6 +33,7 @@ void ImGuiPlotterViewer::paint(MSXMotherBoard* motherBoard)
 {
 	if (!show) return;
 
+	ImGui::SetNextWindowSize({480, 720}, ImGuiCond_FirstUseEver);
 	im::Window("Plotter viewer", &show, [&]{
 		if (!motherBoard) {
 			ImGui::TextUnformatted("No machine present.");


src/imgui/ImGuiPlotterViewer.cc
```diff
@@ -33,6 +33,7 @@ void ImGuiPlotterViewer::paint(MSXMotherBoard* motherBoard)
 {
 	if (!show) return;
 
+	ImGui::SetNextWindowSize({480, 720}, ImGuiCond_FirstUseEver);
```
m9710797, line 36: Not strictly needed, but often we select the initial window size based on the selected font size (default font size = 13). That works well for windows that mostly show text, but that's not the case for this one. On the other hand, if the user prefers a larger font size, he might prefer all his windows to be larger.

What do you prefer? Fixed size is also fine for me (for this window). Let me know and then I'll merge.

Can we merge to master after this?

sndpl: @m9710797 i choose for 480×720 as it is a portrait aspect that comfortably fits A4 paper (paper aspect 1050:1485 ≈ 0.707) plus the control panel, matching the style of other viewer windows (BitmapViewer 528×620, CharacterViewer 692×886, etc.).

and yes, we can merge to master after this 👍🏻

m9710797:
I'll try to do the merge to master soon (tomorrow?).

PR #2110, opened 2026-04-22
fix(plotter): manage pen lift state and real plotter font

sndpl:
- Add flushPendingPenUp method to handle deferred pen lifts.
- Update setPenDown to manage pen state and prevent redundant actions.
- Ensure drawing paths are correctly terminated on mode changes and color selections.


m9710797:
This simplification looks ok.

It's been too long, I forgot the details... What was the status of this PR? Was this the only pending issue? Anything left before we can merge into master branch?

sndpl:
This is the last fix. The font looks good, just like the manual and you can also see the dots of the pen down. So for me the feature is complete now.

PR #2114, opened 2026-04-25
Add missing tcl functions found in help command

pvmm:
Also added new line after each command because it seems to be the standard.


m9710797:
Thanks a lot!

PR #2120, opened 2026-05-05
Search for file with ROM extension in zip file 

pvmm:
When loading a ROM file as a cartridge, the contents of a supplied zip file are checked until a .ROM extension is found. The matching file is then deflated.

### DIFF
--- a/src/file/File.cc
+++ b/src/file/File.cc
@@ -15,7 +15,7 @@ namespace openmsx {
 
 File::File() = default;
 
-[[nodiscard]] static std::unique_ptr<FileBase> init(zstring_view filename, File::OpenMode mode)
+[[nodiscard]] static std::unique_ptr<FileBase> init(zstring_view filename, File::OpenMode mode, zstring_view extension = {})
 {
 	static constexpr std::array<uint8_t, 3> GZ_HEADER  = {0x1F, 0x8B, 0x08};
 	static constexpr std::array<uint8_t, 4> ZIP_HEADER = {0x50, 0x4B, 0x03, 0x04};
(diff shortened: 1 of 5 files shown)


m9710797:
Thanks for your PR. Though this is a strategy I'd like to avoid. I'll try to explain my reasoning:

* You added an extra (optional) parameter to the `File` constructor.
* But now it matters from where in the code you open your .zip file.
* For example, I didn't check the code but I can imagine that in some cases we have slightly different code paths for files passed via the command line, via the console, via the GUI, via savestates or replays, ...
* So for example if on one of these paths we don't list the exact same file extension(s), it could be that at first loading the rom worked fine, but e.g. not when later loading a savestate.

My proposal would be to make opening .zip files less dependent from the context. So no changes in `File.hh` and `File.cc`, `Rom.cc` or even in `ZipFileAdapter.hh`. Only some changes in `ZipFileAdapter.cc`. So an internal implementation detail. And that implementation then is:
* Not always pick the first file (current implementation).
* But pick the first one that has a "known" extension. Or still the first file if there is otherwise no match.
* And then have a **fixed** set of "known" file extensions.

To get that set of known extensions we can call:
* `DiskImageCLI::getExtensions()`
* `MSXRomCLI::getExtensions()`
* `CassettePlayerCLI::getExtensions()`

This fixed set of known extensions is not the most pretty. And it does make our otherwise generic `ZipFileAdapter` class specific to openMSX (and it does introduce a dependency in the "wrong" direction, from `src/file/` onto the more specific `src/{memory,fdc,cassette}/` directories). That's ugly, but I think it's better than trying to list the correct set of extensions on every `File` constructor.

What do you think?

PR #2121, opened 2026-05-05
video: replace GLEW with GLAD

### LINKED ISSUE #1906: Migrate from GLEW to libepoxy or GLAD

cmitu:
GLAD (https://github.com/Dav1dde/glad) is a well known and widely used (E)GL(X|ES)/Vulkan/WGL loader and generator. The changes replace GLEW with ustom header-only GLAD implementation generated via https://gen.glad.sh/ with the following options:

 * APIs
    - GL: 3.3, compatibility
    - GLES: 2.0

 * Options:
    - alias (on)
    - header only (on)
    - merge (on)

NOTE: the options used for the GLAD header creation are part of the `gl.h` file itself. In the future, if need be, other options may be useful (i.e. Vulkan instead of GL/EGL for Android).

Dynamic loading of the GL(ES) libraries is done with the SDL's `SDL_GL_GetProcAddress` function, so the 'loader' option was not used when generating the GLAD header.

Some other notes about the changes:

* the `GLRenderer` component was removed from the list of build components since the build doesn't need a GL library present and openMSX is already requiring a GL enabled renderer since version 19.1. I modified the `meson.build` file, although I understand it's not used right now, but nevertheless I removed the `glew`/`gl` references from it.
* I left the GLES load path even if right now it's not a build option nor a runtime check/configuration options. I did test the path by manually modifying `GLUtil.hh` to force GLES and it worked (tested on a Raspberry Pi4, which suports GLES 3.2 and GL 2.1).

Closes: #1906
Supercedes: #1911 

Build artifacts (except macOS ones) are available from my branch [here](https://github.com/cmitu/openMSX/actions/runs/25356570827).


src/video/VisibleSurface.cc
```diff
@@ -110,35 +112,16 @@ VisibleSurface::VisibleSurface(
...
-	}
-	if (!GLEW_VERSION_2_1) {
-		throw InitException(
-			"Need at least OpenGL version " VERSION_STRING);
-	}
+	// Load the GL APIs with GLAD
+#if OPENGL_VERSION == OPENGL_ES_2_0
+	gladLoadGLES2((GLADloadfunc) SDL_GL_GetProcAddress);
+	if (!GLAD_GL_ES_VERSION_2_0)
+		throw InitException("Need at least OpenGL version " VERSION_STRING);
```
MBilderbeek, line 119: A tiny thing, but shouldn't it report about the OpenGL ES version instead of "normal" OpenGL?

MBilderbeek:
Thanks a LOT for this work! 

I can't really judge whether the correct options were used, so I hope @mthuurne and @m9710797 will take a look at that. But in general this looks good to me. @mthuurne and @m9710797 what do you think?

src/video/VisibleSurface.cc
cmitu: I'm not really sure what you mean here, but I think you mean one of:

* GLAD will populate the `GLAD_GL_ES_VERSION_..` defines when the GLES loader is called, but not the `GLAD_GL_VERSION_..` ones (which will remain set to `0`), so we cannot use (uniformly) a check just for `GLAD_GL_VERSION_2_0`/`GLAD_GL_VERSION_2_0`.

* `VERSION_STRING` is defined in sync with `OPENGL_VERSION`, so it's _openGLES 2.0_ when `OPENGL_VERSION` is `OPENGL_ES_2_0`, thus if an initialization error occurs, the correct version string will be included in the exception message. See here for the define(s): https://github.com/openMSX/openMSX/blob/9824509fbf0286721ec4995d30dc53adf8554fda/src/video/VisibleSurface.cc#L77-L95

cmitu:
While trying to include a license file with the generated headers, I found there are a few incompatibilities betwen GPL v2 and the GLAD generated code's own license.

The GLAD generated code is under _WTFPL or CC0 **and** Apache 2.0_. From GLAD's [project page](https://github.com/Dav1dde/glad/):

> The generated code from glad is any of Public Domain, WTFPL or CC0. Now Khronos has some of their specifications under Apache Version 2.0 license which may have an impact on the generated code, https://github.com/KhronosGroup/OpenGL-Registry/issues/376#issuecomment-596187053 on the Khronos / OpenGL-Specification issue tracker.

The official (from ASF) stance is that GPL v2 is not compatible with Apache 2.0 (https://www.apache.org/licenses/GPL-compatibility.html), though it is compatible with GPL v3 (same doc as earlier). I have checked a few of the open source project I posted as examples and they appear to be GPL v3 (i.e. ScummVM or GearGrafx/GearColeco).

I'm trying to see if there's a way around this, but I think that even if openMSX's license is considered GPL v2+, then including/linking with the generated GLAD code would mean the entire product to be GPL v3.

MBilderbeek:
Well, the problem here is that openMSX is explicitly licensed under GPL2 only. :(

src/video/VisibleSurface.cc
MBilderbeek: I mean I think that the error message for OpenGL ES should mention OpenGL ES ("`Need at least OpenGL ES version `"). So, just a cosmetic thing.

cmitu: OK, so this is already covered, since `VERSION_STRING` is set to _openGL ES 2.0_ on the GLES path, so the resulting message mentions GLES 2.0 as minimum.

cmitu:
@MBilderbeek yes, indeed. 

I'll close this PR, but if you (or anyone else) find a way to include GLAD (or perhaps I'm wrong w.r.t. licensing), feel free to make use of the changes here.

I've looked at a few other similar projects while investigating the licensing issue, I'll just leave here some notes: 

* the [Khronos API headers](https://github.com/KhronosGroup/OpenGL-Registry/tree/main/api)  themselves are distributed under MIT. These are just the headers though, there's no loader included here.

* [glbinding](https://github.com/cginternals/glbinding) is MIT licensed, even if it claims to be generating the API from the same source as GLAD (the `gm.xml` distributed in https://github.com/KhronosGroup/OpenGL-Registry).
  The GLAD license for generated code was WFTPL/CC0 _until_ 2020 when someone pointed out the license for `gm.xml` being Apache 2.0, thus it changed the lices adding Apache 2.0.

* [libepoxy](https://github.com/anholt/libepoxy) is also MIT licensed. The project mentions Khronos' license but I think it's referencing the MIT license included in the headers.

* Imgui has also a minimal GL loader (see https://github.com/ocornut/imgui/issues/4445), but it's internal only and GL API headers are still needed at build time.

MBilderbeek:
Thank you very much for your excellent work on creating the patch, doing a lot of research (also around the license) and communicating so clearly about all of this. I certainly hope to see more contributions from you in the future!

PR #2122, opened 2026-05-06
Fixes library version probing

gilbertfrancois:
When running ./configure, freetype was not returning its version number. The log file `derived/x86_64-linux-opt/config/probe.log` shows the following error:
```
FREETYPE: Found header
FREETYPE: Found lib
Error executing "pkg-config freetype2 --ftversion"
pkgconf: unknown option -- ftversion
Execution failed with exit code 1
```

After applying the fix, the library returns the version number, e.g.:
```
$ ./configure
...
Found libraries:
  ALSA:             version 1.2.15.3
  FreeType:         version 2.5.1
  GLEW:             version unknown
...
```

### DIFF
--- a/build/libraries.py
+++ b/build/libraries.py
@@ -189,7 +189,7 @@ class FreeType(Library):
 	@classmethod
 	def getVersion(cls, platform, linkStatic, distroRoot):
 		configScript = cls.getConfigScript(platform, linkStatic, distroRoot)
-		return '`%s --ftversion`' % configScript
+		return '`%s --version`' % configScript
 
 class GL(Library):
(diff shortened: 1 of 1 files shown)


MBilderbeek:
Apparently something is wrong with internet connectivity of some runners... I'd like to see them all checked before merging. @mthuurne  can you also take a quick look? (It looks fine to me, but I'm not an expert at all.)

gilbertfrancois:
It seems that the runners have been healed.

MBilderbeek:
I'll already merge it. If @mthuurne has any objections, he's free to revert or refine it.

PR #2124, opened 2026-05-10
C++ documentation encounrages const-catching

pvmm:


src/cassette/CassettePlayer.cc
```diff
@@ -376,7 +376,7 @@ void CassettePlayer::insertTape(const Filename& filename, EmuTime time)
 			try {
 				// first try WAV
 				newImage = std::make_unique<WavImage>(filename, filePool);
-			} catch (MSXException& e) {
+			} catch (const MSXException& e) {
 				msgWav = std::move(e).getMessage();
```
m9710797, line 380: Thanks for you effort. Though I'm not sure this is generally a good idea.
* I understand that conceptually exception object should not be modified, and then adding `const` is a good idea.
* However a common pattern is to turn an exception into another, or take the message from one exception and doing something else with it. When the exception object is `const` you're forced to make a copy of the internal `string`. Before we could recycle (=move) that `string`. In other words, currently your PR is a (small) pessimization.

You say "C++ documentation encounrages const-catching". Can you point me to such documentation. Personally, or at least in the projects I've worked on, we never had such a rule. But I can see the point. Anyway before adopting that rule in openMSX (and maybe skip the places where `std::move()` is beneficial?) I'd like to read more about the rationale those other projects have. Can you point me to that "c++ documentation". Thanks.

m9710797 (commented):
Thanks. Though before accepting this I need more convincing this is actually a good idea. (Because it's also a pessimization in some places).

m9710797:
Thanks again. But I must say, I'm not sure this is an improvement.
In general it's good to be const-correct. But in this case it does make it slightly more difficult to correctly (as in: optimally) use the caught exception object: why catch sometimes by const-reference, sometimes by non-const-reference.

What might be a good idea is to reduce the number of `MSXException` subclasses. Then there will be less need to convert between exception types.

pvmm:
I closed this because I cannot find evidence to support the claim I got from this [article](https://www.codestudy.net/blog/catching-exceptions-by-const-value-in-c-compilers-diverge/)

I can't get the same ambiguous results from the article when I compile with gcc against clang.

m9710797:
I only quickly read the article, but it seems to talk about a difference in catching "const **value**" vs "non-const **value**" (not const vs non-const **reference**)?

pvmm:
It has this to say about the issue:

By const reference (catch (const MyException& e)): Binds e to the exception object as a const reference. Avoids copy overhead and prevents modification—**often the safest and most efficient choice.**

But I couldn't find any difference in practice.

PR #2125, opened 2026-05-13
feat(wavegame): add Wave Game audio cartridge support

### LINKED ISSUE #1975: Support WAVE-GAME protocol from MSX-Pico
There is now a popular feature in the MSX Pico, that enables games to play music from WAV files stored on the MSX-Pico.

We shouldn't start emulating that whole MSX-Pico, because it's a complete computer on an MSX cartridge.

But the WAVE-GAME protocol could be emulated at a high level, as a special/specific device (independent from MSX-Pico).

This is the documentation:
```
This document describes the Wave Game interface of the MSX Pico cartridge. This makes it possible to get high quality music during gameplay.

For the control commands, output port 92h is used.


For firmware versions v1.49 / v2.05 / v2.06 this is valid:

Read from RAM address 0hF91E will return 0

Commands (in binary):

00000000 : stop music
0000xxxx : fade out music in xxxx seconds
01xxxxxx : play song xxxxxx once (for instance command 01000101 will play 05.wav once)
10xxxxxx : play song xxxxxx in a loop (for instance command 10000101 will play 05.wav continuously)
11xxxxxx : toggle pause (lower 6 bits are ignored)


For firmware v1.50 / v2.07 and higher this is valid:

Read from RAM address 0hF91E will return 1

Commands (in binary):

00000000 : stop music (actually a 1-second fade out)
0000xxxx : fade out music in xxxx seconds
01xxxxxx : play song xxxxxx once (for instance command 01000101 will play 05.wav once)
10xxxxxx : play song xxxxxx in a loop (for instance command 10000101 will play 05.wav continuously)
11000000 : toggle pause
11110000 : play pause.wav sound (use 11100000 to continue previous song)
11100000 : continue previous song with a 1-second fade in
11010000 : next command received will be temporarily stored and executed after the current song stops playing, or already has stopped

Every xx.wav file can have an optional companion xx.cfg file, which should contain the lines:

Loop offset (used with command 01xxxxxx)
Optional start offset (used with command 01xxxxxx and 10xxxxxx)

When a song does not exist as xx.wav, multi.wav (with multiple songs in sequence) is accessed. Companion file multi.cfg is required to determine the start offsets for each song, and should contain the lines:

Loop offset
Start offset for song 1
Start offset for song 2
Start offset for song 3
Start offset for song 4
Start offset for song 5
Start offset for song 6
Start offset for song 7
etc


Notes:

- The .wav files need to be in the same directory as the .rom or .dsk file
- Tests can be done in BASIC by selecting the directory with the .wav files, and then exiting the MSX Pico menu by pressing <ESC>
- When a file can't be found, the song will just not play
- A song does not need to be stopped before a new one is started. It is actually best to avoid stopping of music.
- When a song is started, the previous song number and position are stored, so these are used to continue the previous song.
- All loop and start offsets are in decimal number of samples from the start of the song.
- When the .cfg file is missing, or the offset is missing in the .cfg file, an offset of 0 is assumed.
- Use WAV files stored as 48kHz mono, 16 bits.
- Commands 11110000 and 11100000 are mainly used to enter/exit pause. This keeps it compatible with command 11xxxxxx of older firmware versions.
- WaveGame and Nextor can never work at the same time.
```

Perhaps we can use our SamplePlayer class to implement this for the biggest part? Just needs a specific MSX device that listens to that port and implements the commands.

sndpl:
- Introduce WaveGame class for MSX Pico audio cartridge emulation
- Implement command handling for audio playback via WaveGameCommand
- Add XML configuration support for loading WAV files and settings
- Register Wave Game as a media provider in the system
- Update project files to include new source files

Implements #1975

### DIFF
--- a/build/msvc/openmsx.vcxproj
+++ b/build/msvc/openmsx.vcxproj
@@ -697,6 +697,8 @@
     <ClCompile Include="$(OpenMSXSrcDir)\sound\VLM5030.cc" />
     <ClCompile Include="$(OpenMSXSrcDir)\sound\WavAudioInput.cc" />
     <ClCompile Include="$(OpenMSXSrcDir)\sound\WavWriter.cc" />
+    <ClCompile Include="$(OpenMSXSrcDir)\sound\WaveGame.cc" />
+    <ClCompile Include="$(OpenMSXSrcDir)\sound\WaveGameCommand.cc" />
     <ClCompile Include="$(OpenMSXSrcDir)\sound\Y8950.cc" />
     <ClCompile Include="$(OpenMSXSrcDir)\sound\Y8950Adpcm.cc" />
     <ClCompile Include="$(OpenMSXSrcDir)\sound\Y8950KeyboardConnector.cc" />
@@ -1305,6 +1307,8 @@
     <None Include="$(OpenMSXSrcDir)\sound\WavAudioInput.hh" />
     <None Include="$(OpenMSXSrcDir)\sound\WavData.hh" />
     <None Include="$(OpenMSXSrcDir)\sound\WavWriter.hh" />
+    <None Include="$(OpenMSXSrcDir)\sound\WaveGame.hh" />
+    <None Include="$(OpenMSXSrcDir)\sound\WaveGameCommand.hh" />
     <None Include="$(OpenMSXSrcDir)\sound\Y8950.hh" />
     <None Include="$(OpenMSXSrcDir)\sound\Y8950Adpcm.hh" />
     <None Include="$(OpenMSXSrcDir)\sound\Y8950KeyboardConnector.hh" />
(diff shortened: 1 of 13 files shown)


MBilderbeek:
Planned changes:
- merge WaveGameCommand into the WaveGame class, like on most places
- do not make that command a mediaprovider
- let the extension automatically find the wav-file folder by default, or let the wavegame command be used to specify another media provider or a specific folder.
- the auto-find heuristic:
  - iterates the current media providers
  - using the 'target' dictionary item of each mediaprovider, looks whether it finds wav files in the folder (e.g. 01.wav or multi.wav) and if it finds that, it sets that folder as the wav folder
  - when playing a (new) song, check whether the current media provider which we used to get the folder is still the same path. If not, re-evaluate to find the proper path.

MBilderbeek:
Superseded by #2126, where I will implement the planned changes and continue the review process.

PR #2126, opened 2026-05-14
Add Wave Game audio cartridge support

### LINKED ISSUE #1975: Support WAVE-GAME protocol from MSX-Pico

MBilderbeek:
- Introduce WaveGame class for MSX Pico audio cartridge emulation
- Implement command handling for audio playback via WaveGameCommand
- Add XML configuration support for loading WAV files and settings
- Register Wave Game as a media provider in the system
- Update project files to include new source files

Closes #1975

### DIFF
--- a/src/sound/WavData.hh
+++ b/src/sound/WavData.hh
@@ -36,6 +36,8 @@ public:
 	explicit WavData(File file, Filter filter = {});
 
 	[[nodiscard]] unsigned getFreq() const { return freq; }
+	/** Get the original number of bits of the data. */
+	[[nodiscard]] unsigned getBits() const { return bits; }
 	[[nodiscard]] size_t getSize() const { return buffer.size(); }
 	[[nodiscard]] int16_t getSample(size_t pos) const {
 		return (pos < buffer.size()) ? buffer[pos] : int16_t(0);
@@ -48,6 +50,7 @@ private:
 private:
 	MemBuffer<int16_t> buffer;
 	unsigned freq = 0;
+	uint16_t bits = 0;
 };
 
 ////
@@ -85,12 +88,15 @@ inline WavData::WavData(File file, Filter filter)
 	    (std::string_view{header->fmtID.data(),    4} != "fmt ")) {
 		throw MSXException("Invalid WAV file.");
 	}
-	unsigned bits = header->wBitsPerSample;
+	bits = header->wBitsPerSample;
 	if ((header->wFormatTag != 1) || (bits != one_of(8u, 16u))) {
 		throw MSXException("WAV format unsupported, must be 8 or 16 bit PCM.");
 	}
 	freq = header->dwSamplesPerSec;
 	unsigned channels = header->wChannels;
+	if (channels == 0) {
+		throw MSXException("Invalid WAV file.");
+	}
 
 	// Skip any extra format bytes
 	size_t pos = 20 + header->fmtSize;
--- a/src/sound/WaveGame.cc
+++ b/src/sound/WaveGame.cc
@@ -1,0 +1,606 @@
+#include "WaveGame.hh"
+
+#include "DeviceConfig.hh"
+#include "EmuDuration.hh"
+#include "EmuTime.hh"
+#include "FileContext.hh"
+#include "FileOperations.hh"
+#include "MSXCliComm.hh"
+#include "MSXCPUInterface.hh"
+#include "MSXException.hh"
+#include "TclObject.hh"
+#include "XMLElement.hh"
+#include "serialize.hh"
+#include "serialize_stl.hh"
+
+#include "strCat.hh"
+
+#include <algorithm>
+#include <cassert>
+#include <cstdint>
+#include <fstream>
+#include <optional>
+
+namespace openmsx {
+
+static constexpr uint16_t FIRMWARE_BYTE_ADDR = 0xF91E;
+
+// The Pico hardware runs at 48 kHz mono 16-bit PCM. Non-conforming
+// files are rejected at load.
+static constexpr unsigned INPUT_RATE = 48'000;
+static constexpr unsigned INPUT_BITS = 16;
+
+// Delay before (re-)writing the BIOS work-RAM firmware byte. We must
+// wait until the MSX BIOS has finished initialising its work-RAM
+// (otherwise BIOS will clobber our value). One second is a comfortable
+// margin on all supported machines.
+static constexpr auto FIRMWARE_BYTE_DELAY = EmuDuration::sec(1);
+
+// 1 second fade-in used by the v2.07+ "continue previous" command and
+// the implicit fade applied to the bare "stop" opcode on new firmware.
+static constexpr float CONTINUE_FADE_SECONDS = 1.0;
+
+// Firmware revision values written into 0xF91E.
+static constexpr byte FW_V207_PLUS = 1;
+
+// NN.wav / NN.cfg filename for song index [0, 64).
+[[nodiscard]] static std::string songFilename(unsigned idx, std::string_view ext)
+{
+	return strCat((idx < 10) ? "0" : "", idx, ext);
+}
+
+[[nodiscard]] static byte getFirmwareValue(const DeviceConfig& config)
+{
+	auto firmware = config.getChildDataAsInt("firmware", FW_V207_PLUS);
+	if (firmware != one_of(0, FW_V207_PLUS)) {
+		throw MSXException("Wave Game firmware value must be 0 or 1.");
+	}
+	return byte(firmware);
+}
+
+WaveGame::WaveGame(const DeviceConfig& config)
+	: MSXDevice(config)
+	, ResampledSoundDevice(
+		config.getMotherBoard(), "wavegame",
+		"MSX-Pico Wave Game device, plays WAV files triggered by MSX software.",
+		/*numChannels=*/1, INPUT_RATE, /*stereo=*/false)
+	, syncFirmwareByte(config.getMotherBoard().getScheduler())
+	, motherBoard(config.getMotherBoard())
+	, firmwareValue(getFirmwareValue(config))
+	, command(motherBoard.getCommandController())
+{
+	registerSound(config);
+	setAutoSamplesDir();
+}
+
+WaveGame::~WaveGame()
+{
+	unregisterSound();
+}
+
+void WaveGame::reset(EmuTime time)
+{
+	stopPlayback();
+	expectQueueByte = false;
+	queuedCmd = {};
+	last = {};
+	syncFirmwareByte.schedule(time);
+}
+
+void WaveGame::writeIO(uint16_t /*port*/, byte value, EmuTime /*time*/)
+{
+	// 0xD0 (v2.07+) stashes the *next* write for deferred execution. The
+	// latched byte is not itself interpreted further - it just arms the
+	// latch.
+	if (expectQueueByte) {
+		expectQueueByte = false;
+		if (current.song != NO_SONG) {
+			queuedCmd = value;
+			return;
+		}
+	}
+	handleCommand(value);
+}
+
+void WaveGame::handleCommand(byte value)
+{
+	const bool v207plus = firmwareValue >= FW_V207_PLUS;
+
+	// 00000000 - stop music
+	if (value == 0) {
+		if (v207plus) {
+			startFadeOut(CONTINUE_FADE_SECONDS);
+		} else {
+			stopPlayback();
+		}
+		return;
+	}
+	// 0000xxxx - fade out in xxxx seconds
+	if ((value & 0b1111'0000) == 0) {
+		startFadeOut(float(value & 0x0F));
+		return;
+	}
+	// 01xxxxxx - play song xxxxxx once
+	if ((value & 0b1100'0000) == 0b0100'0000) {
+		playSong(value & 0x3F, /*loop=*/false);
+		return;
+	}
+	// 10xxxxxx - play song xxxxxx in a loop
+	if ((value & 0b1100'0000) == 0b1000'0000) {
+		playSong(value & 0x3F, /*loop=*/true);
+		return;
+	}
+	// 11xxxxxx - firmware-specific
+	if (!v207plus && (value & 0b1100'0000) == 0b1100'0000) {
+		// v1.49 .. v2.06: any 11xxxxxx toggles pause
+		paused = !paused;
+		return;
+	}
+	switch (value) {
+	case 0b1100'0000: // toggle pause
+		paused = !paused;
+		break;
+	case 0b1101'0000: // queue next command
+		expectQueueByte = true;
+		break;
+	case 0b1110'0000: // continue previous song with 1s fade-in
+		continuePreviousSong();
+		break;
+	case 0b1111'0000: // play pause.wav once
+		playSong(PAUSE_SONG, false);
+		break;
+	default:
+		// Unknown 11xxxxxx opcode on new firmware: ignore, just as the
+		// real Pico does.
+		break;
+	}
+}
+
+void WaveGame::SyncFirmwareByte::schedule(EmuTime time)
+{
+	removeSyncPoints();
+	setSyncPoint(time + FIRMWARE_BYTE_DELAY);
+}
+
+void WaveGame::writeFirmwareByte(EmuTime time)
+{
+	// BIOS work RAM is in page 3, which is normally mapped to main RAM.
+	// If that happens not to be the case at this instant, the write is
+	// silently absorbed by whatever device is mapped there, which is
+	// also what a real Pico cart would experience.
+	getCPUInterface().writeMem(FIRMWARE_BYTE_ADDR, firmwareValue, time);
+}
+
+float WaveGame::getAmplificationFactorImpl() const
+{
+	// The default amp factor (1/32768) scales an int16 sample to [-1, 1].
+	// The PSG uses amp = 1.0 on already-normalized floats, which is why at
+	// matching XML volumes the PSG is noticeably louder. Pre-multiplying
+	// by 3 gives WAV playback comparable headroom without forcing users
+	// to crank master + device sliders all the way up; the slider range
+	// then stays useful (full-scale music stops clipping below ~70%).
+	return 3.0f / 32768.0f;
+}
+
+void WaveGame::generateChannels(std::span<float*> bufs, unsigned num)
+{
+	assert(bufs.size() == 1);
+	
+	if (current.song == NO_SONG || paused) {
+		bufs[0] = nullptr;
+		return;
+	}
+	assert(songs[current.song]);
+
+	auto& song = songs[current.song];
+	auto* wav = &wavs[song->wavIdx];
+	for (unsigned i = 0; i < num; ++i) {
+
+		auto handleEndOfSong = [&] {
+			stopPlayback();
+			if (queuedCmd) {
+				handleCommand(*std::exchange(queuedCmd, {}));
+			}
+			if (current.song == NO_SONG || paused) {
+				// If we're not playing anymore, fill the buf with zeroes.
+				while (i < num) bufs[0][i++] = 0.0f;
+				return true; // finished
+			}
+			// update our pointers
+			song = songs[current.song];
+			wav = &wavs[song->wavIdx];
+			return false;
+		};
+
+		if (current.pos >= song->endSample) {
+			if (current.looping) {
+				current.pos = song->loopSample;
+			} else {
+				// Reached natural end of this song.
+				if (handleEndOfSong()) return;
+			}
+		}
+		auto sample = float(wav->getSample(current.pos++)) * fadeGain;
+		bufs[0][i] = sample;
+
+		if (fadeStep != 0.0f) {
+			fadeGain += fadeStep;
+			if (fadeStep < 0.0f && fadeGain <= 0.0f) {
+				// Fade-out complete.
+				if (handleEndOfSong()) return;
+			}
+			if (fadeStep > 0.0f && fadeGain >= 1.0f) {
+				// Fade-in complete.
+				fadeGain = 1.0f;
+				fadeStep = 0.0f;
+			}
+		}
+	}
+}
+
+void WaveGame::resetSongs()
+{
+	stopPlayback();
+	last = {};
+	wavs.clear();
+	std::ranges::fill(songs, Song{});
+}
+
+void WaveGame::playSong(unsigned idx, bool loop)
+{
+	assert(idx != NO_SONG);
+
+	// Remember last playback.
+	// Only regular songs are eligible for "continue previous" - the
+	// pause.wav jingle is explicitly excluded.
+	// When the same song is started as the current song, it is not
+	// remembered as previous song.
+	if (current.song != one_of(NO_SONG, PAUSE_SONG, idx)) {
+		assert(songs[current.song]);
+		last = current;
+	}
+
+	// if we're in auto dir mode, check for possible media change
+	if (autoSamplesDir) {
+		auto needRefresh = [&] {
+			if (auto* mediaProvider = motherBoard.findMediaProvider(currentMediaProviderName)) {
+				// still exists, does it still have the same target?
+				TclObject result;
+				mediaProvider->getMediaInfo(result);
+				if (auto target = result.getOptionalDictValue(TclObject("target"))) {
+					if (target->getString() == currentMediaProviderTarget) return false;
+				}
+			}
+			return true;
+		}();
+		if (needRefresh) setAutoSamplesDir();
+	}
+
+	if (!songs[idx]) {
+		// Real cart is silent for missing slots.
+		stopPlayback();
+		return;
+	}
+	current = { .pos     = songs[idx]->startSample,
+		    .looping = loop,
+		    .song    = idx};
+	paused = false;
+	fadeGain = 1.0f;
+	fadeStep = 0.0f;
+}
+
+void WaveGame::continuePreviousSong()
+{
+	if (last.song == PAUSE_SONG) return;
+	assert(songs[last.song]);
+	// note that if last.song == NO_SONG, current.song will now also be NO_SONG
+	current = last;
+	paused = false;
+	fadeGain = 0.0f;
+	fadeStep = 1.0f / float(CONTINUE_FADE_SECONDS * INPUT_RATE);
+}
+
+void WaveGame::stopPlayback()
+{
+	current = {};
+	paused = false;
+	fadeGain = 1.0f;
+	fadeStep = 0.0f;
+}
+
+void WaveGame::startFadeOut(float seconds)
+{
+	if (current.song == NO_SONG) return;
+	// Preserve current playback state; the fade runs in generateChannels.
+	fadeStep = -1.0f / float(seconds * INPUT_RATE);
+}
+
+// Parse a per-song NN.cfg. Format:
+//   line 1: loop offset     (sample count)
+//   line 2: start offset    (sample count, optional)
+//  Returns pair with loopOffset and startOffset, respectively
+static std::pair<size_t, size_t> parseCfg(const std::string& path)
+{
+	std::ifstream ifs(path);
+	if (!ifs) return {0, 0};
+	long long tmp = 0;
+	if (!(ifs >> tmp)) return {0, 0};
+	auto loopOff = size_t(std::max(tmp, 0ll));
+	size_t startOff = 0;
+	if (ifs >> tmp) {
+		startOff = size_t(std::max(tmp, 0ll));
+	}
+	return {loopOff, startOff};
+}
+
+// Parse multi.cfg. Format:
+//   line 1: loop offset (sample count, absolute within multi.wav)
+//   line 2+: start offset for song 0, 1, 2, ...
+// Absent or unreadable files leave `loopOff` at 0 and `starts` empty,
+// which matches the Pico's behaviour for a multi.wav without cfg.
+static bool parseMultiCfg(const std::string& path,
+                          size_t& loopOff,
+                          std::vector<size_t>& starts)
+{
+	std::ifstream ifs(path);
+	if (!ifs) return false;
+	long long tmp = 0;
+	if (!(ifs >> tmp)) return false;
+	loopOff = size_t(std::max(tmp, 0ll));
+	while (ifs >> tmp) {
+		starts.push_back((tmp < 0) ? 0 : size_t(tmp));
+	}
+	return true;
+}
+
+static bool isValidWaveGameWav(const WavData& wav)
+{
+	return wav.getFreq() == INPUT_RATE &&
+	       wav.getBits() == INPUT_BITS &&
+	       wav.getSize() > 0;
+}
+
+static void printInvalidWavWarning(MSXCliComm& cli, std::string_view filename, const WavData& wav)
+{
+	cli.printWarning(
+		"Wave Game: ", filename, " is ",
+		wav.getFreq(), " Hz, ", wav.getBits(), "-bit, ",
+		wav.getSize(), " samples in size; expected ",
+		INPUT_RATE, " Hz, 16-bit, size > 0, skipping.");
+}
+
+void WaveGame::loadSongs()
+{
+	resetSongs();
+
+	auto& cli = motherBoard.getMSXCliComm();
+
+	// Try multi.wav + multi.cfg so its indices can fill gaps between
+	// individual NN.wav files.
+	std::optional<size_t> multiIdx;
+	size_t multiLoop = 0;
+	std::vector<size_t> multiStarts;
+	{
+		auto mwav = FileOperations::join(currentDir.getResolved(), "multi.wav");
+		if (FileOperations::isRegularFile(mwav)) {
+			try {
+				WavData w(mwav);
+				if (!isValidWaveGameWav(w)) {
+					printInvalidWavWarning(cli, "multi.wav", w);
+				} else {
+					auto mcfg = FileOperations::join(currentDir.getResolved(), "multi.cfg");
+					parseMultiCfg(mcfg, multiLoop, multiStarts);
+					multiIdx = wavs.size();
+					wavs.push_back(std::move(w));
+				}
+			} catch (const MSXException& e) {
+				cli.printWarning(
+					"Wave Game: multi.wav load failed: ",
+					e.getMessage());
+			}
+		}
+	}
+
+	// Individual NN.wav and pause.wav (optional; v2.07+ plays it on 0xF0).
+	for (unsigned i = 0; i < NUM_SONGS + 1; ++i) {
+		auto filename = i == PAUSE_SONG ? "pause.wav" : songFilename(i, ".wav");
+		auto path = FileOperations::join(currentDir.getResolved(), filename);
+		if (!FileOperations::isRegularFile(path)) continue;
+		try {
+			WavData w(path);
+			if (!isValidWaveGameWav(w)) {
+				printInvalidWavWarning(cli, filename, w);
+				continue;
+			}
+			const auto sz = w.getSize();
+
+			size_t loopOff = 0;
+			size_t startOff = 0;
+			if (i != PAUSE_SONG) {
+				auto cfgPath = FileOperations::join(currentDir.getResolved(), songFilename(i, ".cfg"));
+				std::tie(loopOff, startOff) = parseCfg(cfgPath);
+			}
+			loopOff  = std::min(loopOff,  sz - 1);
+			startOff = std::min(startOff, sz - 1);
+
+			wavs.push_back(std::move(w));
+			songs[i] = Song{
+				.wavIdx      = wavs.size() - 1,
+				.startSample = startOff,
+				.endSample   = sz,
+				.loopSample  = loopOff,
+			};
+		} catch (const MSXException& e) {
+			cli.printWarning(
+				"Wave Game: ", filename,
+				" load failed: ", e.getMessage());
+		}
+	}
+
+	// Fill remaining slots from multi.wav using multi.cfg offsets.
+	if (multiIdx) {
+		const auto sz = wavs[*multiIdx].getSize();
+		for (unsigned i = 0; i < NUM_SONGS; ++i) {
+			if (songs[i]) continue;
+			if (i >= multiStarts.size()) break;
+			size_t start = std::min(multiStarts[i], sz - 1); // TODO: warn?
+			if (start >= sz) continue; // TODO: warn?
+			songs[i] = Song{
+				.wavIdx      = *multiIdx,
+				.startSample = start,
+				.endSample   = sz,
+				.loopSample  = std::min(multiLoop, sz),
+			};
+		}
+	}
+}
+
+bool WaveGame::assignDirFromMediaProviderIfValid(/*const*/MediaProvider& mediaProvider, const std::string& mediaName) // getMedia info is non-const...
+{
+	TclObject result;
+	mediaProvider.getMediaInfo(result);
+	if (auto target = result.getOptionalDictValue(TclObject("target"))) {
+		auto targetString = target->getString();
+		if (targetString.empty()) return false;
+		auto dirName = FileOperations::getDirName(target->getString());
+		if (FileOperations::isRegularFile(FileOperations::join(dirName, "01.wav")) ||
+		    FileOperations::isRegularFile(FileOperations::join(dirName, "multi.wav"))) {
+			currentDir = Filename(dirName);
+			currentMediaProviderName = mediaName;
+			currentMediaProviderTarget = targetString;
+			loadSongs();
+			return true;
+		}
+	}
+	return false;
+}
+
+void WaveGame::setAutoSamplesDir()
+{
+	clear();
+	autoSamplesDir = true;
+	currentMediaProviderName.clear();
+
+	for (const auto& mediaProvider: motherBoard.getMediaProviders()) {
+		if (assignDirFromMediaProviderIfValid(*mediaProvider.provider, std::string(mediaProvider.name))) return;
+	}
+}
+
+bool WaveGame::setSamplesDirByMediaSlot(const std::string& slot)
+{
+	autoSamplesDir = false;
+	if (auto* mediaProvider = motherBoard.findMediaProvider(slot)) {
+		// this is a known media provider!
+		if (assignDirFromMediaProviderIfValid(*mediaProvider, slot)) return true;
+	}
+	return false;
+}
+
+void WaveGame::setSamplesDir(const Filename& dir)
+{
+	autoSamplesDir = false;
+	currentMediaProviderName.clear();
+	currentDir = dir;
+	loadSongs();
+}
+
+void WaveGame::clear()
+{
+	autoSamplesDir = false;
+	currentDir = {};
+	resetSongs();
+}
+
+
+WaveGame::Cmd::Cmd(CommandController& commandController_)
+	: Command(commandController_, "wavegame")
+{
+}
+
+void WaveGame::Cmd::execute(
+	std::span<const TclObject> tokens, TclObject& result)
+{
+	auto& waveGame = OUTER(WaveGame, command);
+	if (tokens.size() == 1) {
+		result = waveGame.autoSamplesDir ? "auto" : std::string_view{waveGame.currentDir.getResolved()};
+
+	} else if (tokens.size() == 2) {
+		if (tokens[1] == "auto") {
+			waveGame.setAutoSamplesDir();
+		} else {
+			// user specified an explicit media provider or directory
+			std::string name(tokens[1].getString());
+			// media provider?
+			if (!waveGame.setSamplesDirByMediaSlot(name)) {
+				// assume it's a filename
+				try {
+					auto dir = Filename(name, userFileContext());
+					if (!FileOperations::isDirectory(dir.getResolved())) {
+						throw CommandException(
+							"Wave Game expects a directory, not a file: ", name);
+					}
+					waveGame.setSamplesDir(dir);
+				} catch (MSXException& e) {
+					throw CommandException(std::move(e).getMessage());
+				}
+			}
+		}
+	} else {
+		throw SyntaxError();
+	}
+}
+
+std::string WaveGame::Cmd::help(std::span<const TclObject> /*tokens*/) const
+{
+	return "wavegame             : show current samples directory (or 'auto')\n"
+	       "wavegame auto        : automatically find samples directory with NN.wav / NN.cfg files based on inserted media\n"
+	       "wavegame <mediaslot> : use the samples directory with NN.wav / NN.cfg files based on inserted media of the given slot ([carta|cartb|diska|...])\n"
+	       "wavegame <path>      : set samples directory with NN.wav / NN.cfg files\n";
+}
+
+void WaveGame::Cmd::tabCompletion(std::vector<std::string>& tokens) const
+{
+	if (tokens.size() == 2) {
+		auto& waveGame = OUTER(WaveGame, command);
+		auto extra = to_vector(std::views::transform(waveGame.motherBoard.getMediaProviders(), &MediaProviderInfo::name));
+		extra.push_back("auto"sv);
+		completeFileName(tokens, userFileContext(), extra);
+	}
+}
+
+template<typename Archive>
+void WaveGame::PlayState::serialize(Archive& ar, unsigned /*version*/)
+{
+    ar.serialize("pos",     pos,
+		 "looping", looping,
+		 "song",    song);
+}
+
+template<typename Archive>
+void WaveGame::serialize(Archive& ar, unsigned /*version*/)
+{
+	ar.template serializeBase<MSXDevice>(*this);
+	ar.serialize("currentDir", currentDir);
+	ar.serialize("autoSamplesDir", autoSamplesDir);
+	if constexpr (Archive::IS_LOADER) {
+		if (autoSamplesDir) {
+			setAutoSamplesDir();
+		} else {
+			currentDir.updateAfterLoadState();
+			if (!currentDir.empty()) {
+				loadSongs();
+			}
+		}
+	}
+	ar.serialize("current",         current,
+	             "paused",          paused,
+	             "fadeGain",        fadeGain,
+	             "fadeStep",        fadeStep,
+	             "last",            last,
+	             "expectQueueByte", expectQueueByte,
+	             "queuedCmd",       queuedCmd);
+}
+INSTANTIATE_SERIALIZE_METHODS(WaveGame);
+REGISTER_MSXDEVICE(WaveGame, "WaveGame");
+
+} // namespace openmsx
--- a/src/sound/WaveGame.hh
+++ b/src/sound/WaveGame.hh
@@ -1,0 +1,189 @@
+#ifndef WAVEGAME_HH
+#define WAVEGAME_HH
+
+#include "MSXDevice.hh"
+#include "MSXMotherBoard.hh"
+#include "ResampledSoundDevice.hh"
+#include "Schedulable.hh"
+#include "WavData.hh"
+#include "Filename.hh"
+
+#include "outer.hh"
+
+#include <array>
+#include <cstdint>
+#include <string>
+#include <vector>
+
+namespace openmsx {
+
+// Emulation of the MSX Pico "Wave Game" audio cartridge.
+//
+// The real cartridge plays 48 kHz mono or stereo 16-bit PCM WAV files from an
+// SD card in response to byte commands written to I/O port 0x92. Songs are
+// named NN.wav (two-digit decimal, 00..63), with optional NN.cfg files
+// giving loop and start offsets in sample counts. A 'multi.wav' +
+// 'multi.cfg' pair is consulted as a fallback for song slots not backed
+// by an individual NN.wav. A firmware byte at BIOS work-RAM address
+// 0xF91E lets games probe the cartridge revision
+// (0 = v1.49 .. v2.06, 1 = v2.07+).
+//
+// Port 0x92 protocol, as documented by the Pico firmware README:
+//   00000000       stop music (instant on v1.49-2.06; 1s fade on v2.07+)
+//   0000xxxx       fade out over xxxx seconds
+//   01xxxxxx       play song xxxxxx once
+//   10xxxxxx       play song xxxxxx in a loop
+//   11xxxxxx       (v1.49-2.06) toggle pause, low bits ignored
+//   11000000       (v2.07+)     toggle pause
+//   11010000       (v2.07+)     store next command, execute after current
+//                                song ends
+//   11100000       (v2.07+)     continue previous song with 1s fade-in
+//   11110000       (v2.07+)     play pause.wav once
+//
+// Detail specs from https://www.msxpico.com/downloads/WaveGameReadme.txt are:
+//
+// Every xx.wav file can have an optional companion xx.cfg file, which should
+// contain the lines:
+// 
+// Loop offset (used with command 01xxxxxx)
+// Optional start offset (used with command 01xxxxxx and 10xxxxxx)
+// 
+// When a song does not exist as xx.wav, multi.wav (with multiple songs in
+// sequence) is accessed. Companion file multi.cfg is required to determine the
+// start offsets for each song, and should contain the lines:
+// 
+// Loop offset
+// Start offset for song 1
+// Start offset for song 2
+// Start offset for song 3
+// Start offset for song 4
+// Start offset for song 5
+// Start offset for song 6
+// Start offset for song 7
+// etc
+// 
+// Notes:
+// 
+// - The .wav files need to be in the same directory as the .rom or .dsk file
+// - Tests can be done in BASIC by selecting the directory with the .wav files,
+//   and then exiting the MSX Pico menu by pressing <ESC>
+// - When a file can't be found, the song will just not play
+// - A song does not need to be stopped before a new one is started. It is
+//   actually best to avoid stopping of music.
+// - When a song is started, the previous song number and position are stored, so
+//   these are used to continue the previous song.
+// - All loop and start offsets are in decimal number of samples from the start
+//   of the song.
+// - When the .cfg file is missing, or the offset is missing in the .cfg file, an
+//   offset of 0 is assumed.
+// - Use WAV files stored as 48kHz mono, 16 bits.
+// - Commands 11110000 and 11100000 are mainly used to enter/exit pause. This
+//   keeps it compatible with command 11xxxxxx of older firmware versions.
+// - WaveGame and Nextor can never work at the same time.
+
+class WaveGame final : public MSXDevice
+                     , public ResampledSoundDevice
+{
+public:
+	explicit WaveGame(const DeviceConfig& config);
+	~WaveGame() override;
+
+	// MSXDevice
+	void reset(EmuTime time) override;
+	void writeIO(uint16_t port, byte value, EmuTime time) override;
+
+	template<typename Archive>
+	void serialize(Archive& ar, unsigned version);
+
+private:
+	struct Song {
+		size_t wavIdx      = 0;   // index into `wavs`
+		size_t startSample = 0;   // first sample of this song in the WAV
+		size_t endSample   = 0;   // one past last sample (exclusive)
+		size_t loopSample  = 0;   // absolute position within the WAV to loop back to
+	};
+
+	// SoundDevice
+	void generateChannels(std::span<float*> bufs, unsigned num) override;
+	[[nodiscard]] float getAmplificationFactorImpl() const override;
+
+	// Used by Cmd.
+	void setSamplesDir(const Filename& dir);
+	bool setSamplesDirByMediaSlot(const std::string& slot);
+	void setAutoSamplesDir();
+
+	bool assignDirFromMediaProviderIfValid(MediaProvider& mediaProvider, const std::string& mediaName);
+
+	void clear();
+	void loadSongs();
+	void resetSongs();
+
+	void handleCommand(byte value);
+	void playSong(unsigned idx, bool loop);
+	void continuePreviousSong();
+	void stopPlayback();
+	void startFadeOut(float seconds);
+
+	void writeFirmwareByte(EmuTime time);
+
+private:
+	struct SyncFirmwareByte final : Schedulable {
+		explicit SyncFirmwareByte(Scheduler& s) : Schedulable(s) {}
+		void schedule(EmuTime time);
+		void executeUntil(EmuTime time) override {
+			auto& wg = OUTER(WaveGame, syncFirmwareByte);
+			wg.writeFirmwareByte(time);
+		}
+	} syncFirmwareByte;
+
+	static constexpr unsigned NUM_SONGS = 64; // 6-bit song index
+	static constexpr unsigned NO_SONG = unsigned(-1);
+	// Reserved index used internally for the pause.wav slot.
+	static constexpr unsigned PAUSE_SONG = NUM_SONGS;
+
+	MSXMotherBoard& motherBoard;
+	Filename currentDir;
+	const byte firmwareValue; // written to BIOS work RAM 0xF91E
+
+	std::vector<WavData> wavs;        // backing WAV buffers
+	std::array<std::optional<Song>, NUM_SONGS + 1> songs{}; // (last index is for PAUSE_SONG)
+
+	// Playback state.
+	struct PlayState {
+		size_t   pos     = 0;       // absolute position inside the song's WAV
+		bool     looping = false;
+		unsigned song    = NO_SONG; // song index (or NO_SONG)
+		template<typename Archive>
+		void serialize(Archive& ar, unsigned version);
+	};
+	PlayState current;
+	// Remembered state for the "continue previous" command.
+	PlayState last;
+
+	bool     paused  = false;
+
+	// Fade. `fadeStep == 0` means no fade in progress.
+	float fadeGain = 1.0f;
+	float fadeStep = 0.0f;
+
+	// Queue. 0xD0 sets `expectQueueByte`; the next write fills `queuedCmd`.
+	// When the current song ends naturally, the queued byte is executed
+	// and the queue cleared.
+	bool expectQueueByte = false;
+	std::optional<byte> queuedCmd;
+
+	bool autoSamplesDir = true;
+	std::string currentMediaProviderName;
+	std::string currentMediaProviderTarget;
+
+	struct Cmd final : Command {
+		explicit Cmd(CommandController& commandController);
+		void execute(std::span<const TclObject> tokens, TclObject& result) override;
+		[[nodiscard]] std::string help(std::span<const TclObject> tokens) const override;
+		void tabCompletion(std::vector<std::string>& tokens) const override;
+	} command;
+};
+
+} // namespace openmsx
+
+#endif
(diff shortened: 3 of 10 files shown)


MBilderbeek:
@m9710797 is this along the lines we discussed? It seems to work well, but I guess more testing would be helpful.

I did not yet implement the `wavegame <mediaprovider>` subcommand. Maybe it's overkill? But OTOH, maybe it's easy to add :)

src/sound/WavData.hh
```diff
@@ -48,6 +51,8 @@ private:
 private:
 	MemBuffer<int16_t> buffer;
 	unsigned freq = 0;
+	unsigned channels = 0;
+	unsigned bits = 0;
```
m9710797: Very much a detail, we could store these two as `uint16_t`, because the .wav file itself also only stores these as 16 bit. In the `getChannels()` and `getBits()` getters returning `unsigned` is fine (keeping this an internal implementation detail).

src/sound/WaveGame.hh
```diff
@@ -0,0 +1,154 @@
...
+//   11100000       (v2.07+)     continue previous song with 1s fade-in
+//   11110000       (v2.07+)     play pause.wav once
+class WaveGame final : public MSXDevice
+                     , public ResampledSoundDevice
+{
+public:
+	static constexpr unsigned NUM_SONGS = 64; // 6-bit song index
+	static constexpr unsigned NO_SONG = unsigned(-1);
+	// Reserved index used internally for the pause.wav slot.
+	static constexpr unsigned PAUSE_WAV_SONG = NUM_SONGS;
```
m9710797: Detail, do these constants need to be `public`?

src/sound/WaveGame.hh
```diff
@@ -0,0 +1,154 @@
...
+
+	// MSXDevice
+	void reset(EmuTime time) override;
+	void writeIO(uint16_t port, byte value, EmuTime time) override;
+
+	// Used by Cmd.
+	void setSamplesDir(const std::string& dir);
+	void setAutoSamplesDir();
+	void clear();
+	[[nodiscard]] const std::string& getSamplesDir() const { return currentDir; }
```
m9710797: Detail: now that `Cmd` is an inner class, these methods can probably become `private`?

src/sound/WaveGame.hh
```diff
@@ -0,0 +1,154 @@
...
+	template<typename Archive>
+	void serialize(Archive& ar, unsigned version);
+
+private:
+	struct Song {
+		size_t wavIdx      = 0;   // index into `wavs`
+		size_t startSample = 0;   // first sample of this song in the WAV
+		size_t endSample   = 0;   // one past last sample (exclusive)
+		size_t loopSample  = 0;   // absolute position within the WAV to loop back to
+		bool   hasLoop     = false;
```
m9710797: Detail: maybe combine `loopSample` and `hasLoop` into a `std::optional`? Or alternatively, we can say `loopSample = -1` means no loop.

src/sound/WaveGame.hh
```diff
@@ -0,0 +1,154 @@
...
+	void startFadeOut(double seconds);
+	void rememberLastPlayback();
+
+	[[nodiscard]] const Song* currentSong() const;
+
+	void writeFirmwareByte(EmuTime time);
+	void scheduleFirmwareByte(EmuTime time);
+
+	struct SyncFirmwareByte final : Schedulable {
+		friend class WaveGame;
```
m9710797: I'm not sure ... is this `friend` declaration needed?

src/sound/WaveGame.hh
```diff
@@ -0,0 +1,154 @@
...
+	void continuePreviousSong();
+	void stopPlayback();
+	void startFadeOut(double seconds);
+	void rememberLastPlayback();
+
+	[[nodiscard]] const Song* currentSong() const;
+
+	void writeFirmwareByte(EmuTime time);
+	void scheduleFirmwareByte(EmuTime time);
+
```
m9710797, line 128: Detail: we prefer to repeat a (redundant) `private:` indicator to separate methods from variables.

src/sound/WaveGame.hh
```diff
@@ -0,0 +1,154 @@
...
+	bool autoSamplesDir = true;
+	std::string currentMediaProviderName;
+	std::string currentMediaProviderTarget;
+
+	struct Cmd final : Command {
+		Cmd(WaveGame& waveGame, CommandController& commandController);
+		void execute(std::span<const TclObject> tokens, TclObject& result) override;
+		[[nodiscard]] std::string help(std::span<const TclObject> tokens) const override;
+		void tabCompletion(std::vector<std::string>& tokens) const override;
+		WaveGame& waveGame;
```
m9710797: See line 101 above. By using `OUTER` we can avoid having to store `WaveGame& waveGame` (a backpointer from the inner object to the enclosing outer object).

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,670 @@
...
+
+// Delay before (re-)writing the BIOS work-RAM firmware byte. We must
+// wait until the MSX BIOS has finished initialising its work-RAM
+// (otherwise BIOS will clobber our value). One second is a comfortable
+// margin on all supported machines.
+static constexpr auto FIRMWARE_BYTE_DELAY = EmuDuration::sec(1);
+
+// 1 second fade-in used by the v2.07+ "continue previous" command and
+// the implicit fade applied to the bare "stop" opcode on new firmware.
+static constexpr double CONTINUE_FADE_SECONDS = 1.0;
```
m9710797: Can we change the type from `double` to `float`? Because it always(?) seems to be used as `float` (please verify). And then similar for the `startFadeOut` parameter.

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,670 @@
...
+
+// Firmware revision values written into 0xF91E.
+static constexpr byte FW_V207_PLUS = 1;
+
+static constexpr static_string_view DESCRIPTION =
+	"MSX-Pico Wave Game device, plays WAV files triggered by MSX software.";
```
m9710797: This is only used once(?). Maybe then just write this string literal there.

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,670 @@
...
+
+static constexpr std::string_view NAME = "wavegame";
```
m9710797: This constant is not used? Remove it?

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,670 @@
...
+// NN.wav / NN.cfg filename for song index [0, 64).
+[[nodiscard]] static std::string songFilename(unsigned idx, std::string_view ext)
+{
+	return strCat((idx < 10) ? "0" : "", idx, ext);
+}
+
+[[nodiscard]] static byte getFirmwareValue(const DeviceConfig& config)
+{
+	auto firmware = config.getChildDataAsInt("firmware", FW_V207_PLUS);
+	if (firmware != 0 && firmware != FW_V207_PLUS) {
```
m9710797: Very much a detail (coding style). if you prefer you could write this as:
`firmware != one_of(0, FW_V207_PLUS)`

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,670 @@
...
+	, firmwareValue(getFirmwareValue(config))
+	, command(*this, motherBoard.getCommandController())
+{
+	registerSound(config);
+	setAutoSamplesDir();
+}
+
+WaveGame::~WaveGame()
+{
+	syncFirmwareByte.removeSyncPoints();
```
m9710797: Not wrong, but redundant. It's not required to remove pending sync points in the destructor.

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,670 @@
...
+		if (v207plus) {
+			startFadeOut(CONTINUE_FADE_SECONDS);
+		} else {
+			stopPlayback();
+		}
+		return;
+	}
+	// 0000xxxx - fade out in xxxx seconds
+	if ((value & 0xF0) == 0x00) {
+		startFadeOut(double(value & 0x0F));
```
m9710797: See above. If we change the parameter to `float` (please verify if that's ok), then also do that here.

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,670 @@
...
+
+const WaveGame::Song* WaveGame::currentSong() const
+{
+	if (current == PAUSE_WAV_SONG) {
+		return pauseWavSong.valid ? &pauseWavSong : nullptr;
+	}
+	if (current < songs.size() && songs[current].valid) {
+		return &songs[current];
+	}
+	return nullptr;
```
m9710797: Just an idea, since `PAUSE_WAV_SONG==64` (one more than last valid song index). Can we make the song array:  `std::array<Song, NUM_SONGS + 1> songs`. And then have no special case for `PAUSE_WAV_SONG`?

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,670 @@
...
+				fadeStep = 0.0f;
+			}
+		}
+	}
+}
+
+void WaveGame::resetSongs()
+{
+	wavs.clear();
+	for (auto& s : songs) s = Song{};
```
m9710797: Detail (if you prefer, coding style):  `std::ranges::fill(songs, Song{});`

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,670 @@
...
+		TclObject result;
+		mediaProvider.provider->getMediaInfo(result);
+		if (auto target = result.getOptionalDictValue(TclObject("target"))) {
+			if (target->getString().empty()) continue;
+			auto dirName = FileOperations::getDirName(target->getString());
+			if (FileOperations::isRegularFile(FileOperations::join(dirName, "01.wav")) ||
+			    FileOperations::isRegularFile(FileOperations::join(dirName, "multi.wav"))) {
+				currentDir = dirName;
+				currentMediaProviderName = mediaProvider.name;
+				currentMediaProviderTarget = target->getString();
```
m9710797: Detail: don't query `target->getString()` twice (line 555 and 562).

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,670 @@
...
+			if (auto* mediaProvider = motherBoard.findMediaProvider(currentMediaProviderName)) {
+				// still exists, does it still have the same target?
+				TclObject result;
+				mediaProvider->getMediaInfo(result);
+				if (auto target = result.getOptionalDictValue(TclObject("target"))) {
+					if (target->getString() == currentMediaProviderTarget) return;
+				}
+			}
+			setAutoSamplesDir();
+		}();
```
m9710797, line 275: It wasn't immediately clear to me why we needed a lambda here .. it's because of the `return` on line 301 to skip `setAutoSamplesDir()`.
Maybe it more clear like this?
```
auto needRefresh = [&]{
    if (...) {
        ...
        if (...) return false;
    }
    return true;
};
if (needRefresh) setAutoSampleDir();
```

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,670 @@
...
+	current = idx;
+	playPos = songs[idx].startSample;
+	looping = loop;
+	paused = false;
+	fadeGain = 1.0f;
+	fadeStep = 0.0f;
+}
+
+void WaveGame::playPauseWav()
+{
```
m9710797: Don't we also need a check for `setAutoSampleDir()` here?
If we make `pauseWavSong` index=64 in `songs[]`, then we could maybe implement this method as:
```
rememberLastPlayback();
playSong(PAUSE_WAV_SONG, false);
```
Though there is a difference: if there is no valid pause song, then the current implementation does nothing, but the alternative implementation via `playSong()` will stop the current song. But isn't that what we actually should do?

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,670 @@
...
+	playPos = 0;
+	looping = false;
+	paused = false;
+	fadeGain = 1.0f;
+	fadeStep = 0.0f;
+}
+
+void WaveGame::startFadeOut(double seconds)
+{
+	if (seconds <= 0.0 || current == NO_SONG) {
```
m9710797: I don't think this can ever be called with `seconds <= 0.0f`. So drop that check?
And if `current == NO_SONG`, then there already shouldn't be any song playing. So the call to `stopPlayback()` (line 361) can be dropped.

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,670 @@
...
+		stopPlayback();
+		return;
+	}
+	// Preserve current playback state; the fade runs in generateChannels.
+	fadeGain = std::max(fadeGain, 0.0f);
```
m9710797: Why this line?  As far as I can see, `fadeGain` is always inside `0..1`. Only inside `generateSamples()` it briefly can get outside this range, but then it immediately gets restored (before we leave that function). Please verify, maybe I missed something.

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,670 @@
...
+	}
+}
+
+std::string WaveGame::Cmd::help(std::span<const TclObject> /*tokens*/) const
+{
+	return "wavegame                : show current samples directory (or 'empty')\n"
+	       "wavegame auto           : automatically find samples directory with NN.wav / NN.cfg files based on inserted media\n"
+	       "wavegame <path>         : set samples directory with NN.wav / NN.cfg files\n"
+	       "wavegame clear          : clear the samples directory\n";
+}
```
m9710797, line 559: Can we simplify this to:
```
wavegame                     <- query, can return `auto`, `<path>`, or `""` (empty string).
wavegame  [auto|<path>|""]   <- set uses the same values as query returns.
```
If you prefer, for the documentation you may still write:
```
wavegame        ...
wavegame auto   ...
wavegame <path> ...
wavegame ""     ...
```

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,670 @@
...
+	, waveGame(waveGame_)
+{
+}
+
+void WaveGame::Cmd::execute(
+	std::span<const TclObject> tokens, TclObject& result)
+{
+	if (tokens.size() == 1) {
+		const auto& dir = waveGame.getSamplesDir();
+		result = dir.empty() ? std::string_view{"empty"} : std::string_view{dir};
```
m9710797: See my comments in `Cmd::help()`. We should also return `auto`, and an empty string instead of `"empty"`.

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,670 @@
...
+
+	} else if (tokens.size() == 2) {
+		if (tokens[1] == "clear") {
```
m9710797: Also here. An empty string instead of `"clear"`.

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,670 @@
...
+			waveGame.clear();
+			result = "Samples directory cleared";
```
m9710797: Personal preference. I don't like these verbose success messages:
```
Sample directory cleared
Samples directory will be looked for in the media directory
Wave Game samples directory set to: mydir
```
These messages don't really contribute anything: if i asked to do something and that command did not give and error message, then that's all I need to know. E.g. in `bash` a command like `rm myfile.txt` also doesn't say `File myfile.txt was deleted`.

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,670 @@
...
+			std::string dir(tokens[1].getString());
+			try {
+				dir = Filename(std::move(dir), userFileContext()).getResolved();
+				if (!FileOperations::isDirectory(dir)) {
+					throw CommandException(
+						"Wave Game expects a directory, not a file: ", dir);
+				}
+				waveGame.setSamplesDir(dir);
+				result = "Wave Game samples directory set to: ";
+				result.addListElement(dir);
```
m9710797: I prefer to NOT set this success message. But this implementation is also wrong. You want to set a single string. But now because of `addListElement` you make the result a list. And then the first part `"Wave Game samples ..."` is converted to a list (6 elements, split on space). In practice this will go wrong when `dir` itself also contains spaces (then escaping will be added to avoid splitting this single `dir` list element in multiple elements).

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,670 @@
...
+	return "wavegame                : show current samples directory (or 'empty')\n"
+	       "wavegame auto           : automatically find samples directory with NN.wav / NN.cfg files based on inserted media\n"
+	       "wavegame <path>         : set samples directory with NN.wav / NN.cfg files\n"
+	       "wavegame clear          : clear the samples directory\n";
+}
+
+void WaveGame::Cmd::tabCompletion(std::vector<std::string>& tokens) const
+{
+	if (tokens.size() == 2) {
+		static constexpr std::array cmds = {"auto"sv, "clear"sv};
```
m9710797: Also here: no `"clear"`.

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,670 @@
...
+			} else {
+				// Reached natural end of this song.
+				current = NO_SONG;
+				fadeGain = 1.0f;
+				fadeStep = 0.0f;
+				if (hasQueued) {
+					byte cmd = queuedCmd;
+					hasQueued = false;
+					queuedCmd = 0;
+					handleCommand(cmd);
```
m9710797: I assume the intention of `handleCommand()` is to seamlessly start the next song? Though the current implementation fills the remainder of `buf[0][]` with zeros. Only on the next invocation of `generateChannels()` will the switch happen.

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,670 @@
...
+				// Fade-out complete.
+				fadeGain = 1.0f;
+				fadeStep = 0.0f;
+				current = NO_SONG;
+				// Also check the queue here - some titles fade-then-queue.
+				if (hasQueued) {
+					byte cmd = queuedCmd;
+					hasQueued = false;
+					queuedCmd = 0;
+					handleCommand(cmd);
```
m9710797: Same problem here.

m9710797 (commented):
Quick initial review comments. Many comments are about details, only a few bigger issues.
Later I'd like to do a more detailed review (e.g. really try to understand the details of song loading).

src/sound/WavData.hh
MBilderbeek: OK, will change.

src/sound/WaveGame.hh
MBilderbeek: Nope, will change.

src/sound/WaveGame.hh
MBilderbeek: Yep, will change.

src/sound/WaveGame.hh
MBilderbeek: I think the whole `hasLoop` is not necessary. It was even a bug: without loop address configuration, the song wouldn't loop at all. That was changed with this boolean, but it's useless, I think.

src/sound/WaveGame.hh
MBilderbeek: yes it is, because `removeSyncPoints()` is protected.

src/sound/WaveGame.cc
MBilderbeek: Agreed, will change.

src/sound/WaveGame.cc
MBilderbeek: Agreed, but was already removed due to previous comment ;-)

src/sound/WaveGame.cc
MBilderbeek: Do you have a suggestion on how to improve this?

Note: I didn't write this code, it was @sndpl with the help of AI... so I'm not sure what the intention was here.

src/sound/WaveGame.cc
MBilderbeek: Same response.

src/sound/WaveGame.hh
MBilderbeek: Check, will change.

src/sound/WaveGame.hh
MBilderbeek: Check, will change. Although I never really understood what the problem is of storing that back-reference. Doesn't everyone do it like that?

src/sound/WaveGame.cc
m9710797: After `handleCommand()`, you need to check if we're now playing a song or not, if not then we indeed need to fill the remainder with zeros. If we are playing a (new) song then we need to continue, but note that `wav` may have have changed (so maybe that needs to become a pointer instead of reference?).
Maybe also try to factor out some repeated code, because now the whole `handleCommand()` stuff is done twice (for end of song and for fade-out finished).

src/sound/WaveGame.cc
MBilderbeek: I think you're right, will change.

src/sound/WaveGame.cc
MBilderbeek: Will change. I missed this one somehow.

src/sound/WaveGame.cc
MBilderbeek: Oops, will change.

src/sound/WaveGame.cc
MBilderbeek: Check, will change.

src/sound/WaveGame.cc
MBilderbeek: OK, will change.

src/sound/WaveGame.cc
MBilderbeek: Will change.

src/sound/WaveGame.cc
MBilderbeek: Good idea, will change.

src/sound/WaveGame.cc
MBilderbeek: Sure, will change.

src/sound/WaveGame.cc
MBilderbeek: OK, will change.

src/sound/WaveGame.cc
MBilderbeek: I admit it was a bit of a nasty trick. I'll change it to your more expressive variant.

src/sound/WaveGame.cc
MBilderbeek: Yes, I was thinking the same when refactoring it, so changed it.

src/sound/WaveGame.cc
MBilderbeek: Yep, will change.

src/sound/WaveGame.cc
MBilderbeek: Agreed, will change. As to why: no idea, wasn't written by me :)

src/sound/WaveGame.cc
MBilderbeek: OK, and as per our discussion: I will just leave out the whole `clear` (empty) thing completely. Who cares.

src/sound/WaveGame.cc
MBilderbeek: Will change.

src/sound/WaveGame.cc
MBilderbeek: Will change.

src/sound/WaveGame.cc
MBilderbeek: Yes, the `handleCommand` may cause the song and thus the `wav` to change, but does it matter? Because the loop will end after `i` is increased up to `num` and even after that `return` exists that loop. So `wav` isn't accessed anymore in the entire loop. 
The comment also says that it silences the buffer until the next one is requested to be filled. 

What is exactly the problem here, except for the duplication, which I will fix?

MBilderbeek:
@m9710797 I was thinking about another simplification: replace the `.valid` field in `Song` by using a `std::array<std::optional<Song>, NUM_SONGS + 1>>` for `songs`.  What do you think?

m9710797:
> @m9710797 I was thinking about another simplification: replace the `.valid` field in `Song` by using a `std::array<std::optional<Song>, NUM_SONGS + 1>>` for `songs`. What do you think?

Good idea. The alternative I had in mind was using `wavIdx == -1` to mean invalid. (You can still have a method to abstract this: `bool valid() const { return wavIdx != size_t(-1); }`.)

src/sound/WaveGame.hh
```diff
@@ -0,0 +1,148 @@
...
+#include "MSXMotherBoard.hh"
+#include "ResampledSoundDevice.hh"
+#include "Schedulable.hh"
+#include "WavData.hh"
+
+#include "outer.hh"
+
+#include <array>
+#include <cstdint>
+#include <memory>
```
m9710797: `#include <memory>` is no longer needed.

src/sound/WaveGame.cc
m9710797: It could be that the song ends relatively soon while filling `buf`, then we insert a (relatively) large pause to fill the full `num` number of samples. But the `num` parameter in `generateSamples()` depends on the `samples` setting. This is a emulator setting, it should not influence how the emulation behaves.

src/sound/WaveGame.hh
m9710797: I would propose to move the implementation of `scheduleFirmwareByte()` from `WaveGame` to `WaveGame::SyncFirmwareByte`. Then we can remove this `friend`.

You can also consider implementing `writeFirmwareByte()` directly in the `SyncFirmwareByte::executeUntil()` method (try it and pick which you like best).

src/sound/WaveGame.hh
```diff
@@ -0,0 +1,148 @@
...
+	// Fade. `fadeStep == 0` means no fade in progress.
+	float fadeGain = 1.0f;
+	float fadeStep = 0.0f;
+
+	// Queue. 0xD0 sets `expectQueueByte`; the next write fills `queuedCmd`
+	// and sets `hasQueued`. When the current song ends naturally, the
+	// queued byte is executed and the queue cleared.
+	bool expectQueueByte = false;
+	bool hasQueued       = false;
+	byte queuedCmd       = 0;
```
m9710797: Detail: if you prefer you can merge `hasQueued` and `queuedCmd` into a `std::optional`.

src/sound/WaveGame.hh
```diff
@@ -0,0 +1,148 @@
...
+
+	// Remembered state for the "continue previous" command.
+	unsigned lastSong    = NO_SONG;
+	size_t   lastPos     = 0;
+	bool     lastLooping = false;
```
m9710797: These 3 are (sort of) a repetition of the `current`, `playPos` and `looping` members above.
I think it's worth to introduce a small struct:
```
struct PlayInfo {
    size_t pos = 0;
    unsigned song = NO_SONG;
    bool looping = false;
};
PlayInfo current;
PlayInfo last;
```
And then to copy between these sets you can just do `last = current` or `current = last`. And to reset you can do `last = {}`.

Detail: notice that i swapped the order of `pos` and `song` (order from large to small), this reduces internal padding because of alignment.

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,610 @@
...
+	unregisterSound();
+}
+
+void WaveGame::reset(EmuTime time)
+{
+	stopPlayback();
+	expectQueueByte = false;
+	hasQueued = false;
+	queuedCmd = 0;
+	paused = false;
```
m9710797: `stopPlayback()` already set `paused = false`.

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,610 @@
...
+		expectQueueByte = false;
+		if (current == NO_SONG) {
+			handleCommand(value);
+		} else {
+			queuedCmd = value;
+			hasQueued = true;
+		}
+		return;
+	}
+	handleCommand(value);
```
m9710797, line 102: Detail, is this simpler?
```
if (expectQueueByte) {
	expectQueueByte = false;
	if (current != NO_SONG) {
		queuedCmd = value;
		hasQueued = true;
        return;
	}
}
handleCommand(value);
```

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,610 @@
...
+		return;
+	}
+	// 10xxxxxx - play song xxxxxx in a loop
+	if ((value & 0xC0) == 0x80) {
+		rememberLastPlayback();
+		playSong(value & 0x3F, /*loop=*/true);
+		return;
+	}
+	// 11xxxxxx - firmware-specific
+	if (!v207plus) {
```
m9710797: Note: values 0x10-0x3F are not handled by the above checks. So then this block also treats them like `11xxxxxx`.

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,610 @@
...
+
+void WaveGame::rememberLastPlayback()
+{
+	// Only regular songs are eligible for "continue previous" - the
+	// pause.wav jingle is explicitly excluded.
+	if (current < songs.size() && songs[current].valid) {
+		lastSong = current;
+		lastPos = playPos;
+		lastLooping = looping;
+	}
```
m9710797, line 261: If was wondering if we can't simplify this to:
```
if (current != PAUSE_SONG) {
    ...
}
```
In other words, also if the current song has finished we should probably remember that (so this might even be a bug fix).

Hmm, now that i think about it: after we included "pause" in the `songs` array, the expression `current < songs.size()` is no longer a valid way to excluded "pause". So this implementation no longer matched the comments.

A remark in general, I think the sub-expression `current < songs.size()` is equivalent with `current != NO_SONG` (I think we do have the invariant that `current` must always point to a valid song or `PAUSE_SONG` or `NO_SONG`). So maybe we should audit the whole code for `current < songs.size()` and simplify to either `current != NO_SONG`, `current != PAUSE_SONG` or `current != one_of(NO_SONG, PAUSE_SONG)`. I think that will be clearer.

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,610 @@
...
+		paused = !paused;
+		break;
+	case 0xD0: // queue next command
+		expectQueueByte = true;
+		break;
+	case 0xE0: // continue previous song with 1s fade-in
+		continuePreviousSong();
+		break;
+	case 0xF0: // play pause.wav once
+		playPauseWav();
```
m9710797: Detail. Maybe we should just inline the implementation of `playPauseWav()` here. Then this snippet looks identical to the `01xxxxxx` and `10xxxxxx` cases above.

My trigger was: shouldn't this also call `rememberLastPlayback()` (because the two cases above do). Then in the implementation of `playPauseWav()` I saw that we do this.

And then I though: apparently all calls to `playSong()` must be proceeded by `rememberLastPlayBack()`. Shouldn't we then just move that call inside `playSong()`? (Or even just inline it there, because there are no other calls left). But before you do, please verify that indeed `playPauseWav` must call `rememberLastPlayBack`.

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,610 @@
...
+	// also what a real Pico cart would experience.
+	getCPUInterface().writeMem(FIRMWARE_BYTE_ADDR, firmwareValue, time);
+}
+
+const WaveGame::Song* WaveGame::currentSong() const
+{
+	if (current < songs.size() + 1 && songs[current].valid) {
+		return &songs[current];
+	}
+	return nullptr;
```
m9710797: I'm wondering, if we reach this function. Can it ever be that `songs[current].valid` is `false`?
Doesn't `playSong()` (and `continuePreviousSong()`) already refuse to set `current` to an invalid song?

Then this simplifies to:
```return (current != NO_SONG) ? &songs[current] : nullptr;```
And with such a simple expression, and only one caller. It's maybe better to just inline this function.

Note that the check `current < songs.size() + 1` is wong (pause is **included** in `songs.size()`). Another reason to audit all uses of `songs.size()`.

File: src/sound/WaveGame.cc [hunk 38]
```diff
@@ -0,0 +1,610 @@
...
+		bufs[0] = nullptr;
+		return;
+	}
+	const auto& wav = wavs[song->wavIdx];
+	for (unsigned i = 0; i < num; ++i) {
+
+		auto handleEndOfSong = [&] {
+			current = NO_SONG;
+			fadeGain = 1.0f;
+			fadeStep = 0.0f;
```
m9710797: Maybe a small simplification, these 3 assignments can be replaced by `stopPlayback()`. This does a bit more, but nothing that harms (I think).

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,610 @@
...
+			if (hasQueued) {
+				byte cmd = queuedCmd;
+				hasQueued = false;
+				queuedCmd = 0;
+				handleCommand(cmd);
+				// this may make wav invalid!
+			}
+			// The queued command (if any) will start from the next
+			// buffer. This introduces at most one buffer of silence,
+			// which is below audible threshold.
```
m9710797: I only half agree. Usually a buffer of silence is below audible threshold. Though with `samples=2048` and `frequency=22050` it does probably cross that threshold.

Low priority, but I'd suggest to try to fix this anyway.

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,610 @@
...
+		if (needRefresh) setAutoSamplesDir();
+	}
+
+	if (idx > songs.size() || !songs[idx].valid) {
+		// Real cart is silent for missing slots.
+		current = NO_SONG;
+		playPos = 0;
+		looping = false;
+		fadeGain = 1.0f;
+		fadeStep = 0.0f;
```
m9710797: Also here we can replace these assignments with a call to `stopPlayback()` (I think).

m9710797 (commented):
I did a 2nd partial review pass. Partial because I ran out of time, I will continue later (I reached line 350).

src/sound/WaveGame.hh
MBilderbeek: Good idea, will change.

src/sound/WaveGame.cc
MBilderbeek: OK, so, do you have an idea how to improve this? Should we directly generate the samples based on the new song instead?

src/sound/WaveGame.hh
MBilderbeek: Will fix.

src/sound/WaveGame.hh
MBilderbeek: Will change.

src/sound/WaveGame.hh
MBilderbeek: Yes, good idea. Will change.

src/sound/WaveGame.cc
MBilderbeek: Will change.

src/sound/WaveGame.cc
MBilderbeek: Yeah, will change.

src/sound/WaveGame.cc
MBilderbeek: I guess that is actually a bug. I'll add just a direct check on these bits for pause toggle.

src/sound/WaveGame.cc
MBilderbeek: I will make some changes along these lines, please check them.

src/sound/WaveGame.cc
MBilderbeek: I think especially the `playPauseWav` is a case where the last play back must be remembered. So I'll implement these simplifications.

src/sound/WaveGame.cc
MBilderbeek: Yeah, it may be redundant. But it's not so easy to see. But as many things were changed, please check this in the version I will push soon.

src/sound/WaveGame.cc
MBilderbeek: Agreed, will change.

src/sound/WaveGame.cc
MBilderbeek: Yeah, I think so too. Will change.

src/sound/WaveGame.hh
```diff
@@ -0,0 +1,143 @@
...
+
+	// Fade. `fadeStep == 0` means no fade in progress.
+	float fadeGain = 1.0f;
+	float fadeStep = 0.0f;
+
+	// Queue. 0xD0 sets `expectQueueByte`; the next write fills `queuedCmd`.
+	// When the current song ends naturally, the queued byte is executed
+	// and the queue cleared.
+	bool expectQueueByte = false;
+	std::optional<byte> queuedCmd = 0;
```
m9710797: That `=0` is wrong, you (probably) want to initialize to an empty optional (not one containing the value `0`). Either fully omit the `= 0` part, or replace it with `= {}`.

src/sound/WaveGame.hh
```diff
@@ -0,0 +1,143 @@
...
+	void writeFirmwareByte(EmuTime time);
+
+	struct SyncFirmwareByte final : Schedulable {
+		explicit SyncFirmwareByte(Scheduler& s) : Schedulable(s) {}
+		void schedule(EmuTime time);
+		void executeUntil(EmuTime time) override {
+			auto& wg = OUTER(WaveGame, syncFirmwareByte);
+			wg.writeFirmwareByte(time);
+		}
+	} syncFirmwareByte;
```
m9710797, line 137: Detail. Not wrong, but this is also a member variable, so it belong in the member-variable-section (below the `private:` separator on line 92 (so it's similar to the `command` member on line 138).

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,575 @@
...
+	// to crank master + device sliders all the way up; the slider range
+	// then stays useful (full-scale music stops clipping below ~70%).
+	return 3.0f / 32768.0f;
+}
+
+void WaveGame::generateChannels(std::span<float*> bufs, unsigned num)
+{
+	assert(bufs.size() == 1);
+	
+	if (current.song == NO_SONG || !songs[current.song] || paused) {
```
m9710797: I think, the sub-expression `!songs[current.song]` can never be `true`. The rest of the code ensures this: we can never have current.song point to an invalid song, instead we set `current.song = NO_SONG` if we attempt to play an invalid song). So no need to check that here (you may replace it with an assert (on line 196) if you want).

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,575 @@
...
+	const auto& song = songs[current.song];
+	const auto& wav = wavs[song->wavIdx];
+	for (unsigned i = 0; i < num; ++i) {
+
+		auto handleEndOfSong = [&] {
+			stopPlayback();
+			if (queuedCmd) {
+				byte cmd = *queuedCmd;
+				queuedCmd = {};
+				handleCommand(cmd);
```
m9710797: If you want to be fancy you could replace these 3 lines with:
```handleCommand(*std::exchange(queuedCmd, {}));```
You decide if you find that more readable or not. It might not be if you're not familiar with `std::exchange()`.

File: src/sound/WaveGame.cc [hunk 45]
```diff
@@ -0,0 +1,575 @@
...
+		if (current.pos >= song->endSample) {
+			if (current.looping) {
+				current.pos = song->loopSample;
+			} else {
+				// Reached natural end of this song.
+				handleEndOfSong();
+				return; // so no problem if 'wav' became invalid
+			}
+		}
+		auto sample = narrow<float>(wav.getSample(current.pos++)) * fadeGain;
```
m9710797: Detail. `narrow<float>(..)` converts the given value (a `int16_t` in this case) to `float` and asserts (so a run-time check) if that conversion was not exact. Though `int16_t` -> `float` conversion is always exact. Maybe the compiler understands that, and fully removes that check, or we could just write this as:
```auto sample = float(wav.getSample(current.pos++)) * fadeGain;```

File: src/sound/WaveGame.cc [hunk 46]
```diff
@@ -0,0 +1,575 @@
...
+	wavs.clear();
+	std::ranges::fill(songs, Song{});
+}
+
+void WaveGame::playSong(unsigned idx, bool loop)
+{
+	// Remember last playback.
+	// Only regular songs are eligible for "continue previous" - the
+	// pause.wav jingle is explicitly excluded.
+	if (current.song != one_of(NO_SONG, PAUSE_SONG) && songs[current.song]) {
```
m9710797: Also here we have the invariant that:
* if `current.song != NO_SONG` then `song[current.song]` must be valid (not a `nullopt`).

So no need to check the latter (you can `assert` if you want).

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,575 @@
...
+				if (auto target = result.getOptionalDictValue(TclObject("target"))) {
+					if (target->getString() == currentMediaProviderTarget) return false;
+				}
+			}
+			return true;
+		}();
+		if (needRefresh) setAutoSamplesDir();
+	}
+
+	if (idx == NO_SONG || !songs[idx]) {
```
m9710797: Do we ever call `playSong()` with `idx == NO_SONG`? That shouldn't happen I think. You may `assert` that (top of the function), but we shouldn't test that here.

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,575 @@
...
+		    .looping = loop,
+		    .song    = idx};
+	paused = false;
+	fadeGain = 1.0f;
+	fadeStep = 0.0f;
+}
+
+void WaveGame::continuePreviousSong()
+{
+	if (last.song == PAUSE_SONG || !songs[last.song]) return;
```
m9710797: Again here `songs[last.song]` should always be valid, thus no need to test it (an `assert` is fine if you want).
My reasoning: `last.song` is always a copy of `current.song`, and the latter will always point to a valid song.

Ah, it could be we load another set of songs, and then `last.song` points to a song that's now no longer valid. On loading other songs we do already clear `current`, but I think we should also clear `last`. Maybe add `last = {};` in `resetSongs()`.
And related: I see that `resetSongs() is always proceeded by `stopPlayback()`, then maybe we should move `stopPlayback()` inside `resetSongs()`?

m9710797: I briefly though we should check for `last.song == NO_SONG`, but setting `current.song = NO_SONG` (copying that from `last`) is fine.

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,575 @@
...
+	current = last;
```
m9710797, line 297: What happens on real hardware, if:
* You continue the previous song and that happens to be the same song as is already playing. Maybe nothing happens? Or maybe we resume the previous (=same) song from the saved position?
* What happens if you continue the previous song multiple times? (So issue the `0xE0` command multiple times in a row). Does that each time resume the song from the last saved position, or does it work only once (e.g. should we clear `last` after having used it).

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,575 @@
...
+static void parseCfg(const std::string& path,
+                                   size_t& loopOff, size_t& startOff)
+{
+	std::ifstream ifs(path);
+	if (!ifs) return;
+	long long tmp = 0;
+	if (!(ifs >> tmp)) return;
+	loopOff = (tmp < 0) ? 0 : size_t(tmp);
+	if (ifs >> tmp) {
+		startOff = (tmp < 0) ? 0 : size_t(tmp);
```
m9710797: Detail (both line 321 and 323, and later 342 and 344):
```loopOff = size_t(std::max(tmp, 0ll));```

File: src/sound/WaveGame.cc [hunk 51]
```diff
@@ -0,0 +1,575 @@
...
+	} else {
+		startOff = 0;
```
m9710797: This assignment is redundant. Because the other `return` statements in this function are only correct if the initial values of `loopOff` and `startOff` are both `0`.

I'm not a fan of in-out parameters (especially not if they need to have a specific value upon entering the function). Some alternatives:
* At the top of the function add: `loopOff = startOff = 0;`
* (My preference) return a `std::pair<size_t, size_t>`, and call with `auto [loopOff, startOff] = parseCfg(path);` IOW replace out-parameters with a proper return (a `pair` in this case).

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,575 @@
...
+	}
+}
+
+// Parse multi.cfg. Format:
+//   line 1: loop offset (sample count, absolute within multi.wav)
+//   line 2+: start offset for song 0, 1, 2, ...
+// Absent or unreadable files leave `loopOff` at 0 and `starts` empty,
+// which matches the Pico's behaviour for a multi.wav without cfg.
```
m9710797, line 340: Question 1: Is there only a single "loop offset" that's shared among all songs? That's weird, how can that work? Is that loop offset valid for all songs, or only for the single song that contains this offset (for which `start <= loop < end`).

Question 2: Later on the code assumes "end of song N" == "start of song N + 1". Does that imply the start offsets **must** be numerically ascending?

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,575 @@
...
+		starts.push_back((tmp < 0) ? 0 : size_t(tmp));
+	}
+	return true;
+}
+
+static bool isValidWaveGameWav(const WavData& wav)
+{
+	return (wav.getFreq() == INPUT_RATE) &&
+	       (wav.getChannels() == INPUT_CHANNELS) &&
+	       (wav.getBits() == INPUT_BITS);
```
m9710797: Maybe we should also check `wav.getSize() > 0`. Because otherwise other stuff may go wrong (see later comments).

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,575 @@
...
+		if (FileOperations::isRegularFile(mwav)) {
+			try {
+				WavData w(mwav);
+				if (!isValidWaveGameWav(w)) {
+					printInvalidWavWarning(cli, "multi.wav", w);
+				} else {
+					auto mcfg = FileOperations::join(currentDir, "multi.cfg");
+					parseMultiCfg(mcfg, multiLoop, multiStarts);
+					wavs.push_back(std::move(w));
+					multiIdx = wavs.size() - 1;
```
m9710797: Very much a detail. My preference is to swap the order of these 2 lines, and then drop the `- 1`.
Also on line 422.

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,575 @@
...
+	// Fill remaining slots from multi.wav using multi.cfg offsets.
+	if (multiIdx) {
+		const size_t mSize = wavs[*multiIdx].getSize();
+		for (unsigned i = 0; i < NUM_SONGS; ++i) {
+			if (songs[i]) continue;
+			if (i >= multiStarts.size()) break;
+			size_t start = std::min(multiStarts[i], mSize);
+			size_t end   = (i + 1 < multiStarts.size())
+			             ? std::min(multiStarts[i + 1], mSize)
+			             : mSize;
```
m9710797: Tradeoff, I'm not sure what's best.

We can simplify the calculation of `end` by, outside this loop (e.g. on line 387, right after `parseMultiCfg()`), doing: `multiStarts.push_back(w.getSize());`

Though the tradeoff is that we then need to change the break-condition on line 439.

BTW, why implement this via a `break`, instead of only iterating up to `multiStarts.size() - 1` (instead of always upto `NUM_SONGS`). Via something like: `for (auto i : xrange(multiStart.size() - 1)) { ...` (then we only calculate this end position once).

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,575 @@
...
+			size_t sz = w.getSize();
+
+			size_t loopOff = 0;
+			size_t startOff = 0;
+			if (i != PAUSE_SONG) {
+				auto cfgPath = FileOperations::join(currentDir, songFilename(i, ".cfg"));
+				parseCfg(cfgPath, loopOff, startOff);
+			}
+			loopOff  = std::min(loopOff,  sz);
+			startOff = std::min(startOff, sz);
```
m9710797: I think this should be
```loopOff = std::min(loopOff, sz - 1);```

Because we have the invariant:
```startSample <= loopSample < endSample```
We have half-open intervals (good): start and loop are inclusive, end is exclusive. The original expression could set start/loop equal to end.

And this is the reason I suggested to reject empty wav files (`wav.getSize == 0`, around line 353). For an empty wav we cannot set start/loop strictly smaller than end.
You could say that start==end means an empty song, but the current implementation in `generateChannels()` (and with `looping == true`) assumes songs with at least 1 sample.

Also should we silently correct invalid loop/start values (via this `std::min()` correction)? Or should we give a warning and reject those invalidly configured songs?

Similar problem on line 449.

File: src/sound/WaveGame.cc [hunk 57]
```diff
@@ -0,0 +1,575 @@
...
+	if (multiIdx) {
+		const size_t mSize = wavs[*multiIdx].getSize();
+		for (unsigned i = 0; i < NUM_SONGS; ++i) {
+			if (songs[i]) continue;
+			if (i >= multiStarts.size()) break;
+			size_t start = std::min(multiStarts[i], mSize);
+			size_t end   = (i + 1 < multiStarts.size())
+			             ? std::min(multiStarts[i + 1], mSize)
+			             : mSize;
+			if (start >= end) continue;
```
m9710797: Because of this condition, I think the start offsets in `multi.cfg` must be (strictly?) ascending.
Should we silently reject songs for which this isn't true, or also give a warning?

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,575 @@
...
+			songs[i] = Song{
+				.wavIdx      = *multiIdx,
+				.startSample = start,
+				.endSample   = end,
+				.loopSample  = std::min(multiLoop, mSize),
```
m9710797: Because `multiLoop` is the same for all songs, for all but one of the songs, `loopSample` will be outside the range `[startSample, endSample)` is that OK?

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,575 @@
...
+			loadSongs();
+		}
+	}
+	ar.serialize("current",         current,
+	             "paused",          paused,
+	             "fadeGain",        fadeGain,
+	             "fadeStep",        fadeStep,
+	             "last",            last,
+	             "expectQueueByte", expectQueueByte);
+	if (queuedCmd) ar.serialize("queuedCmd", *queuedCmd);
```
m9710797: Serialization of `queuedCmd` seems wrong. We should also restore this value, also if currently it is `nullopt`.

I did not check, possibly currently the serialization framework does not have support for serializing `std::optional`. If that's the case then we should extend it (I can help with that, just ask).

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,575 @@
...
+}
+
+template<typename Archive>
+void WaveGame::serialize(Archive& ar, unsigned /*version*/)
+{
+	ar.template serializeBase<MSXDevice>(*this);
+	ar.serialize("currentDir", currentDir);
+	if constexpr (Archive::IS_LOADER) {
+		if (!currentDir.empty()) {
+			loadSongs();
```
m9710797: This works when the saving/loading the savestate on the same system. We also have (limited) support for saving on one system and loading an another with a slightly different directory structure. We should try to do the same here.

m9710797 (commented):
Finished the 2nd review pass.

jeroentaverne:
Hi. Just found out about the work on WaveGame. Let me know if you have any questions. I can share some code pieces if needed.

MBilderbeek:
> Hi. Just found out about the work on WaveGame. Let me know if you have any questions. I can share some code pieces if needed.

Please review the code of this PR and check whether the behaviour will be the same as on the Pico....

src/sound/WaveGame.hh
MBilderbeek: Oops, will fix.

src/sound/WaveGame.hh
MBilderbeek: You're right, will fix.

src/sound/WaveGame.cc
MBilderbeek: OK, agreed, will change.

src/sound/WaveGame.cc
MBilderbeek: I like it :) Will change.

src/sound/WaveGame.cc
MBilderbeek: I'll simplify it without the `narrow`.

src/sound/WaveGame.cc
MBilderbeek: OK, will change.

src/sound/WaveGame.cc
MBilderbeek: Agreed, will change.

src/sound/WaveGame.cc
MBilderbeek: Good suggestions again, will change.

MBilderbeek: OK, so no changes needed, right?

src/sound/WaveGame.cc
MBilderbeek: @jeroentaverne can you comment on this?

src/sound/WaveGame.cc
MBilderbeek: OK, will change.

src/sound/WaveGame.cc
MBilderbeek: OK, will change.

jeroentaverne:
Comments for now:
- MSX Pico ignores the amount of channels. Playing stereo for instance is just possible. Mono is just a recommendation to reduce the SD card data processing because there are some plans to add a way to play 2 wave files at the same time. One for music and one for sound effects.
- For multi.wav the offsets are not required to be ascending.
- Continue previous song just assumes it was looping.
- When the same song is started as the current one, the current one will not be stored as previous one. But I guess this is already managed.

I will have another look later. Thanks for adding this feature!

src/sound/WaveGame.cc
MBilderbeek: Excellent questions!
@jeroentaverne @MauricioBraga can you comment on these details?

src/sound/WaveGame.cc
MBilderbeek: Yes, can you please implement that? When I tried to just put it in the list, I got an error that there is no serialize method for that type indeed.

src/sound/WaveGame.cc
MBilderbeek: How do you propose we could do that? I don't immediately see how.

MBilderbeek:
>     * MSX Pico ignores the amount of channels. Playing stereo for instance is just possible. Mono is just a recommendation to reduce the SD card data processing because there are some plans to add a way to play 2 wave files at the same time. One for music and one for sound effects.

And what about the other properties? Does it require 48kHz, 16-bit? Or will it play anything you throw at it? We now have strict checks on this, but I'm not sure if the actual Pico checks that as well.

src/sound/WaveGame.cc
MBilderbeek: Seems like a good idea, I'll add it.

src/sound/WaveGame.cc
MBilderbeek: OK, if you prefer that, I'll change it.

src/sound/WaveGame.cc
MBilderbeek: @jeroentaverne what does the Pico do with such invalid loop/start values?

src/sound/WaveGame.cc
MBilderbeek: I'll look at this as soon as I have a test case with the multi stuff. But with your suggestion, how do you select the song slot?

src/sound/WaveGame.cc
MBilderbeek: OK, so @jeroentaverne says:
> For multi.wav the offsets are not required to be ascending.

So, I'm not sure what this means for the code.  (I haven't fully understood this AI generated code yet...)

src/sound/WaveGame.cc
MBilderbeek: @jeroentaverne and @MauricioBraga - can you please comment?

MauricioBraga: The idea of multi.wav is to have one  "big song", that has multiple start points (that are also loop offsets), one to each stage of the game. Imagine you are playing King's Valley. The 1st stage "song" starts, but the player is taking too long to finish the stage. So the music will be done playing the first stage song and it will "invade" the region of the 2nd stage music, and if necessary, the region of the 3rd stage song, 4th stage song... Until the user finishes the stage or the wave file ends (which will restart the music at the 1st stage start (loop) offset.  When you reach the 2nd stage, it will now start from the 2nd stage start (loop offset) point and so on.

So, you have multiple start points in multi.cfg, that are also the offset points of each song in the big song (multi.wav). In another words, there's no single loop offset that is valid to all music, I guess the document is not very clear about this.

MauricioBraga: If it helps to clarify, this is the multi.cfg file of the (still unreleased) Boulder Dash wave patch.

0
2648832
5313792
7940864
10588288
13235328
15883360
18525440
21174048
23816896
26466688 
29109504

src/sound/WaveGame.cc
MBilderbeek: So @MauricioBraga clarified: the multi.wav is considered one single song with multiple parts. The loop is just about when the full wav has ended. The start offsets are the starts of the parts.

jeroentaverne:
> > ```
> > * MSX Pico ignores the amount of channels. Playing stereo for instance is just possible. Mono is just a recommendation to reduce the SD card data processing because there are some plans to add a way to play 2 wave files at the same time. One for music and one for sound effects.
> > ```
> 
> And what about the other properties? Does it require 48kHz, 16-bit? Or will it play anything you throw at it? We now have strict checks on this, but I'm not sure if the actual Pico checks that as well.

The samplingrate is actually not checked as well. But should be 48kHz, else the PSG and SCC will run at incorrect frequency. 16 bits signed is a must. Currently on holiday. I will do some more review next week.

src/sound/WaveGame.cc
MBilderbeek: OK, I made a start with it, see my next push.

src/sound/WaveGame.cc
MBilderbeek: Got your patch and next push has it included.

src/sound/WaveGame.cc
MBilderbeek: Yes, so the loopsample is the same for all and the end is just the size. I'll adjust it in my next push.

src/sound/WaveGame.cc
MBilderbeek: See my next push.

src/sound/WaveGame.cc
MBilderbeek: I did an attempt. Please review :)

src/sound/WaveGame.cc
MauricioBraga: If by loop sample you mean the loop offset, each line in that multi.cfg file I posted it is both a start point but also a loop offset (if the music is supposed to be played in loop). if the game tells that a particular region of the multi.wav should be played in loop, it will comeback to to loop offset used from the point (line) where the music started. it will comeback to that point when the file ends. But if the game tells to play a particular index (line) without loop, music playing will end when the file ends.

src/sound/WaveGame.cc
m9710797: At first sight it seems fine.

src/sound/WaveGame.cc
m9710797: I think we should also serialize `autoSamplesDir`, and then on loadState restore either "auto" or "dir".
But the handling of `currentDir` itself seems fine.

src/sound/WaveGame.cc
```diff
@@ -0,0 +1,601 @@
...
+	: Command(commandController_, "wavegame")
+{
+}
+
+void WaveGame::Cmd::execute(
+	std::span<const TclObject> tokens, TclObject& result)
+{
+	auto& waveGame = OUTER(WaveGame, command);
+	if (tokens.size() == 1) {
+		result = waveGame.autoSamplesDir ? "auto" : std::string_view{waveGame.currentDir.getResolved()};
```
m9710797, line 525: Maybe instead of `getResolved()` we should return `getOriginal()`. You want to return the string as was originally passed? Or do you want to see the resolved version? You could argue either way:
* Setting followed by getting should give the save result. (E.g. that allows to get the current value, temporarily change it, and afterwards restore the original).
* Showing the resolved path could also be useful.

MBilderbeek: How to decide? :)

MBilderbeek:
> @m9710797 is this along the lines we discussed? It seems to work well, but I guess more testing would be helpful.
> 
> I did not yet implement the `wavegame <mediaprovider>` subcommand. Maybe it's overkill? But OTOH, maybe it's easy to add :)

Is now done, so this is obsolete.

src/sound/WaveGame.cc
MBilderbeek: But that means there cannot be a real loop offset per song in a multi.wav, only for the first song? Isn't that a bit weird? Also, it means that if you have song 1, 2, 3, 4, start playing at 3, it will then play 4 and then continue with 3. Is that really the intention?

src/sound/WaveGame.cc
MBilderbeek: Should be done in the lastest push.

jeroentaverne:
Find attached the updated readme file. Hope it's clear now. :-)
[WaveGameReadme_updated.txt](https://github.com/user-attachments/files/28197978/WaveGameReadme_updated.txt)

jeroentaverne:
As reference also the updated C code used on MSX Pico. (Function file_load_number returns the number on a certain line in a file. The top line is line 0. When file or line doesn't exist, 0 is returned.)
[wavegame.c](https://github.com/user-attachments/files/28198143/wavegame.c)

MBilderbeek:
My comments:
- why the memory address read? How does that work? It seems a violation of the MSX standard to have some memory address always return a specific value. It also made the openMSX implementation therefore harder. To me, it would sound much more logical to have the I/O port return the version when being read.
- already mentioned in my first comment, but I'd propose to make this less MSX-Pico specific. Suppose someone wants to make a WaveGame compatible separate cartridge. (This is basically what the openMSX implementation is doing.)
- What should the implementation do if:
  - the wav file does not meet the requirements (not 48kHz, not 16-bit or not mono). Apparently not mono is no problem. But what about the others? Should it reject these? Should they just be used? What happens on a Pico? (Should we emulate that where possible?)
  - loop offsets or start offsets are wrong, e.g. beyond the end of file. Ignore them? Clip them? 
- did I understand correctly that there are no loop offsets specified/possible for multi.wav songs? What happens if loop is enabled for these songs, do they loop at the start offset, after the full wav file is played?

jeroentaverne:
- Memory read: your opinion makes sense indeed, I will make IO read possible. No game is currently using the memory read anyway.
- Special wavegame cartridge: perfect idea
- When the frequency is different, the pitch of SCC and PSG emulation are affected. I have some ideas to fix this in future firmware. I would not emulate this behavior on OpenMSX. 😀
- Loop offset for multi.wav: always 0, so starting from first song

MBilderbeek:
@jeroentaverne OK, but what about my other questions? Basically: what to do in case of "illegal" situations. I think when using openMSX one should get warned if things will not work on a "real" WaveGame hardware (like Pico currently). So it's important to know what happens there and then we can replicate it.

jeroentaverne:
I thought I answered all questions. But a proper warning when the wav file isn't 48kHz 16 bits signed should be enough.

MBilderbeek:
@jeroentaverne what about illegal offsets or start positions?

jeroentaverne:
> @jeroentaverne what about illegal offsets or start positions?

Illegal loop offset: wav file will play only once
Illegal start position: wav file will not play at all

MBilderbeek:
@jeroentaverne @m9710797 asked on IRC: 

> <wouterv> Quibus: I also reviewed the attached file "wavegame.c".   The only difference I see is that we implement the stop command according to the spec:   0000'xxxx,   but in wavegame.c   sound_fadeout() is called for all of xx00'0000.     But the implementation if sound_fadeout() is not included, possibly that still rejects values of xx'xxxx larger than 15?   Do you know if the full source code is available somewhere?

Can you comment on this?

jeroentaverne:
The fadeout can be 63 seconds, so there actually is an error in the manual.

MBilderbeek:
> The fadeout can be 63 seconds, so there actually is an error in the manual.

Can you please fix it and paste/upload the update, for completeness sake?

MBilderbeek:
> > The fadeout can be 63 seconds, so there actually is an error in the manual.
> 
> Can you please fix it and paste/upload the update, for completeness sake?

I'm asking, because I'd like to establish a "definitive spec" which we can put in the openMSX implementation. Also including the updated version info byte and all other things discussed here.

MBilderbeek:
@m9710797 any comments and/or questions you still have on that spec in https://github.com/user-attachments/files/28197978/WaveGameReadme_updated.txt?

MBilderbeek:
@jeroentaverne did you see my questions from last week?

MBilderbeek:
@jeroentaverne reminder.... I can't really proceed with this without clarity on what the exact specification is.

MBilderbeek:
@jeroentaverne do you still care about this?

PR #2128, opened 2026-05-24
Add more flexible scaling 

### LINKED ISSUE #1044: Support 4K displays
screenshot from 2016-11-18 09-38-06
I have a WQHD (2560x1440) display on my notebook, and openMSX, even with scalefactor set to 4 is small. And this is not even 4K. Especially the command mode in openMSX which uses much smaller fonts then the MSX itself is almost unreadable. I think scale factors or six and eight are needed to support 4K displays.

And obviously 8K displays are supposed to be just around the corner...

### LINKED ISSUE #1951: Allow 1x scaling again for real CRT output
Some CRT fans (like @chazbc24-oss ) like to use openMSX to enjoy as if it was a real machine connected to a CRT. And for the CRT's to properly render the signal, it needs 1x scaling output (so low res).

This will interfere with the user interface of course.

Perhaps we can still allow it, but then pop the main menu out of the window?

MBilderbeek:
Creating a PR for this to get automatic builds.

@m9710797 wrote about this work:

> Currently the "scale_factor" settings is limited to the values 2, 3, 4. This work extends that to
> * 1, 2, 3, 4, 5, 6, 7, 8
> * integer
> * fixed_aspect_ratio
> * free
> 
> The first group 1-8 are new integer scale factors, straight forward extension on what we already had. The last 3 are new:
> * integer: same as the first group, but the user can just resize the window, and then we fit an integer scale inside.
> * fixed_aspect_ratio: user can freely resize the window, and we fit the largest 4:3 image inside
> * free: use can freely resize the window, no restrictions (expect that in all modes the smallest dimensions are 320x240).

Closes #1951 
Closes #1044


PaulBoss:
I've been trying to scale the window with the mouse with the free and aspect options but it does not seem to work correctly on my setup. If i try to resize the window with the mouse the output jumps in size as if it jumps from 4x scaling to 8x and back. I've made a screenrecording: https://bosselaar.net/tmp/openmsx-scaling.mp4

This is on gnome 50 with sdl2 (not sdl2-compat)

MBilderbeek:
Yep, I found similar issues. Thanks for the video! (Unfortunately it's too big to attach here, but I hope it will stay online for a while longer.)

m9710797:
@PaulBoss Thanks for the video! I already had an email discussion about this with @MBilderbeek. I'll include it at the end. In short: it doesn't happen for me (KDE), and I'm not sure what's happening or how to investigate/solve it. Help would be appreciated.



> On Sun, May 24, 2026 at 1:31 PM Manuel Bilderbeek wrote:
>
>    Hi,
>
>    Wow, great work! I gave it a quick spin:
>    1. works well in general!
>    2. alas is non-integer scaling bound to give awful aliasing effects with scanlines. But not much you can do about >that I guess. ("Use number or integer if you're bothered with this.")


indeed. For smaller factors this is a problem, but nothing(?) we can do about it. For larger factors the problems is a lot less or practically goes away.
 

>    3. things went a bit crazy when I tried to resize the window. In any kind of scale mode. The window grew quickly and seemed to toggle between full screen size and normal as I dragged it larger. That really needs attention. I think this is the biggest issue that is still open as far as I found things so far.


Weird, for me it works as expected. I'm also not sure what to do about it, or even how to start investigating.
But let's start by trying your symptom better, Could you perhaps create a small video that show the problem?
 

>    4. what happens with the "unused borders"? Should we perhaps use them to render the full MSX borders in? (with our 320x240 base we clip some of them). I know that's pretty complex. Just asking anyway.


Horizontally we actually invent new border pixels (depending on the value of 'horizontal_stretch' these may or may not be shown, by default they are not IIRC).
Vertically we indeed hide some lines. But TBH, I don't think there is enough useful info in those lines to spend effort in trying to visualize them. So my preference would be to keep the code simple.

One idea I had in mind was to, when the user is done resizing the window, "snap" the window to a smaller size so that the black borders disappear (instead of black borders, we could draw them as a checkerboard pattern). Practical problems:
* We need some heuristic to detect when the user is still actively resizing, and when he's done.
* Ideally we snap the window on the side that the user was changing (e.g. user might drag any of the 4 corners/sides)

 

>    5. what to do with the 'horizontal_stretch' stuff?


Internally we could do some optimization:
* Currently we always render as-if horizontal_stretch=320 (no stretch), and then in a second pass stretch that image.
* But some (not all!) scale_algorithms could directly render the stretched image.

But I assume your question was more about the UI stuff? I agree there is some overlap between the "free" scaling mode and "horizontal_stretch". But I don't know of a good way to couple these.

BTW I'm also open to dropping the "free" option from the "scale_factor" setting. For now it's exposed because it's almost no extra effort to implement in the current code. But I'm not convinced myself it's really something useful. And if we drop "free" scaling, then the question about the relation with "horizontal_stretch" may disappear.

MBilderbeek:
@PaulBoss do you want to help testing this again? The branch has been updated.

@m9710797 when the `scale_mode` is on integer and you try to resize, it gives quite some flickering. Like it's scaling for one frame and then snapping back. This was when I put it to 5x scaling (which is the largest that fits) and dragging the bottom right corner to make it bigger. 
Not a huge problem, but it looks a bit messy. 

As for removing the `interleave_black_frame`: I didn't use it either, but I guess it was very useful for people with high refresh monitors? And I guess more and more of these will be used? So maybe good to get it back in? 
See also https://www.msx.org/forum/msx-talk/openmsx/how-add-100-120hz-lightboost-enabled-monitors and https://www.google.com/search?q=openmsx+interleave_black_frame gives a few websites where people seem to use it.

MBilderbeek:
https://github.com/user-attachments/assets/d2f80082-2e00-43a3-90c3-d51251b7765b

I'm dragging in the lower right corner here.

MBilderbeek:
This is the result with the test program you sent:

https://github.com/user-attachments/assets/39cdfe9c-7bd6-4c68-bb32-a73c96a1ad52

MBilderbeek:
Attention points for people who want to help testing:
- stuff like video recording screenshots (raw, non-raw, with/without osd)
- laserdisc, the old OSD stuff
- video9000
- different scale_algorithms, display_deform, noise, glow
- changing scale settings via the imgui menu
- general useabilty feedback  (like this smoothness stuff) is also very welcome

MBilderbeek:
My own testing:
- laserdisc seems OK
- video9000 seems OK
- screenshot with several variations seems OK, including OSD
- OSD seems OK (tested MoG overlay)
- scalers: I get this warning:
```
Warning(s) compiling shader "tv.vert":
0:18(15): warning: Float suffixes are invalid in GLSL 1.10
```
  - display deform: seems OK
  - setting horizontal_stretch to 320 gives weird noise bars at the sides: 
image
- noise and glow: seems OK

sndpl:
On MacOS it only renders the MSX screen at a quarter of the viewport: 
image

I asked Claude why this happend, and it says: 

The window is still created with SDL_WINDOW_ALLOW_HIGHDPI (VisibleSurface.cc:211), and the comment there points to VisibleSurface::setViewPort() — a function this PR deleted. That function used SDL_GL_GetDrawableSize to set the MSX glViewport in physical pixels. The new path feeds ImGui::GetWindowSize() (logical points) into OutputDimensions → glViewport. On a Retina Mac the drawable is 2× the logical size, so the MSX image renders into only a quarter (bottom-left) of the window. ImGui's own UI is unaffected (its backend still multiplies by FramebufferScale), so the bug is MSX-layer-only. Non-HiDPI (the Gnome test case in the commit log) is fine because the scale is 1. The leftover dangling comment is itself evidence the supporting code was dropped without replacement.

sndpl:
Ok, also asked Claude if it could review the PR. I didn't verify the points it found, so it can be total crap what it says, but this is the report it gave:

Confirmed bugs (verified against source)

1. share/shaders/tv.vert:16 — invalid GLSL literal, likely breaks the TV scaler
float BIAS = 0.001f;   // 'f' suffix
openMSX compiles shaders as #version 100 (GLSL ES 1.00) or #version 110 (src/video/GLUtil.cc:133/142). The f float suffix wasn't legal until GLSL 1.20 / ES 3.00. It's the only f-suffixed literal in the entire share/shaders/ tree. On strict/ES drivers this fails to compile and the TV scaler breaks; some desktop drivers are lenient, so it may have slipped through CI. Fix: 0.001.

2. src/video/GLImage.cc:156-157 — border axis copy/paste bug (PR-introduced)
This PR promoted borderSize from int to ivec2, but the port is wrong:
int bx = (size.x > 0) ? borderSize.x : -borderSize.y;   // neg branch should be -borderSize.x
int by = (size.y > 0) ? borderSize.x : -borderSize.y;   // pos branch should be borderSize.y
master was the correct scalar ?borderSize:-borderSize. As written, by uses the x border thickness on the normal path. When width≠height border (e.g. OSDRectangle with -relbordersize), the vertical border renders with the horizontal value — exactly the distortion this PR was meant to get right. Fix: bx = (size.x>0)?borderSize.x:-borderSize.x; by = (size.y>0)?borderSize.y:-borderSize.y;

Plausible, worth raising with the author (not fully confirmed)

3. GLDefaultScaler.cc:42 — vertical area filter uses over-allocated texture height (MEDIUM/HIGH). texelCount.y = src.getHeight() (480, the padded texture) instead of srcSize.y (240 valid lines). yRatio corrects the final sample coord but not the texel grid the area filter iterates, so even a clean 2× integer scale blends adjacent source lines (vertical blur). The interlace path (where the two are equal) hides it. Likely fix: srcSize.y. This one is subtle — could conceivably be intentional, so confirm with m9710797.

4. Screenshots mutate state / resize the live window (MEDIUM — two agents flagged independently). SDLVideoSystem::takeScreenShot → Display::paintLayers, which unconditionally calls getScaleFactorSetting().setFloat(factor) and, when windowSize != viewSize, surf->resize(...) on the real window (Display.cc:332 + :338-341). Taking a screenshot can therefore change the scale_factor setting and resize the on-screen window. The write-back-setting-during-paint pattern is also the fragile mechanism behind this; an offscreen render shouldn't touch the visible window or settings.

5. Window scale can exceed the clamped scale_factor (MEDIUM). paintLayers uses the raw computed factor for OutputDimensions while setFloat clamps the setting to [1,8]. Maximize onto a >2560×1920 display and the window renders at e.g. 12× while the setting reports 8× — they diverge above 8.

Low / cleanup

- build/buildinfo2code.py:47-48 still emits dead MIN_SCALE_FACTOR/MAX_SCALE_FACTOR defines; range is now hardcoded 1.0–8.0 (config divergence).
- New integer operator/ paths in gl_vec.hh (which OutputDimensions' /2 now relies on) have no unit-test coverage — the existing test is float-only.
- Stale repaintImpl() reference in Display.hh:56 comment.
- Possible snowLayer/osdLayer null-deref on the dummy-renderer path (Display.cc:344-351) — agent couldn't confirm reachability; worth a glance.

MBilderbeek:
Another issue: when I maximize the window with the WM maximize button, stray pixels appear at the bottom: 
image

balyuf:
I am testing this under Windows 11 on a portable screen.  Overall the scaling and mouse behavior looks good.  However there are a few rough edges.

Maximizing the window only attempts to maximize the title while seeking a maximal scale for the MSX screen, resulting in a title bar that is cut off a few pixels larger than the MSX screen.

Entering full screen (with imgui option or alt-enter) works fine, but exiting full screen leaves the MSX screen at the same scale and results in a too large window for the windows screen.  When moving the window, it only then adapts to a suitable scale.

After removing imgui.ini I get a vertical led bar, which is detached when reducing the scale to its minimum through manual reducing of the window size.  It never gets re-attached when enlarging the main window again.  Only entering full screen and leaving it re-attaches it.  The detached led bar does no show any obvious handle to move or re-attach it.

PaulBoss:
I basicly see the same issues as @MBilderbeek when using the mouse to drag the window bigger . Also when changing the scale factor in setting with the mouse to a scale with results in a window bigger then the monitor resolution (e.g. 6 and higher on 1440p), you also get a lot of flickering. 

HiDPI with a scale factor in Gnome (e.g. 133%) does not seem to work. As if the scaling is being ignored but that is for the openMSX window and imGUI menu's as a whole.

src/imgui/ImGuiManager.cc
```diff
@@ -550,9 +555,10 @@ void ImGuiManager::paintFrame(Display& display)
 			}
 		}
 		// Draw MSX layers into this window via callback (runs during ImGui::Render)
+		gl::vec2 windowSize = ImGui::GetWindowSize();
+		DrawCallbackData data{.display = &display, .windowSize = gl::ivec2(windowSize)};
```
grauw: @m9710797 Does this local not go out of scope before the callback is executed?

m9710797: I'm afraid you're right. This is a bug.
I guess I should make this variable `static` (there's no risk of recursive calls)?

grauw:
@m9710797 @sndpl I pushed a few changes with fixes for:

* PR build failure. (5b88892a06)
* On MacOS it only renders the MSX screen at a quarter of the viewport (41b0d904a2)
* share/shaders/tv.vert:16 — invalid GLSL literal, likely breaks the TV scaler (c383ab08c2)
* Screenshots mutate state / resize the live window (4ac09f1bd8)
* Window scale can exceed the clamped scale_factor (e680ea8fd1)

Additional attention points for testing:

* Mouse interactions with OSD and MSX touchpad. (77be836bd8)
* HiDPI with a scale factor in Gnome @PaulBoss — if 133% still does not work, please try 200%.

One point of feedback is that in integer scaling more I think it would be nicer if it the calculated scaling factor was rounded instead of floored. Now if I resize my window just slightly too small it will snap to a lower integer value rather than the nearest one. However when rounding up it should take care not to exceed the available desktop space, which complicates things.

MBilderbeek:
This is what happens if I stop the dragging (even on float scale mode) and just keep the mouse pressed:

https://github.com/user-attachments/assets/9316a6bf-5e92-4e7f-8f50-af3a38a4d29d

MBilderbeek:
And now with showing the cursor:

https://github.com/user-attachments/assets/4d6303c1-6eb0-4cea-8a1c-0354f016cab0

MBilderbeek:
As you can see, video updates normally:

https://github.com/user-attachments/assets/4f8f4ff9-a245-4bff-ab38-18355983adab

MBilderbeek:
@m9710797 Here's a video with the results of your extra logging and workaround attempt:

https://github.com/user-attachments/assets/9e238167-7a08-47fd-ad67-d42553e39ea8

MBilderbeek:
This is happening when I keep still and the window starts flipping bigger and smaller:
```
paint frame: windowSize=[ 1016 860 ] framebufferScale=[ 1 1 ]
call SDL_SetWindowSize(960,720)
SDL_WINDOWEVENT_RESIZED 1016 860
SDL_WINDOWEVENT_SIZE_CHANGED 1016 860
paint frame: windowSize=[ 1016 860 ] framebufferScale=[ 1 1 ]
call SDL_SetWindowSize(960,720)
SDL_WINDOWEVENT_RESIZED 1016 860
SDL_WINDOWEVENT_SIZE_CHANGED 1016 860
paint frame: windowSize=[ 1016 860 ] framebufferScale=[ 1 1 ]
call SDL_SetWindowSize(960,720)
SDL_WINDOWEVENT_RESIZED 1016 860
SDL_WINDOWEVENT_SIZE_CHANGED 1016 860
paint frame: windowSize=[ 1016 860 ] framebufferScale=[ 1 1 ]
call SDL_SetWindowSize(960,720)
SDL_WINDOWEVENT_RESIZED 1066 947
SDL_WINDOWEVENT_SIZE_CHANGED 1066 947
paint frame: windowSize=[ 1066 947 ] framebufferScale=[ 1 1 ]
SDL_WINDOWEVENT_RESIZED 1016 860
SDL_WINDOWEVENT_SIZE_CHANGED 1016 860
paint frame: windowSize=[ 1016 860 ] framebufferScale=[ 1 1 ]
```

Here another example. All I do is resize slightly larger and keep pressed. Then periodically (every N frames due to the patch I guess) you get:
```
paint frame: windowSize=[ 1281 964 ] framebufferScale=[ 1 1 ]
paint frame: windowSize=[ 1281 964 ] framebufferScale=[ 1 1 ]
paint frame: windowSize=[ 1281 964 ] framebufferScale=[ 1 1 ]
paint frame: windowSize=[ 1281 964 ] framebufferScale=[ 1 1 ]
paint frame: windowSize=[ 1281 964 ] framebufferScale=[ 1 1 ]
call SDL_SetWindowSize(1280,960)
SDL_WINDOWEVENT_RESIZED 1281 964
SDL_WINDOWEVENT_SIZE_CHANGED 1281 964
paint frame: windowSize=[ 1281 964 ] framebufferScale=[ 1 1 ]
call SDL_SetWindowSize(1280,960)
SDL_WINDOWEVENT_RESIZED 1281 964
SDL_WINDOWEVENT_SIZE_CHANGED 1281 964
paint frame: windowSize=[ 1281 964 ] framebufferScale=[ 1 1 ]
call SDL_SetWindowSize(1280,960)
SDL_WINDOWEVENT_RESIZED 1281 964
SDL_WINDOWEVENT_SIZE_CHANGED 1281 964
paint frame: windowSize=[ 1281 964 ] framebufferScale=[ 1 1 ]
call SDL_SetWindowSize(1280,960)
SDL_WINDOWEVENT_RESIZED 1281 964
SDL_WINDOWEVENT_SIZE_CHANGED 1281 964
paint frame: windowSize=[ 1281 964 ] framebufferScale=[ 1 1 ]
call SDL_SetWindowSize(1280,960)
SDL_WINDOWEVENT_RESIZED 1281 964
SDL_WINDOWEVENT_SIZE_CHANGED 1281 964
paint frame: windowSize=[ 1281 964 ] framebufferScale=[ 1 1 ]
call SDL_SetWindowSize(1280,960)
SDL_WINDOWEVENT_RESIZED 1331 1051
SDL_WINDOWEVENT_SIZE_CHANGED 1331 1051
paint frame: windowSize=[ 1331 1051 ] framebufferScale=[ 1 1 ]
SDL_WINDOWEVENT_RESIZED 1281 964
SDL_WINDOWEVENT_SIZE_CHANGED 1281 964
paint frame: windowSize=[ 1281 964 ] framebufferScale=[ 1 1 ]
paint frame: windowSize=[ 1281 964 ] framebufferScale=[ 1 1 ]
paint frame: windowSize=[ 1281 964 ] framebufferScale=[ 1 1 ]
```

MBilderbeek:
Looking at this logging I think it would help if openMSX would ignore `SDL_WINDOWEVENT_RESIZED` and `SDL_WINDOWEVENT_SIZE_CHANGED` if it is the same value as last time.
It looks like GNOME keeps sending these events when the user is resizing, even though the size isn't actually changing (yet).

grauw:
@m9710797 wrote:

> @grauw I only now am reviewing your patch 41b0d904a2c1 "Display: Take frame buffer scale into account".  Thanks for that. Though I wonder if framebufferScale should be `vec2` instead of `ivec2`.

I had that originally but then `updateOutputDimensions` needs a bunch of casts when multiplying with `windowSize` which is not a `vec2` either (arguably should be as well). This looked cleaner.

> IOW is non-integer scale also possible?

I saw a “MUST BE INTEGER VALUES” comment here so I felt it was a safe choice.

https://github.com/openMSX/openMSX/blob/1aac0d944b/src/3rdparty/imgui/imgui.h#L4291

(Replying here since you’re offline on IRC atm.)

m9710797:
> Looking at this logging I think it would help if openMSX would ignore `SDL_WINDOWEVENT_RESIZED` and `SDL_WINDOWEVENT_SIZE_CHANGED` if it is the same value as last time. It looks like GNOME keeps sending these events when the user is resizing, even though the size isn't actually changing (yet).

Note that openMSX does do anything with those 2 events. I **only** added logging on these events to better see what's going on. And KDE/wayland is doing the same, each frame when the user is resizing (also if the mouse is stationary, so consecutive frame have the same size), the 2 events are send.

If I try to interpret your 2nd log file:
* You made the window a little bit larger, from 1280x960 to 1281x964, and then keep the move stationary (while keeping the mouse butten pressed, so user-resize action is still ongoing).
* After some time openMSX sees that the window size hasn't changed for a while (30 frames), and then openMSX tries to snap the window size to 1280x960.
* Because the user is still resizing, the window manager (WM) overrules the size set by openMSX, and then we get events still for the user-size 1281x964. Also the next drawn frame still has this same size.
* This patterns repeats a few times: openMSX sees a stable constant sized window, and tries to snap it. But the WM overrules. I see the exact same kind of logging pattern in KDE/wayland.
* However (in this case) on the 6th repetition, openMSX again requests the same size 1280x960. But now the WM(?) reponds with making the window 1331x1051 in size. I have no idea where this comes from. And this is something that (as far as we know) only happens with gnome. Also the next frame is drawn using that larger size.
* OpenMSX now sees a changing window size, so it will back-off trying to snap the window size. Then, for some reason, the WM(?) again resizes the window to 1281x964. And the next frame that's indeed the size we get.

I have no idea where this "glitch" comes from. I don't think it originates from the openMSX code. For sure not from the openMSX code itself. Maybe from the ImGui code, see src/3rdparty/imgui/imgui_impl_sdl2.cc line 1221? Could you add a print there as well? And does that print ever trigger in this scenarion (my guess is it won't).

So I have no idea how to continue investigating this problem, or how to work around it.

MBilderbeek:
Shoot, i had missed that last event with the different coordinates... 
I added that logging in ImGui, but it didn't trigger indeed in the resizing scenario.

MBilderbeek:
I found that the problem goes away if I start up like this:
```
SDL_VIDEODRIVER=wayland ./openmsx
```
Gemini tells me:

> 
> The Mutter X11 _NET_FRAME_EXTENTS Bug
> 
> The 50x99 pixel jump you are seeing perfectly matches the dimensions of GNOME's X11 window decorations (usually ~50px high for the title bar, plus side borders/shadows totaling ~99px).
> 
> There is a long-standing, well-documented issue in Mutter regarding X11 applications that enforce sizing constraints (like integer scaling steps or aspect-ratio snapping) during a live resize.
> 
>     When you drag the window, Mutter sends an X11 ConfigureNotify event with the new dimensions.
> 
>     openMSX (via SDL) evaluates this and says, "Wait, that doesn't fit my scale multiplier. I need to snap to this specific size instead," and fires back a geometry request.
> 
>     The Bug: Mutter intercepts this new size request but fumbles the math between the client area (your emulator output) and the frame extents (the title bar and borders). It accidentally adds the _NET_FRAME_EXTENTS to your requested dimensions.
> 
>     The window instantly inflates by exactly the size of the title bar and borders (50x99).
> 
> Because KDE's KWin handles X11 extents and geometry constraints differently, the math doesn't desync there, which is why you only see this on GNOME.

Apparently, I'm on XWayland. With native Wayland it doesn't occur.

> 
> If you must remain on XWayland, the only way to stop Mutter from infinitely adding the frame extents is to stop fighting its ConfigureNotify events during the drag.
> 
>     Instead of calling SDL_SetWindowSize() inside your resize event loop (or using strict SDL_SetWindowMinimumSize / aspect ratio hints that apply instantly), let the user drag the window freely.
> 
>     Allow the window to be a non-integer size temporarily, and center the openMSX render buffer inside it with black letterboxing.
> 
>     When the user lets go of the mouse button, then snap the window to the nearest integer scale.

For the latter, it suggest to just have some kind of timeout: if no resize events have come in for a certain amount of time, consider resizing finished and then do the snapping to integer.

balyuf:
With the latest build here on Windows 11, maximize now gives a nice maximized window with maximal scale of MSX screen inside.  The leds now get reattached when there is room for them.  In integer scaling, the smallest scale has no room, but the next scale has room, but the positioning overlaps with the text part of the screen.  From there on, the position of the leds seems fully ok. Returning from full screen and maximized now nicely restores the old scale.  A minor detail: maximizing, then full screen, exit full screen, exit maximize does not restore the original scale, but keeps the maximized scale (though the window is no longer maximized).  I checked with e.g. Chrome and there the original scale is restored.  Again, this is minor.  Going back to the leds: after quitting and restarting, the position has changed and for the worse, for no reason I can think of.  Also with float scaling, the leds may be entirely invisible, though not detached.  Quitting and restarting can then make them partially visible.  Also the timeline bar suffers from all this positioning chaos.

PR #2129, opened 2026-05-26
Fix emulator input focus loss after openMSX startup or ImGui interaction

uniskie:
Add ImGui window flags to the status bar (NoFocusOnAppearing, NoBringToFrontOnFocus) to avoid stealing focus when it appears.

This fixes an issue where the emulator loses input focus upon starting openMSX or after interacting with the ImGui UI. Unless there is a specific reason why the status bar needs input focus, I recommend applying this fix.


m9710797:
I've applied half of this PR. You added both:
* `ImGuiWindowFlags_NoFocusOnAppearing`
* `ImGuiWindowFlags_NoBringToFrontOnFocus`

I've only kept the 1st one. The reason is that after adding the 2nd, tooltips in the status bar no longer work.

@uniskie Can you verify that with just half of your PR the focus problem is also fixed? (Because I don't know how to reproduce the focus problem, for me it just works).

uniskie:
> I've applied half of this PR. You added both:
> 
> * `ImGuiWindowFlags_NoFocusOnAppearing`
> * `ImGuiWindowFlags_NoBringToFrontOnFocus`
> 
> I've only kept the 1st one. The reason is that after adding the 2nd, tooltips in the status bar no longer work.
> 
> @uniskie Can you verify that with just half of your PR the focus problem is also fixed? (Because I don't know how to reproduce the focus problem, for me it just works).


@m9710797
I tested the commit, and it looks good to me. 

If clicking the status bar accidentally shifts the input focus, I think that's fine if it's expected behavior.

uniskie:
Regarding the fix for the other factors (the debug window issue), looking at the issue tracker, users seem to have slightly different requirements. Therefore, I’d like to separate the problems and address them individually. If I find a solution, I will submit it as a separate pull request.

uniskie:
I have closed this pull request as the fix has been incorporated in a better way.

MBilderbeek:
> Regarding the fix for the other factors (the debug window issue), looking at the issue tracker, users seem to have slightly different requirements. Therefore, I’d like to separate the problems and address them individually. If I find a solution, I will submit it as a separate pull request.

thanks a lot for looking into this! There have indeed been several focus related issue reports. Your help on these issues is greatly appreciated!

PR #2131, opened 2026-06-01
Add Namco mapper for A13-not-connected 16kB cartridges

ricbit:
Several Namco MSX arcade ports (Galaga, Dig Dug, Bosconian, King & Balloon, Mappy, Pac-Man) use a 16kB ROM chip wired with the chip's A13 line tied to MSX A15 and MSX A13/A14 left unconnected, so A15 selects which 8kB block is visible and that block mirrors within each 32kB half. These carts are normally distributed as 32kB [b0,b0,b1,b1] overdumps (which happen to load correctly as 'Mirrored'); the de-overdumped 16kB original needs block 1 at 0x8000, which no existing mapper provides.

Add RomType::NAMCO and the RomNamco mapper (block 0 -> pages 0-3, block 1 -> pages 4-7), register it in RomFactory/RomInfo and the MSVC project, and add softwaredb 'Namco' entries for the six boot-verified titles.

Curiously, some overdumps are of common Mirrored type. I verified that some games have two versions, recompiled to each mirror type.

### DIFF
--- a/build/msvc/openmsx.vcxproj
+++ b/build/msvc/openmsx.vcxproj
@@ -618,6 +618,7 @@
     <ClCompile Include="$(OpenMSXSrcDir)\memory\RomMatraCompilation.cc" />
     <ClCompile Include="$(OpenMSXSrcDir)\memory\RomMatraInk.cc" />
     <ClCompile Include="$(OpenMSXSrcDir)\memory\RomMSXDOS2.cc" />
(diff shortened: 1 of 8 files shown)


MBilderbeek:
Thanks!

The softwaredb.xml is actually an export of @Vampier his database. So, I've asked him to put them in there, so the changes won't get lost in the next export.

PR #2140, opened 2026-06-08
fix(fdc): support 40-track 180kB drives in dir-as-disk

### LINKED ISSUE #1146: Dir-as-disk not working with Talent DPF-550 and Sony HB-701FD
I tried to save a simple basic program with dir-as-disk using a Talent DPF-550 (with a Talent DPC-200 MSX). I only got "Disk I/O error" when saving.

sndpl:
DirAsDSK hardcoded an 80-track geometry and could only produce media descriptor 0xF9 (720kB, double-sided) or 0xF8 (360kB, single-sided). The Talent/AVT DPF-550 is a single-sided 40-track 180kB 5.25" drive whose disk ROM only supports the 40-track family and rejects any media descriptor below 0xFC (SUB 0FCH/RET C in GETDPB, CP 0FCH/JR NC in DSKIO) before doing any I/O. As a result every dir-as-disk operation on this drive failed with "Disk I/O error" (FILES, SAVE, ...).

Select one of the standard 9-sectors/track MSX formats from the drive's track count and number of sides instead of assuming 80 tracks:

```
  80t/2s -> 0xF9, spc 2, dir 7  (720kB)
  80t/1s -> 0xF8, spc 2, dir 7  (360kB)
  40t/2s -> 0xFD, spc 2, dir 7  (360kB)
  40t/1s -> 0xFC, spc 1, dir 4  (180kB)
```

sectorsPerCluster and sectorsPerDir become per-instance values derived from the selected format; the 40t/1s layout matches the DPF-550 ROM's 0xFC DPB exactly (dir at sector 5, data at sector 9, 360 sectors).

Add a `<nominal_tracks>` FDC config element (default 80, validated 40 or 80) and thread it MSXFDC -> RealDrive -> DiskChanger -> DirAsDSK, mirroring the existing `<singlesided/>` flow. Set `<singlesided/>` and `<nominal_tracks>40</nominal_tracks>` on Talent_DPF-550.xml and AVT_DPF-550.xml (same hardware/ROM behavior).

The name is deliberately "nominal": the value only selects the dir-as-disk geometry, it is not a hard limit on how far the head can seek, so drives keep accepting seeks beyond it (81-track copy protections and the like still work).

Backward compatible: drives without `<nominal_tracks>` default to 80 and produce the identical geometry and media descriptor as before, so no existing machine or extension changes behaviour.

This fixes: https://github.com/openMSX/openMSX/issues/1146


m9710797 (commented):
Thanks! I only took a quick look yet (not a full detailed review of all code). It seems to fix a real problem in a 100% backwards compatible way.

Though before merging I'd like to bring up these discussion points. Maybe that triggers some changes, or maybe we can merge unchanged.

Real disk drives often support more than 80 tracks. IIRC on my turboR machine I could use 83 tracks. On my Philips NMS8255 it was only 82 tracks. The exact numbers don't matter, but it differs per machine (and possibly it depends on manufacturing tolerances, so maybe not even the same for each machine instance?).

We do emulate more than 80 tracks, but we currently do not emulate an upper limit. And we should emulate more than 80 tracks because some software uses the 81th track as a copy-protection mechanism (81 tracks are supported on most (all?) drives). Though when we add a new tag `<tracks>80</tracks>` (ok, it's the default, so we won't write this in practice), it's a bit weird to then still support more than 80 tracks.

Should we instead make this the number of tracks that are physically supported by the drive? And then maybe make 83 the default? And have the drive clip the track number when you try to seek beyond this value?

A second discussion point contradicts my first point somewhat. A common complaint about DirAsDsk is that it only supports a limited capacity (720kB). Most (but not all) disk roms have no problems with much larger disks. For DirAsDsk we could emulate an artificial disk containing 250 tracks and 10 or even 11 sectors per track (also no problem for most disk roms). But that's weird in combination with the new `<tracks>` tag.

MBilderbeek:
Or rename as "nominal_tracks"? Or maybe some other name that is (even) more to the point?

MBilderbeek:
Does this also fix the Sony HB-701FD case of #1146?

sndpl:
just for reference, the Sony HB-701FD xml was missing the <singlesided/> tag, has nothing to do with this issue.

PR #2141, opened 2026-06-08
Update Yamaha_FD-03.xml

f-fix:
Adding new ROM dumps from https://www.msx.org/forum/msx-talk/software/list-undumped-msx-software-please-help?page=68#comment-487151


MBilderbeek:
Thanks! I'd like to replace all these dumps with register values with clean ones... but it will take a while, I guess. This helps.

PR #2142, opened 2026-06-08
fix(dir-as-disk): don't rewrite host file on no-op data-sector writes

### LINKED ISSUE #1547: DiraskDsk: Multiple files are rewritten again, while only saving another file
When saving files (like in trilotracker or graphsaurus) with drive b as 'diraskdsk', I can see a lot of 'untouched' files, are being saved again.. 

Config: MSX2, PHILIPS NMS8255, OpenMSX on Windows

**01-11-2023  12:39                 3 bladie.bas**
01-11-2023  12:25            10.240 COMMAND.COM
02-09-2023  13:49             1.024 GS2.COM
03-09-2023  20:50           476.160 GS2.DAT
**01-11-2023  12:25               256 kingsqu2.pl5
01-11-2023  12:25            27.143 kingsqu2.sr5
01-11-2023  12:25               256 kingsque.pl5
01-11-2023  12:39            27.143 kingsque.sr5**
03-09-2023  20:50               256 miner2.pl5
03-09-2023  20:50            27.143 miner2.sr5
30-08-2023  14:58             2.432 MSXDOS.SYS

BTW: It also seems trilotracker tracks are being corrupted, but I'm not sure if that is related.

sndpl:
When saving a file in dir-as-disk mode, the MSX disk driver flushes its sector cache and re-writes the data sectors of a previously-saved file with byte-identical content. writeDataSector() propagated those writes to the host file unconditionally (open "rb+" + write), which left the content unchanged but bumped the file's modification time. As a result, saving one file re-stamped earlier, unrelated files: e.g. SAVE"TEST.BAS" followed (later) by SAVE"TEST1.BAS" ended up giving both files the same timestamp.

Before writing, compare the bytes that would actually be written against the sector's previous content and skip the host write when they're equal. Only the [0, writeSize) range that maps to the host file is compared; bytes past EOF are never written, so a change there must not trigger a write. This follows the same "snapshot-old, compare, skip-if-unchanged" pattern already used in writeFATSector() and writeDIRSector().

Fixes: https://github.com/openMSX/openMSX/issues/1547

### DIFF
--- a/src/fdc/DirAsDSK.cc
+++ b/src/fdc/DirAsDSK.cc
@@ -1328,6 +1328,10 @@ void DirAsDSK::writeDataSector(unsigned sector, const SectorBuffer& buf)
 	assert(sector >= firstDataSector);
 	assert(sector < nofSectors);
 
+	// Remember the previous content of this sector so we can detect whether
+	// the MSX actually changes anything (see below).
+	auto oldSector = sectors[sector];
+
 	// Buffer the write, whether the sector is mapped to a file or not.
 	sectors[sector] = buf;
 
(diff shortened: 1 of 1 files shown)


m9710797:
I had completely missed this PR. Next time feel free to send reminders if you don't see any reaction. (Are there any more PRs that I missed?)

The PR didn't apply cleanly anymore, but a manual cherry-pick did work.

Thanks!

PR #2145, opened 2026-07-01
Panasonic FS-JH1 Joy Handle emulation (2026)

theNestruo:
An attempt to support Panasonic FS-JH1 Joy Handle emulation.

image

In the _to do_ list:
- highlight buttons on hover in the input mapping window.
- _analogValue_ in _JoyHandleState_.
- _analogValue_ serialization.

This PR supersedes https://github.com/openMSX/openMSX/pull/1770.


src/imgui/ImGuiSettings.cc
```diff
@@ -306,7 +307,7 @@ void ImGuiSettings::showMenu(MSXMotherBoard* motherBoard)
 						im::Indent([&]{
 							im::Disabled(!unlocked, [&]{
 								float fval = float(value.getInt()) / 1.0e6f;
-								if (ImGui::InputFloat(tmpStrCat("frequency (MHz)##", name).c_str(), &fval, 0.01f, 1.0f, "%.2f")) {
+								if (ImGui::InputFloat(tmpStrCat("frequency (MHz)##", name).c_str(), &fval, 0.01f,   1.0f, "%.2f")) {
```
m9710797: Detail: here and in various other places in this file you (accidentally) added extra spaces in unrelated code. Can you revert that.

src/imgui/ImGuiSettings.cc
```diff
@@ -825,29 +828,2327 @@ static void draw(gl::vec2 scrnPos, std::span<uint8_t> hovered, int hoveredRow)
...
+static constexpr std::array<zstring_view, NUM_BUTTONS> keyNames = { // keys in Tcl dict
+	"UP", "DOWN", "LEFT", "RIGHT",
+	"A", "B",
+	"WHEEL_LEFT", "WHEEL_RIGHT"
+};
+
+static constexpr auto boundingBox = gl::vec2{300.0f, 160.3f};
+static constexpr float spanH = 0.50f;
+// Each row is one filled rectangle: {x0, x1, yTop, yBottom} in local coords.
+static constexpr std::array<std::array<float, 4>, 2239> fill = {{
```
m9710797, line 868: That image looks nice, though also a bit weird in some places (especially the non-axis-aligned lines). Now I understand why, everything is drawn using (+2000) filled rectangles. Including the circles, curves and diagonal lines :-)

It's not high priority, but when everything else is done. Maybe you could try using the other ImGui drawing primitives as well (like circles, non-axis aligned lines (possibly lines with different thickness), bezier curves, ...  That should be a bit lighter to process (not so important, this is probably fast enough). But it will also look nicer because of anti-aliasing (now the image is purely black and white, with AA it will also use shades of grey (for the circles and diagonal lines)).

You anyway had a TODO to revisit the drawing code for highlighting the hovered input.

---

Anyway low priority. I just though it was funny you used this approach :-)

Another possibility is to store a bitmap: 300x160 (?) B/W pixels takes 6000 bytes. Currently you store 2239 x 4 x 4 = 35824 bytes.  Then draw this bitmap as a texture. (Not a serious suggestion).

m9710797 (commented):
Thanks for working on this!

I only had a quick first look. At first sight it looks good, so not much feedback yet. The feedback I do have is all about details.

But then there are also several TODOs still in the code. I'll do a more thorough review when those are resolved. But so far it looks very good.

src/imgui/ImGuiSettings.cc
theNestruo: The drawing code was machine-generated from an actual image.
To be honest, it would be nice if SVG could be used for these images. I think using ImGui drawing primitives like circles, bezier curves, ... results in hard to maintain code (there are a lot of tools to edit SVG files, but I'm not aware of any tool to edit a hard-coded graphic).
The code for the FS-JH1 image has been machine-generated. If there is such a tool to vectorize and convert to ImGui drawing primitives, please let me know and I'll try to feed it the cleaned-up image I have used.

This image was the reason the previous PR for FS-JH1 emulation got stalled. Actually, if it is going to cause a lot of trouble, I prefer to drop the image entirely. One of the remaining TODO is related to hovering/clicking on the image to map the controls; but the mapping can be done without the image.
What I actually care about is to be able to emulate the FS-JH1 using an analog modern controller, and that is already working.

The other TODO are at JoyHandle::signalMSXEvent and JoyHandle::signalStateChange, but I don't know what needs to be done there. Friendly reminder I'm not proficient in C++, and I just hacked this FS-JH1 emulation based on previously existing code for joystick emulation, JoyMega emulation, and paddle emulation.

src/imgui/ImGuiSettings.cc
theNestruo: Fixed at https://github.com/theNestruo/openMSX/commit/9e9b08a2a073617d8a1dd036748bdc2e6f054b4b

src/imgui/ImGuiSettings.cc
theNestruo: For the reference:

The source image:
image

The illustration extracted from the source image:
image

The manually cleaned up illustration:
image

m9710797: >... Friendly reminder I'm not proficient in C++, and I just hacked this FS-JH1 emulation based on previously existing code for joystick emulation, JoyMega emulation, and paddle emulation.

Thanks for the reminder, I had absolutely forgotten that.

If you want I can take over from here? Some people prefer review feedback and then tweak the PR themselves, others (maybe like you?) create the PR as best as they can and then want to leave the final tweaking to us. Just let me know what you prefer.

theNestruo: For JoyHandle::signalMSXEvent and JoyHandle::signalStateChange, I would definitely need help from you.

From my side, what I can do is to further simplify the drawing and try to machine-traslate it to proper ImGui primitives. And maybe the hover/highlight over the resulting graphic.

m9710797:
Hi,

I only now found the time to really start working on this. Yesterday I spend most of the day creating this .svg file (partially based on your simplified image, partially on the original box art, partially on the youtube video that shows the real hardware). Please check, let me know if you think important elements are missing, or if you see mistakes, or other suggestions:
FS-JH1

Here are some properties of this file:
* I've mostly manually written this file (in a text editor). Taking care to only use constructions that should(*1) easily map to Dear ImGui drawing primitives.
* The steering wheel itself is separate from the body. That means we can draw it rotated in any angle. Drawing the wheel clears the background (i mean it fills the wheel-outline (*2)), so it should correctly hide all lines that were drawn behind it. Same for the stick on the left (we can draw that translated).
* The buttons A/B can be highlighted, the current .svg shows butten A in light-blue.
* It is the intention that you download this .svg, open it in a text editor and experiment by changing the documented values (like rotation angle etc).

The next step is to convert this .svg to Dear ImGui c++ code. I'll *try* to make time for that next week.

I also now better understand what you mean with "... help with signalMSXEvent ... signalStateChange ...". I think for now we don't have a good way yet to configure analog inputs. We do need those for this device. I'll have to think about that.


(*1) I did not actually test this yet. But at least when I read the ImGui documentation, it should all be supported.
(*2) If you view this .svg on a dark background it may look a bit weird: the body in dark, the steering wheel and stick in white. But it does clearly show which are the moveable parts. Obviously in openMSX we'll use the proper colors.

theNestruo:
Wow! **Thank you** for working on this! Your SVG is superb. It is definitely way better than anything I've tried to simplify the SVG (without success). 

I was not expecting the steering wheel graphic to rotate, to be honest. Didn't know OpenMSX code supported it; I though the base graphic was drawn only once, then the overlay to highlight buttons was drawn over the base graphic. It would be a nice feeback on which position of the steering wheel is mapped to which angle of the analog stick, but it if is too complex to implement I would drop it.

> The buttons A/B can be highlighted, the current .svg shows butten A in light-blue.

Minor issue: the highlighted button is B. (B is on the left, A is on the right).

> I also now better understand what you mean with "... help with signalMSXEvent ... signalStateChange ...". I think for now we don't have a good way yet to configure analog inputs. We do need those for this device. I'll have to think about that.

Is there anything in the paddle control code that can be reused?

> let me know if you think important elements are missing

Just for the record, there are a couple of things in the original hardware that I left out on my draft implementation:
- Stick/Handle switch: I think it is unnecessary (why would anyone want to disable the handle?)
- Autofire B/A: If autofire hasn't been implemented yet and nobody has complained yet, it definitely looks unnecessary (I'm thinking about joypads with autofire: FS-JS221, FS-JS222, TN-J2, JS-303T, JC-300...)

Again: this is just for the record, as both feature looks unnecessary.

m9710797:
Hi.

I've pushed a new branch "theNestruo-fsjh1-2026" (in the openMSX git repo, not your repo).
It's heavily based on your work. I reworked the drawing(*1), and highlighting code. I implemented the analog binding TODOs.
The code is not yet finished. For sure I still need to review myself and maybe cleanup/simplify some stuff.
But I think all functionality is present, and in my very limited basic tests it also worked fine.
So now I'd like to ask for help with testing. Any bugs? Anything missing? Anything that should still be improved?
One thing I did not yet test myself is replay. Do replay files with this input device work correctly?

Thanks.



(*1) The drawing with occlusion (in semi-transparant ImGui window) was more work than I originally thought. But it was fun, and I learned some new things :)

theNestruo:
I've cloned openmsx/theNestruo-fsjh1-2026 to theNestruo/theNestruo-fsjh1-2026 and I've been trying to build it to do some tests... but I have been unable to successfully build the binaries (see https://github.com/theNestruo/openMSX/actions).
If you have a Windows build I can use, I can do some testing.

m9710797:
I did some research. Apparently my code triggers a known bug in msvc? I've pushed an attempt to solve it. Please try again.

m9710797:
@MBilderbeek any idea why the automatic build actions don't trigger on the "theNestruo-fsjh1-2026" branch?

theNestruo:
> I did some research. Apparently my code triggers a known bug in msvc? I've pushed an attempt to solve it. Please try again.

That did the trick (artifacts available at https://github.com/theNestruo/openMSX/actions/runs/29761859804)

Unfortunately, OpenMSX refuses to start with message:

> info: Failed to initialize default machine: Error in "Philips_NMS_8250" machine: Invalid key: must be one of UP, DOWN, LEFT, RIGHT, A, B, WHEEL,
> info: Using fallback machine: C-BIOS_MSX2+
> Fatal error: Error in "C-BIOS_MSX2+" machine: Invalid key: must be one of UP, DOWN, LEFT, RIGHT, A, B, WHEEL,

(some incompatibility with some configuration previously saved by my version, maybe?)

m9710797:
@theNestruo I can confirm it's fixed now.  E.g. see [here](https://github.com/openMSX/openMSX/pull/2156). I've turned my branch into a PR, so that our build actions pick it up (thanks MBilderbeek for the tip). But let's keep your PR (this page) for the discussion.

m9710797:
> > info: Failed to initialize default machine: Error in "Philips_NMS_8250" machine: Invalid key: must be one of UP, DOWN, LEFT, RIGHT, A, B, WHEEL,
> > info: Using fallback machine: C-BIOS_MSX2+
> > Fatal error: Error in "C-BIOS_MSX2+" machine: Invalid key: must be one of UP, DOWN, LEFT, RIGHT, A, B, WHEEL,
> 
> (some incompatibility with some configuration previously saved by my version, maybe?)

Yes, you probably saved the setting you created with your version. You have both WHEEL_LEFT and WHEEL_RIGHT. I replaced that with a single WHEEL entry. Because a single analog joystick axis can report both positive and negative values (covers both left and right).

So for now, best remove this invalid setting from your `settings.xml` file. But I'll also look into making openMSX robust against invalid settings (openMSX should not refuse to start with whatever settings are saved).

theNestruo:
I've tested it and it works great!
I have an experimental [fsjh1test.rom](https://github.com/theNestruo/msx-msxlib/blob/master/games/experiments/fsjh1test/fsjh1test.rom) where I was testing different ways to detect if the wheel is half or full turned.
Maybe it can be used for tests (or you can use SEGA G.P. World for less boring tests :D )

---

From my tests, I have detected a bug:
- The remove button in the binding window does not work if there are more than one binding (instead of removing one or all bindings, it adds "mouse button1" to the bindings)

---

Also from my test, I have found [this value](https://github.com/openMSX/openMSX/blob/theNestruo-fsjh1-2026/src/input/JoyHandle.cc#L157) `static constexpr int HALF = 20000; // TODO tune?` definitely needs to be tuned. As a reference, In my [old code](https://github.com/theNestruo/openMSX/blob/fsjh1-2026/src/input/JoyHandle.cc#L197), I used the first 99% of the values for "half turned", and only the highest 1% for "fully turned":
```c
int deadZone = getJoyDeadZone(bind.getJoystick()); // percentage 0..100
int threshold = (deadZone * 32768) / 100; // 0..32768
/*
 * Modern gamepad saturate analog values halfway already,
 * so let's use a minimal upper dead zone (99%) for now
 * instead of mirroring the lower dead zone percentage
int halfwayZone = (100 - deadZone); // percentage 0..100
 */
int halfwayZone = 99;
int halfwayThreshold = (halfwayZone * 32768) / 100; // 0..32768
```

I don't know if "analogValue" is still in the range 0..32768 but, it if is, it would be convenient to use a higher value for HALF (like 32000 instead of 20000). At least with my controller, even with the stick halfway to the left/right the value is already sending the maximum value, making it very hard to be able to stay in the "halfway" zone (maybe it is problem with my controller).

---

Thank you so much for the time and effort you are putting into this PR!

m9710797:
> I've tested it and it works great!

Thanks.

> I have an experimental [fsjh1test.rom](https://github.com/theNestruo/msx-msxlib/blob/master/games/experiments/fsjh1test/fsjh1test.rom) where I was testing different ways to detect if the wheel is half or full turned. Maybe it can be used for tests (or you can use SEGA G.P. World for less boring tests :D )

I've used the most boring test (but works great):
```
10 ?STICK(1):GOTO10
```
> 
> From my tests, I have detected a bug:
>     * The remove button in the binding window does not work if there are more than one binding (instead of removing one or all bindings, it adds "mouse button1" to the bindings)

I can reproduce this. I'll investigate and try to fix soon.

> I don't know if "analogValue" is still in the range 0..32768 but, it if is, it would be convenient to use a higher value for HALF (like 32000 instead of 20000). At least with my controller, even with the stick halfway to the left/right the value is already sending the maximum value, making it very hard to be able to stay in the "halfway" zone (maybe it is problem with my controller).

For my controller (Nintendo pro controller) the relation between stick movement and output value is very linear. If I move the stick about halfway I get a value around 16000, to get 32767 I need to move almost the full distance. That's why I picked 20000 (trying to get a middle ground between your comments and how my controller behaves). For me a value 32000 would be unusable (always half-steering, except for the last mm).

Can't you calibrate your controller (maybe windows offers a mechanism for that?). From a programmer point of view it would be, by far, the easiest if you can assume a more or less linear relation between stick movement and analog value (and that's indeed what I see in my case). So if there is any way to calibrate your controller that would be my preferred choice. If there's no way to do that, only then we should offer something in openMSX ...

... I've done some more research, it seems to be common to offer both
* an inner-dead-zone setting: values below this get reported as zero (openMSX already has this).
* on outer-dead-zone setting: values above this get mapped to maximum

The latter helps for controller that don't reach the max output value (anymore due to wear?), then you can still reach the max output in software after adjustment. Though this doesn't help you because your controller reaches max output (32767) much too soon.

Doing some more research: some software also offers a "sensitivity setting", (a multiplier, can be above or below 1.0). That could in theory solve the problem.

Though before adding of these new settings, I'd like to ask that you check your controller settings (or windows settings, or steam input layer settings, ...). Maybe at some level there already is some transformation applied?

> Thank you so much for the time and effort you are putting into this PR!

Thanks you for the initial idea and for testing.

theNestruo:
> Though before adding of these new settings, I'd like to ask that you check your controller settings (or windows settings, or steam input layer settings, ...). Maybe at some level there already is some transformation applied?

The root cause is the low-quality of my controller. The setting I need to change is called "buying a decent controller". :)

---

I've been thinking about the settings. I think inner-dead-zone, outer-dead-zone and sensitivity (or power curve) are valuable when mapping to a true analog value. But for this particular use case (detecting no steering, half steering, full steering), maybe using two threshold percentages would be easier to implement (and maybe also easier to understand and configure for the users).

So for my low-quality controller I could set it to 10%-95% so I have plenty of joystick movement to read as "half-steering" state, and your could set it to 15%-65% to read "full-steering" state without having to move the joystick to the full distance.

m9710797:
I've fixed the 'Remove' button. Please test again. For completeness here's the full(?) list of things to test:
* Hovering over the drawing should highlight that element and the corresponding row in the table (the gear-stick and buttons in the left pane, the steering wheel, the two buttons on the wheel).
* Hovering over a row in the table should highlight the corresponding element in the drawing.
* When something is bound to "Wheel", e.g. horizontal joystick axis, then moving the joystick should turn the wheel in the drawing.
* Test the 'Add' and 'Remove' button to create a custom binding. When there are multiple binding for one row, 'Remove' should open a popup, try removing a single binding or 'all bindings'.
* Try the 'Default bindings...' button. Does it make sensible default bindings? (When you create default bindings for keyboard we cannot bind the analog (wheel) input).
* Are the settings correctly saved/restored (you may need to manually execute `save_settings` in case `save_settings_on_exit=off`).
* Do replays with this input device work correctly?

If all the above works, then the only(?) thing left TODO is tuning the half-way value? Or is there still anything else?

Now about that half-way tuning. Maybe I'd like to split this in two parts:
* A quick temporary solution. Can we try to find an "acceptable" fixed value, that works on most controllers. Maybe 25000? After that we could maybe already push this PR to the master branch so that it's easier for more people to test.
* But I'd still like to continue with a proper solution for the joystick tuning. And preferably via the GUI. So also a GUI option for the current `joystick1_deadzone` setting. In addition to the new setting(s), but I'm not exactly sure which new settings yet.

theNestruo:
Thank you! I will test as soon as possible.

m9710797:
I played around with some ideas for making the analog joystick stuff configurable (allow to set the dead zones and half-way point).

In the "Configure MSX joysticks" window, scroll down and press the (sort of hidden) "Mockup calibrate joystick axis" button. That opens a new window:

THIS IS A MOCKUP ONLY: changing settings in this window doesn't have any effect yet. But I already wanted to ask for feedback. Do you think something like this could be useful? Both for the FS-JH1 emulation, but also for the other joystick emulation (there only(?) the inner dead zone is relevant I think).

theNestruo:
Tests:

- [x] Hovering over the drawing (...)
- [X] Hovering over a row (...)
- [X] (...) moving the joystick should turn the wheel in the drawing.
- [X] Test the 'Add' and 'Remove' button to create a custom binding. (...)
- [X] Try the 'Default bindings...' button. (...)
- [ ] Are the settings correctly saved/restored (...).
- [ ] Do replays with this input device work correctly?

I need to test more carefully the two last items, so please give me a little more time.

> In the "Configure MSX joysticks" window, scroll down and press the (sort of hidden) "Mockup calibrate joystick axis" button.

This configuration window is really visual and has any setting I could think of. I was suggesting something simpler because I though something like this would be too much work... but you nailed it!

m9710797:
Just a quick ping.
I haven't forgotten about this issue. It's just that other items always seem to skip ahead on my TODO list.
I hope to pick this up again soon.

theNestruo:
No problem. Summertime is keeping me mostly AFK, to be honest... I hope to be active again mid-september. I still owe you a couple of tests!

m9710797:
I've (force) pushed a new version of the branch. I think it's close to ready now.

Biggest remaining item I could use help with is testing.
And I do want to do a full review / cleanup of the code myself.

PR #2150, opened 2026-07-11
fix KEYPAD_COMMA

uniskie:
Fix KEYPAD_COMMA input bug introduced in d24d37d and add POSITIONAL mode support

### DIFF
--- a/src/events/SDLKey.hh
+++ b/src/events/SDLKey.hh
@@ -19,17 +19,17 @@ struct SDLKey {
 	SDL_Keysym sym;
 	bool down; // down=press / up=release
 
-	[[nodiscard]] static SDLKey create(SDL_Keycode code, bool down, uint16_t mod = 0) {
+	[[nodiscard]] static SDLKey create(SDL_Keycode code, SDL_Scancode scancode, bool down, uint16_t mod = 0) {
 		SDL_Keysym sym;
-		sym.scancode = SDL_SCANCODE_UNKNOWN;
+		sym.scancode = scancode; // scancode needs for POSITIONAL mapping mode
 		sym.sym = code;
 		sym.mod = mod;
 		sym.unused = 0; // unicode
 		return {.sym = sym, .down = down};
 	}
 
-	[[nodiscard]] static SDLKey createDown(SDL_Keycode code) {
-		return create(code, true);
+	[[nodiscard]] static SDLKey createDown(SDL_Keycode code, SDL_Scancode scancode) {
+		return create(code, scancode, true);
 	}
 
 	[[nodiscard]] static std::optional<SDLKey> fromString(std::string_view name);
--- a/src/input/Keyboard.cc
+++ b/src/input/Keyboard.cc
@@ -253,7 +253,8 @@ static constexpr auto getMSXMapping()
 		M{P{0x74}, K{SDLK_F8},          S{SDL_SCANCODE_F8}}, // STOP
 		M{P{0x75}, K{SDLK_BACKSPACE},   S{SDL_SCANCODE_BACKSPACE}},
 		M{P{0x76}, K{SDLK_F7},          S{SDL_SCANCODE_F7}}, // SELECT
-		M{P{0x77}, K{SDLK_RETURN},      S{SDL_SCANCODE_RETURN}},
+		M{P{0x77}, K{SDLK_RETURN,         SDLK_KP_ENTER},
+				   S{SDL_SCANCODE_RETURN, SDL_SCANCODE_KP_ENTER}},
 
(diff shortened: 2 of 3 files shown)


MBilderbeek:
Thanks! Your change also affects the unit tests. Can you take a look at that please?

uniskie:
Done. I've updated Keys_test.cc. 
Note that I used a dummy value for SCANCODE since it is not used in the unit tests.

MBilderbeek:
In the meantime, I think that you, @uniskie are the expert on the openMSX keyboard code... no one in our team has recently worked on it... So, could you help us understand the changes better? So, can you explain:
- what exactly is broken and how we can test to confirm that it's broken
- how exactly it got broken
- how these changes fix that and how we can then test that fix?

I hope that with that information, we (@m9710797 and me) learn something about this (old) code and can then discuss and approve the PR.

uniskie:
I’m not very good at English, so I might not be able to explain it well, but I’d appreciate your patience.

image

This is an issue regarding the behavior when pressing the ENTER key on the numeric keypad.
1. Even when `kbd_numkeypad_enter_key` is set to `KEYPAD_COMMA`, a comma is not input into the MSX.
2. When `kbd_numkeypad_enter_key` is set to `ENTER`, the ENTER key is input in both CHARACTER mode and KEY mode.
3. In POSITIONAL mode, neither ENTER nor a comma is input.

## The Beginning

It was introduced in commit  d24d37d6 (`[8/10] Switch keyboard events to SDL_Event`)

### Before d24d37d6

Matrix conversion was performed using a table indexed by proprietary internal key codes (K_****).

For instance, K_KP_ENTER (0x10F) was mapped to the numeric keypad comma (0xA6) in the MSX key matrix.

image

image

### After d24d37d6 

As a result of a major overhaul, the code was rewritten into a clear, organized format mapping `keymatrix`, `sdl_keycodes`, and `sdl_scancodes` to one another.

During this process, the key code corresponding to the MSX numeric keypad's comma key (Key matrix 0xA6) was changed from `K_KP_ENTER` to `SDLK_KP_COMMA`.

image
(This screen shot is an excerpt from `getMSXMapping()`)


Although the key codes in the table have changed, the `Keyboard::processKeypadEnterKey` handling for the numeric keypad's ENTER key remains the same.
(It has switched to SDL key codes, but the action of sending `KP_RETURN` is unchanged.)

That is why it is no longer possible to input a comma.

image

## Changes in this update

~~For the table, I reverted the handling of `KP_ENTER` to treat it as the MSX numeric keypad comma (0xA6), as it was previously.~~

In `Keyboard::processQueuedEvent`, I modified the code to toggle the key passed to `processSdlKey` between `KP_ENTER` and `KP_COMMA`, based on the `kbd_numkeypad_enter_key` setting.

Accordingly, I updated the table entries for the SDL_KEYCODE and SDL_SCANCODE corresponding to the MSX RETURN key and the numeric keypad comma key.

Regarding `Keyboard::processQueuedEvent`, I added a change to pass the scancode as well, in order to support POSITIONAL mode.

### KP_ENTER and KP_COMMA

Apparently, some host machines have a comma key on the numeric keypad where the Enter key is usually located (featuring one or the other, but not both).

To address this, I modified `Keyboard::processQueuedEvent` to call `Keyboard::processSdlKey` if the key is either `SDLK_KP_ENTER` or `SDLK_KP_COMMA` (i.e., `if ((key.sym.sym == SDLK_KP_ENTER) || (key.sym.sym == SDLK_KP_COMMA)) {`).

uniskie:
I made a mistake in part of the explanation, so I corrected it by adding a strikethrough.

MBilderbeek:
Thank you very much for your explanation. Don't worry about your English, it is fine. Now I think I should discuss with @m9710797 - as he made the changes. I also need to study...

uniskie:
There are multiple operating modes—such as MSX, SVI, ColecoVision Joystick, and Sega SC-3000/SK-1100—and it appears that the comma key on the numeric keypad is used for the MSX and SVI modes.

m9710797:
Thanks @uniskie for the extra background information. I did not understand the intention of this PR from your initial message, but now it's clear.

@MBilderbeek code changes look fine. But i did not have the time to test yet (I'm also at work ATM :-)
Can you (later) do some tests and merge this PR?

MBilderbeek:
> There are multiple operating modes—such as MSX, SVI, ColecoVision Joystick, and Sega SC-3000/SK-1100—and it appears that the comma key on the numeric keypad is used for the MSX and SVI modes.

Yes, the SVI-328 for instance, has an ENTER key on the numeric keypad, but also a comma key: 
image

MBilderbeek:
I confirmed that it works fine, so let's merge this.

PR #2153, opened 2026-07-15
Fix mouse speed scaling to resolve unnatural movement

uniskie:
Changed the mouse speed correction from a fixed 1/2 multiplier to a ratio based on the display size and the MSX screen internal resolution (240).

### DIFF
--- a/src/PluggableFactory.cc
+++ b/src/PluggableFactory.cc
@@ -61,7 +61,7 @@ void PluggableFactory::createAll(PluggingController& controller,
 	controller.registerPluggable(std::make_unique<ArkanoidPad>(
 		msxEventDistributor, stateChangeDistributor));
 	controller.registerPluggable(std::make_unique<Mouse>(
-		msxEventDistributor, stateChangeDistributor));
+		msxEventDistributor, stateChangeDistributor, display));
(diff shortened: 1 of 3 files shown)


m9710797:
I only had a very quick look yet. I agree this is mostly a good change(*1). I'll try to do a review this evening and then merge.

(*1) So I agree that in MSX programs where the mouse is used to move some cursor this change is probably an improvement.
But i also want to give a counter example: there are also MSX games (like "arkanoid"(?) where the mouse is used to control some paddle. And then it's weird if the controls feel different if you enlarge the openMSX window, or if you switch between fullscreen and windowed mode. I'm not sure what the best solution is (and if we need a solution at all).

uniskie:
You're right, we should probably address games like Arkanoid where mouse input isn't tied to standard cursor movement.One idea is to add a dedicated option that lets users manually override the scaling factor. Arkanoid still has a dedicated fanbase, so we can implement this if requested. This logic would also apply to other games that treat the mouse like a trackball

uniskie:
However, since the host machine's mouse speed is fundamentally adjusted to match the host screen size, this should presumably not cause much of an issue.

Issue #1331, opened 2021-02-20
Code/Kana acting weird (at least on Windows and macOS)


sdsnatcher73:
The Kana key is acting weird on my Windows 10 laptop as well as my macOS 12.1 MacBook.

On Windows 10 with a machine with Code/Kana lock (e.g. Yamaha CX7M) if I press right Alt the Kana light turns on, if I then press a key e.g. e sometimes e is entered sometimes nothing and Kana light turns off. If I press right Alt together with a key noting is entered in the MSX, it actually appears right Alt is handled like Ctrl. For example right Alt+g makes the MSX beep.

On Windows 10 with a machine with no Code/Kana lock (e.g. Yamaha CX5MII) if I press right Alt together with a key noting is entered in the MSX, it actually appears right Alt is handled like Ctrl. For example right Alt+g makes the MSX beep.

On macOS with a machine with Code/Kana lock (e.g. Yamaha CX7M)  if I press right Alt the Kana light turns on, if I then press a key e.g. e sometimes e is entered sometimes nothing and Kana light turns off. If I keep right Alt pressed and type a letter the Kana equivelant is typed.

On macOS with a machine with no Code/Kana lock (e.g. Yamaha CX5MII) if I press right Alt together with a key the corresponding code character is entered. For example right Alt + e enters î, the next character is however also changed once you release right Alt. so:
1. Enter right Alt + e -> î is entered
2. Enter e -> é is entered (for u -> ü)

Colpocorto:
This issue is causing my trouble as well. Any advance? It's almost 4 years old now.

MBilderbeek:
Does the keyboard mode have any influence on this?

MBilderbeek:
It is also interesting to know whether you have international keyboard mode of your OS enabled. That sometimes causes problems, e.g. see #1610 .

Colpocorto:
Still not working.  In my case, I use the Spanish keyboard layout. I don't ever touch the dead key (as mentioned in #1610) so I don't think both issues are related to each other.

Once I press Kana, I can't disable it anymore. Pressing it a second time turns the Kana LED off, but as soon as I press any other key (for instance, any English alphabet letter), the Kana key (and LED) turns itself on again.

The only solution I've found so far is closing OpenMSX totally. This is extremely annoying.

For reproducibility purposes, my current setup is:

-Windows 10 22H2
-Razer Cynosa V2, Spanish layout keyboard
[...]

Issue #974, opened 2015-08-10
Some keys unresponsive when trying to enter GRPH keys on OS X


gseidler:
When kbd_mapping_mode is set to CHARACTER I'm unable to type down many of the graphic characters while pressing the OPTION key. Sometimes triggering a stuck key repeat.  Seems to work fine when having kbd_mapping_mode set to KEY. 
Running openMSX 0.11.0 on Mac OS X 10.10.4

Example: OPT+ D (or G,H,J among others) doesn't print characters. 
Key Repeat: Hold OPT and press N 3 times and then M. Release all keys.

MBilderbeek:
This is what I get on Windows 8, so in principle it could work...
Can you make the same screen for Mac, please? (Leaving stuff blank if there's no output.)
![openmsx0002](https://cloud.githubusercontent.com/assets/12675978/9183167/f4e774d0-3fad-11e5-8e4e-73c5ea5ff5c9.png)

When I enable kbd_trace_key_presses, I get this for ASDF, btw:

```
Key pressed, unicode: 0x0000, keyCode: 0x00134, keyName: LALT
Key pressed, unicode: 0x0061, keyCode: 0x80061, keyName: A+ALT
Key released, unicode: 0x0000, keyCode: 0x480061, keyName: A+ALT,RELEASE
Key pressed, unicode: 0x0073, keyCode: 0x80073, keyName: S+ALT
Key released, unicode: 0x0000, keyCode: 0x480073, keyName: S+ALT,RELEASE
Key pressed, unicode: 0x0064, keyCode: 0x80064, keyName: D+ALT
Key released, unicode: 0x0000, keyCode: 0x480064, keyName: D+ALT,RELEASE
Key released, unicode: 0x0000, keyCode: 0x400134, keyName: LALT,RELEASE
```

Anyway, the output I get here is like how it is on a real MSX. What happens on OS X is definitely wrong.

gseidler:
![Screenshot](https://dl.dropboxusercontent.com/s/nikfnc7y025kmhr/kbd1.png)
This is what I get from all the ASDF line:

```
Key pressed, unicode: 0x0000, keyCode: 0x00134, keyName: LALT
Key pressed, unicode: 0x00e5, keyCode: 0x80061, keyName: A+ALT
Key released, unicode: 0x0000, keyCode: 0x480061, keyName: A+ALT,RELEASE
Key released, unicode: 0x0000, keyCode: 0x400134, keyName: LALT,RELEASE
[...]

Issue #797, opened 2015-06-24
[Bug] Keyboard emulation problems (mostly Mac) [sf#403]


openMSX-import:
**Reported by sd-snatcher on 2010-12-01 22:33 UTC**

The GRAPH and CODE keys of the emulated machine isn't working as expected. It can't produce most of the desired characters. Only a few keys produce output when pressed together with GRAPH or CODE.

Emulated machines: Toshiba HX-10, Sharp Hotbit HB-8000

Testbed:
openMSX-0.8.0 on Mac OS-X v10.6.5

openMSX-import:
**Commented by manuelbi on 2010-12-02 21:43 UTC**
The Hotbit probably won't be working well, because the keyboard matrix is not properly set, as you probably know (int mapping is used).

But the Toshiba should work and actually works here. At least with GRPH key. Which give the wrong output for you? Can you give some concrete examples?

openMSX-import:
**Updated by manuelbi on 2010-12-02 21:43 UTC**
- **labels**:  --> General/misc

openMSX-import:
**Commented by sd-snatcher on 2010-12-05 16:01 UTC**
Test the following on the emulated Toshiba HX-10:

Hold GRAPH (left alt) and press the sequence: FGHJ

On the real machine, by looking at the keyboard tables, this will produce:

Border character: |-
Border character: -|-
Border character: -|
Graphic character: upper half bg-color, lower half fg-color.

On the emulated HX-10: GRAPH+F will give you the "1/4" character and the rest of the sequence will just be ignored.

The same symptom occurs exactly the same way on the emulated Hotbit. Brazilian machines don't seem to differ from the european models when it comes to the {GRAPH,CODE}{,+SHIFT} characters.

openMSX-import:
**Commented by manuelbi on 2010-12-05 22:15 UTC**
Interesting, because I got at least the first 3 border characters and the last one is a graphic character with the left two columns fg color and the rest bg color.

Try to enable the tracing: set kbd_trace_keypresses on

This gives here:

Key pressed, unicode: 0x0000, keyCode: 0x00134, keyName: LALT
Key pressed, unicode: 0x0066, keyCode: 0x80066, keyName: F+ALT
Key released, unicode: 0x0000, keyCode: 0x480066, keyName: F+ALT,RELEASE
Key pressed, unicode: 0x0067, keyCode: 0x80067, keyName: G+ALT
[...]

PR #2159, opened 2026-07-26
Keyboard: ignore host Option-level characters on macOS

sndpl:
Addresses #797 and #974 (and the macOS side of #1331, partly - see the bottom).
Not using a closing keyword, since none of those tickets is only about this.

On macOS, in CHARACTER mapping mode, GRAPH and CODE combinations produce the
wrong MSX character or nothing at all. Reported since 2010, and reproducible
today.

## Cause

Both Option keys are MSX modifier keys for us: left Option is GRAPH
(`processGraphChange` on `SDLK_LALT`), right Option is CODE/KANA by default.
But the host *also* uses Option to select an alternate character from the
keyboard layout, and it does so before openMSX sees the key. So `option+f`
reaches us as U+0192, not as `f`.

In CHARACTER mode that character is looked up in the MSX keymap. Whenever it
happens to exist there, we press the keys that produce *that* character on the
MSX rather than the key the user actually hit.

That explains a detail which had made this look erratic: it depends on whether
the host's Option-level character exists in that machine's character set. On a
Toshiba HX-10, holding GRAPH and pressing F G H J:

```
                             MSX keys pressed
  KEY mode (correct)         CHARACTER mode (before)
  Option+F  [F, GRAPH]       [1, CODE, GRAPH]      <-- wrong
  Option+G  [G, GRAPH]       [G, GRAPH]
  Option+H  [H, GRAPH]       [H, GRAPH]
  Option+J  [J, GRAPH]       [J, GRAPH]
```

Only F is wrong, because U+0192 is part of the MSX character set so the lookup
succeeds. U+00A9, U+02D9 and U+2206 are not, so the lookup fails and the
existing fall-back to key-to-matrix mapping already does the right thing. Which
keys break therefore varies with the host layout and the emulated machine's
character set, which is why the reports each name a different set of keys.

## Fix

Ignore the host's interpretation while such an Option key is held, and use that
same key-to-matrix fall-back. This is exactly what SDL3 does for us through
`SDL_HINT_MAC_OPTION_AS_ALT`, whose `ReplaceEvent()` rebuilds the NSEvent using
`charactersIgnoringModifiers`; that hint does not exist in SDL2. Combinations
with Command or Control are left alone, matching SDL3, so the `Cmd+I` -> Insert
handling and the hotkeys are untouched. KEY and POSITIONAL mode already force
the unicode to zero, so nothing changes there, and on non-Apple builds the test
folds to `false`.

## How it was verified

Driven against a real emulator run with synthetic host key events. Ordinary
`CGEventCreateKeyboardEvent()` cannot reproduce this: it computes the event's
character payload from the keycode alone, so a synthetic option+f carries `f`
and the bug never appears. Using `CGEventKeyboardSetUnicodeString()` to supply
the character macOS actually produces replays the real event exactly - the
resulting `kbd_trace_key_presses` output matches the trace posted from real
hardware in #797 value for value:

```
unicode: 0x0192 (f-hook)  keyCode: F  keyName: F+ALT
unicode: 0x00a9 (c)       keyCode: G  keyName: G+ALT
unicode: 0x02d9 (dot)     keyCode: H  keyName: H+ALT
unicode: 0x2206 (delta)   keyCode: J  keyName: J+ALT
```

The oracle is KEY mapping mode, which both #797 and #974 report as working. The
check reads the emulated key matrix rather than the screen, so it does not
depend on character sets or screen mode. After the change, CHARACTER mode
presses the same MSX keys as KEY mode for all four combinations, over repeated
runs.

Also checked unchanged:

* the CODE/KANA lock traces from #2158, all three mapping modes;
* plain and shifted typing, and accented characters via the `type` command
  (`heloHELO` and `e-acute u-diaeresis` come out correctly).

## Not covered

Host dead keys. `option+e` arms a composition in the host's input context, and
the composed character then arrives with the *next* keystroke, which no longer
has Option held, so nothing on our side can distinguish it. Only the SDL3 hint
prevents the composition being armed at all. That is the remaining macOS half
of #1331.

### DIFF
--- a/src/input/Keyboard.cc
+++ b/src/input/Keyboard.cc
@@ -873,6 +873,21 @@ void Keyboard::syncHostKeyMatrix(EmuTime time)
 	lastUnicodeForScancode.clear();
 }
 
+/*
+ * Is an Option key held that we present to the MSX as a modifier key, and that
+ * the host therefore must not also have used to pick an alternate character?
+ * Combinations with Command or Control are left alone, matching what SDL3 does
+ * for SDL_HINT_MAC_OPTION_AS_ALT.
+ */
+[[nodiscard]] static constexpr bool optionIsMsxModifier([[maybe_unused]] uint16_t mod)
+{
+#ifdef __APPLE__
+	return (mod & KMOD_ALT) && !(mod & (KMOD_GUI | KMOD_CTRL));
+#else
+	return false;
+#endif
+}
+
 uint8_t Keyboard::needsLockToggle(const UnicodeKeymap::KeyInfo& keyInfo) const
 {
 	return modifierIsLock
@@ -1200,6 +1215,18 @@ bool Keyboard::processKeyEvent(EmuTime time, bool down, const KeyEvent& keyEvent
 			// combinations (e.g. digit on main keyboard or TAB/DEL).
 			// Use unicode to handle the more common combination and use direct
 			// matrix to matrix mapping for the exceptional cases (keypad).
+			unicode = 0;
+		} else if (optionIsMsxModifier(key.sym.mod)) {
+			// On macOS both Option keys are MSX modifier keys for us: left
+			// Option is GRAPH, right Option is CODE/KANA by default. But the
+			// host also uses Option to pick an alternate character from the
+			// keyboard layout, and it does so before we see the key: option+f
+			// arrives as U+0192 rather than 'f'. Whenever such a character
+			// happens to exist in the MSX character set, looking it up presses
+			// the keys that produce *that* character on the MSX instead of the
+			// key the user actually hit. Ignore the host's interpretation and
+			// fall back to key-to-matrix mapping, which is what SDL3 does for
+			// us via SDL_HINT_MAC_OPTION_AS_ALT.
 			unicode = 0;
 		} else {
 			unicode = keyEvent.getUnicode();


sndpl:
Latest stable release:
latest stable release
With fix from this PR:
fixed release

MBilderbeek:
I understand that (part of?) this fix comes from SDL3 code. We are also preparing to migrate to SDL3. Would it make sense to first migrate to SDL3 and then look at the keyboard situation again?
And if we don't, we should probably remove/revise this fix in the SDL3 branch. (And probably first upgrade to the latest SDL2 release?) @m9710797 what do you think?

MBilderbeek:
As for this fix: the problem seems that we use a host key as modifier key, that is also used on the host side to compose characters with (i.e. functions as a different modifier key on the host). 
As far as I understand this: this patch says that the host key must be first seen as MSX modifier key and not as host composer/modifier key. But suppose you want to type in an MSX character (in CHARACTER mode) that requires that host modifier key ("Option") to type it, how would that work then, @sndpl ? Isn't that now blocked with this patch?

sndpl:
Yes you are correct, there seems to be a fix for this issue in SDL3, but it seems I missed that there is a SDL3 branch (no draft PR for it?)

For you second question, you mean if you want to type an ö on the MSX? because then you still need to use CODE + F like on a real MSX with an international keyboard.

m9710797:
I'm only following this thread from a distance.

But I think the example is invalid. The point of CHARACTER mode is that if you want to type "ö", you need to type the **host** key combination, not the **MSX** key combination (CODE + F). So I don't immediately see it as a problem that CODE + F doesn't work to produce "ö".

And maybe that means not all MSX characters can be typed (when there is no corresponding host key combination). But that's fine for CHARACTER mode IMHO.

MBilderbeek:
> Yes you are correct, there seems to be a fix for this issue in SDL3, but it seems I missed that there is a SDL3 branch (no draft PR for it?)

There is only this branch: https://github.com/openMSX/openMSX/tree/sdl3 - no linked PR.

> For you second question, you mean if you want to type an ö on the MSX? because then you still need to use CODE + F like on a real MSX with an international keyboard.

Yeah, as @m9710797 also said: the point of CHARACTER mode is that you can use the host key combination (so as you are used to it on a host) to type the character and that openMSX will perform the MSX keyboard combination to get that character.

Suppose you have a QWERTZ host keyboard, then it is very logical that you can directly type the ö from the keyboard even on an English MSX keyboard.

It *seems* that there are situations of host and MSX keyboard in which both CHARACTER and POSITIONAL modes are not ideal. Not sure what the best solution is then. But to me it seems good to keep these modes consistent.

PR #2160, opened 2026-07-26
Update to SDL 2.32.10.

MBilderbeek:
This seems to build. Didn't test running.... @m9710797 feel free to merge. I suggest to just merge it and if problems arise, revert it or try another release instead. I expect no issues as it contains only bug fixes.


m9710797:
I never use the "3rdparty" build. I always use the build with system libraries. And via that way I was already using SDL 2.32.10 for some time (multiple months?). I didn't see any problem, so let's indeed just merge this.

Are there other libraries that need to be updated?

MBilderbeek:
I don't know, didn't check... In most cases, like this case, a version bump with only bug fixes is fairly trivial. Especially if we don't apply patches.
So, it's a matter of manually (or AI) checking if there are newer versions available.

PR #2161, opened 2026-07-28
Add ASCII8Flash ROM type: ASCII8 mapper with AM29F040B flash

plasticbugs:
Emulates homebrew 512KB flash cartridges that pair a standard ASCII8 mapper (bank registers at 0x6000/0x6800/0x7000/0x7800) with an AMD AM29F040B flash chip, fully writable, no SCC. Games on such carts can persist data (e.g. high scores) by erasing/programming the flash through the AMD command set; content persists via the usual SRAM file.

Implemented as a RomManbow2 variant since that class already pairs bank-switching mappers with AmdFlash; ASCII8FLASH selects ASCII8 register decoding and skips SCC/PSG creation.

Tested with a game that saves its high-score table to the last 64KB sector: autoselect ID reads, timed sector erase with status polling, byte program, and persistence across emulator restarts all behave like the real cartridge.


MBilderbeek:
Thanks!

Can you name some examples of such cartridges? Can they be bought
somewhere? Or is there some public design file people can use to build
their own?

Kind regards,
Manuel

Op di 28 jul. 2026 03:48 schreef Scott Moschella ***@***.***>:

plasticbugs:
Hi @MBilderbeek ! This is for a new cartridge that is not yet available but is currently being programmed. I am not certain if the board uses the actual AMD chip or an FPGA implementation yet. I am only the game programmer so I don’t have all the details. But we needed this change to test the game’s high score saving feature.

I will find out from the board developer and will get back to you regarding the specs and a programming guide. So for now, we should leave this as a draft until I can get you all the details. 

I can add implementation details in the source file as commented code or is there a better way to provide it?

MBilderbeek:
Okay, thanks for the clarification. In general we do not want to emulate non existing hardware. That's not emulation, but fantasy.
But once it is really available, it's most welcome. In the meantime you want to use the emulation of the hardware-to-be as development tool, of course. Luckily you can do that already. And when the hardware is finalised, let's make sure it's emulated properly.

I am usually a fan of documenting as closely as possible to the code, so that it is at least tempting to keep it updated with code changes...

plasticbugs:
> Okay, thanks for the clarification. In general we do not want to emulate non existing hardware. That's not emulation, but fantasy. But once it is really available, it's most welcome. In the meantime you want to use the emulation of the hardware-to-be as development tool, of course. Luckily you can do that already. And when the hardware is finalised, let's make sure it's emulated properly.
> 
> I am usually a fan of documenting as closely as possible to the code, so that it is at least tempting to keep it updated with code changes...

Here is the hardware, which my partner tells me that openMSX has support for in the beta. https://github.com/RBSC/MultiMapperCartridge

Though would a specific combination of Ascii8 + the AMD chip require the change on this branch?

m9710797:
Hi plasticbugs. Thanks for your PR.

I only now found the time to look at it. Sorry about the delay.

I think technically it works, but the approach might not be the best:
* You started from `RomManbow2` which is a Konami+SCC with optional PSG mapper.
* Then you disable SCC and PSG (was already possible in the existing code).
* And in addition you change the bank switching logic from Konami to Ascii8.

So basically almost nothing of the functionality in `RomManbow2` remains. And you made an already complex class even more complex.

Since this will be the first Ascii8 flash ROM mapper, I think it might be cleaner to create a fully new class for it. In total it will be more lines of code, but it will be much simpler.

I think you can start from these examples, and then combine them (copy/paste certain parts):
* `RomAscii8`: the basic ascii8 mapper implementation (but with a plain ROM instead of flash ROM).
* `RomMatraInk`: the simplest example I could find that uses a flash ROM.

You can also get inspiration from:
* `RomAscii16X`: that's a ascii16 mapper with 8MB flash. But because the flash is so big, the bank values required 12 bits, and more than 8 bits required some complications .. you don't need any of that.

----

Ah, I only now see your last comment:
> Here is the hardware, which my partner tells me that openMSX has support for in the beta. https://github.com/RBSC/MultiMapperCartridge

It seems the hardware is much more than only a Ascii8 flash ROM cartridge. And do I understand correctly that openMSX already does support that mapper? Then which mapper? Maybe Yamanooto??

Like MBilderbeek already said above: we prefer to only emulate existing hardware (not all hardware that could exist). Does that mean:
* We should close this PR because openMSX already does emulate the mapper? (Then my question is: which mapper is it?)
* We should extend your implementation to fully match the functionality of the "Multi Mapper Cartridge"?

PR #2163, opened 2026-08-01
workaround: add option to ignore Grave key after SDL scancode change

uniskie:
An SDL update changed the scancode for the "Zenkaku/Hankaku" key on Windows Japanese keyboards from `SDL_SCANCODE_UNKNOWN` to `SDL_SCANCODE_GRAVE`. This key causes major issues on Windows Japanese layouts because it never fires a Release event, causing the key state to get stuck.

Ideally, SDL should return `SDL_SCANCODE_LANG5`, but it practically returns `SDL_SCANCODE_GRAVE`. Since the Grave key is critical for US/English layouts, we cannot ignore it globally.

As a workaround, I added a boolean config option `kbd_ignore_grave_key` so users can toggle this behavior depending on their keyboard setup.

### DIFF
--- a/src/events/Event.hh
+++ b/src/events/Event.hh
@@ -411,6 +411,8 @@ class Rs232TesterEvent           final : public SimpleEvent {};
 class Rs232NetEvent              final : public SimpleEvent {};
 class ImGuiDelayedActionEvent    final : public SimpleEvent {};
 
+/** Send when changed host keyboard layout. */
+class KeyMapChangedEvent          final : public SimpleEvent {};
 
 // --- Put all (non-abstract) Event classes into a std::variant ---
 
@@ -452,7 +454,8 @@ using Event = std::variant<
 	Rs232TesterEvent,
 	Rs232NetEvent,
 	ImGuiDelayedActionEvent,
-	ImGuiActiveEvent
+	ImGuiActiveEvent,
+	KeyMapChangedEvent
 >;
 
 template<typename T>
@@ -501,6 +504,7 @@ enum class EventType : uint8_t
 	RS232_NET                = event_index<Rs232NetEvent>,
 	IMGUI_DELAYED_ACTION     = event_index<ImGuiDelayedActionEvent>,
 	IMGUI_ACTIVE             = event_index<ImGuiActiveEvent>,
+	KEYMAP_CHANGED           = event_index<KeyMapChangedEvent>,
 
 	NUM_EVENT_TYPES // must be last
 };
--- a/src/events/InputEventGenerator.cc
+++ b/src/events/InputEventGenerator.cc
@@ -323,6 +323,10 @@ void InputEventGenerator::handle(SDL_Event& evt)
 		handleKeyDown(evt.key, 0);
 		break;
 
+	case SDL_KEYMAPCHANGED:
+		event = KeyMapChangedEvent();
+		break;
+
 	case SDL_MOUSEBUTTONUP:
 		event = MouseButtonUpEvent(evt);
 		break;
--- a/src/input/Keyboard.cc
+++ b/src/input/Keyboard.cc
@@ -955,6 +955,13 @@ bool Keyboard::processQueuedEvent(const Event& event, EmuTime time)
 	auto key = keyEvent.getKey();
 
 	// Check if the host keyboard is a Japanese keyboard. If so, ignore the GRAVE key.
+	// Reason:
+	// On Windows with a Japanese keyboard layout, the "Half-width/Full-width" key 
+	// does not send a release event. This causes the key to get stuck in a pressed state.
+	// This key is for switching input modes, but its physical location is the GRAVE key.
+	// Japanese MSX does not have a GRAVE key and does not use it, so we simply ignore it.
+	// Previously, we could easily ignore it because the scancode was UNKNOWN.
+	// Recently, it started reporting as a GRAVE key scancode, so we changed the condition here.
 	if (hostKeyMap == HostKeyMap::UNKNOWN) {
 		if (SDL_GetKeyFromScancode(SDL_SCANCODE_LEFTBRACKET) == SDLK_AT) {
 			hostKeyMap = HostKeyMap::JP;
--- a/src/input/Keyboard.cc
+++ b/src/input/Keyboard.cc
@@ -670,6 +670,7 @@ Keyboard::Keyboard(MSXMotherBoard& motherBoard,
 	, commandController(commandController_)
 	, msxEventDistributor(msxEventDistributor_)
 	, stateChangeDistributor(stateChangeDistributor_)
+	, eventDistributor(eventDistributor)
 	, keyCodeTab (defaultKeyCodeMappings [matrix])
 	, scanCodeTab(defaultScanCodeMappings[matrix])
 	, modifierPos(modifierPosForMatrix[matrix])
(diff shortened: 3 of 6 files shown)


MBilderbeek:
Instead of this workaround, can we also fix the root cause somehow?

Did I understand correctly that this is caused by SDL2?
Is it better in SDL3, perhaps?

uniskie:
To explain this problem, I need to try a few variations until I can confirm the connections between the information through an accurate automatic translation. Please give me some time.

uniskie:
(This is a test post for explanation purposes.)

### 1. Can this be resolved fundamentally?

A fundamental solution is not feasible within the scope of a single pull request.

*   **Architectural Limitation:** openMSX emulates the MSX physical keyboard matrix by mapping high-level OS key events (`KEYDOWN` and `KEYUP`). When the Japanese IME is activated, the OS hooks the "Half-width/Full-width" (Grave) key and consumes the corresponding `KEYUP` event. Because openMSX never receives this `KEYUP` event, the key remains permanently stuck. Furthermore, while the IME is active, almost all other regular key inputs (such as W, A, S, D, and even ESC) are captured by the IME input buffer and fail to register correctly in the emulator.
*   **Scope of a Fundamental Fix:** Resolving this fundamentally would require abandoning high-level OS key events entirely and implementing a custom input system based on raw key-scanning APIs (such as Raw Input) via physical polling. Doing so requires a major rewrite of the entire cross-platform input architecture, which is too broad for this PR.
*   **Why a manual toggle is preferred over state tracking:** While it is possible to implement state tracking similar to the Macintosh Caps Lock workaround, the physical key at the GRAVE position is not used in the Japanese JIS MSX layout. Ignoring the event completely via a manual toggle is a cleaner and less complex solution than implementing state tracking.
*   **Why automatic layout detection is impractical:** Dynamic layout auto-detection is highly prone to regressions. Key mapping behavior varies heavily across different operating systems, window managers, and keyboard layouts (such as US, UK ISO, German QWERTZ, or French AZERTY). The verification cost required to guarantee that automatic detection does not break other layouts is too high to expect from a PR contributor.

Therefore, the manual setting to completely ignore the GRAVE key events is the most practical workaround.

---

### 2. Is SDL2 the cause of this issue?

No, SDL2 is not the cause.

The shift in the scancode from `UNKNOWN` to `GRAVE` is due to the Text Services Framework (TSF) based IME introduced in Windows 10 v2004 and Windows 11.
*   Under the legacy IME, the OS sent a scancode of 0 (`VK_PROCESSKEY`), which SDL classified as `UNKNOWN`.
*   The new IME directly sends the physical scancode `0x29` (the GRAVE position), which SDL then maps correctly to the `GRAVE` keycode.

---

### 3. Is this behavior improved in SDL3?

No, the behavior is identical in SDL3.

The issue is caused by the OS-level IME consuming the `KEYUP` event. Because this OS-level behavior does not change, the key will still get stuck in SDL3 when the IME is active. The only difference is that the keycode name changes from `SDLK_BACKQUOTE` to `SDLK_GRAVE`.

m9710797:
Technically the code change is fine. The only problem is that users will encounter this problem, but still have no idea what to do about it: they won't know about this setting, or how/when to change it.
* So my preference would be to "somehow" automatically solve it (no need for a setting). But maybe that's not possible?
* Or else we should clearly document this in the manual. We should clearly describe the symptom, and then what to do (which setting to change to what value). And maybe we should also describe what happens if you wrongly activated this setting?

uniskie:
The reason auto-detection was not adopted is that the conditions could not be definitively guaranteed, and because this was a highly urgent issue among Japanese users, a faster and more reliable method was chosen. However, as you pointed out, clarity and proper documentation are essential.

I will do some further research to see if there is a better approach.

If it still seems too risky, implementing a suppression switch would be the logical next step. (In that case, adding an explanatory note will definitely be necessary.)

---

Currently, under international standards, an incorrect code (GRAVE) is being transmitted for the key defined as LANG5. Therefore, it is necessary to investigate the actual real-world behavior rather than just relying on the specifications.

Changes such as this notification keycode modification might cause auto-detection to become unreliable in the future. To prepare for such cases, providing a feature to configure a user-defined "ignore key" setting can also be considered as an alternative option. (Though it would require a detailed explanation for users.)

MBilderbeek:
I was wondering what the cause is that GRAVE is getting emitted instead of LANG5. Is that an SDL2 bug?

uniskie:
To clarify the situation: the reason it is treated as GRAVE instead of LANG5 is due to a legacy convention in USB guidelines designed to maintain PS/2 compatibility for physical keyboard mapping. Thus, it is not an SDL issue. (This topic is always quite confusing.)

As a separate matter, the fact that the key event's scancode changed from UNKNOWN to GRAVE seems to be caused by a recent specification change in the IME/FEP.

In language environments that use an IME or FEP, there are certain keys that shouldn't be handled directly via raw key events. Consequently, typical applications in these environments are usually designed to handle only text input and a limited set of control codes through key events. This is why I hadn't looked into the finer details of this issue before.

MBilderbeek:
Is it somehow possible to detect whether an IME is being active or used? So that we can take measures for that situation, like managing the GRAVE event.

uniskie:
I'm also looking into whether we can query the host environment to determine the layout and handle the key state correctly.

My findings suggest that the root cause is not the presence of an IME, but rather the active keyboard layout itself. For example, even on a US language profile (with no IME installed), if a Japanese keyboard layout is specified, the GRAVE key fails to trigger a RELEASE event. I also recall seeing reports that CAPS key behavior varies by layout as well. 

I'll do some more research to see if we can use layout detection for a precise workaround.

MBilderbeek:
Thanks a lot! Looking forward to the results.

uniskie:
New commit introduces a cross-platform layout detection method without
relying on Win32 APIs. By calling SDL_GetKeyFromScancode(SDL_SCANCODE_LEFTBRACKET),
we check if it maps to SDLK_AT, which is a characteristic of Japanese layouts.

To avoid performance overhead, the layout is lazily evaluated on the first
key event and cached using std::optional<bool>. The cache is reset
whenever SDL_KEYMAPCHANGED is received.

(`ignoreGraveKey` switch was deleted.)

m9710797:
Thanks!
I like this solution much better. From a user perspective it now works automatically.

My only request is to add a little more comments in the source code, so that in a few years from now we still know why it was done like this (why and when ignoring `SDL_SCANCODE_GRAVE` is needed). Maybe just adding a single line that refers to this PR (`https://github.com/openMSX/openMSX/pull/2163`) is sufficient?

After that (later today when I'm back at home) I can merge this PR.

uniskie:
Indeed, I might have cut out the comments during the revision. Yesterday, I collapsed with a high fever and was unable to work, so I plan to add the comments tonight.

uniskie:
Added the following comment:

// Check if the host keyboard is a Japanese keyboard. If so, ignore the GRAVE key.
// Reason:
// On Windows with a Japanese keyboard layout, the "Half-width/Full-width" key 
// does not send a release event. This causes the key to get stuck in a pressed state.
// This key is for switching input modes, but its physical location is the GRAVE key.
// Japanese MSX does not have a GRAVE key and does not use it, so we simply ignore it.
// Previously, we could easily ignore it because the scancode was UNKNOWN.
// Recently, it started reporting as a GRAVE key scancode, so we changed the condition here.

m9710797:
Thanks!

PR #2165, opened 2026-08-14
Add 'Recent machines' submenu to the Machine menu 

### LINKED ISSUE #2137: Move recently used from select machine to the main machine menu
The recently used machines is now a dropdown in the "Select MSX Machine" panel. I think it would be better to move this to the main machine menu below the "Select MSX Machine" item like this:
 
Image

Many programs have this (editors, word, excel, etc). If this is ok, I will create a PR for it.

sndpl:
Switching to a recently used machine required opening the 'Select MSX machine' window first. This adds a Recent machines submenu directly below Select MSX machine..., like most applications have. Entries show the machine's display name, carry the usual config tooltip on hover, and are drawn in the error color when the machine fails its hardware test (and are then not selectable).

The existing recent-machines list inside the 'Select MSX machine' window is kept as-is, because that window is also reachable via the Quick setup editor.

The submenu is grayed out when no machine has been selected via the GUI yet, and also when none of the remembered machines still exists — the names are remembered by config name in imgui.ini and never validated, so they can all point at machines that are no longer installed, which previously left the submenu enabled but empty.

Fixes #2137.

**How the machine info is cached**

ImGuiMachine keeps a machineInfo cache: one entry per installed machine config, filled by running openmsx_info machines <config> for every config found on disk. With 200+ configs that's about 13ms, so it's built once and reused. findMachineInfo() is a plain lookup in that cache.

Previously a lookup miss wiped and refilled the whole cache, so that configs added or renamed while openMSX was running were still picked up. Since findMachineInfo() is called from painting code, one stale name in the recent list was enough to trigger that rescan every frame. As discussed in review, openMSX now simply assumes machine config files don't change while it is running, and that refresh is gone.

**Trade-off**

Machine configs added, renamed or removed while openMSX is running are not picked up until you hit Machine → Test MSX hardware → Rerun test (or restart).

### DIFF
--- a/src/imgui/ImGuiMachine.cc
+++ b/src/imgui/ImGuiMachine.cc
@@ -152,6 +152,7 @@ void ImGuiMachine::showMenu(MSXMotherBoard* motherBoard)
 		const auto& hotKey = reactor.getHotKey();
 
 		ImGui::MenuItem("Select MSX machine...", nullptr, &showSelectMachine);
+		showRecentMachinesMenu();
 
 		auto showSetupDepthLevelSelector = [&](const std::string& displayText, const bool includeNone, SetupDepth currentDepth) {
 			static constexpr array_with_enum_index<SetupDepth, zstring_view> helpText = {
@@ -404,6 +405,34 @@ void ImGuiMachine::showMenu(MSXMotherBoard* motherBoard)
 			previewSetup.motherBoard.reset();
 		});
 	}
+}
+
+void ImGuiMachine::showRecentMachinesMenu()
+{
+	// grayed-out as long as no machine has been selected via the GUI
+	im::Menu("Recent machines", !recentMachines.empty(), [&]{
+		// don't switch machine (and thus modify 'recentMachines') while iterating over it
+		std::string selectedMachine;
+		for (auto [i, item] : enumerate(recentMachines)) {
+			auto* info = findMachineInfo(item);
+			if (!info) continue;
+			bool ok = getTestResult(*info).empty();
+			im::StyleColor(!ok, ImGuiCol_Text, getColor(imColor::ERROR), [&]{
+				if (ImGui::MenuItem(tmpStrCat(info->displayName, "##", i).c_str()) && ok) {
+					selectedMachine = info->configName;
+				}
+			});
+			if (ImGui::IsItemHovered(ImGuiHoveredFlags_DelayNormal | ImGuiHoveredFlags_NoSharedDelay | ImGuiHoveredFlags_Stationary)) {
+				im::ItemTooltip([&]{
+					printConfigInfo(*info);
+				});
+			}
+		}
+		if (!selectedMachine.empty()) {
+			manager.executeDelayed(makeTclList("machine", selectedMachine));
+			addRecentItem(recentMachines, selectedMachine);
+		}
+	});
 }
 
 void ImGuiMachine::signalQuit()
--- a/src/imgui/ImGuiMachine.cc
+++ b/src/imgui/ImGuiMachine.cc
@@ -409,8 +409,12 @@ void ImGuiMachine::showMenu(MSXMotherBoard* motherBoard)
 
 void ImGuiMachine::showRecentMachinesMenu()
 {
-	// grayed-out as long as no machine has been selected via the GUI
-	im::Menu("Recent machines", !recentMachines.empty(), [&]{
+	// grayed-out as long as no machine has been selected via the GUI, and also
+	// when none of the remembered machines exists (anymore)
+	bool anyMachine = std::ranges::any_of(recentMachines, [&](const auto& item) {
+		return findMachineInfo(item) != nullptr;
+	});
+	im::Menu("Recent machines", anyMachine, [&]{
 		// don't switch machine (and thus modify 'recentMachines') while iterating over it
 		std::string selectedMachine;
 		for (auto [i, item] : enumerate(recentMachines)) {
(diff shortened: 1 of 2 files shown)


src/imgui/ImGuiMachine.cc
```diff
@@ -406,6 +407,38 @@ void ImGuiMachine::showMenu(MSXMotherBoard* motherBoard)
...
+{
+	// grayed-out as long as no machine has been selected via the GUI, and also
+	// when none of the remembered machines exists (anymore)
+	bool anyMachine = std::ranges::any_of(recentMachines, [&](const auto& item) {
+		return findMachineInfo(item) != nullptr;
+	});
+	im::Menu("Recent machines", anyMachine, [&]{
+		// don't switch machine (and thus modify 'recentMachines') while iterating over it
+		std::string selectedMachine;
+		for (auto [i, item] : enumerate(recentMachines)) {
```
m9710797: I'm not sure because of the `enumerate`, but should we use `const auto&` instead of plain `auto` to avoid making a copy of the `std::string`? If it compiles then I'd change it to `const auto&` (even if it's not strictly needed).

src/imgui/ImGuiMachine.cc
```diff
@@ -1166,15 +1205,20 @@ bool ImGuiMachine::printConfigInfo(MachineInfo& info)
...
+	};
+	if (auto* info = search()) return info;
+
+	// Not found, maybe a config was added or renamed on disk. Refresh and retry,
+	// but only once per name: refreshing parses all (>200) config files and we're
+	// called while painting (compare getTestResult()).
+	if (contains(unknownMachines, config)) return nullptr;
+	unknownMachines.emplace_back(config);
+	machineInfo.clear();
+	return search();
```
m9710797: I understand why you added this. In the GUI we call `findMachineInfo()` every frame, and we absolutely need to avoid re-scanning all machine configs every frame.

However I think this is the wrong approach. The existing code that did the re-scan was already wrong IMHO. It's introduced in a25e595. That commit does 2 things:
* Print a fallback name (instead of crashing) in case of no longer existing machine configs.
* (Code in this function) clear the cache in case `findMachineInfo()` failed.

The first bullet was the main point of that commit. The second bullet was a drive-by tweak, but it's wrong IMHO. 

I think openMSX should be allowed to assume that the machine/extension config files don't change while openMSX is running. And as noted in this PR's description: there's still the `Rerun test` button in the `Test MSX hardware` window if you *really* don't want to restart openMSX.

So what I'd propose is to drop the old lines 1171-1176 (the lines that a25e595 added in this function). And then also revert large parts of this PR that deal with working around this mistake. That means dropping the new `unknownMachines` variable, and removing the `resetMachineInfo()` function.

@MBilderbeek since you wrote a25e595, can you also comment. Do you agree that the main point of your commit was the 1st bullet. And now (with hindsight) do you agree that the 2nd bullet was a mistake?
EDIT: possibly at the time I did review your code and approved it. But I didn't realize the negative effects of clearing the cache.

m9710797 (changes requested):
Thanks for your PR!

It looks good. Though I'd like to ask for some changes. Not because any of your code is wrong, but because you rightfully work around a mistake in the existing code. But instead of working around it, I'd prefer to fix that mistake instead.

More details in the comments.

src/imgui/ImGuiMachine.cc
MBilderbeek: @m9710797 can you be a bit clearer about what the problem is with the folder rescan? I would expect it would happen only once in the presented use case and I don't understand the downsides of it at the moment.

m9710797: Suppose the list of "recent machines" contains a config name that no longer exists on disk. Or another scenario is as described in the commit message of a25e5959250d7d57e008ce12435596a3ddef6c1c.

As part of ImGui rendering (each frame again and again) we call `findMachineInfo()`. But if the file is really no longer present on disk, then clearing the cache won't help in finding it. So on the next frame we call `findMachineInfo()` again, this again does all full parse of all machine configs, and again on the next frame etc.

Also, as far as I understand, clearing the cache in a25e5959250d7d57e008ce12435596a3ddef6c1c was not needed to fix the problem described in the commit message. I think it was just something "extra" you did?

MBilderbeek:
```
<Quibus> wouterv: I think the idea was that if there is a bad entry in the cache, it should be refreshed
<Quibus> It was of course not the idea to refresh that each frame
<Quibus> IIRC the idea was that it would refresh only once
<Quibus> So, perhaps the condition of the refresh was wrongly implemented?
<Quibus> IIRC the idea was: if machine found in cache, but it is invalid, clear the cache to rebuild it.
<Quibus> But I didn't dive into it (yet)
<wouterv> what was implement is:   lookup not found in cache:  throw away the cache
<wouterv> i propose to just drop this stuff.  So let openMSX assume that config files it once read remain unchanged.         (some clarification:  1) the cache is only about the meta data in the config file   2) if we later actually load the machine and the file is then missing we should not crash)
```
Well, if the current implementation of that cache stuff is only causing trouble and does nothing useful, then please drop it.

src/imgui/ImGuiMachine.cc
sndpl: updated it

src/imgui/ImGuiMachine.cc
sndpl: Updated the code, it indeed became a lot simpeler to add this menu item now :)

m9710797:
Thanks!

PR #2167, opened 2026-08-15
Add ASCII8-R mapper for reversed ASCII 8kB bank registers

sndpl:
The retail cartridge of XeGrader uses the regular ASCII 8kB scheme, but with the four bank registers wired to the memory regions in the opposite order:

  0x4000 - 0x5fff <- register at 0x7800 - 0x7fff
  0x6000 - 0x7fff <- register at 0x7000 - 0x77ff
  0x8000 - 0x9fff <- register at 0x6800 - 0x6fff
  0xa000 - 0xbfff <- register at 0x6000 - 0x67ff

Loading such a ROM as plain ASCII8 puts the banks in the wrong place, and no existing mapper covers this wiring, so add RomType::ASCII8R, named "ASCII8-R" in the database and the GUI.

It reuses RomAscii8kB with a new 'reversedRegs' constructor flag that mirrors the register-to-region mapping (5 - reg instead of reg + 2); reset, write cache lines and serialization are unaffected.

Also add the softwaredb entry for the dump of the retail cartridge (will also send the sha1 to the MSX ROM database)

### DIFF
--- a/share/softwaredb.xml
+++ b/share/softwaredb.xml
@@ -13493,6 +13493,7 @@ The Database is located at https://romdb.vampier.net/
     <rom sha1="6ea5bb55d3f5c242d3cd1e8af74d2078b568f5b9" type="ASCII8" status="Author" remark="Portuguese" />
   </software>
   <software title="XeGrader" system="MSX2" company="Tokihiro Naito" year="2025" country="JP">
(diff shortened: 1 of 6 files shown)


m9710797:
Thanks!

PR #2168, opened 2026-08-15
HD: register with DiskManipulator, so SD cards and SCSI disks are usable

### LINKED ISSUE #1795: Support disk manipulator for MegaFlashROM SCC+SD
Diskmanipulator doesn't list the partitions of my virtual SD card of the MegaFlashROM SCC+ SD. AFAIK Nextor partitions are supposed to be supported, but maybe I'm wrong.

Anyways, it would be very helpful to be able to use diskmanipulator with this extension. Very hard to get stuff on it, otherwise.

sndpl:
Registration lived in IDEHD, so only IDE hard disks showed up in the 'diskmanipulator' command and in the Disk Manipulator window. The SD cards of the MegaFlashROM SCC+ SD and SCSI hard disks are plain HD objects and were therefore invisible, even though HD already implements DiskContainer and the partition-table code understands Nextor. Move register/unregisterDrive() to HD and update the (now stale) note in the manual.

Fixes #1795

image

### DIFF
--- a/doc/manual/diskmanipulator.html
+++ b/doc/manual/diskmanipulator.html
@@ -29,10 +29,21 @@
   of this document you will find some <a class="internal" href="#examples">examples of typical use</a>.</p>
 
   <div class="note">
-    Note: For now, this command does not work with SCSI/SD disks and drives!
-    So, you cannot create valid formatted SCSI hard disk images, SD card images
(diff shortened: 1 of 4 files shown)


m9710797:
Thanks!

Issue #1659, opened 2024-06-04
Add slot map GUI view to Tools


MBilderbeek:
Make a graphical view (accessible from the main menu somewhere) of the information presented by the `slotmap` command on the console.

sndpl:
I have created an initial version for this, see the PR for some screenshots.

MBilderbeek:
Implemented via #2169 - thanks a lot!

PR #2169, opened 2026-08-16
feat(imgui): add Slot map window to Tools

sndpl:
Adds a graphical version of the 'slotmap' console command, requested in issue #1659.

The window shows a grid of slots against the four 16kB pages, in two layouts, switchable with a dropdown and remembered in imgui.ini:

- vertical (default): pages as rows with the highest address on top and slots as columns, matching the memory maps in the MSX wiki and in most manuals
- horizontal: the transpose, slots as rows and pages as columns

The layout is fixed at 4x4 slots, so a non-expanded primary slot takes the same space as four sub-slots and machines with different slot layouts still look alike. Cells are colored empty / occupied / external cartridge slot, a device covering several pages is drawn as one merged cell, and long device names wrap over the taller cells (falling back to an ellipsis plus a tooltip when even that doesn't fit).

MSXCPUInterface registers a device that covers less than a whole page through an MSXMultiMemDevice, which knows the exact address ranges. Add an accessor for those ranges so such a page can be split proportionally, e.g. the 8kB RAM at 0xE000 of a Casio PV-7 only fills the upper half of its page instead of the whole page as 'slotmap' reports it.

Vertical:
image

Horizontal:
image

### DIFF
--- a/src/CartridgeSlotManager.cc
+++ b/src/CartridgeSlotManager.cc
@@ -193,13 +193,9 @@ void CartridgeSlotManager::createExternalSlot(int ps, int ss)
 
 unsigned CartridgeSlotManager::getSlot(int ps, int ss) const
 {
-	for (auto slot : xrange(MAX_SLOTS)) {
-		if (slots[slot].exists() &&
-		    (slots[slot].ps == ps) && (slots[slot].ss == ss)) {
-			return slot;
-		}
-	}
-	UNREACHABLE; // was not an external slot
+	auto slot = findSlot(ps, ss, false);
+	if (!slot) UNREACHABLE; // was not an external slot
+	return *slot;
 }
 
 void CartridgeSlotManager::testRemoveExternalSlot(
@@ -351,10 +347,7 @@ void CartridgeSlotManager::freeSlot(
 
 bool CartridgeSlotManager::isExternalSlot(int ps, int ss, bool convert) const
 {
-	return std::ranges::any_of(xrange(MAX_SLOTS), [&](auto slot) {
-		int tmp = (convert && (slots[slot].ss == -1)) ? 0 : slots[slot].ss;
-		return slots[slot].exists() && (slots[slot].ps == ps) && (tmp == ss);
-	});
+	return findSlot(ps, ss, convert).has_value();
 }
 
 std::string CartridgeSlotManager::insertCartridge(
--- a/src/CartridgeSlotManager.cc
+++ b/src/CartridgeSlotManager.cc
@@ -194,7 +194,7 @@ void CartridgeSlotManager::createExternalSlot(int ps, int ss)
 unsigned CartridgeSlotManager::getSlot(int ps, int ss) const
 {
 	auto slot = findSlot(ps, ss, false);
-	if (!slot) UNREACHABLE; // was not an external slot
+	assert(slot); // was an external slot
 	return *slot;
 }
 
--- a/src/cpu/MSXMultiMemDevice.hh
+++ b/src/cpu/MSXMultiMemDevice.hh
@@ -30,7 +30,7 @@ public:
 	[[nodiscard]] std::vector<MSXDevice*> getDevices() const;
 
 	/** The address ranges actually in use, in no particular order.
-	  * The sentinel at the back is not included. */
+	  * The sentinel is not included. */
 	[[nodiscard]] std::span<const Range> getRanges() const {
 		return {ranges.data(), ranges.size() - 1};
 	}
@@ -48,7 +48,10 @@ private:
 	[[nodiscard]] const Range& searchRange(unsigned address) const;
 	[[nodiscard]] MSXDevice* searchDevice(unsigned address) const;
 
-	std::vector<Range> ranges; // ordered (sentinel at the back)
+	// In no particular order, except that the sentinel is at the back: it
+	// covers the whole address space, so searching front-to-back finds the
+	// real ranges first and always finds something.
+	std::vector<Range> ranges;
 };
 
 } // namespace openmsx
--- a/src/cpu/MSXMultiMemDevice.hh
+++ b/src/cpu/MSXMultiMemDevice.hh
@@ -2,12 +2,23 @@
 #define MSXMULTIMEMDEVICE_HH
 
 #include "MSXMultiDevice.hh"
+#include <span>
 #include <vector>
 
 namespace openmsx {
 
 class MSXMultiMemDevice final : public MSXMultiDevice
 {
+public:
+	struct Range {
+		Range(unsigned base_, unsigned size_, MSXDevice& device_);
+		[[nodiscard]] constexpr bool operator==(const Range&) const = default;
+
+		unsigned base;
+		unsigned size;
+		MSXDevice* device;
+	};
+
 public:
 	explicit MSXMultiMemDevice(HardwareConfig& hwConf);
 	~MSXMultiMemDevice() override;
@@ -17,6 +28,12 @@ public:
 	void remove(MSXDevice& device, unsigned base, unsigned size);
 	[[nodiscard]] bool empty() const { return ranges.size() == 1; }
 	[[nodiscard]] std::vector<MSXDevice*> getDevices() const;
+
+	/** The address ranges actually in use, in no particular order.
+	  * The sentinel at the back is not included. */
+	[[nodiscard]] std::span<const Range> getRanges() const {
+		return {ranges.data(), ranges.size() - 1};
+	}
 
 	// MSXDevice
 	[[nodiscard]] const std::string& getName() const override;
@@ -28,15 +45,6 @@ public:
 	[[nodiscard]] uint8_t* getWriteCacheLine(uint16_t start) override;
 
 private:
-	struct Range {
-		Range(unsigned base_, unsigned size_, MSXDevice& device_);
-		[[nodiscard]] constexpr bool operator==(const Range&) const = default;
-
-		unsigned base;
-		unsigned size;
-		MSXDevice* device;
-	};
-
 	[[nodiscard]] const Range& searchRange(unsigned address) const;
 	[[nodiscard]] MSXDevice* searchDevice(unsigned address) const;
 
--- a/src/imgui/ImGuiSlotMap.cc
+++ b/src/imgui/ImGuiSlotMap.cc
@@ -2,6 +2,7 @@
 
 #include "ImGuiCpp.hh"
 #include "ImGuiManager.hh"
+#include "ImGuiMedia.hh"
 #include "ImGuiUtils.hh"
 
 #include "CartridgeSlotManager.hh"
@@ -42,9 +43,9 @@ void ImGuiSlotMap::loadLine(std::string_view name, zstring_view value)
 	loadOnePersistent(name, value, *this, persistentElements);
 }
 
-// The letter of the external cartridge slot at (ps, ss), if any.
+// The external cartridge slot at (ps, ss), if any.
 // 'ss' must already be normalized: 0 for a non-expanded primary slot.
-[[nodiscard]] static std::optional<char> getCartridgeLetter(
+[[nodiscard]] static std::optional<unsigned> findCartridgeSlot(
 	const CartridgeSlotManager& slotManager, int ps, int ss)
 {
 	for (auto slot : xrange(CartridgeSlotManager::MAX_SLOTS)) {
@@ -52,7 +53,7 @@ void ImGuiSlotMap::loadLine(std::string_view name, zstring_view value)
 		auto [slotPs, slotSs] = slotManager.getPsSs(slot);
 		if (slotSs == -1) slotSs = 0; // not expanded
 		if ((slotPs == ps) && (slotSs == ss)) {
-			return char('A' + slot);
+			return slot;
 		}
 	}
 	return {};
@@ -109,22 +110,6 @@ static void getBlocks(std::vector<Block>& result, std::vector<Block>& scratch,
 	if (pos < ADDRESS_SPACE) result.emplace_back(pos, ADDRESS_SPACE, nullptr);
 }
 
-// Mirror of ImGuiUtils' leftClip(): keep the start of the string and replace
-// the part that doesn't fit with an ellipsis.
-[[nodiscard]] static std::string rightClip(std::string_view s, float maxWidth)
-{
-	maxWidth -= ImGui::CalcTextSize("..."sv).x;
-	if (maxWidth <= 0.0f) return "...";
-
-	// The first length that is too wide, minus one, is the longest prefix that
-	// still fits. Length 0 always fits, so that first length is never 0.
-	auto lengths = std::views::iota(0uz, s.size() + 1);
-	auto it = std::ranges::upper_bound(lengths, maxWidth, {},
-		[&](size_t n) { return ImGui::CalcTextSize(s.substr(0, n)).x; });
-	auto num = (it == lengths.end()) ? s.size() : (*it - 1);
-	return strCat(s.substr(0, num), "...");
-}
-
 // Draw 'text' centered in the rectangle, wrapped over as many lines as fit.
 // Falls back to a single clipped line, and to nothing at all when even that
 // doesn't fit - the tooltip is what makes those cells readable.
@@ -168,7 +153,7 @@ static void drawText(ImDrawList* drawList, std::string_view text,
 			y += lineHeight;
 		}
 	} else {
-		draw(rightClip(text, width), min.y + 0.5f * ((max.y - min.y) - lineHeight));
+		draw(ImGui::rightClip(text, width), min.y + 0.5f * ((max.y - min.y) - lineHeight));
 	}
 }
 
@@ -177,6 +162,7 @@ namespace {
 // overlap after being enlarged to a hoverable size.
 struct HoveredBlock {
 	Block block{0, 0, nullptr};
+	std::optional<unsigned> cartridgeSlot; // when it's in an external slot
 	float height = 0.0f;
 	bool valid = false;
 };
@@ -219,12 +205,12 @@ static constexpr float MIN_BLOCK_HEIGHT = 4.0f;
 static void drawSlot(DrawContext& ctx, int ps, int ss, bool expanded,
                      float x0, float x1, float headerTop, float top, float bottom)
 {
-	auto cartridge = getCartridgeLetter(ctx.slotManager, ps, ss);
+	auto cartridge = findCartridgeSlot(ctx.slotManager, ps, ss);
 	auto labelHeight = 0.5f * (top - headerTop);
 	drawText(ctx.drawList, expanded ? strCat(ps, '-', ss) : strCat(ps),
 	         gl::vec2(x0, headerTop), gl::vec2(x1, headerTop + labelHeight), ctx.textColor);
 	if (cartridge) {
-		drawText(ctx.drawList, tmpStrCat("Slot ", *cartridge),
+		drawText(ctx.drawList, tmpStrCat("Slot ", char('A' + *cartridge)),
 		         gl::vec2(x0, headerTop + labelHeight), gl::vec2(x1, top), ctx.dimTextColor);
 	}
 
@@ -273,20 +259,35 @@ static void drawSlot(DrawContext& ctx, int ps, int ss, bool expanded,
 		if ((x0 <= ctx.mouse.x) && (ctx.mouse.x < x1) &&
 		    (blockTop <= ctx.mouse.y) && (ctx.mouse.y < blockBottom) &&
 		    (!ctx.hovered.valid || (blockHeight < ctx.hovered.height))) {
-			ctx.hovered = HoveredBlock{block, blockHeight, true};
+			ctx.hovered = HoveredBlock{block, cartridge, blockHeight, true};
 		}
 	}
 }
 
 static void drawDeviceToolTip(DrawContext& ctx)
 {
-	const auto& [block, height_, valid_] = ctx.hovered;
+	const auto& block = ctx.hovered.block;
 	im::Tooltip([&]{
 		im::TextWrapPos(ImGui::GetFontSize() * 35.0f, [&]{
+			bool header = false;
+			// For an external slot, what is plugged into it. That's not the
+			// same thing as the device below, which is a part of it, though
+			// for a simple extension the two names are the same.
+			if (auto slot = ctx.hovered.cartridgeSlot) {
+				auto content = ctx.manager.media->displayNameForSlotContent(
+					ctx.slotManager, *slot);
+				if (block.device && (content == block.device->getName())) {
+					ImGui::StrCat("Slot ", char('A' + *slot));
+				} else {
+					ImGui::StrCat("Slot ", char('A' + *slot), ": ", content);
+				}
+				header = true;
+			}
 			if (block.device) {
 				ImGui::TextUnformatted(block.device->getName());
-				ImGui::Separator();
+				header = true;
 			}
+			if (header) ImGui::Separator();
 			auto size = block.end - block.begin;
 			ImGui::StrCat("address: 0x", hex_string<4, HexCase::upper>(block.begin),
 			              " - 0x", hex_string<4, HexCase::upper>(block.end - 1));
--- a/src/imgui/ImGuiSlotMap.cc
+++ b/src/imgui/ImGuiSlotMap.cc
@@ -53,8 +53,10 @@ struct Block {
 };
 
 // Draw 'text' centered in the rectangle, wrapped over as many lines as fit.
-// Falls back to a single clipped line, and to nothing at all when even that
-// doesn't fit - the tooltip is what makes those cells readable.
+// The last line ends in an ellipsis if the text needs more lines than that.
+// Falls back to a single clipped line when a word doesn't fit on a line of its
+// own, and to nothing at all when even that doesn't fit - the tooltip is what
+// makes those cells readable.
 static void drawText(ImDrawList* drawList, std::string_view text,
                      gl::vec2 min, gl::vec2 max, ImU32 color)
 {
@@ -76,26 +78,35 @@ static void drawText(ImDrawList* drawList, std::string_view text,
 	auto* font = ImGui::GetFont();
 	auto fontSize = ImGui::GetFontSize();
 	static_vector<std::string_view, 16> lines;
+	auto maxLines = std::min(size_t(height / lineHeight), lines.max_size());
+	std::string_view rest; // what didn't fit, goes on the last line
 	bool wrapped = true;
 	const char* pos = text.data();
 	const char* end = pos + text.size();
 	while (pos < end) {
-		if (lines.size() == lines.max_size()) { wrapped = false; break; }
+		if ((lines.size() + 1) == maxLines) {
+			// Nowhere left to wrap to, so the last line takes the remainder.
+			rest = std::string_view(pos, size_t(end - pos));
+			break;
+		}
 		const char* stop = font->CalcWordWrapPosition(fontSize, pos, end, width);
 		if ((stop == pos) || ((stop != end) && (*stop != ' '))) { wrapped = false; break; }
 		lines.push_back(std::string_view(pos, size_t(stop - pos)));
 		pos = stop;
 		while ((pos < end) && (*pos == ' ')) ++pos; // eat the wrap point
 	}
-	if (wrapped && !lines.empty() && (float(lines.size()) * lineHeight <= height)) {
-		auto y = min.y + 0.5f * ((max.y - min.y) - float(lines.size()) * lineHeight);
-		for (auto line : lines) {
-			draw(line, y);
-			y += lineHeight;
-		}
-	} else {
+	if (!wrapped) {
 		draw(ImGui::rightClip(text, width), min.y + 0.5f * ((max.y - min.y) - lineHeight));
+		return;
 	}
+	auto numLines = lines.size() + (rest.empty() ? 0 : 1);
+	auto y = min.y + 0.5f * ((max.y - min.y) - float(numLines) * lineHeight);
+	for (auto line : lines) {
+		draw(line, y);
+		y += lineHeight;
+	}
+	// rightClip() adds the ellipsis, and returns 'rest' unchanged when it fits.
+	if (!rest.empty()) draw(ImGui::rightClip(rest, width), y);
 }
 
 namespace {
@@ -187,6 +198,44 @@ static void getBlocks(DrawContext& ctx, int ps, int ss)
 	if (pos < ADDRESS_SPACE) result.emplace_back(pos, ADDRESS_SPACE, nullptr);
 }
 
+namespace {
+// Everything that only depends on the font and the style, shared by the layout
+// and by the default size of the window.
+struct Metrics {
+	float rowHeight;
+	float headerHeight;
+	float labelWidth;
+	float minSlotWidth;
+	float minMapHeight;
+	gl::vec2 gap;
+	float overhang;
+};
+}
+
+static Metrics getMetrics()
+{
+	const auto& style = ImGui::GetStyle();
+	auto lineHeight = ImGui::GetTextLineHeight();
+	auto rowHeight = lineHeight + 2.0f * style.CellPadding.y;
+	return {
+		.rowHeight = rowHeight,
+		.headerHeight = 2.0f * rowHeight, // external slot name, slot number
+		// Wide enough for every label on the address axis. Letters and digits
+		// don't have the same width, so check both kinds.
+		.labelWidth = 2.0f * style.CellPadding.x +
+		              std::max({ImGui::CalcTextSize("0000"sv).x,
+		                        ImGui::CalcTextSize("C000"sv).x,
+		                        ImGui::CalcTextSize("FFFF"sv).x}),
+		// A primary slot always gets the same width, expanded or not, so
+		// machines with a different slot layout still look alike.
+		.minSlotWidth = 4.0f * 3.0f * lineHeight,
+		.minMapHeight = 10.0f * rowHeight,
+		.gap = 2.0f * gl::vec2(style.ItemSpacing),
+		// The labels of the lowest and highest address stick out half a line.
+		.overhang = 0.5f * rowHeight,
+	};
+}
+
 // Below this a block can't be seen, let alone hovered. A single byte is a
 // hundredth of a pixel on a 64kB axis, so those are drawn out of scale.
 static constexpr float MIN_BLOCK_HEIGHT = 4.0f;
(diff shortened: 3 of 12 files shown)


src/imgui/ImGuiSlotMap.cc
```diff
@@ -0,0 +1,448 @@
...
+	for (auto slot : xrange(CartridgeSlotManager::MAX_SLOTS)) {
+		if (!slotManager.slotExists(slot)) continue;
+		auto [slotPs, slotSs] = slotManager.getPsSs(slot);
+		if (slotSs == -1) slotSs = 0; // not expanded
+		if ((slotPs == ps) && (slotSs == ss)) {
+			return char('A' + slot);
+		}
+	}
+	return {};
+}
```
MBilderbeek, line 66: This functionality should already be in other GUI parts like ImGuiMedia.... 

Can't we reuse that?

src/imgui/ImGuiSlotMap.cc
```diff
@@ -0,0 +1,448 @@
...
+		auto end = std::min(range.base + range.size, pageBegin + PAGE_SIZE);
+		if (begin < end) {
+			result.emplace_back(begin - pageBegin, end - pageBegin, range.device);
+		}
+	}
+	std::ranges::sort(result, {}, &Segment::begin);
+}
+
+// Mirror of ImGuiUtils' leftClip(): keep the start of the string and replace
+// the part that doesn't fit with an ellipsis.
```
MBilderbeek, line 103: How about moving it there as well?

MBilderbeek (commented):
I didn't do a full review, just a few remarks when glancing over the code.

MBilderbeek:
Great that you are working on this! Thanks!

Some remarks about the GUI design (as requested on Discord):
1. Maybe make the unexpanded slots a bit narrower and the expanded slots a bit wider? I don't like that unexpanded slots take the same space as expanded slots... Why not make all slots take the same space? 
2. For the Vertical layout, maybe it's better to not put everything next to each other but put 2 primary slots in a second row? (Or are we OK with the very wide window,?)
3. Instead of "cart A" just use "slot A" or simply "A" - also saves space. This also already makes it clear that it's an external cartridge slot and thus that the content is an external cartridge/device. 
4. Also the address range could be but below the page number and the openMSX slot name below the slot number (and I would render it like the rest of the GUI, "Slot A") 
5. The memory address axis can also be made continuous. I.e. the GUI code just uses the start and end address of the devices and marks that at that address according to the scale of the view. Then there is no need to hardcode 16kB page boundaries anywhere and the exact addresses can be viewed with a tooltip on the edge. Any kind of weird device memory range is then automatically working.
6. Optional idea: clicking on an external cartridge slot opens the dialog to insert a cartridge or extension in it 
7. What if rescaling the window vertically would also rescale the slot viewer? Currently only horizontal scaling is working, 
8. Vertical layout: I'm not sure why in default vertical layout "MSX BIOS with BASIC ROM" isn't word-wrapped. Maybe the vertical scaling would help? (I'm guessing the text doesn't fit vertically.)
9. Optional idea: if the user hovers on the main RAM and it is a mapper, show that and its size.
10. Optional idea (generic version of previous comment): when hovering a device show some more info about it.
11. Check the Victor HC-95A, it has a tiny device "System Control Register" - I guess this needs point 5 to show properly.

m9710797:
Hi. Great idea, I like it!

I haven't tried this PR yet. I also haven't read any of the code yet. I first want to discuss the functionality from a high level point of view.

About the layout:
* I do like the "vertical layout" better than "horizontal". And you said yourself that most documentation also tends to use that layout. So I'd propose to drop "horizontal".
* But one disadvantage is that with "vertical layout" the window becomes very wide. So I'd propose to dynamically "reflow" the layout. Depending on the size of window (and the user can resize of course). If the window is very wide, then put the 4 slots next to each other (4x1, as in the current screenshot). If the window is narrow then put the slots below each other (1x4 layout), or in between use a 2x2 layout. I seem to remember that such a 2x2 layout was also quite common in old paper(!) MSX magazines.

Maybe we can show extra info as tooltips. The command "machine_info device <device-name>" returns a dict (key-value pairs). Maybe we can show these as a tooltip on hovering a rectangle?

100% optional. I have a WIP branch "theNestruo-f1jh1-2026" where I implemented a helper function 
[`verticalText()`](https://github.com/openMSX/openMSX/commit/ba25847993cacd10c810d6bdcfbb9a7c868b16c7#diff-9b107e5d7e21ac3af491eb8b608b5d2dc067ff9defa7e0fd21456ef8957bd7cdR1379). Feel free to copy that helper (e.g. place it somewhere in `ImGuiUtils`?).

----

The obvious next feature request then becomes: do the same for `iomap`.

src/imgui/ImGuiSlotMap.cc
sndpl: Moved the function

src/imgui/ImGuiSlotMap.cc
sndpl: I refactored this code to a findSlot and two other loops are now also using the same function.

src/cpu/MSXMultiMemDevice.hh
```diff
@@ -18,6 +29,12 @@ public:
 	[[nodiscard]] bool empty() const { return ranges.size() == 1; }
 	[[nodiscard]] std::vector<MSXDevice*> getDevices() const;
 
+	/** The address ranges actually in use, in no particular order.
+	  * The sentinel at the back is not included. */
```
m9710797: This comment contradicts the comment on line 51 which says `ranges` *is* ordered.
Though when I check the code I think you're right and the comment on line 51 is wrong.

src/CartridgeSlotManager.cc
```diff
@@ -193,13 +193,9 @@ void CartridgeSlotManager::createExternalSlot(int ps, int ss)
...
 {
-	for (auto slot : xrange(MAX_SLOTS)) {
-		if (slots[slot].exists() &&
-		    (slots[slot].ps == ps) && (slots[slot].ss == ss)) {
-			return slot;
-		}
-	}
-	UNREACHABLE; // was not an external slot
+	auto slot = findSlot(ps, ss, false);
+	if (!slot) UNREACHABLE; // was not an external slot
```
m9710797: That's not wrong. But a more usual way to write this (line 197) is:
`assert(slot);`
I'm not sure why the original code was written with `UNREACHABLE`. Possibly to avoid a warning about falling of the end of a non-void function.

src/imgui/ImGuiTools.cc
```diff
@@ -146,6 +147,9 @@ void ImGuiTools::showMenu(MSXMotherBoard* motherBoard)
 		ImGui::MenuItem("Audio channel viewer", nullptr, &manager.waveViewer->show);
 		ImGui::MenuItem("Plotter viewer", nullptr, &manager.plotterViewer->show);
 		simpleToolTip("The plotter should be plugged into the printer port");
+		ImGui::MenuItem("Slot map", nullptr, &manager.slotMap->show);
+		simpleToolTip("Show which devices are visible in which slot/page. "
+		              "Same information as the 'slotmap' console command.");
```
m9710797, line 152: Does this belong in the main menu bar below `Tools` (current location), or below `Debugger`?
I'm not sure, I think you can make valid arguments for both locations.

src/imgui/ImGuiSlotMap.cc
```diff
@@ -0,0 +1,437 @@
...
+	loadOnePersistent(name, value, *this, persistentElements);
+}
+
+// A continuous address range in one slot, covered by a single device.
+// 'device' is nullptr for the parts where nothing is mapped.
+struct Block {
+	unsigned begin;
+	unsigned end;
+	const MSXDevice* device;
+};
```
m9710797, line 53: My first reaction when reading this was: "I've seen this before .. in `MSXMultiMemDevice.hh`". Only later I saw the difference: here we use `begin` + `end`, the other used `base` + `size`.

Both are equivalent of course. If in this context `begin` + `end` is easier then that's fine. Or an alternative (middle ground) could be: store `base` + `size`, and have a method `end()`, so you can always use the most convenient way. Anyway, doesn't really matter.

Maybe one suggestion: can you add a comment documenting that `end` is exclusive.

src/imgui/ImGuiSlotMap.cc
```diff
@@ -0,0 +1,437 @@
...
+// about page boundaries.
+static void getBlocks(std::vector<Block>& result, std::vector<Block>& scratch,
+                      MSXCPUInterface& cpuInterface, const MSXDevice* dummyDevice,
+                      int ps, int ss)
+{
+	scratch.clear();
+	for (auto page : xrange(4)) {
+		auto pageBegin = PAGE_SIZE * unsigned(page);
+		const auto* device = cpuInterface.getMSXDevice(ps, ss, page);
+		if (!device || (device == dummyDevice)) continue;
```
m9710797: I believe `getDevice()` will never return a `nullptr` (but please verify).

One of the goals of `dummyDevice` was that we can eliminate nullptr checks from the hot code paths (during emulation). This is not a hot code path. But nevertheless we can exploit this here as well.

Of course `getDevice()` should then document this property (missing documentation, that's my bad).

src/imgui/ImGuiSlotMap.cc
```diff
@@ -0,0 +1,437 @@
...
+			for (const auto& range : multi->getRanges()) {
+				auto begin = std::max(range.base, pageBegin);
+				auto end = std::min(range.base + range.size, pageBegin + PAGE_SIZE);
+				if (begin < end) scratch.emplace_back(begin, end, range.device);
+			}
+		} else {
+			scratch.emplace_back(pageBegin, pageBegin + PAGE_SIZE, device);
+		}
+	}
+	std::ranges::sort(scratch, {}, &Block::begin);
```
m9710797: In principle we only need to sort the sub-range(s) inserted in the `MSXMultiMemDevice` branch (Remember the end-iterator before the loop on line 69, then on line 73 sort the sub-range between that iterator and the current end-iterator).
And then in practice we most often don't need to sort at all. But maybe that optimization is not worth it?

src/imgui/ImGuiSlotMap.cc
```diff
@@ -0,0 +1,437 @@
...
+	// and fill what's left with 'empty' blocks.
+	result.clear();
+	unsigned pos = 0;
+	for (const auto& block : scratch) {
+		if (!result.empty() && (result.back().device == block.device) &&
+		    (result.back().end == block.begin)) {
+			result.back().end = block.end;
+		} else {
+			if (pos < block.begin) result.emplace_back(pos, block.begin, nullptr);
+			result.emplace_back(block.begin, block.end, block.device);
```
m9710797: Just checking: line 90 is equivalent with `result.push_back(block)`, right?

src/imgui/ImGuiSlotMap.cc
```diff
@@ -0,0 +1,437 @@
...
+};
+
+// Split the address space of one slot into blocks. MSXCPUInterface stores a
+// device per 16kB page, and wraps devices that cover less than a page in an
+// MSXMultiMemDevice, which knows the exact address ranges. Recombining both
+// gives the real extent of every device, so that nothing here has to care
+// about page boundaries.
+static void getBlocks(std::vector<Block>& result, std::vector<Block>& scratch,
+                      MSXCPUInterface& cpuInterface, const MSXDevice* dummyDevice,
+                      int ps, int ss)
```
m9710797: About the `getBlock()` function signature:
* Semantically `result` is the output, so you could make it the return value instead of an output parameter.
* And `scratch` could technically be a local variable in this function.

So you could eliminate these two parameters. However if you do that, then each function invocation will need to allocate 2 new vectors. And this function gets called (up to) 16 times per frame. So I tend to agree with this optimization (so OK to keep these as parameters). This reduces the number of allocation to only 1 per frame (I mean 1 per vector per frame).

Optionally you could go a step further, and replace `vector<Block>` with `small_buffer<Block, 8>`. Then you don't have any allocations per frame (unless in the very rare case of more than 8 Blocks per slot, but then still only 1 per frame).

I'd also suggest to replace the 4 parameters `result`, `scratch`, `cpuInterface` and `dummyDevice` with a single `DrawContext` parameter (like is done for most other functions in this file).

src/imgui/ImGuiSlotMap.cc
```diff
@@ -0,0 +1,437 @@
...
+
+// Draw 'text' centered in the rectangle, wrapped over as many lines as fit.
+// Falls back to a single clipped line, and to nothing at all when even that
+// doesn't fit - the tooltip is what makes those cells readable.
+static void drawText(ImDrawList* drawList, std::string_view text,
+                     gl::vec2 min, gl::vec2 max, ImU32 color)
+{
+	const auto& style = ImGui::GetStyle();
+	auto width = (max.x - min.x) - 2.0f * style.CellPadding.x;
+	auto height = (max.y - min.y) - 2.0f * style.CellPadding.y;
```
m9710797: Detail, just a suggestion, use vector-math:
`auto size = (max - min) - 2.0f * style.CellPadding;`
(possibly you need to convert `CellPadding` to a `gl::vec2` for this to compile .. try it).

OTOH, then you have a single `size` variable, and the rest of the code benefits more from separate `width` and `height`. You could use `size.x` and `size.y`. But another alternative is:
`auto [width, height] = (max - min) - 2.0f * style.CellPadding;`
(I think this works, but I'm not 100% sure .. also just try it).

src/imgui/ImGuiSlotMap.cc
```diff
@@ -0,0 +1,437 @@
...
+		                  color, line.data(), line.data() + line.size());
+	};
+
+	// Wrap at word boundaries. ImGui breaks mid-word when a word is wider than
+	// the cell, which reads as garbage in a diagram, so treat that as 'doesn't
+	// fit' and fall back to clipping.
+	auto* font = ImGui::GetFont();
+	auto fontSize = ImGui::GetFontSize();
+	std::array<std::string_view, 16> lines;
+	size_t numLines = 0;
```
m9710797: The combination of `lines` and `numLines` is like our `static_vector<std::string_view, 16>`.

If you don't want a hard upper limit, you can also consider `small_buffer<std::string_view, 16>` (this will store up-to 16 elements in-place, and for more use heap-allocation). But I agree 16 should be more than sufficient.

src/imgui/ImGuiSlotMap.cc
```diff
@@ -0,0 +1,437 @@
...
+	const char* end = pos + text.size();
+	while (pos < end) {
+		if (numLines == lines.size()) { wrapped = false; break; }
+		const char* stop = font->CalcWordWrapPosition(fontSize, pos, end, width);
+		if ((stop == pos) || ((stop != end) && (*stop != ' '))) { wrapped = false; break; }
+		lines[numLines++] = std::string_view(pos, size_t(stop - pos));
+		pos = stop;
+		while ((pos < end) && (*pos == ' ')) ++pos; // eat the wrap point
+	}
+	if (wrapped && (numLines != 0) && (float(numLines) * lineHeight <= height)) {
```
m9710797: I assume this was a deliberate choice, but I want to check anyway:
* Currently when the text would require more than 16 lines, we fall back to the `wrapped=false` case (show as a single clipped line). In practice this case will likely never happen.
* Or when the wrapped text requires more lines than fit in the box, we also fall back to a single clipped line. I think this might happen in practice.

An alternative could be to show as many lines as fit, but replace the last line with `...`.

Maybe not worth the extra complexity?

src/imgui/ImGuiSlotMap.cc
```diff
@@ -0,0 +1,437 @@
...
+			ImGui::StrCat("address: 0x", hex_string<4, HexCase::upper>(block.begin),
+			              " - 0x", hex_string<4, HexCase::upper>(block.end - 1));
+			ImGui::StrCat("size: ", (size < 1024) ? strCat(size, " bytes")
+			                                      : strCat(size / 1024, "kB"));
+			if (!block.device) return;
+
+			const auto& name = block.device->getName();
+			// Whatever the device wants to tell about itself, e.g. the mapper
+			// type of a ROM or the file it was loaded from.
+			if (auto info = ctx.manager.execute(makeTclList("machine_info", "device", name))) {
```
m9710797: This code is only triggered in a tool tip. So it's good enough, no need to change. But let me give my remark anyway.

In general in GUI code, because it's re-executed every frame, we care about performance. Here we:
* Construct a Tcl list for the command `machine_info device ...`
* Let Tcl interpreter decode that command.
* That then delegates to our c++ code for the `machine_info` command.
* That delegates to the `DeviceInfo` class which handles the `device` query.
* That looks up the device by name, and then calls `MSXDevice::getDeviceInfo()`

But in this code we already have a pointer to the `MSXDevice`. So we could directly call `block.device->getDeviceInfo()`.

---

Not fully the same but below we execute the `debug size ...` command. We could also take a shortcut and skip the Tcl interpreter. We can obtain the `Debuggable` via `Debugger::findDebuggable()`, and if that returns a non-nullptr, we can call `getSize()`.

There are still several other places in the GUI where we execute Tcl commands rather than more directly querying c++ code. So this just repeats that pattern. Ideally we'd optimize them all. But low priority.

src/imgui/ImGuiSlotMap.cc
```diff
@@ -0,0 +1,437 @@
...
+					auto value = info->getListIndexUnchecked(i + 1).getString();
+					if (value.empty()) continue;
+					ImGui::StrCat(key, ": ", value);
+				}
+			}
+			// Devices with a debuggable of the same name (memory mappers, RAM)
+			// can also tell how much memory they actually have.
+			if (auto total = ctx.manager.execute(makeTclList("debug", "size", name))) {
+				if (auto bytes = total->getOptionalInt(); bytes && (*bytes > 0)) {
+					ImGui::StrCat("total memory: ", *bytes / 1024, "kB");
```
m9710797: Detail: above we also print size, and there we handle sizes less than 1024 bytes special (here we'd print 0kB). In practice it's probably fine, but since you already have the code, you could maybe factor it out in a small lambda (maybe taking a prefix string_view parameter?) and reuse that code here.

src/imgui/ImGuiSlotMap.cc
```diff
@@ -0,0 +1,437 @@
...
+}
+
+static void drawSlotMap(DrawContext& ctx)
+{
+	const auto& style = ImGui::GetStyle();
+	auto lineHeight = ImGui::GetTextLineHeight();
+	auto rowHeight = lineHeight + 2.0f * style.CellPadding.y;
+	auto headerHeight = 2.0f * rowHeight; // slot number, external slot name
+	auto gapX = 2.0f * style.ItemSpacing.x;
+	auto gapY = style.ItemSpacing.y;
```
m9710797: In the 2x2 or 1x4 layout, I personally think the vertical spacing between two slots is not enough: the labels like `3.0` are closer to the slot above than the slot below (that confused me for a moment the first time I saw it). Here I see that for the horizontal spacing you take `2 * ItemSpacing`, but for vertical only `1 *`. Maybe it'll help to also make that `2 *`?

Or maybe we can swap the order of the `2-0` and `Slot B` labels? There are more labels of the former category, and then more are closer to their target? It'll especially help for slot 3 which is often expanded and not often has external slots.

Detail: if you'd change the factor to `2 *` for both dimensions, then you could simplify the code to:
`auto gap = 2.0f * gl::vec2(style.ItemSpacing);`
And then use `gap.x` and `gap.y`.

src/imgui/ImGuiSlotMap.cc
```diff
@@ -0,0 +1,437 @@
...
+			x += slotWidth + gapX;
+		}
+	}
+
+	// Outlines in one go, after all the fills, so that the fill of one block
+	// can't paint over the outline of the block next to it. This is also what
+	// keeps a device that is too small to see as a rectangle visible as a line.
+	for (const auto& [min, max] : ctx.outlines) {
+		ctx.drawList->AddRect(min, max, ctx.outlineColor);
+	}
```
m9710797, line 459: Small tweak: every device's outline is drawn as a rectangle. That means all interior lines are overdrawn twice. We could draw the exterior lines and vertical sub-slot lines separately, and then per device only draw a single horizontal line (if it's not for address=0x10000). That'll eliminate all overdraw.

Actually if I look more closely to the result, it seems that the interior lines have double the line-width of the exterior lines (maybe a consequence of how ImGui draws rectangles?). Then the above tweak is not only a performance tweak, but it'll also influence (fix?) the look.

Maybe the intended look was to have lines with width=2, but then IMHO also the exterior lines should have width=2. Then we can still use the above tweak, but explicitly draw lines with width=2.  (Personally I'd first try the default width=1)

The `tinyBlocks` below are an exception, those should probably still be drawn as a full rectangle?

src/imgui/ImGuiSlotMap.cc
```diff
@@ -0,0 +1,437 @@
...
+		.externalColor = getColor(imColor::YELLOW_BG),
+		.outlineColor = ImGui::GetColorU32(ImGuiCol_Text),
+		.textColor = ImGui::GetColorU32(ImGuiCol_Text),
+		.dimTextColor = ImGui::GetColorU32(ImGuiCol_TextDisabled),
+		.mouse = gl::vec2(ImGui::GetIO().MousePos),
+	};
+	ctx.outlines.reserve(32);
+
+	auto fontSize = ImGui::GetFontSize();
+	ImGui::SetNextWindowSize(ImVec2(64.0f * fontSize, 26.0f * fontSize), ImGuiCond_FirstUseEver);
```
m9710797: This default window size results in a 4x1 layout. Personally I like the 2x2 layout better. But that's just my opinion.

m9710797 (commented):
Thanks for working on this! I like it a lot.

Several review comments. Mostly about details, small tweaks or suggestions. Nothing major.

src/cpu/MSXMultiMemDevice.hh
sndpl: You're right that they contradicted, and mine was the accurate one: add() does ranges.insert(begin(ranges), …), so the vector is in reverse insertion order. Reworded the old comment to say what actually holds.

src/CartridgeSlotManager.cc
sndpl: Changed to assert(slot). Returning *slot right after is enough to keep the compiler happy about the return path.

src/imgui/ImGuiSlotMap.cc
sndpl: Dead indeed: the MSXCPUInterface constructor fills all of slotLayout with the DummyDevice. Check dropped, and I added the guarantee to the declaration of getMSXDevice() so the next reader doesn't have to work it out.

src/imgui/ImGuiSlotMap.cc
sndpl: Added // exclusive to end. Kept begin/end rather than base/size as you concluded it doesn't really matter.

src/imgui/ImGuiSlotMap.cc
sndpl: Narrowed. Most slots now sort nothing at all.

src/imgui/ImGuiSlotMap.cc
sndpl: Done, it takes the DrawContext& like the rest of the file.

src/imgui/ImGuiSlotMap.cc
sndpl: It does compile — gl::vecN<2, T> has public x/y members. Adopted here, and I used the same form for the layout metrics in drawSlotMap().

src/imgui/ImGuiSlotMap.cc
sndpl: Done, static_vector<std::string_view, 16> — the numLines counter is gone.

src/imgui/ImGuiSlotMap.cc
sndpl: Agreed, it happens often. The wrap loop now stops when one line is left and gives that line the remainder, clipped with an ellipsis — e.g. "MSX Opening ROM" in a two-line cell shows MSX / Openi.... The single-clipped-line fallback is still there for the case a word doesn't fit on a line of its own, where wrapping can't help anyway.

src/imgui/ImGuiSlotMap.cc
sndpl: Good point, it was doing a lot of work for something we already have. Both lookups are direct now: block.device->getDeviceInfo(info) instead of machine_info device, and Debugger::findDebuggable(name) + Debuggable::getSize() instead of debug size. Output is unchanged — checked against the Tcl route (FMPAC gives type FMPAC either way, Main RAM total memory: 512kB for its 524288 bytes). Devices without a debuggable (MSX Music) simply print no line.

src/imgui/ImGuiSlotMap.cc
sndpl: Factored both lines through one small lambda, so the memory of a debuggable is formatted like a block size. Worth noting the 0kB was latent rather than visible: on the machines I checked (FS-A1GT, Victor HC-95A, Boosted_MSX2_EN) no debuggable under 1kB shares its name with a memory-mapped device. While in there I also made it say 1 byte instead of 1 bytes, which the HC-95A's System control register does hit

src/imgui/ImGuiSlotMap.cc
sndpl: Both done. The gap is now 2.0f * gl::vec2(style.ItemSpacing) so the vertical spacing matches the horizontal, and the two header rows are swapped: the dim Slot A on top and the slot number directly above its own grid.

src/imgui/ImGuiSlotMap.cc
sndpl: Right, and that explains the thick interior lines. Each edge is now drawn exactly once: one AddRect around a primary slot, one AddLine per sub-slot separator, one AddLine per interior block boundary (a block only draws the line at its top, so 0x0000 and 0xFFFF are left to the exterior). Blocks drawn out of scale keep a full rectangle, since those overlap their neighbours on purpose. Thickness left at the default 1.

src/imgui/ImGuiSlotMap.cc
sndpl: Also like the 2x2, so changed it.

src/imgui/ImGuiSlotMap.cc
sndpl: I didn't get the small_buffer working, when I asked Claud about it, it says:

I looked at small_buffer, but I don't think it fits here. Its size is fixed by the constructor — there's no push_back or clear() — while these two vectors are grown a block at a time and then reused for the next slot. Its header says as much: "Typically objects of this type are used as local variables. Not as member variables of a larger class."

The goal behind the suggestion is already met though: clear() keeps the capacity, so these allocate during the first frame and never again. That's exactly why they're members instead of locals — I've added a comment on getBlocks() saying so, since it wasn't obvious.

If you'd prefer a fixed-capacity container anyway, static_vector is the one with the right API (I'm using it for the wrapped text lines). But I don't think it works for the blocks: the count isn't bounded by a small constant, because a page can hold an arbitrary number of MSXMultiMemDevice sub-ranges — the HC-95A's sub-slot 0-1 is already at five blocks. It would need a heap fallback, which is what std::vector gives us for free.

build/packages.py
```diff
@@ -185,10 +185,12 @@ def getMakeName(cls):
 		return 'VORBIS'
 
 class ZLib(DownloadablePackage):
-	downloadURL = 'https://zlib.net/fossils/'
 	niceName = 'zlib'
 	sourceName = 'zlib'
 	version = '1.3.1'
+	# Note: zlib.net serves an HTML page instead of the tarball to some hosts,
+	#       among them the GitHub Actions runners, so take it from GitHub.
+	downloadURL = f'https://github.com/madler/zlib/releases/download/v{version}'
```
m9710797: Maybe this is a good change, but IMHO it doesn't belong in the same PR as the slotmap stuff.
Can you remove it here. Please do submit as a separate PR. Same for the change in `thrirdparty_download.py`.

src/imgui/ImGuiSlotMap.cc
m9710797: Just checking ... you say "adopted", but I still see 2 separate lines for `width` and `height`.

src/imgui/ImGuiSlotMap.cc
```diff
@@ -0,0 +1,514 @@
...
+			rest = std::string_view(pos, size_t(end - pos));
+			break;
+		}
+		const char* stop = font->CalcWordWrapPosition(fontSize, pos, end, width);
+		if ((stop == pos) || ((stop != end) && (*stop != ' '))) { wrapped = false; break; }
+		lines.push_back(std::string_view(pos, size_t(stop - pos)));
+		pos = stop;
+		while ((pos < end) && (*pos == ' ')) ++pos; // eat the wrap point
+	}
+	if (!wrapped) {
```
m9710797: I wonder if the logic related to the `bool wrapped` variable is still correct. It's set to false when at some point `CalcWordWrapPosition()` return some error(?) (that function is not very well documented, so I'm not sure), or at least when it could not wrap any word??? But then what with all the previous iterations where `CalcWordWrapPosition()` could make progress? (I do understand that in case `wrapped=false` we do print the full string, so we don't loose those previous iterations).

Maybe my comment boils down to: the behavior related to `wrapped` is unclear. Either it needs a line of comments, or something needs to be simplified or fixed.

src/imgui/ImGuiSlotMap.cc
```diff
@@ -0,0 +1,514 @@
...
+		.dummyDevice = &cpuInterface.getDummyDevice(),
+		.drawList = nullptr, // only valid inside the window
+		.emptyColor = ImGui::GetColorU32(getColor(imColor::GRAY), 0.4f),
+		.occupiedColor = ImGui::GetColorU32(ImGuiCol_Header),
+		.externalColor = getColor(imColor::YELLOW_BG),
+		.outlineColor = ImGui::GetColorU32(ImGuiCol_Text),
+		.textColor = ImGui::GetColorU32(ImGuiCol_Text),
+		.dimTextColor = ImGui::GetColorU32(ImGuiCol_TextDisabled),
+		.mouse = gl::vec2(ImGui::GetIO().MousePos),
+	};
```
m9710797, line 511: Compilation with gcc-15.2 gives these warnings:
```
src/imgui/ImGuiSlotMap.cc: In member function ‘virtual void openmsx::ImGuiSlotMap::paint(openmsx::MSXMotherBoard*)’:
src/imgui/ImGuiSlotMap.cc:499:9: warning: missing initializer for member ‘openmsx::{anonymous}::DrawContext::hovered’ [-Wmissing-field-initializers]
  499 |         };
      |         ^
src/imgui/ImGuiSlotMap.cc:499:9: warning: missing initializer for member ‘openmsx::{anonymous}::DrawContext::outlines’ [-Wmissing-field-initializers]
src/imgui/ImGuiSlotMap.cc:499:9: warning: missing initializer for member ‘openmsx::{anonymous}::DrawContext::separators’ [-Wmissing-field-initializers]
src/imgui/ImGuiSlotMap.cc:499:9: warning: missing initializer for member ‘openmsx::{anonymous}::DrawContext::tinyBlocks’ [-Wmissing-field-initializers]
src/imgui/ImGuiSlotMap.cc:499:9: warning: missing initializer for member ‘openmsx::{anonymous}::DrawContext::blocks’ [-Wmissing-field-initializers]
src/imgui/ImGuiSlotMap.cc:499:9: warning: missing initializer for member ‘openmsx::{anonymous}::DrawContext::scratch’ [-Wmissing-field-initializers]
```
I think none of these need to be initialized (i mean the default initialization is fine). But to silence the warning we should maybe nevertheless do something like: `.outlines = {}`

src/imgui/ImGuiSlotMap.cc
```diff
@@ -0,0 +1,514 @@
...
+	       gl::vec2(0.0f, ImGui::GetFrameHeight()); // title bar
+}
+
+void ImGuiSlotMap::paint(MSXMotherBoard* motherBoard)
+{
+	if (!show) return;
+	if (!motherBoard) return;
+
+	auto& cpuInterface = motherBoard->getCPUInterface();
+	DrawContext ctx{
```
m9710797, line 488: Somewhere you replied that using small_buffer for `blocks` and `scratch` wasn't possible because it can't dynamically grow. And you're right: we'd need something more like llvm's `SmallVector`, currently we don't have such a utility in openMSX. And only for this it's not worth adding.

But in your reply you say, it doesn't matter because `ctx` is anyway a member variable and reused over frames (so the `std::vector`s are anyway only allocated once). I'd agree if that were true. But here I see that `ctx` is a local variable in `paint()`, not a member variable.

But it's only a minor tweak. I don't mind if you keep the code as is.

src/imgui/ImGuiSlotMap.cc
```diff
@@ -0,0 +1,514 @@
...
+	}
+
+	// Outlines in one go, after all the fills, so that the fill of one block
+	// can't paint over the outline of the block next to it. This is also what
+	// keeps a device that is too small to see as a rectangle visible as a line.
+	// Each edge is drawn exactly once, otherwise the shared ones would come out
+	// twice as thick as the rest.
+	for (const auto& [p1, p2] : ctx.separators) {
+		ctx.drawList->AddLine(p1, p2, ctx.outlineColor);
+	}
```
m9710797, line 455: Thanks, this solves the issue of "overdraw / thick lines".

However what I notice now is that sometimes the horizontal separators between devices, or the vertical separators between sub-slots stick out 1 pixel on the right/bottom compared to the exterior outline.

I did some investigation, I think I found the cause. This is how Dear ImGui implements `AddLine()` and `AddRect`:
```
void ImDrawList::AddLine(const ImVec2& p1, const ImVec2& p2, ImU32 col, float thickness)
{
    ...
    const ImVec2 points[2] = { ImVec2(p1.x + 0.5f, p1.y + 0.5f), ImVec2(p2.x + 0.5f, p2.y + 0.5f) };
    AddPolyline(points, 2, col, thickness);
}
...
// p_min = upper-left, p_max = lower-right
// Note we don't render 1 pixels sized rectangles properly.
void ImDrawList::AddRect(const ImVec2& p_min, const ImVec2& p_max, ImU32 col, float rounding, float thickness, ImDrawFlags flags)
{
    ....
     PathRect(p_min + ImVec2(0.50f, 0.50f), p_max - ImVec2(0.50f, 0.50f), rounding, flags);
    ....
```
Notice that `AddLine` does `+ 0.5f` on both end points. But `AddRect` does `+ 0.5f` on `p_min` but `- 0.5f` on `p_max`. So it seems the `p_max` coordinates are exclusive, while for line both end points are inclusive?? (And it also seems to assume that `p_min` is top-left and `p_max` is bottom-right, at least my expectation was that `p_min` and `p_max` would be interchangeable. 

I see a few possible fixes:
* Pass an adjusted "exclusive" `p_max` to `AddRect()`
* Draw manually as 4 lines.
* Use `AddPolyLine()` (with 4 points and `ImDrawFlags_Closed`).

I have a slight preference for the 3rd option, but maybe you can check if there isn't still any better solution. Because this seems like a common pitfall, with maybe a well known solution?


EDIT:
This inclusive/exclusive stuff also explains why before the lines appeared to have line-width=2. If `AddRect()` behaved as expected (both points inclusive) then the overdrawn lines would perfectly overlap, and then the overdraw would not be visible.

m9710797 (commented):
Thanks. I still have a few remarks. But It's getting close to ready.

m9710797:
Merged. Thanks!

I'll try to review the IO ports PR later today.

Issue #2170, opened 2026-08-16
Add io map GUI view to Tools


sndpl:
Make a graphical view (accessible from the main menu somewhere) of the information presented by the iomap command on the console.

I made a small mockup for this:
<img width="400" height="166" alt="Image" src="https://github.com/user-attachments/assets/26270b4f-1e68-4d1f-ae68-ef7bd10dfd12" />

It contains a table with:
- port range
- I/O 
- name
- values written to it (real time? depending on the range of ports it contains one or more values)

or should it be more static with just the first 3 items of the table, because with the real time values it becomes more a debugger tool :)

MBilderbeek:
Duplicate of #2012 ?

m9710797:
Dear ImGui supports columns that can be hidden (via right-click context menu in the table header). The `Value` column could be like that, possibly hidden by default.

All Dear ImGui views show real-time info. Better said: using real-time updated info is easier than not-real-time:
* real-time info: you query the info each frame, the GUI remains stateless (nothing can get out-of-date or inconsistent).
* non-real-time: requires making a snapshot and keeping the info available and update it when needed. So more complex.

Via `Main menu bar > Debugger > Add hex editor > ioports` we do already have a real-time updated view on IO ports. So strictly speaking the proposed `Value` column is redundant. Though this view still has the advantage of clearly grouping the ports per device.

Note:
* Multiple devices can share the same IO port(s). Normally that's something that should be avoided, so it doesn't happen often. But for example you could insert two FM-PAC extensions.
* The set of ports that belong to a single device don't necessarily need to be a contiguous range. The `Philips_NMS_1205` extension is a good example of that.

sndpl:
implemented with https://github.com/openMSX/openMSX/pull/2171

PR #2171, opened 2026-08-22
Add 'I/O ports' window to the Tools menu

sndpl:
Shows which device is connected to which I/O port, the equivalent of the 'iomap' script but always up to date and without leaving the GUI.

Two views: flat, sorted by port number with consecutive ports of the same device(s) coalesced into one range; and grouped per device, because a device doesn't necessarily occupy a contiguous range (e.g. the Philips NMS-1205). Ports shared between devices are highlighted in yellow, unless the machine sets <devices overlap_warning="false">, matching what MSXCPUInterface already does for its startup warning.

The optional 'Value' column peeks the input ports, so opening the window doesn't change emulation state. It's hidden by default since the hex editor on the 'ioports' debuggable already shows those values.

image
image

Solves issue #2170

### DIFF
--- a/src/imgui/ImGuiIoPorts.cc
+++ b/src/imgui/ImGuiIoPorts.cc
@@ -29,18 +29,6 @@ namespace openmsx {
 
 using namespace std::literals;
 
-static constexpr auto OVERLAP_TOOLTIP =
-	"These ports are shared with another device. Reading such a port returns "
-	"the bitwise AND of all values, writing goes to every device."sv;
-
-static constexpr int TABLE_FLAGS =
-	ImGuiTableFlags_RowBg |
-	ImGuiTableFlags_BordersInnerV |
-	ImGuiTableFlags_Resizable |
-	ImGuiTableFlags_Reorderable |
-	ImGuiTableFlags_Hideable |
-	ImGuiTableFlags_ContextMenuInBody;
-
 /** A maximal range of ports with the same device(s) in the same direction(s).
   */
 struct Row {
@@ -200,7 +188,15 @@ static void drawTable(ImFont* fontMono, MSXCPUInterface& cpuInterface,
 	};
 	auto arrowWidth = ImGui::GetFontSize() + style.ItemInnerSpacing.x;
 
-	im::Table("ports", 4, TABLE_FLAGS | ImGuiTableFlags_Sortable, [&]{
+	static constexpr int TABLE_FLAGS =
+		ImGuiTableFlags_RowBg |
+		ImGuiTableFlags_BordersInnerV |
+		ImGuiTableFlags_Resizable |
+		ImGuiTableFlags_Reorderable |
+		ImGuiTableFlags_Hideable |
+		ImGuiTableFlags_ContextMenuInBody |
+		ImGuiTableFlags_Sortable;
+	im::Table("ports", 4, TABLE_FLAGS, [&]{
 		// 'Value' is hidden by default: the hex editor on the 'ioports'
 		// debuggable already shows these values. Un-hide it via the
 		// right-click context menu on the table header.
@@ -240,7 +236,12 @@ static void drawTable(ImFont* fontMono, MSXCPUInterface& cpuInterface,
 				im::StyleColor(conflict, ImGuiCol_Text, getColor(imColor::YELLOW), [&]{
 					ImGui::TextUnformatted(row.device);
 				});
-				if (conflict) simpleToolTip(OVERLAP_TOOLTIP);
+				if (conflict) {
+					simpleToolTip(
+						"These ports are shared with another device. Reading "
+						"such a port returns the bitwise AND of all values, "
+						"writing goes to every device."sv);
+				}
 			}
 			if (ImGui::TableNextColumn()) { // value(s)
 				drawValue(fontMono, cpuInterface, row, time);
--- a/src/imgui/ImGuiIoPorts.cc
+++ b/src/imgui/ImGuiIoPorts.cc
@@ -157,60 +157,6 @@ std::vector<ImGuiIoPorts::Row> ImGuiIoPorts::getRows(MSXCPUInterface& cpuInterfa
 	return result;
 }
 
-std::vector<ImGuiIoPorts::Group> ImGuiIoPorts::getGroups(MSXCPUInterface& cpuInterface)
-{
-	auto names = collectNames(cpuInterface);
-	const auto& in = names.in;
-	const auto& out = names.out;
-
-	// Per device, which ports it occupies (bit 0 = input, bit 1 = output).
-	struct Acc {
-		std::string_view device;
-		std::array<uint8_t, 256> mask = {};
-	};
-	std::vector<Acc> accs;
-	auto mark = [&](std::string_view device, unsigned port, uint8_t bit) {
-		auto it = std::ranges::find(accs, device, &Acc::device);
-		if (it == accs.end()) {
-			accs.push_back(Acc{.device = device});
-			it = accs.end() - 1;
-		}
-		it->mask[port] |= bit;
-	};
-	// Ports are visited in ascending order, so the devices end up ordered by
-	// their lowest port. That keeps the grouped view roughly address ordered.
-	for (auto port : xrange(256)) {
-		for (const auto& device : in [port]) mark(device, port, 1);
-		for (const auto& device : out[port]) mark(device, port, 2);
-	}
-
-	std::vector<Group> result;
-	for (const auto& acc : accs) {
-		Group group{.device = std::string(acc.device)};
-		unsigned port = 0;
-		while (port < 256) {
-			if (acc.mask[port] == 0) { ++port; continue; }
-			unsigned end = port + 1;
-			while ((end < 256) && (acc.mask[end] == acc.mask[port])) ++end;
-			bool isIn  = (acc.mask[port] & 1) != 0;
-			bool isOut = (acc.mask[port] & 2) != 0;
-			bool shared = std::ranges::any_of(xrange(port, end), [&](unsigned p) {
-				return (isIn  && (in [p].size() > 1)) ||
-				       (isOut && (out[p].size() > 1));
-			});
-			group.ranges.push_back(Row{
-				.begin = narrow<uint8_t>(port),
-				.end = narrow<uint8_t>(end - 1),
-				.in = isIn,
-				.out = isOut,
-				.overlap = shared});
-			port = end;
-		}
-		result.push_back(std::move(group));
-	}
-	return result;
-}
-
 void ImGuiIoPorts::drawValue(MSXCPUInterface& cpuInterface, const Row& row, EmuTime time)
 {
 	im::ScopedFont sf(manager.fontMono);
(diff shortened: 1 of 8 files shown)


m9710797:
Thanks for working on this.

I have not yet looked at the code, nor have I actually tried to run it. Here's my first reaction:

You can consider making (some of) the columns sortable (and default sort is ascending on `Ports`).
And then if you sort on `Device` you have something that approached the `Group by device` view. But maybe it won't look as good, so maybe it's still worth keeping this .. or maybe not, we'll have to experiment.

src/imgui/ImGuiIoPorts.cc
```diff
@@ -0,0 +1,375 @@
...
+{
+	Names result;
+	auto add = [&](const std::string& name) {
+		if (!name.empty()) result.emplace_back(name);
+	};
+	auto* device = cpuInterface.getIODevice(port, isIn);
+	if (auto* multi = dynamic_cast<MSXMultiIODevice*>(device)) {
+		for (const auto* dev : multi->getDevices()) {
+			add(dev->getName());
+		}
```
m9710797: Minor tweak:

I believe that sub-devices of a `MSXMultiIODevice` will never be the `DummyDevice` and then never have an empty name. I'm 99% sure of that, but not 100% (so please verify). Actually we could then also simplify the implementation of `MSXMultiIODevice::getNameList()` (ok if you change that, but not strictly needed for this PR). In this case that then allows to simplify the `add` lambda again (only called from one place, no need to factor it out anymore).

But the bigger motivation than skipping the check for empty string is that you then get a loop that does nothing more than `push_back` of all elements of some `vector` onto another `vector`. And than can be written as: `append(result, multi->getDevices());`. And that is a bit faster. Not that performance is so important here, but still.

src/imgui/ImGuiIoPorts.cc
```diff
@@ -0,0 +1,375 @@
...
+
+[[nodiscard]] static std::string join(const Names& names)
+{
+	std::string result;
+	for (const auto& name : names) {
+		if (!result.empty()) strAppend(result, ", ");
+		strAppend(result, name);
+	}
+	return result;
+}
```
m9710797, line 75: See `utils/join.hh` we do have a function for this already, also named `join()`.
You call it like this:
```
auto myString = join(myRange, ", ");
```
Where `myRange` can be any range of "printable" types (e.g. could also be `int`, those are then converted to string before joined into the larger string).

src/imgui/ImGuiIoPorts.cc
```diff
@@ -0,0 +1,375 @@
...
+	// 'SpecsDirty' is set.
+	const auto* sortSpecs = ImGui::TableGetSortSpecs();
+	if (!sortSpecs || (sortSpecs->SpecsCount == 0)) return;
+	assert(sortSpecs->Specs);
+
+	// 'rows' is generated in ascending port order, and all sorts below are
+	// stable. So e.g. sorting on device keeps each device's ports ordered.
+	switch (sortSpecs->Specs->ColumnIndex) {
+	case 0: // ports
+		sortUpDown_T(rows, sortSpecs, &Row::begin);
```
m9710797: Maybe this isn't needed (will only complicate things). But since `rows` is rebuilt every frame and already sorted on ports:
* For the default (sorted on ascending ports) we don't need to do anything here (skip `O(n log(n))` operation).
* For sorted on descending port number we could just `std::ranges::reverse(rows)` (which is `O(n)`).

src/imgui/ImGuiIoPorts.cc
```diff
@@ -0,0 +1,375 @@
...
+		break;
+	case 1: // direction
+		sortUpDown_String(rows, sortSpecs, [](const Row& row) { return dirText(row); });
```
m9710797: Here we could skip the conversion to `string` and instead use a projection that extracts the `bool in, out` into a `std::pair<bool, bool> or into an integer 0,1,2.

But frankly: is sorting on the `Dir` column useful? Maybe we could just drop this?

src/imgui/ImGuiIoPorts.cc
```diff
@@ -0,0 +1,375 @@
...
+	auto names = collectNames(cpuInterface);
+	const auto& in = names.in;
+	const auto& out = names.out;
+
+	std::vector<Row> result;
+	unsigned port = 0;
+	while (port < 256) {
+		// coalesce consecutive ports with identical input and output device(s)
+		unsigned end = port + 1;
+		while ((end < 256) && (in[end] == in[port]) && (out[end] == out[port])) {
```
m9710797, line 128: I wonder, here we group by doing string comparisons. What if `collectNames()` would collect `MSXDevice*` instead of already converting the device via `getName()` to `string`. Then this grouping could be done via cheaper pointer comparisons (more specifically vector of pointers).

And then also, later on, we only need to call `getName()` once per group instead of for all 256 ports (in + out).

src/imgui/ImGuiIoPorts.cc
```diff
@@ -0,0 +1,375 @@
...
+		} else {
+			add(in [port], true, false);
+			add(out[port], false, true);
+		}
+		port = end;
+	}
+	return result;
+}
+
+std::vector<ImGuiIoPorts::Group> ImGuiIoPorts::getGroups(MSXCPUInterface& cpuInterface)
```
m9710797: I played around with the `Group by device` check box. To be honest I personally don't think `Group by device` has much(*1) added value over sorting on the `Device` column. The latter is also more compact (the TreeNodes do take up a lot of vertical space).

So personally I'd drop the `Group by device` code. But feel free to disagree.

For this first review pass, I therefor reviewed the `getGroups()` and `drawGroups()` functions in less detail.


  (*1) I'm not saying *no* added value, but *not a lot* of added value.

src/imgui/ImGuiIoPorts.cc
```diff
@@ -0,0 +1,375 @@
...
+{
+	auto names = collectNames(cpuInterface);
+	const auto& in = names.in;
+	const auto& out = names.out;
+
+	// Per device, which ports it occupies (bit 0 = input, bit 1 = output).
+	struct Acc {
+		std::string_view device;
+		std::array<uint8_t, 256> mask = {};
```
m9710797: This is fine, an alternative could have been 2 `std::bitset<256>`, 1 for in, 1 for out.

src/imgui/ImGuiIoPorts.cc
```diff
@@ -0,0 +1,375 @@
...
+		strAppend(values, hex_string<2, HexCase::upper>(
+			cpuInterface.peekIO(narrow<uint16_t>(row.begin + i), time)));
+	}
+	if (num <= MAX_SHOWN_VALUES) {
+		ImGui::TextUnformatted(values);
+	} else {
+		// every value takes 3 characters ("XX "), minus the trailing space
+		ImGui::StrCat(std::string_view(values).substr(0, MAX_SHOWN_VALUES * 3 - 1), " ...");
+		simpleToolTip(values);
+	}
```
m9710797, line 169: Why not always show all values, so not limit to `MAX_SHOWN_VALUES = 8`. Then if the user makes the column wide enough he sees all values.

Ideally would be to use our utility `rightClip()` to draw the text with clipping. Though there might be some technical limitations. I'm not sure Dear ImGui allows to query the width of a column (because it also tries to size the columns based on the width of the elements it contains).

Currently even if there are e.g. only 4 values they might not fit (and then you don't even see those `...`).

In practice only(?) the `gfx9000` device has more than 8 ports. So maybe we shouldn't care too much about this corner case?

So, if you agree, let's simplify and just draw all values (and let Dear ImGui deal with clipping). Maybe not even show the tool tip, the user can always resize the window/columns. And the `Value` column is even hidden by default, making it even less important.

src/imgui/ImGuiIoPorts.cc
```diff
@@ -0,0 +1,375 @@
...
+		// 'Value' is hidden by default: the hex editor on the 'ioports'
+		// debuggable already shows these values. Un-hide it via the
+		// right-click context menu on the table header.
+		// It's also not sortable: the values change while the emulation
+		// runs, so sorting on them would reshuffle the rows every frame.
+		ImGui::TableSetupColumn("Ports", ImGuiTableColumnFlags_WidthFixed |
+		                                 ImGuiTableColumnFlags_DefaultSort,
+		                        sortableWidth("Ports", monoWidth("FF-FF")));
+		ImGui::TableSetupColumn("Dir", ImGuiTableColumnFlags_WidthFixed,
+		                        sortableWidth("Dir", ImGui::CalcTextSize("I/O"sv).x));
```
m9710797: As I suggested above, maybe the `Dir` column should not be sortable?
Should it be resizable?

Ah, and then maybe the `Ports` column should also not be resizable?

And then we could experiment with the following (when `Ports` and `Dir` are not resizable). Instead of only giving `Device` a stretchable column width, do the same for `Value`, then if the user enlarges the window (because some of the values are clipped), then the extra width is evenly(?) divided over these two columns.

src/imgui/ImGuiIoPorts.cc
```diff
@@ -0,0 +1,375 @@
...
+static constexpr auto OVERLAP_TOOLTIP =
+	"These ports are shared with another device. Reading such a port returns "
+	"the bitwise AND of all values, writing goes to every device."sv;
+
+static constexpr int TABLE_FLAGS =
+	ImGuiTableFlags_BordersInnerV |
+	ImGuiTableFlags_Resizable |
+	ImGuiTableFlags_Reorderable |
+	ImGuiTableFlags_Hideable |
+	ImGuiTableFlags_ContextMenuInBody;
```
m9710797: Many (most?) of our tables also add `ImGuiTableFlags_RowBg`. See if you like that or not.

src/imgui/ImGuiIoPorts.hh
```diff
@@ -0,0 +1,70 @@
...
+	  */
+	struct Group {
+		std::string device;
+		std::vector<Row> ranges;
+	};
+	[[nodiscard]] static std::vector<Row> getRows(MSXCPUInterface& cpuInterface);
+	[[nodiscard]] static std::vector<Group> getGroups(MSXCPUInterface& cpuInterface);
+	[[nodiscard]] static std::string portsText(const Row& row);
+	[[nodiscard]] static std::string_view dirText(const Row& row);
+	static void sortRows(std::vector<Row>& rows);
```
m9710797: `static` + `private` methods can often be removed from the header file: put them directly in the `.cc` file without declaring them in the `.hh` file. That's because if they are `static` they anyway can't access any member variables. They could still require types defined in the header. But I think if we move the methods, we can also move the `Row` and `Group` types.

So I think the only thing that must remain here is `show` and `persistentElements` (and `groupByDevice` if you want to keep that).

m9710797 (commented):
Hi. Thanks for working on this!

I did a first review pass, I focused most on usability feedback, and then often by removing/simplifying stuff.

I also have a few suggestions for how to improve performance. Some important, other very minor (feel free to ignore those).

src/imgui/ImGuiIoPorts.cc
```diff
@@ -0,0 +1,283 @@
...
+
+namespace openmsx {
+
+using namespace std::literals;
+
+static constexpr auto OVERLAP_TOOLTIP =
+	"These ports are shared with another device. Reading such a port returns "
+	"the bitwise AND of all values, writing goes to every device."sv;
+
+static constexpr int TABLE_FLAGS =
```
m9710797: `OVERLAP_TOOLTIP` and `TABLE_FLAGS`:
After the recent changes these two are both used exactly once. Maybe it's good to declare them closer to their only use.

m9710797 (commented):
For me this PR is fine.  My only review comment was a minor detail, I would have been happy to merge without that. But before merging I anyway still wanted to ask if this is also "done" from your side?

PR #2172, opened 2026-08-23
 feat: Make the 3rd-party download robust in CI

sndpl:
Two independent fixes:

- thirdparty_download.py had no retry: a single failed download aborts the whole Windows-VC build with a bare exit code 1. It now retries with a delay, and reports which URL failed.
- zlib is fetched from zlib.net, which serves a ~12 KB HTML block page instead of the tarball — so the checksum check fails, consistently and only in CI. The download URL now points at the byte-identical GitHub release asset.

### DIFF
--- a/build/packages.py
+++ b/build/packages.py
@@ -185,10 +185,12 @@ class Vorbis(DownloadablePackage):
 		return 'VORBIS'
 
 class ZLib(DownloadablePackage):
-	downloadURL = 'https://zlib.net/fossils/'
 	niceName = 'zlib'
 	sourceName = 'zlib'
 	version = '1.3.1'
+	# Note: zlib.net serves an HTML page instead of the tarball to some hosts,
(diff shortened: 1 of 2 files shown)


MBilderbeek:
Weird that the zlib change is needed. Why did it work fine so far?

Hmm, it was recently (Jun 6 18:11:57 2025, c9d7862f465715c80f863018f14708e2c4aa7a21) changed to zlib.net, and used to be coming from `downloads.sourceforge.net/libpng`... Not sure if that's a good source nowadays.

sndpl:
no clue, but each time the Windows VC build failed was because the download of zlib failed. I first added the retry, but after 3 retries it still failed until I changed the download location. The strange thing is that locally I couldn't reproduce it, it downloads the correct file from the zlib.net. But somehow it fails to download it in the CI.

PR #2173, opened 2026-08-23
Document debug subcommands and fix stale examples in manuals

pvmm:
## Description

The console command reference (`doc/manual/commands.html`) deferred six
`debug` subcommand families to the built-in `help` command, and several of
its examples used syntax that no longer exists. This PR brings the manuals
back in sync with current openMSX behavior.

### What changed

**commands.html**
- Dedicated sections (instead of "Type help ... for details") for all six
  `debug` families:
  - `breakpoint` — list/create/configure/remove, properties `-address`,
    `-condition`, `-command`, `-enabled`, `-once`
  - `watchpoint` — same subcommands plus `-type`
    (read_io/write_io/read_mem/write_mem) and begin/end address ranges
  - `condition` — list/create/configure/remove, continuous Tcl condition
  - `probe` — list/desc/read/set_bp/remove_bp/list_bp
  - `watchexpr` — list/create/configure/remove, `-description`,
    `-expression`, `-format`
  - `symbols` — types/load/remove/files/lookup, incl. the `$::sym(<name>)`
    shortcut
- Rewrote 5 examples using removed positional syntax (e.g.
  `debug breakpoint create 0xF37D {cond}` → syntax error today) into the
  current property style
- Fixed wrong setting name: `too_fast_vram_access_callback` →
  `VDP.too_fast_vram_access_callback`
- Added availability notes to `laserdiscplayer`/`autorunlaserdisc`:
  PX-7/PX-V60 machines only, and only when built with Laserdisc support

**user.html**
- Obsolete `debug watchpoint add read_io/write_io` → `debug watchpoint
  create -type read_io/write_io` (§9.5 Programmable Device)

## Verification

Every documented form was executed against a running openMSX instance
before writing it down: create/list/configure/remove cycles for
breakpoints, watchpoints, conditions, watch expressions and symbol files,
plus probe set_bp/remove_bp/list_bp. The old positional syntax was
confirmed broken at runtime (`Syntax error`).

## Out of scope

- Documenting the ~121 runtime commands missing from the reference
  (gap list included in the report).
- Removing legacy files (e.g. `doc/openmsx.sgml`, a 2004 man-page source);
  intentionally kept per project decision, see report §3.

### DIFF
--- a/doc/manual/commands.html
+++ b/doc/manual/commands.html
@@ -180,7 +180,7 @@
         <li><a class="internal" href="#soundchip_volume">&lt;soundchip&gt;_volume</a></li>
         <li><a class="internal" href="#throttle">throttle</a></li>
         <li><a class="internal" href="#too_fast_vram_access">too_fast_vram_access</a></li>
-        <li><a class="internal" href="#too_fast_vram_access_callback">too_fast_vram_access_callback</a></li>
+        <li><a class="internal" href="#too_fast_vram_access_callback">VDP.too_fast_vram_access_callback</a></li>
         <li><a class="internal" href="#touchpad_transform_matrix">touchpad_transform_matrix</a></li>
         <li><a class="internal" href="#turborpause">turborpause</a></li>
         <li><a class="internal" href="#umr_callback">umr_callback</a></li>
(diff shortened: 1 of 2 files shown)


MBilderbeek:
Thanks a lot!

Just a question: what's the point of documenting in a file which fixes we applied?

pvmm:
> Just a question: what's the point of documenting in a file which fixes we applied?

Because it was AI generated, so it does it by default. I will clean up unnecessary stuff before changing it to "ready for review"

m9710797:
I had a quick look (not yet in detail), looks reasonable. Thanks for working on this.

But indeed, as MBilderbeek already said, we shouldn't add `obsolete-docs-report.md`.

One question. In the description of the PR you (or AI) mention that the old positional syntax now gives syntax errors. That should NOT be the case. The old syntax is deprecated, and no longer mentioned in the help text. But for backwards compatibility it should remain supported. Did you encounter any examples that indeed produce syntax errors now? (Or was this some AI hallucination?) If so, then I'd like to know, because then we should fix that.

Currently we have two sets of parallel documentation. The help text in openMSX itself (queried via the `help` command). And the `html` manual. I would love to see some kind of mechanism to reduce this duplication. Maybe some way to generate (parts of) the manual from the openMSX help text?

pvmm:
> One question. In the description of the PR you (or AI) mention that the old positional syntax now gives syntax errors. That should NOT be the case. The old syntax is deprecated, and no longer mentioned in the help text. But for backwards compatibility it should remain supported. Did you encounter any examples that indeed produce syntax errors now? (Or was this some AI hallucination?) If so, then I'd like to know, because then we should fix that.

Yes, it found examples that cause syntax errors. In **command.html** since e40f401cc520c97737c4cc81e48b8f5988b9b760, lines 706-717:
```
debug breakpoint create 0xF37D {[reg C] == 0x2F}
debug watchpoint create write_io 0x99 {[reg A] == 0x81}
debug condition create {[reg hl] == 1234}
```
It also found an instance of a `debug watchpoint` command in user.html that doesn't work in the Programmable Device section:
```
debug watchpoint add read_io/write_io
```

m9710797:
Thanks! Merged.

The commands with syntax errors appeared to be a mix between the old positional syntax and the new command names. I missed that. Thanks for correcting those.

Anyway, I think it would be good if we could generate (more of) the documentation from a single source. Then it can never get out of sync and PRs like this won't be needed anymore.

PR #2164, opened 2026-08-04
Host RS232 / TCP-IP UNAPI 

ducasp:
Resolves #88 

(It originally was asking for network support, with this support to TCP-IP and UDP through UNAPI is supported)

Resolves #161 

(adds RS232raw as option to RS232 connector, works with MSXRS232 like OpenMSX Team RS232 adapter)

Reasoning: basically I develop a lot of UNAPI softwares. Also, develop ESP firmwares that have UNAPI built-in. I was using BlueMSX so far, with a custom build and rs232 and sm-x emulation I made myself. With the advent of AI assistance, I was able to migrate it to OpenMSX, and thought that it would be nice if merged with the oficial code. This will benefit developers like me, NataliaPC or Leo Manes... This will benefit developers that might want to try making network applications or games but the lack of a good debugging environment.

Why have an option needing an ESP plugged over serial (or its own serial to usb converter)? That was the base anyway, but also because using the host connection might be limitating... I.e.: you might have issues with reserved ports, firewalls, nagle settings, etc... ESP over a serial connection will give you total freedom of such nuisances. On Windows it seems to work fine even for server applications like OBSOFTP, while on Linux you have quite some steps to do and might want to turn firewall to make your life easier or adding rules, whatever...

Testing:

Testing was done on Windows 64 bits, Windows 11, and on Linux 64 bits, Ubuntu. Code should also work on MacOS but I have no Mac device to compile and test, so it might need some tweaking.

How to test:

For the Generic UNAPI it is pretty simple, just plug it... On any free slot, use one of my UNAPI roms, I recommend this: [ESPUNAPI_IO.rom](https://github.com/ducasp/MSX-Development/blob/master/UNAPI_BIOS_CUSTOM_ESP_V2/ESPUNAPI_IO.rom), reboot, you can use any TCP IP application like TELNET, TFTP, OBSOFTP, HGET, FTP, etc...

If wanting to test the SM-X (ESP connected) you can simply buy a cheap ESP32-C6 module that has usb-c connection and 16MB flash, and flash the firmware (source code and instructions on building it can be found on https://github.com/ducasp/ESP32-UNAPI-Firmware.

I'll make the following hard disk image available for a while, that already has UNAPI applications on it for testing: 
[new-image0001.zip](https://github.com/user-attachments/files/30717758/new-image0001.zip)

Changes descriptions:

- changes on some vcxproj so it compiles without error on Visual Studio 2022
- changes on some vcxprojs for the new files
[...]

PR #2174, opened 2026-08-24
rs232raw

### LINKED ISSUE #161: Map RS232 Devices to host RS-232C devices [sf#137]
**Reported by cdsg on 2006-04-15 18:11 UTC**
Emulation of rs232 port tru the COM port of the host
system, real or emulated host port, expecially Sunrise
RS232.

This makes available the use of internestor software
for example.

Greets.

ducasp:
Resolves https://github.com/openMSX/openMSX/issues/161

(adds RS232raw as option to RS232 connector, works with MSXRS232 like OpenMSX Team RS232 adapter)

This is part 1 of 3 pull requests, let's start with the easier one. Basically with it you can use openmsx team rs232 adapter or any other compatible adapter with RS232 port on your host. This has already passed for a first pass review on #2164. You can on the MSX side baud rate, bits, parity, etc... DTR and RTS should work as well, other signals are rarely used so did not consider refactoring the code to also allow the other signals to be changed from openmsx side.

Tested on Windows and Linux, should build fine on Windows, Linux, MAC and cross compilation MINGW

I consider it might be a driver to close #161


MBilderbeek:
There seems to be a serious (not some transient issue) build error on MacOS x86_64 at least. Please check that.

ducasp:
> There seems to be a serious (not some transient issue) build error on MacOS x86_64 at least. Please check that.

Yeah, seems macosx bukid stack does not support safer jthreead, will fall back th regular thread and send again

ducasp:
Yeah, investigation showed that jthread is not available normally on xcode version being used here. As such, just refactored it to use std::thread instead, re-tested, all looks fine and now should compile without issues. If we get a SonarQube request to use jthread, well, sorry, can't do it without breaking the MAC build

build/msvc/openmsx.vcxproj
```diff
@@ -1700,8 +1704,9 @@ popd</Command>
   <ItemGroup>
     <ClInclude Include="$(OpenMSXSrcDir)\sound\fake_windows.h" />
     <ClInclude Include="$(OpenMSXSrcDir)\MSXPiDevice.hh" />
+    <ClInclude Include="..\..\src\imgui\ImGuiPluggable.hh" />
```
m9710797: This seems wrong. All other tags like this use `$(OpenMSXSrcDir)`, this is the only one that uses a relative path.

build/msvc/openmsx.vcxproj.filters
```diff
@@ -1,15 +1,6 @@
...
   <ItemGroup>
-    <Filter Include="3rdparty\ImGuiFileDialog">
-      <UniqueIdentifier>{04a6bee4-b7ea-42af-ac09-cbdafe897ae5}</UniqueIdentifier>
-    </Filter>
-    <Filter Include="3rdparty\imgui">
-      <UniqueIdentifier>{3b13637c-13a0-4e77-a5aa-d68d68fdf5a9}</UniqueIdentifier>
-    </Filter>
-    <Filter Include="3rdparty\imgui_memory_editor">
-      <UniqueIdentifier>{305ed602-57dc-4f3c-9fdd-e372d1403920}</UniqueIdentifier>
-    </Filter>
```
m9710797, line 12: I see a lot of changes in this file, while I was only expecting the addition of a few new files.

Maybe the change is still correct and/or even intentional? Can you comment?

src/serial/RS232Device.hh
```diff
@@ -15,7 +15,7 @@ public:
 	// input
 	virtual void signal(EmuTime time) = 0;
 
-	// SerialDataInterface (part) (output)
+	// **COMMENT** SerialDataInterface (part) (output)
```
m9710797: I assume this is still to do?

src/utils/SerialPort.cc
```diff
@@ -0,0 +1,604 @@
...
+		return std::unexpected(ec);
+	}
+
+	return Handle(h);
+}
+
+Handle::Handle(serial_handle_t handle_)
+	: handle(handle_)
+{
+}
```
m9710797: This and the following methods (lines 114-145) seem to be identical(?) between the windows and POSIX version. Can you move them outside the `#ifdef`.

src/utils/SerialPort.cc
```diff
@@ -0,0 +1,604 @@
...
+		.dcd = (status & MS_RLSD_ON) != 0,
+		.ri  = (status & MS_RING_ON) != 0,
+	};
+}
+
+std::vector<std::string> list_ports()
+{
+	std::vector<std::string> ports;
+	for (int i = 1; i <= 256; ++i) {
+		std::string name = "COM" + std::to_string(i);
```
m9710797: Detail: this is indeed the typical standard c++ way. But in openMSX we have a utility to simplify this (and slightly more efficient, not that it matters here):
```
auto name = strCat("COM", i);
```
With `strCat()` you can concatenate all kinds of things (in this case a string-literal and an integer) into a `std::string`, and if needed each component is first converted to a string. (At least conceptually that's how it works, the actual implementation avoids the intermediate string conversions).

src/utils/SerialPort.cc
```diff
@@ -0,0 +1,604 @@
...
+			CloseHandle(h);
+			ports.push_back(name);
+		}
+	}
+	return ports;
+}
+
+#else // POSIX (Linux / macOS / BSD)
+
+static constexpr int BAUDRATE_MAP[][2] = {
```
m9710797: We try to avoid plain C-arrays in openMSX, instead we use std::array.

Also, could you use stronger types here, so that we don't need the casts in the 2 `baudToXXX()` functions below. Maybe something like this:
```
struct BaudMapItem {
    unsigned baud;
    speed_t speed;
};
static constexpr auto BAUDRATE_MAP = std::to_array<BaudMapItem>({
    {50, B50},
    {75, B75},
    ...
});
```

src/utils/SerialPort.cc
```diff
@@ -0,0 +1,604 @@
...
+	}
+	return B0;
+}
+
+#ifndef __linux__
+static unsigned baudToNearest(unsigned baud)
+{
+	unsigned bestRate = 0;
+	unsigned bestDiff = 0;
+	bool first = true;
```
m9710797: Detail, minor tweak.

A way to avoid this `bool first` is to initialize `bestDiff` to a very large value (like `UINT_MAX`). That way you don't need to test/set `first`.
And if you then initialize `bestRate` to `115200` you can also drop the test on line 331.

src/utils/SerialPort.cc
```diff
@@ -0,0 +1,604 @@
...
+		cfsetispeed(&tio, speed);
+		cfsetospeed(&tio, speed);
+	} else {
+#ifdef __linux__
+		cfsetispeed(&tio, BOTHER);
+		cfsetospeed(&tio, BOTHER);
+		tio.c_ispeed = static_cast<speed_t>(baud);
+		tio.c_ospeed = static_cast<speed_t>(baud);
+#elif defined(__APPLE__)
+		unsigned nearest = baudToNearest(baud);
```
m9710797: It seems that all (2) callers of `baudToNearest()` are immediately followed by `baudToSpeedExact()`.
Should we then maybe make `baudToNearest()` dircetly return a `speed_t`? That also avoids walking the `BAUDRATE_MAP` twice.

src/utils/SerialPort.cc
```diff
@@ -0,0 +1,604 @@
...
+	return ioctl(fd, IOSSIOSPEED, &speed) >= 0;
+#else
+	return true;
+#endif
+}
+
+std::expected<Handle, ErrorCode> open(std::string_view portName, const SerialParams& params)
+{
+	auto name = std::string(portName);
+	int fd = ::open(name.c_str(), O_RDWR | O_NOCTTY | O_NONBLOCK);
```
m9710797: The first (and only) thing you do with the `std::string_view portName` parameter is convert it to `std::string` so that it has a guaranteed 0-terminator (and you can call `.c_str()`).

That's an indication that `std::string_view` is the wrong type for the parameter. Better would be our `zstring_view` type. That's like `std::string_view` but with the additional guarantee that it refers to a 0-terminated string. (This type is proposed for a future c++ version, originally via the name `std::zstring_view`, later updated to `std::cstring_view`, in openMSX we still use the old proposed name).

If you change the type here, for consistency, we should also change it in most other places in this file that pass a `portName`. Including the Windows variants (even if they don't strictly need the 0-terminator).

src/utils/SerialPort.cc
```diff
@@ -0,0 +1,604 @@
...
+});
+
+std::vector<std::string> list_ports()
+{
+	std::vector<std::string> ports;
+	ReadDir dir("/dev");
+	while (auto* entry = dir.getEntry()) {
+		for (auto pfx : portPrefixes) {
+			if (std::string_view(entry->d_name).starts_with(pfx)) {
+				ports.push_back("/dev/" + std::string(entry->d_name));
```
m9710797: Also here our `strCat()` utility is useful:
`ports.push_back(strCat("/dev", entry->d_name));`
It can concatenate 2 `const char*`. And that avoids the cost of having to construct that intermediate `std::string`.

src/utils/SerialPort.cc
```diff
@@ -0,0 +1,604 @@
...
+		static_cast<DWORD>(buffer.size()),
+		nullptr
+	);
+
+	if (char_count == 0) {
+		return "Unknown error (" + std::to_string(ec.value) + ")";
+	}
+	std::wstring_view sv(buffer.data(), char_count);
+
+	while (!sv.empty() && (sv.back() == L'\n' || sv.back() == L'\r')) {
```
m9710797: Detail, small tweak:
`while (!sv.empty() && sv.back() == one_of(L'\n', L'\r')) { ...`

src/serial/RS232Raw.cc
```diff
@@ -0,0 +1,431 @@
+// Keep these before "RS232Raw.hh": that header pulls in <windows.h> (via
+// SerialPort.hh), whose macros (TRANSPARENT, BLACK, WHITE, YELLOW, ...) would
+// otherwise clash with the imColor enum in ImGuiUtils.hh.
```
m9710797: Is this still needed, even after you added `-DWIN32_MEAN_AND_LEAN`? Or is there any other way to prevent namespace pollution by the windows header? If not then this workaround is fine (but it's always better to try to fix the root-cause rather than work around the secondary problems).

src/serial/RS232Raw.cc
```diff
@@ -0,0 +1,431 @@
...
+	poller.emplace();
+	thread = std::thread([this, portName, p]() { openAndRun(portName, p); });
+#else
+	auto p = buildParamsImpl(currentBaud, currentDataBits, currentStopBits,
+	                         currentParityEnabled, currentParity);
+	(void)handle->set_params(p);
+	(void)handle->set_dtr(DTR);
+	(void)handle->set_rts(RTS);
+#endif
+}
```
m9710797, line 95: I see quite a bit of code duplication between:
* the destructor
* this function (`applyParams()`)
* `plugHelper()`
* `unplugHelper()`

Maybe we should try to factor out `setup()` and `tearDown()` (or better names) functions that handle cleaning-up/re-create the `thread`, `poller` and `buildParamsImpl` stuff.


EDIT:
And these have the same duplication:
* update()
* serialize()

src/serial/RS232Raw.cc
```diff
@@ -0,0 +1,431 @@
...
+	handle.reset();
+	if (thread.joinable()) {
+		thread.join();
+	}
+	{
+		std::scoped_lock lock(mutex);
+		queue.clear();
+	}
+	auto portName = std::string(rs232RawPortSetting.getString().c_str());
+	if (portName.empty()) return;
```
m9710797, line 81: `rs232RawPortSetting.getString()` return a zstring_view IIRC. Do we really need to convert that into a `std::string`? Can't we pass a `zstring_view` directly to `openAndRun()` (I mean change the code so that it does). Because I did already discuss changing serial::open() to take a `zstring_view` parameter.

Ah wait! `openAndRun` runs in a different thread! We indeed **must** pass an **owning** string between threads (the string_view would be dangling).

Maybe still a small tweak: the test for empty (line 99 and also 119) is safe to do on the `zstring_view`, so we could postpone creating the `string` till after that check.

src/serial/RS232Raw.cc
```diff
@@ -0,0 +1,431 @@
...
+}
+
+void RS232Raw::handleImGuiExtraMenuItems()
+{
+	std::string cur(rs232RawPortSetting.getString().c_str());
+	if (cur != currentPort) {
+		ports = serial::list_ports();
+		currentPort = cur;
+	}
+	ImGui::SameLine();
```
m9710797: I'd like discuss this `ImGui::SameLine()` thing. And then compare it with the alternative of wrapping the code below in a `im::Indent()` block. So something like:
```
// remove this:  ImGui::SameLine();
im::Indent([&]{
    ImGui::SetNextItemWidth(...);
    im::Combo(
       ...
   });
   ...
   ImGui::SameLine();
   if (ImGui::SmallButton("Refresh")) { ...
       ...
   }
});
```

Both are correct, the only difference is visual:
* Put everything on a single line.
* Put the new combo box + button (both) on the next line, but indented.

The single line is very clear that it belongs together. Hopefully because of the indent it's still clear enough.
But everything on a single line makes the menu very wide (with large empty areas). When using a small screen (and e.g. scale_factor=2) that makes the Connector menu no longer fit inside the main openMSX window, that's not necessarily a problem, but it doesn't look nice.

What do you think?

src/serial/RS232Raw.cc
```diff
@@ -0,0 +1,431 @@
...
+}
+
+// Opens the port (retrying briefly) and then runs the read loop. Runs on
+// the worker thread: the retry delays sleep this host thread, never the
+// emulation thread (a just-closed Windows COM port can stay owned by the
+// OS for a few ms, so an immediate reopen can transiently fail).
+void RS232Raw::openAndRun(std::string portName, serial::SerialParams params)
+{
+	auto h = serial::open(portName, params);
+	for (int attempt = 1; !h.has_value() && attempt < 5 && !poller->aborted();
```
m9710797: This is not necessarily wrong, but it's unusual. This loop tries 4 times (not 5) as the `attempt < 5` might suggest. I'd suggest to initialize as `int attempt = 0`, to get the typical counting loop pattern.

src/serial/RS232Raw.cc
```diff
@@ -0,0 +1,431 @@
...
+		char b;
+		auto r = handle->read(std::span<char>(&b, 1));
+		if (!r.has_value()) {
+			handle.reset();
+		} else if (*r == 0) {
+#ifdef _WIN32
+			Sleep(1);
+#endif
+		} else if (direct) {
+			conn->recvByte(static_cast<uint8_t>(b), EmuTime::zero());
```
m9710797: In general it's a good idea to try to minimize the number of casts that are needed. My initial idea was: "if we change the type of `b` from `char` to `uint8_t` then we can avoid this cast here. But then `serial::Handle::read()` currently expects `char` instead of `uint8_t`. Would it make sense to change the API to `uint8_t` instead (and then have the cast in the `SerialPort` code probably).

Technically it's the same (because there's only a single user of this API yet): have a cast in (all?) users of the API, or once in the implementation. But IMHO the question is more about semantics. Do we send *bytes* over the serial port or *characters*? Personally I'm leaning towards "bytes" (thus change to API to `uint8_t`).

---

Similar remark around line 277. If we change `queue` from `cb_queue<char>` to `cb_queue<uint8_t>` we can eliminate some casts.

m9710797 (commented):
I've done a 2nd round of review (1st was done on a different PR).
Looks good. Most of my comments are about details or small tweaks.

build/msvc/openmsx.vcxproj
ducasp: Will fix it on next commit

build/msvc/openmsx.vcxproj.filters
ducasp: Yeah, somehow it changed a lot more than needed, will fix in next commit and make sure it only add the n

src/serial/RS232Device.hh
ducasp: This was a comment sonnarqube was thinking to be code commented, will change it in a better way so it doesn't look like code :D

src/utils/SerialPort.cc
ducasp: Agree!

src/utils/SerialPort.cc
ducasp: Gotcha, will use it in the next commit

src/utils/SerialPort.cc
ducasp: Ok, will make sure it is updated next commit

src/utils/SerialPort.cc
ducasp: Will get it on next commit

src/utils/SerialPort.cc
ducasp: Yeah, makes sense and can merge the identical apple part of code

src/utils/SerialPort.cc
ducasp: Will get it on next commit

src/utils/SerialPort.cc
ducasp: Will get it on next commit

src/serial/RS232Raw.cc
ducasp: Yeah, DWIN32_LEAN_AND_MEAN will not trim those defines that comes from wingdi.h and can clash if imported in different order, so thought on just documenting that keeping this order avoid issues. Alternatives possible would be to rename inColor enum, which might be a tad excessive for this, or, undefining the unused macros, which also might be a tad too much in this case... Let me know what you think about this.

src/serial/RS232Raw.cc
ducasp: Will check and adjust acordingly, serialize() might prove a little bit more difficult but will work on it as well :)

src/serial/RS232Raw.cc
ducasp: Yeah, agree it doesn't look exactly nice and will try to adjust according to your suggestion :)

src/serial/RS232Raw.cc
ducasp: The thought was on trying up to 5 times, so if it doesn't work, retry 4 more times....  Will change to be = 0 and < 4 so it doesn't look weird / like it was unintentional mistake

src/serial/RS232Raw.cc
ducasp: Agree and will get it on the next commit

src/serial/RS232Raw.cc
m9710797: Changing the names in the `imColor` enum is indeed a worse solution. Instead we should fix the problem (as close as possible) at the root cause.
* The best place would be fixing `windows.h` itself. But that's of course not possible. (This header is a notoriously problematic header, as in: defining various macros with common names).
* The next best thing is to limit the damage that `windows.h` can do. For example by not including it in `SerialPort.hh` but only in `SerialPort.cc`. Then the "problems" are at least limited to `SerialPort.cc` and do not propagate to other (transitive) users of `SerialPort.hh`.
    * But currently we need `windows.h` in `SerialPort.hh` for `HANDLE`. We had the same problem in `MappedFile.hh`, these we solved it via `using HANDLE = void*` (see `MappedFile.hh` line 12).

src/utils/SerialPort.cc
```diff
@@ -0,0 +1,584 @@
...
+	auto n = ::read(handle, buf.data(), buf.size());
+	if (n > 0) return IoResult(n);
+	if (n == 0) return IoResult(0);
+	return std::unexpected(currentError());
+}
+
+IoResult Handle::write(std::span<const uint8_t> buf) const
+{
+	auto n = ::write(handle, buf.data(), buf.size());
+	if (n > 0) return IoResult(n);
```
m9710797: I overlooked this in my previous review. And maybe it's not a problem, so just asking ...

In `read()` we check for `n == 0`, but here in `write()` we do not (an asymmetry). Should we check for 0? Can `::write()` ever return 0? Is it valid to pass a `buf` parameter with `size()=0`?

Detail: in `read()` we can combine the check of `n > 0` and `n ==0` into `n >= 0` (because both branches anyway do the same).

src/serial/RS232Raw.cc
```diff
@@ -0,0 +1,422 @@
...
+
+zstring_view RS232Raw::getDescription() const
+{
+	return "RS232 raw serial port pluggable. Connects the RS232 port to "
+	       "a host serial port, selected with the 'rs232-raw-port' setting.";
+}
+
+void RS232Raw::handleImGuiExtraMenuItems()
+{
+	std::string cur(rs232RawPortSetting.getString().c_str());
```
m9710797: Detail, small tweak:
Do we (always) need to convert the result of `rs232RawPortSetting.getString()` (a `zstring_view`) to `std::string`? I think keeping it as `auto cur = rs232rawPortSetting.getString();` (so keep as `zstring_view`) is fine, we only need to convert to `std::string` on line 153.

Maybe another suggestion:
```
if (auto cur = rs232RawPortSetting.getString(); cur != currentPort) {
    ports = serial::list_ports();
    currentPort = std::string(cur);
}
... rest of the code uses `currentPort`, not `cur`
```
This limits the scope of the `cur` variable to this if-block, and then in the rest of the code there's no discussion/confusion about whether we should use `cur` or `currentPort` (they are the same).

m9710797 (commented):
Thanks for all the fixes, looks good.

My only major remaining remark is about the hack for `windows.h` in `RS232Raw.hh`. If this is fixed we can merge this PR.

The other stuff are details (small stuff I missed during my previous review pass).

src/utils/SerialPort.cc
ducasp: Indee, write() can actually return 0. It can't happen with our current callers (we always write exactly 1 byte), but treating this as an error felt wrong, so it's fixed now.

Also, read() checks will be combined into n >= 0, write() is now symmetric with the same check.

src/serial/RS232Raw.cc
ducasp: cur is now a zstring_view scoped to the refresh-if via an init-statement, and the rest of the menu uses currentPort — combo label, selected check, all of it. No more wondering which one is which.

About the windows.h hack, did some changes so the order of headers do not matter, please check if that is a good way of doing it, suggestions accepted if not :)

m9710797: I had a very quick look, all seems fine. Your hack for the `windows.h` stuff is exactly what I had in mind.

I'll do a final check when I'm back at home, and then merge it.
Thanks!

m9710797:
Merged. Thanks!

Just checking .. originally this was part of a larger PR. Do you still have plans to work on those other parts?

ducasp:
Yeah, absolutely, just trying to not overwhelm you guys, so doing it one at a time, and this feature has its use independently of the other two. I'm checking if I can simplify the other two so it could be one where you can either choose to try an esp over rawrs232 or using emulated esp, that would surely diminish the lines of code as interface is the same both cases, just what it talks to would change

PR #2177, opened 2026-08-26
Add debug_log / debug_log_read utility functions

pvmm:
Add a simple debug message buffer to the utils namespace for use by Tcl scripts that need to send diagnostic output through the MCP server. Standard puts/stderr output is not visible to the MCP tools, so these functions accumulate messages in a buffer that can be read and cleared via `utils::debug_log` and `utils::debug_log_read`.

### DIFF
--- a/share/scripts/_utils.tcl
+++ b/share/scripts/_utils.tcl
@@ -168,6 +168,25 @@ namespace export file_completion
 namespace export filename_clean
 namespace export get_next_numbered_filename
 
+#
+# debug_log / debug_log_read
+#
+# Accumulates debug messages in a buffer that can be read via
+# debug_log_read. This is useful for sending diagnostic output
+# through the MCP server, which cannot see puts/stderr output.
+#
+variable _debug_log {}
+proc debug_log {msg} {
+	variable _debug_log
+	lappend _debug_log $msg
+}
+proc debug_log_read {} {
+	variable _debug_log
+	set result [join $_debug_log "\n"]
+	set _debug_log {}
+	return $result
+}
+
 } ;# namespace utils
 
 # Don't import in global namespace, these are only useful in other scripts.


share/scripts/_utils.tcl
```diff
@@ -168,6 +168,25 @@ namespace export file_completion
...
+# through the MCP server, which cannot see puts/stderr output.
+#
+variable _debug_log {}
+proc debug_log {msg} {
+	variable _debug_log
+	lappend _debug_log $msg
+}
+proc debug_log_read {} {
+	variable _debug_log
+	set result [join $_debug_log "\n"]
```
m9710797, line 185: Rather than forcing a `join` upon the user you could just return the list as is.
* Then the user could still do the `join` if he wants to.
* Or he can iterate over the original list. (And not pay the cost of `join` soon after (probably) followed by `split`).

This can make a difference when the messages send via `debug_log` can themselves contain `\n` characters. When you return the plain list the user can still see the original message boundaries. In the current code those boundaries are lost.

Maybe those boundaries don't matter. But I wonder why you collected in a list. If you know you always want to `join` you could instead do `append _debug_log $msg "\n"`  (notice `append` not `lappend`).

I don't know what's best. You know the use case better, you should decide (I'm only presenting 3 alternatives).

---

But taking a step back. Basically what you created is a thin wrapper around a global list variable. When I originally read the name `debug_log` I was thinking about very different functionality (so maybe we should not use the name `debug_log` here?).

And maybe rather than teaching the MCP about `debug_log` and `debug_read_log` we could teach it about Tcl global variables? So:
* replace `debug_log` with `lappend ::mcp_log $msg`.
* replace `debug_log_read` with `$::mcp_log` (+clear if needed).

Wouldn't that also be more flexible? Just one example: maybe what the MCP wants is a counter rather than messages (each with an increasing value).

m9710797 (commented):
Hi. Thanks for your PR.

Functionally this is correct. But I present 3 possible alternatives, You should choose.
* keep as is
* change towards list-based approach
* change towards string-based approach

But maybe more important: do we actually need this? Aren't there already more flexible mechanisms available in the Tcl language itself?

pvmm:
Not necessary anymore. I solved this on mcp-openmsx side.

PR #2179, opened 2026-08-28
feat(macOS): updated the installer so it contains an Applications folder and a background

sndpl:
- Window sized exactly to the background art (600x403 points), with toolbar and status bar hidden.
- App icon and Applications shortcut centred in the two boxes drawn in the background. Because Finder renders an item's label beneath its icon, the icon+label block is centred as a unit, leaving even margins above and below.
- Titled openMSX <version> installer, shown in the title bar and Finder's path bar. This is the volume name, and it uses the plain release version rather than the git revision so it stays stable between builds.
- Background is Retina-ready: dmg_bg.png (600x403) and dmg_bg@2x.png (1200x806) are combined at build time into a multi-resolution TIFF via tiffutil. Finder lays the window out at the 1x size in points while drawing the 2x representation on Retina displays. (use the background from the site, also the same font)
- .background, .fseventsd and friends are pushed out of view.
- README.html is no longer included in the disk image (only contained obsolete information)

Previous installer looked like:
Screenshot 2026-08-28 at 09 39 18

New installer looks like:
Screenshot 2026-08-28 at 09 38 45


m9710797:
Description looks good, screenshots look good.
I had a brief look at the changed code: looks good to me. But I'm not really qualified to understand/review that code in detail.

@MBilderbeek can you review, or ask someone to do it? As shall we just merge this?

MBilderbeek:
@grauw or @gilbertfrancois can you review this?

gilbertfrancois:
I checked out the source, built it, and tested the DMG installer on macOS 26. It works great and gives a much more native-feeling install experience on Mac. Nice work!

One minor, purely subjective note: the background artwork is a bit taste-sensitive. Right now it uses a warm/creme-toned background, which can clash with a user's own desktop settings (dark/light mode, colored, tinted, or monochrome icon themes, etc.) — especially since this is still a Finder window, not the running app itself. A neutral option, transparent background with neutral gray lines and text, would adapt cleanly to whatever look the user already has, rather than imposing one.

To be clear, this is just my personal preference and not a blocker for merging.

PR #2181, opened 2026-08-29
fix(input): scale MSX mouse movement in host window coordinates

sndpl:
Since commit 396a7b0fa the host-to-MSX mouse scale was derived from OutputSurface::getPhysicalSize(), which is expressed in physical pixels. SDL reports mouse motion in window coordinates ('points'), and on macOS openMSX creates its window with SDL_WINDOW_ALLOW_HIGHDPI, so on a retina display those differ by the backing scale factor. The MSX pointer moved at half the speed of the host pointer. Measured with scale_factor 4: a 1280x960 point window has a 2560x1920 drawable, so the scale was 8 points per MSX pixel where it should be 4.

Two smaller problems in the same expression, and these do affect all platforms:
- it was an integer division, so in full screen the ratio got truncated. At 1080p with scale_factor 2 that gave 4 instead of 4.5;
- it used one scale for both axes, while 'horizontal_stretch' means MSX pixels are not drawn square. At the default of 280 the X axis needs to be about 14% faster than the Y axis to track the host pointer.

Add VisibleSurface::getMsxPixelSize(), which returns the on-screen size of a single MSX pixel per axis, in the same unit SDL uses for mouse coordinates. It converts the viewport from physical pixels to points via the window size, and takes 'full_stretch' and 'horizontal_stretch' into account. Mouse now uses that, keeping the sub-pixel remainder in a float so slow movements are still not lost.

Note that this scales for equal distance on screen, which is not the same as equal distance in MSX pixels when horizontal_stretch is not 320: the MSX pointer then crosses fewer MSX pixels per centimetre horizontally than vertically.

Trackball and Paddle keep their fixed SCALE of 2: they never used the display size, and neither is a 1:1 pointer device.

I only tested this on my macbook, so maybe somebody can check it on windows/linux. The software I used to test with was Dot Designer Club of T&E Soft.

### DIFF
--- a/src/input/Mouse.cc
+++ b/src/input/Mouse.cc
@@ -10,14 +10,16 @@
 #include "serialize_meta.hh"
 
 #include "Display.hh"
-#include "OutputSurface.hh"
 
-#include "Math.hh"
+#include "gl_vec.hh"
+
+#include "narrow.hh"
 #include "unreachable.hh"
 
 #include <SDL.h>
 
 #include <algorithm>
+#include <cmath>
 
 namespace openmsx {
 
@@ -32,6 +34,21 @@ static constexpr int PHASE_XLOW2  = 5;
 static constexpr int PHASE_YHIGH2 = 6;
 static constexpr int PHASE_YLOW2  = 7;
 static constexpr int STROBE = 0x04;
+
+/** Convert host mouse movement along one axis to a whole number of MSX
+  * pixels. The part that doesn't (yet) add up to a full MSX pixel is
+  * remembered in 'fraction', so that slow movements don't get lost.
+  * Note: rounding in a uniform direction (instead of towards zero) gives
+  * smoother output.
+  */
+static int scaleHostMotion(int delta, float scale, float& fraction)
+{
+	if (scale <= 0.0f) return 0;
+	float value = narrow_cast<float>(delta) + fraction;
+	float result = std::floor(value / scale);
+	fraction = value - (result * scale);
+	return narrow_cast<int>(result);
+}
 
 
 Mouse::Mouse(MSXEventDistributor& eventDistributor_,
@@ -87,7 +104,6 @@ void Mouse::unplugHelper(EmuTime /*time*/)
 	stateChangeDistributor.unregisterListener(*this);
 	eventDistributor.unregisterEventListener(*this);
 }
-
 
 // JoystickDevice
 uint8_t Mouse::read(EmuTime /*time*/)
@@ -239,21 +255,16 @@ void Mouse::signalMSXEvent(const Event& event, EmuTime time) noexcept
 	visit(overloaded{
 		[&](const MouseMotionEvent& e) {
 			if (e.getX() || e.getY()) {
-				// Note: regular C/C++ division rounds towards
-				// zero, so different direction for positive and
-				// negative values. But we get smoother output
-				// with a uniform rounding direction.
-				float scale = SCALE;
-				if (const auto* output = display.getOutputSurface()) {
-					scale = output->getPhysicalSize().y / 240;
-				}
-				auto qrX = Math::div_mod_floor(e.getX() + fractionalX, scale);
-				auto qrY = Math::div_mod_floor(e.getY() + fractionalY, scale);
-				fractionalX = qrX.remainder;
-				fractionalY = qrY.remainder;
+				// Scale host mouse movement such that the MSX pointer
+				// travels the same distance on screen as the host
+				// pointer. The scale differs per axis because MSX
+				// pixels are not drawn square (see 'horizontal_stretch').
+				auto scale = display.getMsxPixelSize().value_or(gl::vec2(SCALE, SCALE));
+				auto deltaX = scaleHostMotion(e.getX(), scale.x, fractionalX);
+				auto deltaY = scaleHostMotion(e.getY(), scale.y, fractionalY);
 
 				// Note: hostXY is negated when converting to MsxXY
-				createMouseStateChange(time, -qrX.quotient, -qrY.quotient, 0, 0);
+				createMouseStateChange(time, -deltaX, -deltaY, 0, 0);
 			}
 		},
 		[&](const MouseButtonDownEvent& e) {
--- a/src/input/Mouse.cc
+++ b/src/input/Mouse.cc
@@ -13,13 +13,12 @@
 
 #include "gl_vec.hh"
 
-#include "narrow.hh"
 #include "unreachable.hh"
 
 #include <SDL.h>
 
 #include <algorithm>
-#include <cmath>
+#include <cassert>
 
 namespace openmsx {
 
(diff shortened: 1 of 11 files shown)


uniskie:
Tested it, looks great.

src/input/Mouse.cc
```diff
@@ -33,6 +35,21 @@ static constexpr int PHASE_YHIGH2 = 6;
...
 
+/** Convert host mouse movement along one axis to a whole number of MSX
+  * pixels. The part that doesn't (yet) add up to a full MSX pixel is
+  * remembered in 'fraction', so that slow movements don't get lost.
+  * Note: rounding in a uniform direction (instead of towards zero) gives
+  * smoother output.
+  */
+static int scaleHostMotion(int delta, float scale, float& fraction)
+{
+	if (scale <= 0.0f) return 0;
```
m9710797: Can at this point `scale` be `<= 0.0f`? Can this be replaced with `assert(scale > 0.0f);`?

To me, at first sight, the `value_or()` protection on line 262, together with when `getMsxPixelSize()` return`nullopt` should make this condition impossible?

src/input/Mouse.hh
```diff
@@ -57,7 +57,7 @@ private:
 	int phase;
 	int xRel = 0, yRel = 0;               // latched X/Y values, these are returned to the MSX
 	int curXRel = 0, curYRel = 0;         // running X/Y values, already scaled down
-	int fractionalX = 0, fractionalY = 0; // running X/Y values, not yet scaled down
+	float fractionalX = 0.0f, fractionalY = 0.0f; // running X/Y values, not yet scaled down
```
m9710797: Detail: was like already like this in the old code, but now that we're changing it ...
Maybe it makes sense to combine these two into `gl::vec2 fractional`, and then use `fractional.x` and `fractional.y`. (And then maybe still in a later refactoring we could replace more scalar math with vector math?)

src/video/VisibleSurface.cc
```diff
@@ -353,6 +353,28 @@ std::optional<gl::ivec2> VisibleSurface::getMouseCoord() const
 	return gl::ivec2{mouseX, mouseY};
 }
 
+std::optional<gl::vec2> VisibleSurface::getMsxPixelSize() const
+{
+	auto physSize = gl::vec2(getPhysicalSize());
+	if ((physSize.x <= 0.0f) || (physSize.y <= 0.0f)) return {};
```
m9710797: I was wondering: "is it ever possible for a `VisibleSurface` to have no physicalSize"?
(If that's the case then we could return a plain `gl::vec2` (drop the `optional` part).

But then when I read the code, I can't easily prove it always has:
* `OutputSurface::m_physSize` seems to be filled on `calculateViewPort()`.
* The reason is that on every frame we allow for the window dimensions to have changed.
    * This is also a preparation for later to allow the MSX window to be docked in a larger ImGui window, and then that dock-space can be easily resized (without and SDL window resize events).
* But if we'd call `getMsxPixelSize` before `calculateViewPort()` is called then maybe this could happen?
    * Ah, but the `VisibleSurface` constructor calls `setViewPort()` and in turn calls `calculateViewPort()`. So now I think `m_physSize` will always be correctly initialized?

Could you confirm the above reasoning. It's a minor tweak. But it's always nice to eliminate impossible corner cases from the code: then we can count of `VisibleSurface::getMsxPixelSize()` to always give a meaningful result.

src/video/VisibleSurface.cc
```diff
@@ -353,6 +353,28 @@ std::optional<gl::ivec2> VisibleSurface::getMouseCoord() const
...
+
+	const auto& renderSettings = display.getRenderSettings();
+	auto viewSize = gl::vec2(renderSettings.getFullStretch() ? getPhysicalSize()
+	                                                         : getViewSize());
+	auto viewInPoints = viewSize * gl::vec2(narrow<float>(windowW), narrow<float>(windowH))
+	                  / physSize;
+
+	// The viewport shows 'horizontal_stretch' of the 320 MSX pixels per
+	// line, and always all 240 lines.
+	return viewInPoints / gl::vec2(renderSettings.getHorizontalStretch(), 240.0f);
```
m9710797: I think this is correct but I'd like to make two tweaks:

1) The full expression is something like:
```
auto t = a * b / c;
return t / d;
```
In general it's nice to transform that into:
```
return (a * b) / (c * d);
```
Reason is: division is more expensive than multiplication. (And we're compiling without `-ffast-math` so the compiler is not allowed to make this transformation for us).

2) In case of `getFullStretch() == true` we have:
```
auto t = getPhysicalSize() * gl::vec2(...) / physSize;
```
With `physSize == getPhysicalSize()`. So have a repeated call to `getPhysicalSize()` and then both multiply and divide by the same value. Try to eliminate that.

m9710797 (commented):
Looks good. I only have a few suggestions for possible simplifications/tweaks. (Some need a bit of investigation before they can be applied).

src/input/Mouse.cc
```diff
@@ -239,21 +255,16 @@ void Mouse::signalMSXEvent(const Event& event, EmuTime time) noexcept
...
-				}
-				auto qrX = Math::div_mod_floor(e.getX() + fractionalX, scale);
-				auto qrY = Math::div_mod_floor(e.getY() + fractionalY, scale);
-				fractionalX = qrX.remainder;
-				fractionalY = qrY.remainder;
+				// Scale host mouse movement such that the MSX pointer
+				// travels the same distance on screen as the host
+				// pointer. The scale differs per axis because MSX
+				// pixels are not drawn square (see 'horizontal_stretch').
+				auto scale = display.getMsxPixelSize().value_or(gl::vec2(SCALE, SCALE));
```
m9710797: Detail: our `gl::vecN` types have "broadcast" constructors (many vector-math libraries have).
So `gl::vec2(SCALE, SCALE)` can be simplified to `gl::vec2(SCALE)`.

src/video/VisibleSurface.cc
sndpl: Confirmed, with one correction to the premise. `setViewPort()` → `calculateViewPort()` does run in the `VisibleSurface` constructor (line 145), and the object cannot be reached from outside before that: `Subject::attach` only appends to a vector, `distributeEvent` queues rather than delivering (delivery is in `deliverEvents()` from the Reactor main loop), and `Display::videoSystem` is only assigned once `SDLVideoSystem`'s constructor has completed, until then `Display::getMsxPixelSize()` returns `nullopt` anyway.

The correction: "uninitialized" was never the risk. `gl::ivec2`'s default constructor zero-initializes (`gl_vec.hh:41`), so before `setViewPort()` it is a deterministic `(0,0)`, not garbage. 

So I've dropped the `optional` from `VisibleSurface::getMsxPixelSize()` and turned the guard in `scaleHostMotion` into an `assert`, as you suggested. I kept the `optional` one layer up on `VideoSystem`/`Display`, since `DummyVideoSystem` genuinely has no window — same reason `getOutputSurface()` returns `nullptr` there.

PS: I did let claude write a program to test this, it wrote a standalone SDL program with ALLOW_HIGHDPI and it reports drawable=1280x960 while minimized and while hidden, never 0×0 :)

m9710797:
Merged. Thanks!

PR #2182, opened 2026-08-31
Fix saving a setup for machines with a '/' in their name (#2071)

### LINKED ISSUE #2071: Save Machine Configuration :  Invalid Filename Characters in Machine Names
# Bug Report: Invalid Filename Characters in Machine Names

## Summary

Two issues were discovered where machine names containing path separators or invalid filename characters cause save operations to fail:

1. **Session Management** - Session save fails with `"Uncaught exception: could not write"` 
2. **Setup Configuration** - Setup save fails when using "Save setup" menu button

Both affect machines with names like "Philips VG 8235/39" that contain forward slashes (`/`) or other invalid filename characters.

### Symptom 1: Session Management
```
Uncaught exception: could not write "Philips VG 8235/39.oms"
```

### Symptom 2: Setup Configuration
When using "Machine → Save setup" menu button with machine "Philips VG 8235/39":
```
Setup could not be saved: Cannot create file path with "/" in name
```

### Root Cause

Machine identifiers can contain forward slashes (e.g., "Philips VG 8235/39"), which are valid for openMSX but **invalid as filename components** on all operating systems. The slash is interpreted as a directory separator, causing the system to attempt creating non-existent subdirectories.

### Affected Machines

At least the following machines have a slash in their name:
- Philips VG 8235/39
- Possibly other machines in `share/machines/` with a `/` in their identifier

## Proposed Solution

Replace unsafe characters for filenames (particularly `/` and other invalid characters) before using the machine name in file paths.

### Files to Modify

**Files:** 
1. `share/scripts/session_management.tcl` - TCL script for session management
2. `src/imgui/ImGuiMachine.cc` - ImGui interface for setup configuration

#### 1. In `share/scripts/session_management.tcl` - `save_session` function (line ~85)

**Before:**
```tcl
	# save using ID as file names
	foreach machine [list_machines] {
		append result "Saving machine $machine ([get_machine_representation $machine])...\n"
		store_machine $machine [file join $directory ${machine}.oms]
	}
	# save in a separate file the currently active machine
	set fileId [open [file join $directory active_machine] "w"]
	puts $fileId [format "%s.oms" [machine]]
	close $fileId
```

**After:**
```tcl
	# save using ID as file names
	foreach machine [list_machines] {
		append result "Saving machine $machine ([get_machine_representation $machine])...\n"
		set safe_machine [string map {/ _} $machine]
		store_machine $machine [file join $directory ${safe_machine}.oms]
	}
	# save in a separate file the currently active machine
	set fileId [open [file join $directory active_machine] "w"]
	set active_machine [machine]
	set safe_active_machine [string map {/ _} $active_machine]
	puts $fileId [format "%s.oms" $safe_active_machine]
	close $fileId
```

#### 2. In `src/imgui/ImGuiMachine.cc` - `save_setup` menu (line ~213)

**Before:**
```cpp
			if (!saveSetupOpen) {
				// on each re-open of this menu, create a suggestion for a name
				auto configName = motherBoard->getMachineName();
				const auto* info = findMachineInfo(configName);
				auto initialSaveSetupName = info ? info->displayName : configName;
				saveSetupName = initialSaveSetupName;
```

**After:**
```cpp
			if (!saveSetupOpen) {
				// on each re-open of this menu, create a suggestion for a name
				auto configName = motherBoard->getMachineName();
				const auto* info = findMachineInfo(configName);
				auto initialSaveSetupName = info ? info->displayName : configName;
				// Sanitize the setup name by replacing invalid filename characters
				std::string sanitized;
				for (auto c : initialSaveSetupName) {
					// Replace path separators and other invalid filename characters with underscore
					if (c == '/' || c == '\\' || c == ':' || c == '*' || c == '?' ||
						c == '"' || c == '<' || c == '>' || c == '|') {
						sanitized += '_';
					} else {
						sanitized += c;
					}
				}
				saveSetupName = sanitized;
```

#### 2. Impact on `load_session`

The `load_session` function doesn't need modification because it enumerates files present on disk. The mapping from raw name to safe filename is implicit.

### Short Term
Apply both fixes above and recompile openMSX.

### Long Term
Consider creating a **centralized filename sanitization utility function** for use throughout the project:

```cpp
// In FileOperations.hh
[[nodiscard]] std::string sanitizeFilename(std::string_view name);

// In FileOperations.cc
std::string sanitizeFilename(std::string_view name)
{
	std::string result;
	for (auto c : name) {
		if (c == '/' || c == '\\' || c == ':' || c == '*' || c == '?' ||
			c == '"' || c == '<' || c == '>' || c == '|') {
			result += '_';
		} else {
			result += c;
		}
	}
	return result;
}
```

**Benefits:**
- Consistent behavior across the entire project
- Single point of maintenance
- Easy to update with additional characters if needed
- Covers all operating systems (Windows, Linux, macOS)

### Suggested Tests
After applying fixes, verify:
1. ✅ Saving setup with "Philips VG 8235/39" works → File saved as `Philips_VG_8235_39.oms`
2. ✅ Saving session with mixed machines including those with "/" works
3. ✅ Restoring saved sessions works correctly
4. ✅ Other machines without special characters unaffected
5. ✅ Old sessions can still be restored

## Test Environment

- **System:** Linux x86_64
- **openMSX Version:** 21.0
- **Compiler:** g++
- **Flavour:** opt

### ✅ COMPLETED AND TESTED

| Component | File | Status | Verification |
|-----------|------|--------|--------------|
| Session Management | `share/scripts/session_management.tcl` | ✅ Applied & Tested | Sessions save successfully with `/` → `_` conversion |
| Setup Configuration | `src/imgui/ImGuiMachine.cc` | ✅ Applied & Tested | Setup saves as `Philips_VG_8235_39.oms` |
| Compilation | Full project rebuild | ✅ Successful | Binary updated: `openmsx` v21.0 |

### Test Results

```
openmsx
info: Setup saved to Philips VG 8235_39
```

**Result:** ✅ **SUCCESS** - Both issues resolved without regressions.

sndpl:
Fixes #2071 

- added FileOperations::cleanFilename(), the C++ counterpart of the existing Tcl utils::filename_clean, mirroring its per-platform disallowed sets so both halves of openMSX produce the same filename for the same text. Only bytes < 0x80 are touched, so UTF-8 passes through unchanged.
- Applied to the four generated name suggestions: the setup name in ImGuiMachine, and the three guess_title suggestions in ImGuiTools (screenshot) and ImGuiReverseBar (save state, save replay). The latter three already had the cleaning step in their Tcl equivalents; the GUI just never got it.
- Caught the save exception in ImGuiMachine, like the Load setup button already does.

What the user types is deliberately left alone, these boxes accept a path on purpose which is why the fix is a cleaned suggestion plus an error report, not validation of the input.

Cleaning is applied where initialSaveSetupName is created rather than to the copy in the text box, which also fixes a second bug: the raw name reaching getNextNumberedFileName() made stem() split on the / and silently truncate the suggestion to 39 0001.

### DIFF
--- a/src/file/FileOperations.cc
+++ b/src/file/FileOperations.cc
@@ -281,6 +281,28 @@ std::string_view stem(std::string_view path)
 	return path.substr(pos2 + 1, pos1 - pos2 - 1);
 }
 
+// Characters that are not allowed (or not desirable) in a filename.
+// Keep in sync with 'utils::filename_clean' in share/scripts/_utils.tcl.
+static constexpr std::string_view disallowedInFilename =
+#ifdef _WIN32
+	R"(/\?*:|"<>+[])";
+#else
+	"/";
+#endif
+
+std::string cleanFilename(std::string_view filename)
+{
+	std::string result{filename};
+	std::ranges::replace_if(result, [](char c) {
+		// UNIX does allow 0x01-0x1f and 0x7f, but we consider them
+		// undesirable. Note that bytes >= 0x80 are left alone, so
+		// UTF-8 multi-byte sequences pass through unchanged.
+		auto u = static_cast<unsigned char>(c);
+		return (u < 0x20) || (u == 0x7f) || disallowedInFilename.contains(c);
+	}, '_');
+	return result;
+}
+
 std::string join(std::string_view part1, std::string_view part2)
 {
 	if (part1.empty() || isAbsolutePath(part2)) {
--- a/src/file/FileOperations.hh
+++ b/src/file/FileOperations.hh
@@ -151,6 +151,21 @@ namespace openmsx::FileOperations {
 	 */
 	 [[nodiscard]] std::string_view stem(std::string_view path);
 
+	/**
+	 * Replace characters that are not allowed (or not desirable) in a
+	 * filename by '_'. The result is a single path component: path
+	 * separators are replaced as well, so this is not suitable to clean
+	 * a whole path.
+	 * Mostly useful to turn free-form text (e.g. an MSX machine display
+	 * name or a guessed software title) into something that can be used
+	 * as a filename.
+	 * This is the C++ equivalent of the Tcl proc 'utils::filename_clean'
+	 * (see share/scripts/_utils.tcl), keep the two in sync.
+	 * @param filename The filename to clean
+	 * @result The cleaned filename, always the same length as the input.
+	 */
+	[[nodiscard]] std::string cleanFilename(std::string_view filename);
+
 	/** Join two paths.
 	 * Returns the equivalent of 'path1 + '/' + path2'. If 'part2' is an
 	 * absolute path, that path is returned ('part1' is ignored). If
--- a/src/imgui/ImGuiMachine.cc
+++ b/src/imgui/ImGuiMachine.cc
@@ -12,6 +12,7 @@
 #include "Debuggable.hh"
 #include "Debugger.hh"
 #include "FileException.hh"
+#include "FileOperations.hh"
 #include "GlobalSettings.hh"
 #include "HardwareConfig.hh"
 #include "MSXCommandController.hh"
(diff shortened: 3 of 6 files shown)


src/file/FileOperations.cc
```diff
@@ -281,6 +281,28 @@ std::string_view stem(std::string_view path)
...
+static constexpr std::string_view disallowedInFilename =
+#ifdef _WIN32
+	R"(/\?*:|"<>+[])";
+#else
+	"/";
+#endif
+
+std::string cleanFilename(std::string_view filename)
+{
+	std::string result{filename};
```
m9710797, line 295: Detail. We pass a `string_view` parameter, and the only thing we do with it is convert that to a `string`. It would be more efficient to pass a `string` parameter (by value):
* Because the parameter is passed by value, the caller needs to create (a copy of) the `string`. 
* But if the caller happens to have an r-value string, then we can `move` that string instead of making a `copy`. So in the worst case the number of copies remains the same (but shifted from callee to caller), in the best case we eliminate 1 copy.

This is very minor. And it does make (some of) the caller code more ugly. So I'll merge without this tweak.

PR #2147, opened 2026-07-07
Add UnapiNet: MSX-UNAPI TCP/IP bridge device (host networking for MSX software)

antxiko:
# Add UnapiNet: an MSX-UNAPI TCP/IP bridge device (host networking for MSX software)

## What this is

UnapiNet lets MSX networking software running inside openMSX reach the real
network. It is a device on I/O ports 0x28/0x29 that implements the
[MSX-UNAPI TCP/IP specification](https://github.com/Konamiman/MSX-UNAPI-specification)
functions — DNS, TCP (active/passive), UDP and ICMP echo — by bridging to the
host's sockets. MSX programs don't talk to these ports directly: they discover
a standard UNAPI TCP/IP implementation via EXTBIO, provided by a companion Z80
driver (`UNAPINET.COM`) that forwards each UNAPI call over the two ports.

```
MSX app --UNAPI/EXTBIO--> UNAPINET.COM (Z80) --I/O 0x28/0x29--> UnapiNet device --> host sockets
```

Any MSX software written against the UNAPI TCP/IP API is a potential client
(HGET, telnet/BBS clients, UNAPI-based multiplayer games). See "Testing
status" below for what I have actually exercised so far.

The device and driver were developed in a fork
([openMSXnet](https://github.com/antxiko/openMSXnet)) against openMSX 21.0;
this PR ports the device to current master (it compiled and ran unchanged).
It follows up on a conversation with one of the openMSX developers, who
suggested integrating it into openMSX itself and asked for two changes that
are included here: moving the generic socket plumbing into `Socket.cc`, and
replacing manual byte-shuffling with `Endian::UA_*`-typed structs.

All new code was written by me for this project and is contributed under
openMSX's GPL v2. The Z80 driver is maintained in the companion repo under the
same license.

## What's in the PR (2 commits)

1. **events: add small generic host-socket helpers to Socket** —
   `sock_setNonBlocking`, `sock_setIntOption`, `sock_readable`,
   `sock_makeIPv4`, `sock_localIPv4`. Generic host-socket plumbing; UnapiNet
   is the first user.
2. **Add UnapiNet** — `src/unapinet/{UnapiNet.cc,UnapiNet.hh,UnapiNetWire.hh}`,
   `share/extensions/unapinet.xml`, DeviceFactory registration, meson
   registration (sources, include dir, unit test) and a Catch2 unit test.

Build integration: the classic make build picks `src/unapinet` up
automatically via its source glob; meson is registered explicitly;
[...]

Issue #88, opened 2015-06-22
[Feature] Obsonet/DUMAS support in openMSX [sf#64]


openMSX-import:
**Reported by vampier on 2003-11-16 01:41 UTC**
This is not a big thing, I don't know even if it has any 
good applications for it (execpt Uzix)... but it would be 
nice to have network support in openMSX just for the 
fun of it.

openMSX-import:
**Commented by manuelbi on 2003-11-17 20:41 UTC**
Logged In: YES 
user_id=78178

There's no network adapter for MSX (yet).

openMSX-import:
**Updated by manuelbi on 2003-11-17 20:41 UTC**
- **status**: open --> wont-fix

openMSX-import:
**Commented by cdsg on 2006-04-15 18:14 UTC**
Logged In: YES 
user_id=1502833

Now we have an Ethernet adapter, Obsonet. And software that
can be run with it, IRC Client, FTP, TFTP, Http Server,
Telnet, etc..

openMSX-import:
**Commented by manuelbi on 2006-04-16 21:52 UTC**
Logged In: YES 
user_id=78178

Although Obsonet is hardly available, I'm still reopening
this feature request.

openMSX-import:
**Updated by manuelbi on 2006-04-16 21:52 UTC**
- **summary**: network (TCP/IP) support in openMSX --> Obsonet/DUMAS support in openMSX
- **status**: wont-fix --> open

openMSX-import:
**Commented by manuelbi on 2007-03-04 22:08 UTC**
Logged In: YES 
user_id=78178
Originator: NO

Note: there is emulation code for this in blueMSX, which could be reused...

openMSX-import:
**Commented by sd-snatcher on 2011-05-13 21:52 UTC**
ObsoNet emulation could help to solve this bug:

http://www.msx.org/forumtopicl12812.html

And since new batches of ObsoNets are available, adding this feature would be very welcome:
http://www.msx.org/ObsoNET2.newspost5799.html

openMSX-import:
**Commented by sd-snatcher on 2011-05-13 21:53 UTC**
If possible, add the ObsoNet2 support too (has some differences from the original one), line:

\- Has no mirroring
- Has IRQ support

frrfabio:
This feature will encourage the development of new games (and other softwares) using the network.

m9710797:
[...]

PR #2185, opened 2026-09-01
Network Device: SM-X WiFi

ducasp:
SM-X WiFi is a device designed by me that is living on every second generation OCM devices (like SM-X, SX2, etc) and also on quite a few first generation OCM devices (like MSX, hmmm, FPGA Book, TRH MSX, etc). It is basically three things:

1. Interface: I/O uart with large receiving (MSX receiving from ESP) FIFO, custom made
2. Connected to: ESP8266 or ESP32 with a custom firmware that provides TCP-IP UNAPI abstraction
3. Minimal ROM BIOS or RAM Driver: since the ESP already implements UNAPI over serial port, driver is just minimal, forward proper command and registers over serial, move the response to the correct registers and memory area 

Optimal way of using it is through ROM BIOS, and this is a driver to have this device along with UnapiNET. UnapiNET works with RAM Driver, and this one with ROM BIOS. The way to access ROM UNAPI is different so developers can benefit of having an alternative device to validate their applications work also with ROM UNAPI. Also, another difference is that SM-X WiFi allows the usage of an active TCP-IP connection using TLS/SSL (https).

This device will automatically emulate the ESP as well, for TCP-IP UNAPI emulation, and it has a RS232 connector. If the RS232 connector is plugged to a RS232Raw port and a valid comm port is opened, it will automatically stop using the emulated ESP and start using the comm port to try to communicate with a real ESP. This also is good for developers, that can use a separate device that will not have the limitations of the host device (i.e.: windows reserving SNTP port, Linux requiring binding for ports below 1000, firewalls configuration, etc)

Most of the code in here had received a first pass / review in the PR #2164, differences:

- RS232Raw has been already merged through a separate PR, #2174 
- A small addition was made to RS232Raw to figure out if the commport is connected to allow automatic changing from emulated ESP to ESP over serial
- Instead of having GenericUNAPI (fully emulated) and SMXWiFi (partially emulated, connected to real ESP over RS232Raw), we have a single SMXWiFi (since the underlying interface is the same) and the ESP emulation has moved to its own files, similar to RS232Raw, making it very simple for SMXWiFi to switch back and forth fully emulated/partially emulated and esp connected over serial
- Improvements based on first pass review and SonarQube report

Idea was to make 2 PRs, one for software only and other for hybrid, but in the end I thought it would be more beneficial to have a single device (so no code duplication) that is able to work both on hybrid or software only mode. This might be a little bit large, sorry, but really could not think on a good way to split into separate pull requests... (perhaps OpenSSL utils could be a separate pull request? but a bit weird since on its own it does nothing...), if you have an idea on a possible split, let me know :)

Before I forget, once approved, this should solve #88 , it is not Obsonet but also provides TCP-IP UNAPI/Network features which was what the requester really wanted. I believe that #2147 also solves #88....


ducasp:
About SonarQubeCloude review:

std::byte - In most cases it is using uint8_t to be inline with what most of openMSX code uses and following wouter advice. There are some instances where we are really manipulating 16 bits data in 16 bits data type, so those I do not see as an error or issue
std::thread being replaced by std::jthread - inline with most of openMSX code AND std::jthread won't compile for MacOS, this would require changing the tool set for MacOS bulding in github
pseudo random number generator - reviewed this before with wouter, using a little bit better option, still pseudo random, but for our purposes more than adequate
non-portable file path Windows.h - yeah, code on openmsx uses windows.h, following what it does... Also, seems that using Windows.h might cause issues with MINGW cross builds...
Rafactor of SSL to not have more than 20 fields - changed it so it has more of an api/interface like following advice, and it would be difficult and probably not worth the effort..
SMWWiFi adding default in switch / case line 104: really not needed, it is registered for IO 0x06 with a range of 2 (so 0x06 and 0x07) and the result port & 0x01  will be either 0 or 1...
EmulatedESP32.hh refactor to get down from 23 to 20 fields... Like refactor of ssl, this seems adequate for the functionality, do not see a need for this refactor

m9710797:
Thanks! I'll try to make time for a review soon, but it might take a few days.

About your comments on the automatic SonarQube review, I agree. Occasionally these automatic reviews are useful, but sometimes it's also just noise.

build/msvc/openmsx.vcxproj
```diff
@@ -116,7 +116,7 @@
...
...)\include;$(ThirdPartySrcDir)\$(LibNameLibPng);$(ThirdPartySrcDir)\$(LibNameSDL)\include;$(ThirdPartySrcDir)\$(LibNameSDL_ttf);$(ThirdPartySrcDir)\$(LibNameTheora)\include;$(ThirdPartySrcDir)\$(LibNameVorbis)\include;$(ThirdPartySrcDir)\$(LibNameOgg)\include;$(ThirdPartySrcDir)\$(LibNameTcl)\generic;$(ThirdPartySrcDir)\$(LibNameZlib);$(OpenMSXSrcDir);$(OpenMSXSrcDir)\3rdparty\ImGuiFileDialog;$(OpenMSXSrcDir)\3rdparty\imgui;$(OpenMSXSrcDir)\3rdparty\imgui_memory_editor;$(OpenMSXSrcDir)\cassette;$(OpenMSXSrcDir)\commands;$(OpenMSXSrcDir)\config;$(OpenMSXSrcDir)\console;$(OpenMSXSrcDir)\cpu;$(OpenMSXSrcDir)\debugger;$(OpenMSXSrcDir)\events;$(OpenMSXSrcDir)\fdc;$(OpenMSXSrcDir)\file;$(OpenMSXSrcDir)\ide;$(OpenMSXSrcDir)\imgui;$(OpenMSXSrcDir)\input;$(OpenMSXSrcDir)\laserdisc;$(OpenMSXSrcDir)\memory;$(OpenMSXSrcDir)\network;$(OpenMSXSrcDir)\resource;$(OpenMSXSrcDir)\security;$(OpenMSXSrcDir)\serial;$(OpenMSXSrcDir)\settings;$(OpenMSXSrcDir)\sound;$(OpenMSXSrcDir)\thread;$(OpenMSXSrcDir)\unapinet;$(OpenMSXSrcDir)\utils;$(OpenMSXSrcDir)\video;$(OpenMSXSrcDir)\video\scalers;$(OpenMSXSrcDir)\video\ld;$(OpenMSXSrcDir)\video\v9990;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
```
m9710797, line 119: Meta-question, not specific about this PR, but more general.

We recently added the `src/unapinet/` directory. Now we add `src/network`. Both only contain a handful of files. Maybe it's a good idea to unify these two? So rename `unapinet` to `network`?

This is a general question. And if we want this, then I don't know if we should do this as part of this PR or separate (and then before or after this PR).

doc/smxwifi.md
```diff
@@ -0,0 +1,284 @@
...
+  RS232 connector with `rs232-raw`, or
+* **an emulated ESP32** inside openMSX, when no real ESP32 is reachable.
+
+The device switches between the two automatically: while the RS232 connector
+has no `rs232-raw` device plugged in (or its host serial port is not open,
+e.g. no port selected), the MSX talks to the emulated ESP32. As soon as a
+valid `rs232-raw` selection is present and its port can be opened,
+communication is switched over to the real ESP32 and the current UART
+parameters are applied to the port. Switching in either direction closes all
+emulated network connections and starts the emulation clean.
```
m9710797, line 23: High level question about this automatic switching (I only just started the review I have not yet seen any code).

Is automatic switching a good idea. I understand it can be convenient. But can it also be misleading? Suppose, as a user, I want to use a real `ESP32`, but I make a mistake and because of that openMSX is using the emulated `ESP32`. Will I notice that mistake?  (E.g. maybe the emulation isn't perfect and therefor I do want to also test with a real `ESP32` but because of my mistake I always conclude it's working fine, the same as emulation).

doc/smxwifi.md
```diff
@@ -0,0 +1,284 @@
...
+UART status bits:
+
+| Bit | Meaning                                            |
+| --- | -------------------------------------------------- |
+| 0   | RX data available in the FIFO                      |
+| 3   | Quick receive supported                            |
+| 4   | Underrun (read of an empty FIFO), cleared by read  |
+
+Reading the data port waits a short time (about 30 ms) for the ESP module to
+provide data; if none arrives, `0xFF` is returned and the underrun bit is set.
```
m9710797, line 63: Is it useful to explain that "waits for a short time" means "insert wait cycles". At least to me this was surprising behavior. I don't know if your previous version of this document already contained this, but if it did, then I missed it (I mean I didn't realize it meant wait states).

Especially because `30 ms` is a long time (almost 2 NTSC VDP frames), that's atypical for wait cycles (more typical is just a few cycles).

doc/smxwifi.md
```diff
@@ -0,0 +1,284 @@
...
+**active** connections. The emulation uses the OpenSSL runtime library
+**installed on the host operating system**, which is loaded dynamically at
+startup.
+
+* If OpenSSL is found, the "_Use TLS in TCP active connections_" capability
+  bit is advertised and TLS connections can be opened. The OpenSSL version
+  is printed to the openMSX console at startup.
+* If OpenSSL is not found, the capability bit is **not** advertised and a
+  `TCPIP_TCP_OPEN` with the TLS flag returns `ERR_NOT_IMP`. A message is
+  printed to the openMSX console at startup.
```
m9710797, line 132: (I only started the review, I haven't read the code yet, maybe this comment is premature).

As a regular openMSX user, most of the time I won't be using any `ESP32` stuff. And then it would be annoying if I do get a warning about a missing `OpenSSL` library. Maybe for some users it's not obvious what that means? Is this important? What will go wrong because of this? ...  (I do understand that for Linux and macos(?) this library will practically always be found).

Also when I'm not using `ESP32`, Do I still need to pay the cost of searching/opening the `openSSL` library? It's probably only a few(?) milliseconds. But maybe it's an idea to only try to load this library the first time it's needed? That would also "solve" my concern about the warning. Then you only see that warning when it actually matters.

doc/smxwifi.md
```diff
@@ -0,0 +1,284 @@
...
+  privileged like port 21 (see below).
+
+  For a quick test the firewall can be disabled temporarily with
+  `sudo ufw disable` and re-enabled with `sudo ufw enable`. iptables- or
+  nftables-based firewalls need equivalent rules.
+
+* **Privileged ports:** on Linux and macOS, binding a port below 1024
+  requires root privileges. If an MSX application uses such a port (for
+  example an FTP server on port 21), either run openMSX as root, or grant
+  the capability once so regular users can bind these ports:
```
m9710797, line 270: I would discourage running openMSX as root. Maybe leave out that sub-sentence? So only recommend the `setcap` stuff?

src/utils/OpenSSL.hh
```diff
@@ -0,0 +1,116 @@
...
+};
+
+// Handle to the loaded OpenSSL runtime. Obtained from load(); its members
+// may only be called after load() succeeded.
+struct LibHandle {
+	// Version string, e.g. "OpenSSL 3.1.4 24 Oct 2023".
+	[[nodiscard]] zstring_view version() const;
+
+	// Description of the last OpenSSL error (for console diagnostics).
+	[[nodiscard]] zstring_view last_error() const;
```
m9710797, line 91: Maybe we should follow the example from `SerialPort` here.
* In `SerialPort` functions that can return an error do immediately return that error code. So in this context that means they immediately call `api.errGetError`, and then we need to make the availability of this function mandatory?
* And then instead of some `last_error()` function we have function like `std::string to_string(ErrorCode)`.

That eliminates the possible mistake of "last" error being replaced by another "last" error before we got a change to query that "previous last" error.

Also because errors should be rare, an performance in the error path is less important, I'd suggest to return the string by value (no `zstring_view` that can become dangling or gets overwritten on the next call).

src/utils/OpenSSL.hh
```diff
@@ -0,0 +1,116 @@
...
+struct SessionHandle {
+	SessionHandle(const SessionHandle&) = delete;
+	SessionHandle& operator=(const SessionHandle&) = delete;
+	SessionHandle(SessionHandle&& other) noexcept;
+	SessionHandle& operator=(SessionHandle&& other) noexcept;
+	~SessionHandle();
+
+	// Drives the TLS handshake. Returns 1 when completed, 0 when it needs
+	// read readiness, 2 when it needs write readiness, -1 on failure.
+	[[nodiscard]] int handshake() const;
```
m9710797, line 57: I assume these values `0`, `1`, `2`, `-1` come from the underlying library? Could we instead return some enum? You can still map the enum-items to the same numeric value.

src/utils/OpenSSL.hh
```diff
@@ -0,0 +1,116 @@
...
+
+	// Certificate validation result after a completed handshake (only
+	// meaningful when the session was created with verify=true). Returns
+	// 0 when the certificate is valid, otherwise a TCP-IP UNAPI close
+	// reason code in the range 9..19 (spec 4.5.4).
+	[[nodiscard]] int verifyResult() const;
```
m9710797, line 63: I admit I don't fully understand this API. Does `0` mean success, and the other values mean some kind of error code? I mean does a non-zero value always need some special handling? In that case it might be better to change the return type to `std::expected<void, int>`?

src/utils/OpenSSL.hh
```diff
@@ -0,0 +1,116 @@
...
+	// WouldBlock, or Closed on a clean TLS close (close_notify).
+	[[nodiscard]] IoResult read(std::span<char> buf) const;
+
+	// Encrypts and sends data. Success: number of bytes written. Failure:
+	// WouldBlock, or Closed when the peer sent close_notify.
+	[[nodiscard]] IoResult write(std::span<const char> buf) const;
+
+	// Plaintext bytes already decrypted and buffered by the SSL layer,
+	// readable without waiting for socket readiness.
+	[[nodiscard]] int pending() const;
```
m9710797, line 75: Since both `read()` and `write()` return `size_t` to indicate the number of bytes, maybe `pending()` should do the same?  It could be that the underlying library only supports 32-bit sizes, but maybe we can still offer a more uniform API? (Or alternatively also return `int` from `read()` and `write()`, but my personal preference is `size_t`.)

src/utils/OpenSSL.hh
```diff
@@ -0,0 +1,116 @@
...
+struct X509StoreCtx;
+struct SslInitSettings;
+
+// Result of a (non-blocking) read() or write(): success carries the
+// number of bytes transferred; on failure IoError says why.
+enum class IoError {
+	WouldBlock, // the call would block; retry later (either readiness)
+	Closed,     // clean TLS close (peer sent close_notify)
+	Failed      // real error
+};
```
m9710797, line 42: See the discussion on `last_error()` vs `to_string(ErrorCode)`. Maybe this `IoError` should then be changed?

src/utils/OpenSSL.hh
```diff
@@ -0,0 +1,116 @@
...
+
+// Opaque handles for OpenSSL runtime types, kept forward-declared so that
+// openMSX itself never needs the OpenSSL development headers.
+struct SslCtx;
+struct Ssl;
+struct X509;
+struct X509VerifyParam;
+struct SslMethod;
+struct X509StoreCtx;
+struct SslInitSettings;
```
m9710797, line 34: Some of these forward declarations aren't used in this `.hh` file. Do we need them? Can't we remove them? Or move them to the `.cc` file?

src/utils/OpenSSL.cc
```diff
@@ -0,0 +1,562 @@
...
+	return dlopen(name, RTLD_NOW | RTLD_LOCAL);
+}
+constexpr auto libraryCandidates = std::to_array<const char*>({
+	// Homebrew (Apple Silicon / Intel) and system-wide installs
+	"/opt/homebrew/opt/openssl@3/lib/libssl.3.dylib",
+	"/opt/homebrew/opt/openssl@1.1/lib/libssl.1.1.dylib",
+	"/usr/local/opt/openssl@3/lib/libssl.3.dylib",
+	"/usr/local/opt/openssl@1.1/lib/libssl.1.1.dylib",
+	"/usr/local/lib/libssl.dylib",
+	"/opt/homebrew/lib/libssl.dylib",
```
m9710797, line 51: I already commented on this in an earlier review:

I don't like that on Linux we try these macos specific libraries. We know those cannot work. Isn't it easy to put them in some `#ifdef`?

And maybe similar the names below, can they ever work on macos. Ah wait I see now that below are also some `.dylib` values.

Shouldn't we have a list of only `.dylib` on macos, and another list of only `.so.*` on Linux? I'm not sure I'm not familiar with macos, but I've never seen `.dylib` on Linux.

src/utils/OpenSSL.cc
```diff
@@ -0,0 +1,562 @@
...
+	long (*getVerifyResult)(const Ssl* ssl) = nullptr;
+	X509* (*get1PeerCertificate)(const Ssl* ssl) = nullptr;
+	void (*freeX509)(X509* x) = nullptr;
+	unsigned long (*errGetError)() = nullptr;
+	void (*errErrorStringN)(unsigned long e, char* buf, size_t len) = nullptr;
+	const char* (*version)(int type) = nullptr;
+};
+
+// All mutable runtime state, held in a function-local singleton so there are
+// no non-const namespace-scope variables (SonarCloud cpp:S1132).
```
m9710797, line 112: I'm not sure I agree with this SonarCloud comment. It does complicate the code, and I don't know if it's for a good reason.

In general it is good advice to avoid non-const global state. But in this specific case there really is only one global OpenSsl library. And we really only attempt to load it once (and return the same already loaded value on a 2nd attempt). So IMHO that's an acceptable use of global variables. We could still document that calling `OpenSSL::load()` is only allowed from the main thread, or at least cannot be done concurrently from multiple threads.

For a clean solution we could make `OpenSSL::load()` not return a plain pointer, but some reference counted RAII object. And then when the last user goes away we close (unload) the library again. But i don't think that's worth the trouble in our case.

---

UPDATE:
Now that I've seen more of the code ... can't these variables live inside the `LibHandle` struct? That is already a singleton (so no need to here make another `Runtime` singleton?)

src/utils/OpenSSL.cc
```diff
@@ -0,0 +1,562 @@
...
+#ifdef _WIN32
+	// The Windows system store provides the full root CA list
+	loadWindowsTrustStore(ctx);
+#else
+	// macOS keeps the root certificates at a well-known location that is
+	// not part of OpenSSL's default search path (libressl uses /etc/ssl
+	// with a symlink, but Homebrew builds do not)
+	if (api.ctxLoadVerifyLocations) {
+		api.ctxLoadVerifyLocations(ctx, "/etc/ssl/cert.pem", nullptr);
+	}
```
m9710797, line 223: You have windows specific code (line 216), fine. Then this comments makes me think lines 221-223 are macos specific?

But then I wonder "what about Linux", or are these lines also fine (and needed) for Linux?

---

A second thing I wonder applies to many locations in this file. But I'll only make this comment once. (But do interpret it in general, not only this specific case).

We test that `api.ctxLoadVerifyLocations` is not `nullptr`, and only if it's not call the function. So apparently it's fine if this function is missing? But then why call it if it's available?

For functions that are truly optional we should document why it's OK to only optionally call them. But to be honest I think we should make most functions mandatory. So refuse to load the library when they are missing. (Again interpret this as a general remark, not only for `ctxLoadVerifyLocations`).

EDIT: sorry I had to add more examples:
In `getSharedContext()` below we test for `api.ssl23ClientMethod` and `api.ctxNew`. If either is missing `getSharedContext()` returns `nullptr`. But if it returns `nullptr` then `createCientSession()` cannot return a valid `SessionHandle`. And then this library is basically unusable. So again I'd say: make most (all?) functions mandatory.

It will also simplify the code. Also because right now it's impossible to test all those error handling code paths.

src/utils/OpenSSL.cc
```diff
@@ -0,0 +1,562 @@
...
+			return p;
+		}
+	}
+	return nullptr;
+}
+
+// Name of the libcrypto DLL that belongs to a libssl DLL
+std::string cryptoNameFor(const char* sslName)
+{
+	std::string n = sslName;
```
m9710797, line 280: If you need `sslName` as a `string` (not a `const char*`), then maybe pass it as a `string` parameter (by value).

src/utils/OpenSSL.cc
```diff
@@ -0,0 +1,562 @@
...
+bool openLibrary()
+{
+	for (const char* name : libraryCandidates) {
+		if (auto h = openSslLib(name)) {
+			runtime().libHandle = h;
+			// Load the matching libcrypto as well: the X509/ERR entry
+			// points are exported by libcrypto, and GetProcAddress does
+			// not search DLL dependencies (best effort; the TLS session
+			// functions themselves all live in libssl).
+			if (auto ch = openSslLib(cryptoNameFor(name).c_str())) {
```
m9710797, line 357: Maybe this is correct (does lifetime extension kick-in here?), but I'm always unsure:
* `cryptoNameFor()` returns a `std::string`.
* This is a temporary variable (we don't assign it to any named variable).
* We call `.c_str()` on this temporary. We pass that to `openSslLib()`.
* At some point the temporary (unnamed) variable goes away, then the `.c_str()` reference to it becomes invalid.
* I don't know at what point the temporary goes away. Is that before or after the call to `openSslLib()`?

Maybe I should investigate this, so that I know if this construct is safe or now. What I usually do instead is split this line in two:
* First assign the result of `cryptoNameFor()` to a (named) local variable. 
* Then in a second statement use `.c_str()` on that local variable.

At least to me it's then clear when that local variable goes away. But maybe I should investigate and learn the C++ rules better, is your version fine or not?

src/utils/OpenSSL.cc
```diff
@@ -0,0 +1,562 @@
...
+		return 10; // invalid server certificate
+	}
+}
+
+IoResult SessionHandle::read(std::span<char> buf) const
+{
+	const OpenSSLApi& api = runtime().api;
+	using enum IoError;
+	if (!api.read || !api.getError) return std::unexpected(Failed);
+	int r = api.read(ssl, buf.data(), static_cast<int>(buf.size()));
```
m9710797, line 475: I said I would only comment about this once, but sorry i can't help it.

If `api.read` or `api.getError` are missing then this whole `openmsx::OpenSSL` stuff is basically unusable. So please make the presence of these functions mandatory.

src/utils/OpenSSL.cc
```diff
@@ -0,0 +1,562 @@
...
+	if (r > 0) return IoResult(r);
+	int e = api.getError(ssl, r);
+	if (e == SSL_ERROR_WANT_READ || e == SSL_ERROR_WANT_WRITE) {
+		return std::unexpected(WouldBlock);
+	}
+	if (e == SSL_ERROR_ZERO_RETURN) {
+		return std::unexpected(Closed); // clean TLS close
+	}
+	return std::unexpected(Failed);
```
m9710797, line 484: I already hinted at this before, but now I can more clearly see it:

I think instead of returning a general `Failed` status code, we should return the actual error (the error code that can then be passed to `to_string(ErrorCode)`. And here I see that this error is already available via `getError()`. And then we don't need to re-query the error later via `errGetError()`.

Same for `write()`.

src/network/SMXWiFi.hh
```diff
@@ -0,0 +1,93 @@
...
+
+	// Whether the real serial endpoint is currently active (tracked to
+	// detect endpoint switches in updateEndpoint()).
+	bool realEndpointActive = false;
+
+	// ROM at 0x4000-0x7FFF (ESPUNAPI_IO.rom)
+	Rom rom;
+
+	// Cached CPU reference for emulated-time waits (VDPIODelay pattern)
+	MSXCPU& cpu;
```
m9710797, line 88: Detail: something to consider, you judge if it's worthwhile.

Currently you have the member variables in this order:
```
mutex
bool             ; followed by 3 bytes padding
unsigned    ; 0 or 4 bytes padding
unique_ptr
bool            ; 7 bytes padding
Rom
MSXCPU&
```

If you instead sort the member variables from large to small (actually larger to smaller alignment) then most of the padding disappears
```
MSXCPU&
Rom
unique_ptr
mutex
unsigned
bool
bool
```

But there are other reasons to order the member variable in a particular way. An important reason is to keep related variables together (for understandability), or because some variables need to be initialized in a particular order in the constructor initializer list.

So you decide what order you want them (I'm only saying from an internal padding point of view it's sub-optimal, but there might be more important reasons).

src/network/SMXWiFi.cc
```diff
@@ -0,0 +1,362 @@
...
+		config, [this](uint8_t b) { pushToFifo(b); });
+}
+
+SMXWiFi::~SMXWiFi() = default;
+
+// ====================================================================
+//  Memory-mapped ROM
+// ====================================================================
+
+uint8_t SMXWiFi::readMem(uint16_t address, EmuTime /*time*/)
```
m9710797, line 38: Usually if you implement `readMem()` you also want `peekmem()` (to make the device visible in the openMSX debugger).

In this specific case you might be lucky because the default implementation for `peekMem()` is based on `getReadCacheLine()` and that happens to already do the correct thing.

Nevertheless may I suggest this structure:
```
const uint8_t* SMXWiFi::getReadCacheLine(uint16_t start) const
{ ... exactly as you have it now ...}

uint8_t SMXWifi::peekMem(uint16_t address) const
{
    return *getReadCacheLine(address);
}

uint8_t SMXWifi::readMem(uint16_t address, EmuTime /*time*/)
{
    return peekMem(address);
}
```
That makes it very clear that all 3 are in sync. Now I had to verify that `readMem()` and `getReadCacheLine()` are indeed consistent. And I had to lookup how the default `peekMem()` behaves.

src/network/SMXWiFi.cc
```diff
@@ -0,0 +1,362 @@
...
+
+void SMXWiFi::powerUp(EmuTime time)
+{
+	reset(time);
+}
+
+void SMXWiFi::powerDown(EmuTime /*time*/)
+{
+	emulated->stop();
+}
```
m9710797, line 62: Maybe this is correct. But because it's unusual I'd ask anyway:

* Not many devices need `powerDown()` (the once that do are typically devices with LEDs that need to be turned off). Instead most devices do cleanup instead in their destructor. I wonder if here that also wouldn't be a better choice.
* The default implementation for `powerUp()` already calls `reset()`, so technically the current implementation isn't needed. However for symmetry if you implement `powerDown()` it might be logical to also implement `powerUp()`, but then I'd also expect the implementation to be symmetrical, so something like `emulated->start()` (in powerUp) and `emulated->stop()` (in powerDown). (But see also previous point about maybe not needing powerDown, and then I'd also remove powerUp).

src/network/SMXWiFi.cc
```diff
@@ -0,0 +1,362 @@
...
+	uint8_t v;
+	if (tryPop(v)) return v;
+
+	// FIFO empty: give the producer (the emulated ESP32 thread or a real
+	// ESP over the serial port) a short chance to provide data by
+	// stalling only the CPU in emulated-time slices — the rest of the
+	// machine keeps running.
+	auto deadline = time + FIFO_TIMEOUT;
+	while (time < deadline) {
+		cpu.wait(time + FIFO_POLL);
```
m9710797, line 242: Here we stall the *emulated* time, good. But I assume we're (possibly) waiting for a real `ESP32`? Then shouldn't we also add some `sleep()` in *real time*?

I mean if via this loop we advance the emulated time 50ms (on a very fast host machine), then most likely not much real time has passed, and then the wait in emulated time was for nothing.

m9710797 (commented):
I started reviewing, but then ran out of time. I'll try to continue one of the following days (hopefully tomorrow).
* I did not review `EmulatedEsp32` at all yet.
* I only had a quick look at `SMXWiFi`.
* The other code is reviewed in reasonable detail (for a first pass).

build/msvc/openmsx.vcxproj
ducasp: Yeah, I was thinking about it, I can move unapinet components to NETWORK, I think it makes lots of sense given the structure of the project, keep all network related components in the same filter

src/network/EmulatedEsp32.cc
```diff
@@ -0,0 +1,2441 @@
...
+	case 13: case 14: case 15: case 16:        // TCP OPEN/CLOSE/ABORT/STATE
+	case 17: case 18:                          // TCP SEND/RCV
+	case 25: case 26:                          // CONFIG_AUTOIP, CONFIG_IP
+	case 206:                                  // TCPIP_DNS_Q_NEW
+		return true;
+	default:
+		break;
+	}
+	if (cmd >= 129 && cmd <= 143) return true; // SSH UNAPI
+	if (cmd >= 200 && cmd <= 202) return true; // HTTP client
```
m9710797, line 75: Very much a detail, coding style, personal preference.

An alternative is:
```
if (cmd == one_of(1, 2, 3, 6,       // GET_CAPAB, GET_IPINFO, NET_STATE, DNSQ,
                  8, 9, 10, 11, 12, // ...
                  ...
                  206) {            // TCPIP_DNS_Q_NEW
    return true;
}
if (129 <= cmd && cmd <= 143) return true;
...
```
Notice `cmd >= 129` vs `129 <= cmd` because this is closer to the mathematical syntax: `129 <= cmd <= 143` (and there are proposals to allow this mathematical syntax in future c++ versions).

Even better would be if you define constants like:
```
static constexpr uint8_t GET_CAPAB = 1;
static constexpr uint8_t GET_IPINFO= 2;
...
```
and then use those, and drop the comments (because now those comments basically define those constants in an ad-hoc way).

You could also make this function `[[nodiscard]]`.
This comment applies to both `isQuickCustomCommand()` and `isDataCommand()`.

src/network/EmulatedEsp32.cc
```diff
@@ -0,0 +1,2441 @@
...
+		auto v = StringOp::stringToBase<10, unsigned>(rest.substr(0, dot));
+		if (!v || *v > 255) return false;
+		octets[n++] = *v;
+		if (dot == std::string_view::npos) break;
+		rest.remove_prefix(dot + 1);
+	}
+	if (n != 4) return false;
+	ipOut = (octets[0] << 24) | (octets[1] << 16) | (octets[2] << 8) | octets[3];
+	return true;
+}
```
m9710797, line 124: The `UnapiNet.cc` code has the same functionality. We could consider factoring out common code (maybe put generic network helper functions in `Socket.hh`, or a new header).

OTOH, if I understood correctly the implementation is only a few lines:
```
struct in_addr addr;
if (inet_pton(AF_INET, hostname.c_str(), &addr) == 1) { ...
```
So maybe not worth factoring out?

doc/smxwifi.md
ducasp: I think that from a user point of view, there are a few ways to verify it. None of them are directly a report of EMULATED, but you can figure it out:

One is just a matter of checking if Serial RS232 Connector is Raw and has a port selected:

image

Other one, at Wi-Fi BIOS Level (F1 pressed while booting), the AP is a mock one:

image

While with ESP it will be your real AP:

image

Using term8266.com, you can type R to reset the module and see the firmware welcome message, which also tells it is an emulated device:

image
image

The final one is the flag to disconnect the network link, which works only on the emulated device, if you type TCPIP S usually you see network open:

image

You can uncheck the box:
image

Real device is kept OPEN:
image

Emulated device shows CLOSED:
image

One note is, if you choose a COM port and it doesn't have an ESP in the other side with the proper firmware, it will work exactly as in real life, and tell ESP is disconnected and nothing will work, so when a COM port is selected it will never fall back to the emulated device, unless you unplug the RS232Raw from the connectors...

Would you think that this could be solved by documenting those ways of checking? Or are you thinking on something different?

src/network/EmulatedEsp32.cc
```diff
@@ -0,0 +1,2441 @@
...
+
+#ifndef _WIN32
+// Read the first two DNS servers from /etc/resolv.conf
+static void readResolvConfDns(uint32_t& dns1, uint32_t& dns2)
```
m9710797, line 128: Detail:
This helper function is only used in the else-part of the `#ifdef _WIN32 / #else / #endif` block below. Consider moving this function inside that block. This then requires fewer `#ifdef _WIN32` blocks (and the helper functions are closer to their use). Because now I was searching for "how is this implemented for windows".

src/network/EmulatedEsp32.cc
```diff
@@ -0,0 +1,2441 @@
...
+{
+	std::ifstream f("/etc/resolv.conf");
+	std::string line;
+	int n = 0;
+	while (n < 2 && std::getline(f, line)) {
+		auto pos = line.find_first_not_of(" \t");
+		if (pos == std::string::npos) continue;
+		std::string_view entry(line);
+		entry.remove_prefix(pos);
```
m9710797, line 137: I'm not sure if it really helps, but you can consider using:
```
StringOp::trim(entry, " \t");
```
to remove whitespace from both ends of the `string_view`.

src/network/EmulatedEsp32.cc
```diff
@@ -0,0 +1,2441 @@
...
+		L"wsl", L"vethernet", L"hyper-v", L"virtual", L"vmware",
+		L"virtualbox", L"vbox", L"tailscale", L"zerotier",
+		L"docker", L"tap-", L"isatap", L"teredo"
+	});
+	std::wstring lower;
+	for (auto c : std::wstring(name)) {
+		lower.push_back(std::towlower(c));
+	}
+	return std::ranges::any_of(markers, [&](const auto& m) {
+		return lower.find(m) != std::wstring::npos;
```
m9710797, line 171: `lower.contains(m)`

src/network/EmulatedEsp32.cc
```diff
@@ -0,0 +1,2441 @@
...
+	// Pick the preferred adapter: the highest-scoring one with a usable
+	// IPv4 address (ties keep the first one).
+	int bestScore = -1;
+	std::optional<HostNetInfo> bestInfo;
+	for (auto* a = adapters; a; a = a->Next) {
+		int score = adapterScore(*a);
+		if (score < 0 || score < bestScore) continue;
+		auto info = hostNetInfoFromAdapter(*a);
+		if (!info) continue;
+		if (score > bestScore) {
```
m9710797, line 246: I think I'm missing something:
* On line 243, when `score == bestScore`, then
* On line 244 we call `hostNetInfoFromAdapter()`
* But then on line 246 we skip because `score` is not strictly better.

Then why query `hostNetInfoFromAdapter()`? Shouldn't line 243 be `score <= bestScore`, and then drop the test on line 246?

Same remark for line 312, 322.

src/network/EmulatedEsp32.cc
```diff
@@ -0,0 +1,2441 @@
...
+		if (line.starts_with("Iface")) continue; // header
+		std::istringstream iss(line);
+		std::string iface;
+		std::string dest;
+		std::string gw;
+		std::string flags;
+		iss >> iface >> dest >> gw >> flags;
+		if (dest != "00000000") continue; // not the default route
+		if (auto f = StringOp::stringToBase<16, unsigned long>(flags);
+		    !f.has_value() || !(*f & 1)) continue;
```
m9710797, line 270: This line (test on `flags`) could use a short comment. What is the meaning of bit 0?

src/network/EmulatedEsp32.cc
```diff
@@ -0,0 +1,2441 @@
...
+
+// ---- connection structure ----
+
+// Passive TCP listener registry entry: multiple passive connections may
+// share a listener for the same local port; the socket is owned by the
+// registry and closed when the last referencing connection is freed.
+struct EmulatedEsp32::ListenSocket {
+	SOCKET sock = OPENMSX_INVALID_SOCKET;
+	uint16_t port = 0;
+	int refs = 0;
```
m9710797, line 358: Instead of manually implementing reference counting, can't you use `std::shared_ptr` (instead of ``std::unique_ptr<ListenSocket>`)?

Your code is a tiny bit more efficient because it's an intrusive reference count and it's not thread-safe (you can't shared across threads). But that shouldn't matter here?

doc/smxwifi.md
ducasp: While really atypical, also, really not what happens.... Most of the time one character arrives by 12.5us at 80KBps (800Kbps), or, when we talk about USB to Serial adapters, it generally is about 16ms because they like to batch data to fill a report, so there is a chance that a few bytes arrive at the end of a transfer, doesn't fill a report, and it waits up to 16ms before sending the report with the data it has... 

I agree that documenting this properly is a good way of doing it. Not sure about many adapters, but this approach has been used on the FPGA devices succesfully for quite a while. While in one hand this might seem as something that might slow down the whole thing, UNAPI call is blocking for READs, and considering the general cases:

1. Small command: the  bios/driver always check if there is fifo data for all commands at least until the data size from protocol is determined, thus, for those commands, this will never be hit

2. RCV commands (where the QuickRCV is implemented) for TCP or UDP or SSH: what is said on 1 still is true, and if just a small quantity of data, like up to a couple dozen bytes, it will all be grouped in one USB report and not hit the WAIT cycle... If a large block of data, usually we are receiving blocks of 1KB of size, and this is where the advantage comes, instead of doing IN to check if data, IN to get data, you just go... On slower regular z80 speeds the gains are quite high and the CPU would not be doing anything anyway when it is in WAIT state (perhaps this can delay the interrupt routine to enter, but in applications like FTP/TFTP/ HGET or even Telnet when receiving large screen updates the gain is very sensible. The halt cycle might be hit a few times

Also, other than documenting this, if you think it is worth it, the BIOS is ready for both scenarios, quick receive being supported or not, perhaps like the LINK setting, we can have a QuickRCV setting to allow it to be disabled? The device itself reports when it is able to work with QuickRCV or not, so it is a possibility.

src/network/EmulatedEsp32.cc
```diff
@@ -0,0 +1,2441 @@
...
+		recvBuffer.push_back(b);
+	}
+	[[nodiscard]] size_t recvAvail() const {
+		std::scoped_lock lock(recvMutex);
+		return recvBuffer.size();
+	}
+	void popRecv(std::vector<uint8_t>& out, size_t max) {
+		std::scoped_lock lock(recvMutex);
+		size_t toRead = std::min(recvBuffer.size(), max);
+		out.reserve(2 + toRead);
```
m9710797, line 381: Why `+ 2`? From only reading this method that's not clear, so maybe add a short comment.

doc/smxwifi.md
ducasp: The message goes on top, either that you have or that you don't have SSL support, for a brief period, when connecting the device only:
image

So the check is done only on device plugging and not anymore.

Not having OpenSSL doesn't allow you to use SSL with the emulated ESP, so HGET to HTTPS addresses won't work. Nothing will go wrong, you lose a capability... I think it should be fine not having the message about SSL being available (or not), thought on the message more on the way of perhaps encouraging the user to read the documentation if he wants https and see that message, but also see your point of view, most users won't care about it so the message every plug might be excessive. 

About loading it only when needed, this would cause a conflict, because then we might run into a scenario where the adapter is advertising to have SSL capability, HGET will try to open a secure connection, and only then we will try to find it and load it... And if it is not available, we already told we have it. So instead of a nice message telling the adapter doesn't support SSL, you have another failure not related (even if returning NOT_IMP hget will not tell not available because it rely on the advertised capability). 

Let me know your thoughts after reading this (the cost being only when plugged, the possibility of having an hget error that is not exactly telling what is hapenning). I consider that removing the message should be done (existing or not existing SSL), will wait your input for the way to handle SSL finding / availability.

src/network/EmulatedEsp32.cc
```diff
@@ -0,0 +1,2441 @@
...
+		for (size_t i = 0; i < toRead; ++i) {
+			out.push_back(recvBuffer.front());
+			recvBuffer.pop_front();
+		}
```
m9710797, line 385: Small performance tweak. Currently we move byte-per-byte. We could do the whole block at once with something like:
```
out.assign(recvBuffer.begin(), recvBuffer.begin() + toRead);
```
Then you also don't need the call to `out.reserve()`. (This version replaces the content of `out`, if that's wrong then instead use `out.insert(out.end(), ...)`).

But then we still need a loop for `revcBuffer.pop_front()`. We can get rid of that loop if we add a small extension: `cb_queue::remove_prefix(n)` (similar to std::string_view::remove_prefix).

Feel free to implement or ignore this performance tweak.

doc/smxwifi.md
ducasp: Ok, will remove it :)

src/utils/OpenSSL.hh
ducasp: Will check about the return type. That information is part of TCP IP UNAPI specification, the reason why the SSL connection was not opened, informing what was the error (i.e.: no certificate shown, expired certificate, certificate not trusted by the truststore, etc..). So the non-zero value doesn't need special handling, it is just extra information on why the secure connection to the remote end failed

src/network/EmulatedEsp32.cc
```diff
@@ -0,0 +1,2441 @@
...
+
+	// TLS (TCP active connections, TCP-IP UNAPI spec 1.1): opaque OpenSSL
+	// session. The handshake runs in the reader thread; the connection is
+	// only considered ESTABLISHED once it has completed (spec: TCP_STATE
+	// must not report ESTABLISHED before the handshake is finished).
+	std::optional<OpenSSL::SessionHandle> ssl;  // null when not using TLS
+	bool tlsVerify = false;     // validate the server certificate
+	std::atomic<uint8_t> handshakePhase{0}; // 0=no TLS, 1=handshaking, 2=done
+
+	std::unique_ptr<std::thread> readerThread;
```
m9710797, line 419: The combination of `unique_ptr` and `thread` is weird. And I also don't think it's needed.

* `std::thread` has a default constructor that creates a null-thread (no thread, not running).
* You can assign a running thread to an object holding a null-thread.
* And you can query `joinable()` on a null-thread (it will return `false`).

And I think that's all you need here. So drop the `unique_ptr`.

src/network/EmulatedEsp32.cc
```diff
@@ -0,0 +1,2441 @@
...
+	std::unique_ptr<Poller> poller;
```
m9710797, line 420: Similar remark here. I think we can drop the `unique_ptr` and instead have a default-constructed Poller object which we re-initialize via `Poller::reset()`.

HOWEVER. I think `poller` isn't really used at all (we never call `Poller::poll()` only `Poller::abort()` and `Poller::reset()`, but without `Poller::poll()` that's useless). So I'd suggest to remove `poller` completely.

src/network/EmulatedEsp32.cc
```diff
@@ -0,0 +1,2441 @@
...
+			cp->readerActive = false;
+			if (cp->poller) cp->poller->abort();
+			if (cp->readerThread && cp->readerThread->joinable()) {
+				cp->readerThread->join();
+			}
+			cp->readerThread.reset();
+			cp->poller.reset();
+	if (cp->ssl) {
+		cp->ssl.reset();
+	}
```
m9710797, line 486: Detail: lines 484-486 have wrong indentation.

src/network/EmulatedEsp32.cc
```diff
@@ -0,0 +1,2441 @@
...
+// ====================================================================
+
+// Boot sequence text to the MSX side (greeting, "Ready"): delivered
+// byte-by-byte through the sink, like RS232Raw delivers bytes from the
+// real ESP32 through the connector.
+void EmulatedEsp32::pushBootText(std::string_view text) const
+{
+	for (char c : text) {
+		sink(static_cast<uint8_t>(c));
+	}
```
m9710797, line 581: Maybe it's worth changing the interface of `sink` to accept a `std::span<const uint8_t>` instead of a single byte. I also checked the rest of the code and most often we want to send a whole block (not a single byte).

The immediate advantage is then that we don't acquire/release the `mutex` in `SMWWiFi` multiple times. 

A follow-up optimization could be to extend `cb_queue` to accept multiple items at once. But it's OK if you don't do that (as part of this PR).

src/network/EmulatedEsp32.cc
```diff
@@ -0,0 +1,2441 @@
...
+	if (haveClient) {
+		maxFd = std::max(maxFd, static_cast<int>(conn->clientSock));
+	}
+	if (int sel = select(maxFd + 1, &rfds,
+	                     inHandshake ? &wfds : nullptr, nullptr, &tv);
+	    sel <= 0) {
+#endif
+		// Nothing to read yet (or transient select error); re-check the
+		// shutdown flag and try again
+		std::this_thread::sleep_for(std::chrono::milliseconds(5));
```
m9710797, line 826: I haven't studied the code in sufficient detail yet (I can't claim to understand most of it). But this `sleep_for()` in combination with `select()` is unexpected.

Can't `select()` itself wait for a certain time?  (Actually the point of `select` is that you can eliminated busy-wait loops (without or without sleep) from your code).

src/network/EmulatedEsp32.cc
```diff
@@ -0,0 +1,2441 @@
...
+bool EmulatedEsp32::tcpReadData(Connection* conn, int connIdx)
+{
+	std::array<char, 1024> buf;
+	ptrdiff_t n = 0;
+	if (conn->ssl) {
+		// TLS: decrypt; a clean TLS close is a close_notify, a real
+		// error fails the connection, WouldBlock is just "try again
+		// later".
+		auto r = conn->ssl->read(
+			std::span<char>(buf.data(), buf.size()));
```
m9710797, line 928: It should be possible to directly pass a `std::array` to a function expecting a `std::span`. So it should be possible to simplify this to:
```
auto r = conn->ssl->read(buf);
```

Same on line 970 (and maybe others).

src/network/EmulatedEsp32.cc
```diff
@@ -0,0 +1,2441 @@
...
+			// Socket closed by the remote side (EOF) or error
+			conn->state = 3; // remote closed -> CLOSE_WAIT
+			return true;
+		}
+	}
+	if (n > 0) {
+		{
+			for (ptrdiff_t i = 0; i < n; ++i) {
+				conn->pushRecv(buf[static_cast<size_t>(i)]);
+			}
```
m9710797, line 954: Similar remark as above: currently this takes/releases the mutex for each byte. We should avoid that. E.g. via an API that takes a buffer (a `std::span`) instead of a single byte.

Same for line 982 and 1001 (and maybe others).

src/network/EmulatedEsp32.cc
```diff
@@ -0,0 +1,2441 @@
...
+			if (r.error() == OpenSSL::IoError::Closed) {
+				conn->state = 3; // remote closed -> CLOSE_WAIT
+			}
+			if (r.error() == OpenSSL::IoError::Failed) {
+				tcpTlsFail(conn, connIdx, 19);
+			}
+			// WouldBlock: nothing more to drain
+			return;
+		}
+		for (ptrdiff_t i = 0; i < *r; ++i) {
```
m9710797, line 982: This triggers a compiler warning:
```
src/network/EmulatedEsp32.cc: In member function ‘void openmsx::EmulatedEsp32::tcpDrainSsl(Connection*, int)’:
src/network/EmulatedEsp32.cc:982:41: warning: comparison of integer expressions of different signedness: ‘ptrdiff_t’ {aka ‘long int’} and ‘long unsigned int’ [-Wsign-compare]
  982 |                 for (ptrdiff_t i = 0; i < *r; ++i) {
      |                                       ~~^~~~
```

Before I suggested to drop this loop. But if you need to keep this loop for some reason you could avoid these type mismatches with a construction like:
```
for (auto i : xrange(*r)) { ...
```
Now (because of `auto`) the compiler deduces the correct type for `i`. So we don't have to know (and don't care) if `i` should be `ptrdiff_t` or `size_t`.

src/network/EmulatedEsp32.cc
```diff
@@ -0,0 +1,2441 @@
...
+		}
+		closeReason[i].store(0);
+		return connections[i].get();
+	}
+	return nullptr;
+}
+
+// Resets a previously used connection slot to its fresh state.
+void EmulatedEsp32::resetConnectionSlot(int i)
+{
```
m9710797, line 1027: Suggestion. Add `auto& c = *connections[i];`, and then simplify all later lines in this function. Or alternatively accept a `Connection& c` parameter instead of `int i`.

src/network/EmulatedEsp32.cc
```diff
@@ -0,0 +1,2441 @@
...
+	default:
+		sendQuickResponse(cmd, 4); // ERR_INV_PARAM
+		break;
+	}
+}
+
+void EmulatedEsp32::handleUnapiCommand(uint8_t cmd, std::span<const uint8_t> data)
+{
+	// Unimplemented command codes of the UNAPI spec: handled before the
+	// switch to keep its case count low.
```
m9710797, line 1186: I don't understand this comment. Wouldn't these all be handled by the `default` case?

m9710797 (commented):
2nd part of the review. Now I only looked at `EmulatedEsp32.{cc,hh}`.

This is a lot of code (+2k lines). I only had a few hours and that's not sufficient for a detail review. So I only reviewed each function is isolation. For sure I do not yet understand the full code (and much less how the functions work together). (Also near the end of the file my attention was drained, so that's reviewed in less detail).

I'll first let you address the current remarks. Then I'll do a 2nd pass review.

Note that some of my remarks are only suggestion. You don't have to address them all. And maybe sometimes I'm just wrong.

@antxiko this PR has some overlap with your work on `UnapiNet` https://github.com/openMSX/openMSX/pull/2147. If you want, could you maybe also review this PR. Possibly exchange ideas: suggests thing to take from your PR, or the other way around, you improve `UnapiNet` with ideas from this PR.

doc/smxwifi.md
m9710797: I'm not really familiar with network stuff on MSX. So I'm going to trust your judgment. If you think what we have now is already sufficient, then that's fine. Or if you think we can resolve this with extra documentation, then also fine.

I only expressed a concern. I don't know this stuff well enough to know if my concern is valid or not.

doc/smxwifi.md
m9710797: My remark was only about the documentation, One extra sentence should be sufficient.

About QuickRCV. Again, I think you're better qualified than me to judge that. My non-very-informed answer would be: "extra options are always useful". And in the hypothetical(?) case of a multi-process MSX operating system (UZIX or SymbOS) it would be nice to be able to switch to a different process instead of blocking (at the hardware level) when a read has no data available yet. (I don't know if QuickRCV allows that, even in theory).

doc/smxwifi.md
m9710797: I've read more of the code now ... This message only appears when you use the `smxwifi` extension. That's fine. Not when you're using openMSX without this extension (that was my main question in this remark).

doc/smxwifi.md
ducasp: I'll think on this a little bit, you got a good point, in general most won't be plugging an ESP and that will be the exception, but on the other hand if someone wants to use the ESP an easy way to make sure it is correct might be better. Probably gearing towards the ROM being updated and showing whether the device is OpenMSX ESP or a real ESP, so the boot message will be very clear and not requiring steps to determine anything.

doc/smxwifi.md
ducasp: Yeah, I will work on having a way to disable this (since ROM is able to work in both ways) :)

doc/smxwifi.md
MBilderbeek: Indeed, I guess 99.9% of the openMSX runs will not involve an ESP. So best not make it be in the way somehow. (But maybe better to ignore me, I had too little time to dive in and am just commenting from the side here ;-))

ducasp: Yeah, I agree with you both, I will work on the ROM/BIOS so it can detect and tell it is on emulated ESP32 at boot, not intrusive for who uses (already has a splash screen) emulated, help who might want to use an ESP32 to overcome  host network limitations, if any. :)

src/network/EmulatedEsp32.cc
ducasp: I think maybe not worth, will check how many occurrences there are and if there are other stuff we could move to Socket

src/network/EmulatedEsp32.cc
ducasp: Ah, the comment is because that change was to achieve a lower count on switch case, should not be there, this is to try to minimize SonarQube complaining about too much cases in a switch... And yes, you are absolutely right, should be handled by the default cause, and just left a comment that the following TCP-IP spec commands are not implemented and fall on it...

src/utils/OpenSSL.hh
ducasp: Will follow the SerialPort example: read()/write() capturing the error at the moment of failure (ERR_get_error() called immediately), errGetError/errErrorStringN are part of the mandatory set now, and last_error() is replaced by OpenSSL::to_string(ErrorCode), returning std::string by value. EmulatedEsp32 now logs TLS failures with it on the console (e.g. SMXWiFi: TLS failure: certificate verification: certificate has expired when doing hget https://expired.badssl.com).

src/utils/OpenSSL.hh
ducasp: handshake() will return a HandshakeResult with a nested Type enum mapped to the previous values (NeedRead = 0, Done = 1, NeedWrite = 2, Failed = -1), and, inline with the read()/write() discussion, a Failed result also will carry the OpenSSL error code captured at that exact moment via ERR_get_error(), so it can be formatted with to_string(ErrorCode) as well.

src/utils/OpenSSL.hh
ducasp: std::expected<void, int> would not fit in this case, I believe. Reworking this so the specific UNAPI close reason (e.g. 11 hostname mismatch, 13 self-signed) is reported and also described on the console. The old check after a completed handshake was dead code and was removed.

src/utils/OpenSSL.hh
ducasp: pending() will return size_t, uniform with read()/write(). SSL_pending() returns an int, but the value is always positive, so the cast is safe.

src/utils/OpenSSL.hh
ducasp: Agree, IoError will be modified to be a small struct { Type type; ErrorCode code; }. A failure will carry the OpenSSL error code captured at the moment, so it can be formatted with to_string(ErrorCode). WouldBlock and Closed stay separate because the non-blocking control flow needs them.

src/utils/OpenSSL.hh
ducasp: struct Ssl will remain in OpenSSL.hh (used by SessionHandle). All the rest will be moved to OpenSSL.cc, where the entry-point table is defined (all of them are needed there).

src/utils/OpenSSL.cc
ducasp: Sorry about that, probably got lost amongst all other stuff... Will split the candidate list per platform. Also, simplified a bit as there was some legacy support for older versions of SSL that no one should be using anyway, so ssleay32.dll no longer searched.

src/utils/OpenSSL.cc
ducasp: Yep, separate Runtime singleton will be removed, will move the load state to LibHandle

src/utils/OpenSSL.cc
ducasp: Entry points will be made mandatory and load() will refuse the library when anything essential is missing, so nullptr guards will be removed. About the trust store: /etc/ssl/cert.pem load is POSIX-wide, not macOS-only — on Linux it covers distros whose CA bundle is not on OpenSSL's compiled-in default path, and it is harmless when the file is missing since the default paths are tried right after.

src/utils/OpenSSL.cc
ducasp: Makes sense, it will take std::string by value.

src/utils/OpenSSL.cc
ducasp: Agreed, cryptoNameFor() result will be assigned to a named local before .c_str() is passed to openSslLib(), so the string's lifetime will be explicit instead of relying on temporary-lifetime rules.

src/utils/OpenSSL.cc
ducasp: Yeah, both will part of the mandatory set and guards will be gone.

src/utils/OpenSSL.cc
ducasp: Yeah, it will be covered by what was discussed on the last_error() comment: read()/write() returning the OpenSSL error code captured at the moment of failure, so it can be formatted with to_string(ErrorCode) without re-querying.

ducasp:
Just a note, this has not been forgotten or abandoned, work and study has taken its toll on my time this week, but will be back soon :)

src/network/SMXWiFi.hh
ducasp: Hey, that is good learning, other than fifo and its mutex that are in the order needed, I thing I could get a good order that will not waste with padding.

src/network/SMXWiFi.cc
ducasp: Yeah, good catch about powerUp... About powerDown, indeed, this device needs a different power down handling because the emulated ESP32 would be powered off with the MSX, thus it would automatically kill the emulated ESP32 and any pending connections, etc that are dealt by a thread... I think this would be clear, and at the power down, if for example anyone called the telnet client to connect to hispamsx and then power down the emulation without closing it, the thread is properly stopped and any ESP related thread activity will be stoped.

src/network/SMXWiFi.cc
ducasp: Ok, yeah, this is though pill to swallow, did not think (or to be honest, check the wait code :D ) exactly about wait advancing emulation time and that perhaps explain why the 15ms timeout that was originally planned was not working so well (thought about USB - UART latency, but this explains it a lot better).

I think I found a way to get a good compromise on trying to keep emulation not skiping the emulated time and still being able to hold it and wait for the data when that comes.

m9710797: Please also study this: https://github.com/openMSX/openMSX/pull/2194 and/or discuss with @costarc.
To me it seems this is trying to solve a very similar problem. I hope you can (both) learn from each other's approach.

src/network/SMXWiFi.cc
ducasp: Worked on it, think it will be correct in the next round

src/network/SMXWiFi.cc
ducasp: Checked it, seems a little bit different on some points, where it has an expected 20us timing, and on the variable route, we are doing similar approaches (I mean, the new approach for the smxwifi that I'm testing before making a new commit). There is a difference where it is using RealTime::timeLeft to determine how long to sleep, and that would not work very well in this specific case as if emulation is accelerated, the time it could take to answer from the external ESP that is not accelerated will not speed up... So the calculation is a little bit different, but the principle and what we are doing seems quite aligned in general terms.

src/network/EmulatedEsp32.cc
ducasp: Yeah, checked it and don't think it is worth the trouble, msxpi network doesn't use it, and two occurrences only here.

src/network/EmulatedEsp32.cc
ducasp: Yeah, makes sense and will work on it

src/network/EmulatedEsp32.cc
ducasp: Will add it!

src/network/EmulatedEsp32.cc
ducasp: Got it!

src/network/EmulatedEsp32.cc
ducasp: Yeah, you are right, this was used at some point in time and left, will remove

src/network/EmulatedEsp32.cc
ducasp: Will fix it

src/network/EmulatedEsp32.cc
ducasp: Makes sense and will try to implement in that way

src/network/EmulatedEsp32.cc
ducasp: Ok, will do it and look for similar occurrences

src/network/EmulatedEsp32.cc
ducasp: Will address this in the next commit

src/network/EmulatedEsp32.cc
ducasp: It would not work immediately as the code is, but also thinking on it made me find a possible issue when having multiple passive connections and one is closed and other is not... Will think on a way to use std::shared_ptr and not have this possible issue

doc/smxwifi.md
ducasp: Ok, pretty easy solution, the ROM will tell the firmware type in the ESP device, openMSX emulated reports:

image

While a real ESP32 C6 would be something like this:

image

So it is easy on a boot to figure out if emulated esp or real esp

doc/smxwifi.md
ducasp: This also has been added, by default On, but user can set it off:
image
Also checked and could see the expected performance hit on receiving data when quick receive is not used, so it is enabling / disabling it correctly, also the ROM that has the UNAPI will tell at boot when it is on:
image

Or when it is off:
image

src/network/EmulatedEsp32.cc
ducasp: Thanks for the suggestions, next commit should follow it.

src/network/EmulatedEsp32.cc
ducasp: I think it will be clearer on next commit

src/network/EmulatedEsp32.cc
ducasp: Thanks for the suggestion, will use it

src/network/EmulatedEsp32.cc
ducasp: Again, thanks for the suggestion, will use it :)

src/network/EmulatedEsp32.cc
ducasp: Ok, next commit will have a comment about it, good catch!

src/network/EmulatedEsp32.cc
ducasp: I think this is a great performance improvement and next commit should have some changes to address this

src/network/EmulatedEsp32.cc
ducasp: Liked the idea to extend cb_queue, next commit will have it and address the change of interface

src/network/EmulatedEsp32.cc
ducasp: Yeah, agreed

src/network/EmulatedEsp32.cc
ducasp: Should be addressed on the next commit

ducasp:
Ok, before commiting the changes I will first compile and test Linux build, and check that mingw continues to compile fine, should make a new commit soon

PR #2186, opened 2026-09-01
Report applied IPS patches from cart<slot> query

pvmm:
Currently the no-argument form of the cartridge commands (`cart`, `carta`, `cartb`, `cartc`, ...) reports only the inserted cartridge name:

```
carta
carta: /path/to/rom.rom
```
When a ROM cartridge has IPS patches applied (e.g. `carta <rom> -ips patch1.ips` `-ips patch2.ips`), there is no way to see that from the emulator state — the applied patches only live in the internal auto-generated config.

This PR makes the plain query also report the list of applied IPS patch files, so the full media state is verifiable:
```
> carta
carta: /path/to/rom.rom {patch.ips}
```
For a ROM cartridge with no patches an empty list is reported:
```
> carta
carta: /path/to/rom.rom {}
```

For compatibility, the `empty` response was maintained when no cart is found, like in:
```
carta: {} empty
```
This request completes the tool set of MCP servers that need to expose an IPS patch interface to AI agents to both apply patches (`carta <rom> -ips <file>`) and query the current patch state of a cartridge so the agent can inspect what was applied, round-trip the state across sessions, or clean up the list of applied patches.

### DIFF
--- a/src/CartridgeSlotManager.cc
+++ b/src/CartridgeSlotManager.cc
@@ -19,6 +19,8 @@
 #include <array>
 #include <cassert>
 #include <memory>
+#include <string>
+#include <vector>
 
 namespace openmsx {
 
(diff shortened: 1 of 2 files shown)


MBilderbeek:
You're right that `carta` won't give you that info. But `machine_info media carta` will:
image

m9710797:
The `cart<x>` command has a long history (it's one of the older commands). Today we would design it in a different way.

E.g. now when we want to return multiple results we return a `dict` (key-value pairs). That way we can later add more info without breaking backwards compatibility.

Because your current patch actually does break backwards compatibility. E.g. I know there are external programs(*1) that query `carta`, then strip of the first 7 characters and interpret the rest of the string as the filename (with escaping rules like `{}` and `\`).

(*1) Some that I wrote +10 years ago. But if I did it like that, then other people (maybe less familiar with Tcl list syntax) may have done the same.

Backwards compatibility is important, but it's not easy. I'm afraid we can't just change the format of the `carta` result. I haven't really thought about this yet, but the first idea that comes to mind is:
* Keep `carta` (without parameters) unchanged, for backwards compatibility.
* But add a new sub-command, maybe `carta info`, that then does return a `dict`.

HOWEVER isn't this the same as the `machine_info media` stuff @MBilderbeek mentioned above, which is already available. And that one is also more general: it offers a uniform interface for all media slots. And it offers a way to query which media slots are available etc.

---

With the above info. Should we close this PR? Is the `machine_info media` stuff sufficient for you?

pvmm:
Closing this because since `machine_info media` is enough. Updating the mcp-openmsx tools to consider it.

PR #2189, opened 2026-09-04
CI: fix and split the 3rdparty cache

andete:
The key hashed only packages.py, not 3rdparty.mk, creating builds using stale 3rd party files.

### DIFF
--- a/.github/actions/build/action.yml
+++ b/.github/actions/build/action.yml
@@ -49,7 +49,7 @@ runs:
         path: |
           derived/3rdparty
           derived/*/3rdparty
-        key: ${{ github.ref }}-${{ inputs.os }}-${{ inputs.cpu }}-${{ inputs.flavour }}-${{ hashFiles('build/packages.py') }}
+        key: ${{ github.ref }}-${{ inputs.os }}-${{ inputs.cpu }}-${{ inputs.flavour }}-${{ hashFiles('build/packages.py', 'build/3rdparty.mk') }}
     - name: Download 3rd party packages for Windows VC build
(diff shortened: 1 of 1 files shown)


andete:
CI build logs show it works. Some cases will only trigger for real when we actually make changes of course.

MBilderbeek:
Looks like a good step forwards. Wouldn't it make sense to actually only include the patch files explicitly in the filter, like `build/3rdparty/*.diff`?

andete:
> Looks like a good step forwards. Wouldn't it make sense to actually only include the patch files explicitly in the filter, like `build/3rdparty/*.diff`?

I did not actually change that part of the filter. Currently it'll also trigger if e.g. visual studio build files have changed, I think?

MBilderbeek:
> > Looks like a good step forwards. Wouldn't it make sense to actually only include the patch files explicitly in the filter, like `build/3rdparty/*.diff`?
> 
> I did not actually change that part of the filter. Currently it'll also trigger if e.g. visual studio build files have changed, I think? 

Well, wasn't it first only hashing on build/packages.py? And the commit message says to also hash on changed patches (which sounds like a very good idea). So that seems not to match the new hashfiles argument. 
But yeah, I guess changed build instructions is also a good reason to rebuild. Shouldn't then the other build files also be included?

andete:
Hey @MBilderbeek, I'm happy with the way it works now, are you okay with me merging it in?

MBilderbeek:
okay, and as always I appreciate a clean commit history before you merge.

Issue #1106, opened 2018-05-08
Vertical adjust does not affect maximum horizontal interrupt line


grauw:
In openMSX changing the vertical adjust affects the maximum screen split line, but on a real MSX it does not affect it.

This hung my game project’s code when I changed the adjust to a non-zero value, and also breaks VGMPlay on an MSX2 @ 50 Hz with a vertical adjust > 5.

Test program, should hang at 234:

    10 VDP(10)=0     ' +2 for 50 Hz, +128 for 212 lines
    20 VDP(19)=&H70  ' changing this should have no effect
    30 FOR I=255 to 0 STEP -1
    40 PRINT I
    50 VDP(20)=I
    60 VDP(0)=VDP(0) OR 16
    70 I$=INPUT$(1)
    80 VDP(0)=VDP(0) AND 239
    90 NEXT

Verified on an MSX turboR GT.

Context: https://www.msx.org/forum/msx-talk/development/grauw%E2%80%99s-rpg-in-development?page=14#comment-338647

grauw:
Given that the maximum interrupt line doesn’t change with the vertical adjustment, I thought maybe IL wasn’t affected by vertical adjust, but it is and openMSX emulates that correctly. But I did notice on my real MSX that the top lines repeat when the adjust is 7 while overscanning (except for the last), openMSX doesn’t do this.

I think this could be explained by the VDP resetting its internal line counter up to seven lines later than usual (so the first few get an old value), rather than immediately after top erase with an offset. Which would explain the maximum interrupt line not changing.

In other words, vertical adjust changes the line on which the line counter is reset, rather than the value the line counter is reset to.

Here’s a test ROM; change the vertical adjustment with u and j. When adjusting to the most downwards position, observe the difference for the topmost displayed lines between openMSX and a real MSX2+.
[tiletile.rom-adjust.zip](https://github.com/openMSX/openMSX/files/1985579/tiletile.rom-adjust.zip)

MBilderbeek:
I wonder if this could be related to #471 ...

grauw:
> I wonder if this could be related to #471 ...

I think it is. That report says the issue presents itself depending on the vertical adjust value, which is exactly what this bug is about.

MBilderbeek:
So, any idea what should be changed in the code to test your theory? :) @m9710797 perhaps? Or @mthuurne ?

m9710797:
I completely forgot about this bug. Sorry about that.

[...]

PR #2191, opened 2026-09-05
VDP: latch the display line counter at its reset, not at frame start

### LINKED ISSUE #471: [Bug] Sex Bomb Bunny vertically shaking image at certain vertical adjust values [sf#170]
**Reported by manuelbi on 2004-11-21 18:50 UTC**
In Sex Bomb Bunny the whole screen shakes during the
game if the vertical set adjust is -7, -6, -5, -4, -3,
-2, 6, 7 or 8.
AFAIK the game uses overscan, so it could be related to
that.
Ark-A-Noah doesn't have the problem, though, and that
game definately uses overscan.

sndpl:
Fixes #471 — Sex Bomb Bunny's vertically shaking image.

> **This PR was rewritten.** The first version added a `prevDisplayStart` variable to stop a line interrupt from firing twice across a frame boundary. @m9710797 pointed out that this added emulator-only state rather than modelling the chip, and suggested modelling the single display line counter instead. That model has since been confirmed on real hardware (FS-A1GT and FS-A1ST) and against the V9958 die-shot reconstruction in [IKA9958](https://github.com/ika-musume/IKA9958), and it is what this PR now implements. The discussion below the first comments refers to the old version.

## The model

The V9938 has one display line counter. It is reset once per frame, in the top border, when the raster line plus the vertical adjust equals 15 — i.e. at line `8 + getVerticalAdjust()` — to a negative value that depends on the 50/60Hz and 192/212-lines bits of R#9 **as they are at that moment**:

| | 60Hz | 50Hz |
|---|---:|---:|
| 192 lines | −27 | −54 |
| 212 lines | −17 | −44 |

From there it counts up once per line: display starts when it reaches 0, the display area ends when it equals 192 or 212 compared against the *current* R#9 bit 7, and the line interrupt fires when it equals `R#19 − R#23`. Nothing about it changes until the next reset, so it keeps counting into the top border of the next frame.

## What was wrong

openMSX re-derived the display start at every frame start, moved it on every R#9 bit 7 write up to the display start itself, and re-derived the line interrupt position from the *new* display start at each frame start. Software that moves the display start mid-frame for overscan — Sex Bomb Bunny sets LN=1 in the top border and LN=0 after display start — therefore got two line interrupts for a single counter match: one at the end of a frame, one in the top border of the next. The resulting ISR phase shift moved the picture 10 lines up and down on alternate frames.

## What changes

- The counter reset is a sync point of its own, at line `8 + vadj`, moving with the vertical adjust.
- `displayStart` is latched there, from R#9 sampled at that moment, and DISPLAY_START / VSCAN / HSCAN are scheduled from it. Nothing else touches it.
- Across a frame start, `displayStart` and `horizontalScanOffset` are only re-expressed relative to the new frame, so until the reset they still describe the previous frame's counter. The carry-over into the next frame's top border needs no extra state.
- An R#9 bit 7 write only reschedules the display end — a comparison against the current bit that never matches if the counter already passed it. That *is* the overscan; it falls out of the model rather than being a special case.
- A match that would land at or after the next reset never happens, which replaces the `lineCountResetTicks` special case (and its "unverified assumption" comment: it is measured now).
- Savestate version 12 → 13. Older states schedule the reset for the current frame on load, which re-latches the value they already had.

All numbers openMSX already used — the reset line, the four display start positions, the "no match beyond the reset" behaviour — turn out to be right; only *when* R#9 is sampled changes, and with it the double fire disappears by construction.

## Evidence

Test program `T471.COM` (source + binary attached in the comments; run from MSX-DOS on an MSX2 or higher, switches a turbo R to Z80 mode itself). It reads the latched FH flag at chosen raster positions and reports, per R#19 value, whether the match landed in this frame (`1`), the next frame's top border (`2`), or both (`B`), for LN=0, LN=1 and the SBB toggle; a fourth row sweeps the *line* at which LN=1 is written and reports whether it took effect (`E`) or came too late (`L`).

Real FS-A1ST, `SET ADJUST(0,0)` and `(0,−4)` — and openMSX with this PR, which now reproduces both **character for character**:

```
                    vadj=7 (R#19 214..239)          vadj=3 (R#19 218..243)
LN=0                1111111222222222222222.....     1111111222222222.........
LN=1                11111111111111111222222222      11111111111111111222222222
TOGGL               11111111111111111222222222      11111111111111111222222222
LATCH               EEEEEELLLLLLLLLLLLLLLLLLLL      EEEEEELLLLLLLLLLLLLLLLLLLL
display start       LN=0 line 41, LN=1 line 31      LN=0 line 37, LN=1 line 27
LN sampled before   line 13 (model 8+vadj = 15)     line 9  (model 8+vadj = 11)
                    OK: every match fired once      OK: every match fired once
```

- `TOGGL` equals `LN=1` with no `B`: moving the display start mid-frame produces one interrupt per match.
- The `LATCH` edge stays at the same *index* while the display start shifts with the adjust: LN is sampled at `8 + vadj`, not at display start. (The printed line is ~1–2 lower than the model's because the test's positions are labelled ~1 line early — the sync lands 0.95 line into its line — so the edge falls between lines 14 and 15, i.e. at 15.)
- The `.` run (no match) starts at the same R#19 as in openMSX: the counter never reaches values beyond the reset.

openMSX 21.0 on the same test shows `TOGGL 111111BBBBBBBBBB22222.....` and `LATCH EEEEEEEEEEEEEEEEEEEEEEEEEE`. grauw's earlier FS-A1GT run with T471 v1 is consistent with all of the above; v1 wrote LN two lines *after* the reset, which is why its LN=1/TOGGL rows showed no `1`s — itself a confirmation of the sampling point.

Also verified on this build: SBB cold-booted at `SET ADJUST` 0, 5 and −4 is steady; a version-10 savestate loads, re-saves and reloads cleanly; plain BASIC boot on PAL (NMS 8250) and NTSC (HB-F1XDJ) machines is unchanged.

## Known approximations (pre-existing, unchanged in spirit)

- A vertical adjust write inside the top border that moves the reset line *behind* the current position: the real chip skips the reset for that frame; openMSX keeps the pending one. A follow-up test row for this is in preparation.
- A 50/60Hz switch: frame length is still latched at frame start; the reset offset is sampled live; the "never" decision for a match in the next frame assumes the same setting there.

Neither affects any known software.

---

*Transparency note: this analysis, the code, the test program and this description were produced by Claude (Opus), an AI assistant, working under Sandy's account and supervision. The hardware measurements were done by Sandy (FS-A1ST) and grauw (FS-A1GT).*


MBilderbeek:
Can you show how it matches the observation that "I can reproduce with the following set-adjust values (Y): -1, 0, 1, 2, 3, 4 and 5. (but no others)"?

sndpl:
Yes — it falls out of the model exactly, and I've now verified it empirically for all 15 values.

## The derivation

Everything below is NTSC (SBB switches to 60 Hz), so 262 lines per frame. Display start line:

```
lineZero = 3 (sync) + 13 (top erase) + 9 (NTSC top border) + (LN ? 0 : 10) + vadj
         = 25 + vadj   with LN=1 (212 lines)
         = 35 + vadj   with LN=0 (192 lines)
```

where `vadj = getVerticalAdjust() = (R#18 >> 4) ^ 7`, so 0..15. The line interrupt match sits at `lineZero + IL` lines, plus `getRightBorder()` which is ~0.95 line.

SBB writes `R#19 = 230` late in the frame, when `displayStart` is still latched at the **LN=1** value:

* **LN=1** → match at line `255 + vadj` (+0.95). It stays inside the frame iff `255 + vadj + 0.95 < 262`, i.e. **`vadj <= 6`** — and then it fires there.
* At the next `frameStart()` R#9 bit 7 is back to 0, so `scheduleHScan()` recomputes with `lineZero = 35 + vadj`: `265 + vadj >= 262` → wraps → lands at line `3 + vadj` (+0.95) of the new frame. That is before the counter reset at line `8 + vadj`, so it is not discarded → **it fires a second time**.

So the duplicate interrupt — and hence the shaking — happens **exactly when `vadj <= 6`**.

For `vadj >= 7` the LN=1 computation *already* wraps, so nothing fires at the end of the frame; it is simply scheduled once, early in the next frame. One interrupt, no shaking. That's the whole reason it's a contiguous band of adjust values with a sharp edge.

## Mapping that back to SET ADJUST

Two steps, both measured rather than assumed:

1. The BIOS stores `R#18` high nibble = `(-y) & 15`. Measured on the NMS 8250 by typing `SET ADJUST(0,y)` in BASIC and reading R#18: `y=-7 → 0x70`, … `y=0 → 0x00`, `y=1 → 0xF0`, … `y=7 → 0x90`.
2. **SBB does not use that value directly** — it adds 6 to the nibble, i.e. it moves its own display 6 lines up relative to the user's setting. @grauw already spotted this in the issue ("with the default adjust (0,0), Sex Bomb Bunny uses a vertical adjust of -6, r#18 is 96" — 96 = 0x60). Confirmed here: `SET ADJUST(0,0)` → BIOS writes `R#18 = 0x00`, and once the game is running `R#18 = 0x60`.

So `R#18_game_hi = ((-y) + 6) & 15` and `vadj = R#18_game_hi ^ 7`, and `vadj <= 6` works out to **y ∈ {-1, 0, 1, 2, 3, 4, 5}** — precisely your list, with the complement `{-7, -6, -5, -4, -3, -2, 6, 7}` being the workaround values.

## Measurement

For each `y` in -7..7: boot to BASIC with no disk, type `SET ADJUST(0,y)`, insert the disk, reset, let the game load and start, then log the display start latched for each frame. Run once with this PR's change disabled and once with it enabled.

| SET ADJUST y | R#18 (BIOS) | R#18 (game) | vadj | predicted | measured, before | after |
|---:|---|---|---:|---|---|---|
| -7 | 0x70 | 0xD0 | 10 | steady | steady — line 35 | steady — 35 |
| -6 | 0x60 | 0xC0 | 11 | steady | steady — line 36 | steady — 36 |
| -5 | 0x50 | 0xB0 | 12 | steady | steady — line 37 | steady — 37 |
| -4 | 0x40 | 0xA0 | 13 | steady | steady — line 38 | steady — 38 |
| -3 | 0x30 | 0x90 | 14 | steady | steady — line 39 | steady — 39 |
| -2 | 0x20 | 0x80 | 15 | steady | steady — line 40 | steady — 40 |
| **-1** | 0x10 | 0x70 | **0** | **SHAKE** | **SHAKE — lines 25/35** | steady — 25 |
| **0** | 0x00 | 0x60 | **1** | **SHAKE** | **SHAKE — lines 26/36** | steady — 26 |
| **1** | 0xF0 | 0x50 | **2** | **SHAKE** | **SHAKE — lines 27/37** | steady — 27 |
| **2** | 0xE0 | 0x40 | **3** | **SHAKE** | **SHAKE — lines 28/38** | steady — 28 |
| **3** | 0xD0 | 0x30 | **4** | **SHAKE** | **SHAKE — lines 29/39** | steady — 29 |
| **4** | 0xC0 | 0x20 | **5** | **SHAKE** | **SHAKE — lines 30/40** | steady — 30 |
| **5** | 0xB0 | 0x10 | **6** | **SHAKE** | **SHAKE — lines 31/41** | steady — 31 |
| 6 | 0xA0 | 0x00 | 7 | steady | steady — line 32 | steady — 32 |
| 7 | 0x90 | 0xF0 | 8 | steady | steady — line 33 | steady — 33 |

Prediction matches measurement on all 15 values, and with the change every value is steady.

One thing worth flagging while reading this: poking R#18 at runtime on a running game is *not* a valid way to reproduce the sweep. The shake is a self-sustaining two-frame cycle, so changing the adjust mid-game can leave it in either phase and gives misleading results — the adjust has to be set before the game loads. (Same trap with `after frame` + `screenshot`: it samples every other emulated frame and hides the alternation completely. Use `record`.)


Transparency note: this analysis, the code and this comment were produced by Claude
(Opus), an AI assistant, posting under Sandy's account and supervision.

grauw:
Can you please make a minimal test case to repro? Preferably a com file + source which reproduces the issue and also tests different variations to demonstrate the upper and lower bounds of the timing and conclusively shows that this is the correct and only conclusion.

Testing just with a full game without verifying conclusions using an isolated test on real hardware has previously proven to lead to incorrect conclusions.

m9710797:
Hi, thanks a lot for this PR!

Before merging I do want to ask some questions. And I do share grauw's concern, so a standalone testcase would indeed be useful. @grauw do you then volunteer to run experiments on real hardware?

My concern is about the extra state we introduce: before we had a single `int`, now we have two (`displayStart` and `prevDisplayStart`). You already explained that likely the real VDP has a single counter (and then does (cheaper in hardware) equality-checks, rather than larger-or-equal-checks). But we have two non-counter variables. Whenever the emulation model differs from the hardware model there's a risk of an inconsistency.

There's a tradeoff, we can try to model the hardware as closely as possible (so in this case model the increasing counter). But that would require also emulating line-by-line. For efficiency, in openMSX, we emulate the CPU instruction-by-instruction (but not cycle-by-cycle), and the other peripherals (like the VDP) in large time-chunks (via the `Scheduler` class and `Schedulable::executeUntil()` methods (the VDP implements several of these `Schedulables`).

Sometimes this difference in emulation model (cycle-by-cycle vs in-large-chunks) also warrants a different internal representation of the "state" (the line counter in this case). But that must be an intentional decision. And especially if we introduce *more* state (not just a different representation of the same state) we should pause and think if this is indeed the right approach.

I have *NOT* yet worked out all the details, but here's my current thinking:
* The real VDP likely has a single counter.
* It likely does equality comparisons against this counter to detect line-IRQ and end-of-display-area.
* Likely this counter is reset at a single position in the frame:
  * Could be reset to zero at the first display line .. but then how to know we're at the first display line?
  * Could be reset to some negative value (a value depending on 192 vs 212 lines) at the top of the frame.
  * Or reset to zero at the top of the frame, but then the later comparison become more difficult (more complex hardware circuit I mean). E.g. not check for zero for the start of display line, similar for end-of-display and line-IRQ.

If my assumptions above are correct, and if we can figure out where the hardware resets the counter. Then maybe we can model it in a similar way in openMSX. So not have a running counter, but a (single) fixed time (represented as `EmuTime` or as ticks since `frameStart`) that we latch at the same position in the frame as the hardware would reset the counter. And then given this latched time, for any later time we can compute the value that the real hardware counter would have.

Since this can reconstruct the hardware state (without needed line-by-line updates) I think, in theory, it should also be sufficient for a correct emulation.
* But it does mean a more invasive change than this PR.
* For example on changing VDP register R#9 we should no longer eagerly change the value of `displayStart` (i think that's the root cause of the bug), see `VDP.cc` line 1241.
* See also the comment in `VDP.cc` line 463.

Other things I haven't looked at yet:
* Maybe we should then also rework `scheduleVScan()` and `scheduleHScan()`, because these are currently triggered from `scheduleDisplayStart()`.

sndpl:
@grauw Test case attached: `t471.zip` contains `t471.asm` (sjasmplus, fully commented) and `t471.com`. It runs from MSX-DOS on an MSX2 or higher.

## What it does

IE1 (R#0 bit 4) is enabled but the Z80 stays in `DI`, so S#1 bit 0 (FH) is a latched, read-to-clear "a line match happened" flag. Each measurement first syncs the raster on a match of a known R#19, then reaches every position with a 228 T-state per line delay loop. For each R#19 the flag is then read at two points — late in *this* frame, and in the top border of the *next* frame — in three configurations:

| | |
|---|---|
| `LN=0` | R#9 bit 7 stays 0 |
| `LN=1` | R#9 bit 7 stays 1 |
| `TOGGL` | R#9 bit 7 set to 1 in the top border and back to 0 after display start — what SBB does |

Output is one character per R#19 value: `1` the match landed in this frame, `2` it landed in the next frame's top border, `B` **both** (one match, two interrupts), `.` no match at all, `?` that entry lost sync.

It forces 60Hz so the numbers are the same everywhere (262 lines, 228 T/line), and it verifies its own timing before measuring anything — two matches of one R#19 are exactly one frame apart — and refuses to trust its own results if the delay loop doesn't fit the machine. Delays include the wait state the MSX adds to every M1 cycle, so don't run it in R800 mode or with an accelerator.

## What it shows

Philips NMS 8250, `SET ADJUST(0,0)`:

```
openMSX 21.0                          with this PR
R#18=00 vadj=7                        R#18=00 vadj=7
R#19 214..239                         R#19 214..239

LN=0  111111222222222222222.....      LN=0  111111222222222222222.....
LN=1  11111111111111112222222222      LN=1  11111111111111112222222222
TOGGL 111111BBBBBBBBBB22222.....      TOGGL 11111111111111111222222222

measured display start:               measured display start:
  LN=0 line 42, LN=1 line 32            LN=0 line 42, LN=1 line 32
BUG: two interrupts for R#19=220..229 OK: every match fired exactly once.
```

Three things fall out of this, all measured rather than assumed:

**1. Where the display start is.** The R#19 at which `1` turns into `2` gives it directly: line 42 with LN=0 and line 32 with LN=1, exactly 10 apart.

**2. The `B` band, bounded on both sides.** Its lower edge is where the LN=0 computation starts to wrap into the next frame; its upper edge is the LN=1 transition, i.e. the last R#19 whose match still falls inside the frame. Checked at three different `SET ADJUST` values, and the band moves exactly with the display start:

| SET ADJUST y | vadj | display start LN=0 / LN=1 | `B` band | `[262−ds₀ … 262−ds₁−1]` |
|---:|---:|---|---|---|
| 0 | 7 | 42 / 32 | 220..229 | 220..229 |
| −4 | 3 | 38 / 28 | 224..233 | 224..233 |
| 7 | 14 | 49 / 39 | 213..222 | 213..222 |

The `TOGGL` counter was started at the LN=1 display start, so on hardware its transition should coincide with the `LN=1` row — which is what this PR makes openMSX do. Any `B` at all means one match produced two interrupts.

**3. Where the display line counter is reset.** The `.` run is where the match would fall *after* the reset in the top border, so the counter never reaches that value. Its first R#19 measures the reset line, and openMSX's assumption of line 8+vadj is reproduced exactly at each adjust — worth having, since the source comment next to `lineCountResetTicks` flags that value as unverified and points at #1106.

## What I would like checked on hardware

The `TOGGL` row. If real hardware gives the same row as `LN=1` (no `B` anywhere), the counter is monotonic and this PR is right. If it shows a `B` band, then openMSX 21.0 was right and the model needs rethinking — and SBB's shaking has another cause.

The `LN=0` / `LN=1` transitions and the start of the `.` run are worth writing down too, since they are direct measurements of the display start position and of the counter reset line.

Fair warning on reproducing this yourself: poking R#18 into a *running* SBB is not a valid way to sweep the adjust — the shake is a self-sustaining two-frame cycle, so changing the adjust mid-game can drop it into either phase. It has to be set before the game loads. Same trap with `after frame` + `screenshot` in openMSX: it samples every other emulated frame and hides the alternation completely; use `record` instead.

[t471.zip](https://github.com/user-attachments/files/31877015/t471.zip)


Transparency note: this analysis, the code and this comment were produced by Claude
(Opus), an AI assistant, posting under Sandy's account and supervision.

sndpl:
You're right about the extra state, and I think the reframing you're reaching for actually makes the patch look worse than it needs to.

`prevDisplayStart` is not really a second piece of state. It is a clumsy way of saying *"at this point in the frame the display line counter has not been reset yet, so its origin is still the one from the previous frame"*. That is a statement about **when** the origin is latched, not about there being two origins. Expressed the way you describe — a single latched origin, updated at the point where the hardware resets the counter — the second variable disappears.

## Your model, checked against what SBB does

Taking a single counter reset once per frame at line `8+vadj` (the value already in `scheduleHScan()`), to `-(17)` or `-(27)` lines depending on R#9 bit 7 *sampled at that moment*:

* Display start = origin. SBB sets LN=1 at line ~11.9, i.e. before the reset at line 14 (this trace is at vadj=6), so the frame is sampled as 212 lines and display starts at line 31. That is exactly what we observe.
* Display end = counter equal to 192 or 212, compared against the *current* R#9. SBB flips to LN=0 at line ~233, when the counter is already at ~202, so the equality with 192 never happens and the display runs to the end of the frame. That is the overscan, and it drops out of the model rather than needing a special case.
* Line IRQ = counter equal to `R#19-R#23`. The counter reaches 230 at line 261, once. It then keeps counting into the next frame until the reset, so it cannot reach 230 again. **The double fire becomes impossible by construction**, rather than being patched afterwards.

So I agree with your diagnosis of the root cause: the problem is that `displayStart` is updated eagerly on an R#9 write for as long as `time < displayStartSyncTime`. If instead the origin is latched at the counter reset, then `scheduleHScan()` re-anchoring at `frameStart()` is simply wrong — between the frame start and line `8+vadj` the origin still belongs to the previous frame. Re-anchor at the reset point instead and my `prevDisplayStart` is not needed.

## The question that decides it, and it is close

Your model and the current code disagree about **when R#9 bit 7 stops affecting the display start**:

* today: any time before display start, so line `25+vadj` or `35+vadj`
* your model: only before the counter reset, line `8+vadj`

That is the `TODO: Find out exactly when display start is fixed` at `VDP.cc:1236`, and SBB nearly distinguishes the two. Its line-230 interrupt fires at line `255+vadj`, the ISR runs about 12 lines, so the LN=1 write lands at roughly line `vadj+5` — about 2 to 3 lines before `8+vadj`, and because both positions shift with the adjust that margin is roughly the same at every `SET ADJUST` value. So your model predicts SBB is stable everywhere, which matches the 15 measurements I posted, but with only a couple of lines to spare. Worth measuring rather than assuming.

## A second prediction worth testing, which the current structure cannot express at all

If the counter really is a single 8-bit counter reset once per frame, then in 60Hz there are 262 lines between resets and only 256 counter values, so **about 6 values of R#19 must match twice in the same frame**. At neutral adjust with LN=0 (origin line 42, reset line 15) the counter runs 229..255 over lines 15..41, 0..219 over lines 42..261, and 220..234 over lines 0..14 of the next frame — so R#19 = 229..234 is hit twice, about 6 lines apart, both times in the top border.

That is a real double interrupt, and openMSX cannot produce it today in any circumstance, because there is at most one HSCAN sync point per frame. If hardware shows it, the counter model is strongly confirmed and it also tells you the reset line precisely. If hardware does *not* show it, then it is not a plain monotonic 8-bit counter and the model needs more thought before it is worth restructuring the code around.

My test case does not detect this yet — it samples the flag once in the top border, so two matches 6 lines apart both look like one. Sampling at a series of points across the top border instead would measure the reset line and this double match in one go. I can extend it that way if it is useful; the third row of the current output (where the `.` run starts) already measures the reset line as `8+vadj` at three different adjust values, in openMSX at least.

## How would you like to proceed?

Either is fine by me:

1. **Keep this PR as a stopgap.** It is a no-op whenever the display start does not move between consecutive frames, so it changes nothing for software that does not do this, and it does not foreclose the rework. I would add a comment pointing at the counter model as the proper fix.
2. **Do it properly here.** Latch a single origin at the counter reset, derive display start, display end and the line IRQ from it, stop updating it eagerly on R#9 writes, and rework `scheduleVScan()` / `scheduleHScan()` so they are not driven from `scheduleDisplayStart()`. Happy to attempt that, but I would rather have grauw's hardware numbers for the reset line and the latch point first, since both feed straight into it.

I would lean towards 2 with the hardware measurements in hand, and I have no objection to this PR sitting unmerged until then.

Transparency note: this analysis, the code and this comment were produced by Claude
(Opus), an AI assistant, posting under Sandy's account and supervision.

sndpl:
Some of this can be answered without hardware. @ika-musume's [IKA9958](https://github.com/ika-musume/IKA9958) is a die-shot based schematic and Verilog reconstruction of the V9958, and its screen timing module is marked complete. It does **not** contain the line interrupt itself — the register file is still in progress and Misc is not started, and the ST sheet has nets for R#18 `V0..V3` and `R#9 b7 LN` but none for R#19 or R#23 — but it does contain the counter that the line interrupt compares against.

From `src/IKA9958_st.sv`, the "vertical-auxiliary counter":

```systemverilog
logic [8:0] vacntr;
wire vadd_eq15 = (vcntr + {{5{REG.Data.V[3]}}, REG.Data.V}) == 9'd15;
always_comb vacntr_rst = vacntr_ci & vadd_eq15;
...
if(vacntr_rst)
    case({REG.Data.LN, REG.Data.NT_n})
        2'b00: vacntr <= 9'd485;   // -27
        2'b01: vacntr <= 9'd495;   // -17
        2'b10: vacntr <= 9'd458;   // -54
        2'b11: vacntr <= 9'd468;   // -44
    endcase
else vacntr <= vacntr + vacntr_ci;
```

with `vapla[0] = vacntr == 0` for display start and `vapla[1] = vacntr == 192 && !LN` / `vapla[2] = vacntr == 212 && LN` for display end.

## Against @m9710797's three bullets

* **A single counter, with equality comparisons.** Yes, on both counts.
* **Reset at one position per frame, moving with the vertical adjust.** Yes: `(vcntr + V) == 15`, with `V` the sign-extended R#18[7:4].
* **"Could be reset to some negative value (a value depending on 192 vs 212 lines) at the top of the frame."** That is exactly what it does, and those are the four constants.

So the third of your three options is the one the silicon took.

## openMSX's numbers are confirmed

`15 - signed(V)` is *identically* openMSX's `8 + getVerticalAdjust()` for all 16 adjust values, since `(U^7) + 8 == 15 - signed(U)`. That includes the "I would expect the counter to reset at line 16 ... measurements on NMS8250 show it is one line earlier" note in `scheduleHScan()`: hardware resets at 15 for neutral adjust, so the measurement behind that comment was right.

The four reset offsets give these display starts at neutral adjust, matching `3 + 13 + (PAL ? 36 : 9) + (LN ? 0 : 10) + vadj` in every mode:

| mode | reset value | offset | display start | openMSX `lineZero` |
|---|---:|---:|---:|---:|
| 60Hz, 192 | 485 | −27 | 42 | 42 |
| 60Hz, 212 | 495 | −17 | 32 | 32 |
| 50Hz, 192 | 458 | −54 | 69 | 69 |
| 50Hz, 212 | 468 | −44 | 59 | 59 |

## And it answers the TODO at VDP.cc:1236

The reset value is a combinational `case` on the register bits at the reset edge, so **R#9 bit 7 is sampled at the counter reset, not at display start**. That is the `TODO: Find out exactly when display start is fixed`, and it confirms your suspicion that the eager `scheduleDisplayStart()` on an R#9 write is the root cause: the window should close at line `15-V`, which is 17 lines earlier than openMSX currently allows in 60Hz/212 and up to 54 lines earlier in 50Hz/192.

Two details for whoever implements the counter model:

* The reset value depends on the 50/60Hz bit as well as on LN, so both need sampling at the reset. openMSX currently latches `palTiming` at `frameStart()` instead.
* Display *end* compares against the **current** LN, not a latched one (`vacntr == 192 && !LN`). So SBB's overscan — flip LN after the counter has passed 192 and the equality never happens — falls out of the model rather than needing a special case. Same for the `TODO` at `VDP.cc:463`.

## A correction to my previous comment

`vacntr` is **9 bits**, not 8. My claim that about 6 values of R#19 must match twice per frame assumed an 8-bit counter, and that was stated too strongly. The reset values 458..495 all have bit 8 set, so that bit effectively marks "before display start", and whether those values double-match depends entirely on whether the line interrupt comparator includes bit 8 — which is the part not in the repo.

For 60Hz/192 at neutral adjust the counter covers 485..511 over lines 15..41 and 0..234 over lines 42..276. So:

* comparator **includes** bit 8 → R#19 in 0..234 matches once, R#19 >= 235 never matches. This is what openMSX implements today, and what `T471` measures (the `.` run starts at 235).
* comparator **ignores** bit 8 → R#19 in 229..234 matches twice per frame about 6 lines apart, and 235..255 matches once, in the top border.

Given that bit 8 exists at all and cleanly marks the pre-display region, the first seems more likely, so I would no longer bet on the double match.

## What is still worth hardware time

Two narrow questions, both of which `T471` can be pointed at:

1. **Does R#19 >= 235 ever fire?** If the `.` run exists on hardware, the comparator is bit-8 aware and openMSX's "never" is right. If instead those values fire once in the top border, it is not.
2. **The LN sampling point.** Both models keep SBB stable, but the margin is only 2-3 lines, so this is worth measuring rather than inferring. Sweeping the line at which LN is flipped, and watching where the display start stops moving, would pin it down.

Caveats on the above: it is a work in progress, and there is an apparent operand-order slip in that `case` — the constants only line up with the modes if the selector is `{NT_n, LN}` rather than `{LN, NT_n}` as written; cases `00` and `11` are correct and `01`/`10` are exchanged, which is the signature of a swapped pair. So this is strong evidence rather than gospel, but it is die-shot derived, which is a step up from black-box probing. Worth someone with more V9958 knowledge than me sanity-checking the reading.

---

*Transparency note: this analysis, the code and this comment were produced by Claude (Opus), an AI assistant, posting under Sandy's account and supervision.*

ika-musume:
Honestly, I'd rather have a human submit a PR for this issue. I think everyone would agree on that.

**Now, a question for the agent:** I haven't read ANY of the messages you've sent so far. So, from a circuit-level perspective, what exactly is still needed? Look at the KiCad schematic I drew and read the V9938/58 datasheet. Tell me what additional work needs to be done to close this case. Keep your answer very concise. Always think at the circuit level, and ask the Fable 5.1 subagent for help when necessary.

grauw:
# T471 test results

## Panasonic FS-A1GT (Z80 mode)

t471-fsa1gtz80-0
t471-fsa1gtz80-3
t471-fsa1gtz80-7
t471-fsa1gtz80-11
t471-fsa1gtz80-15

## openMSX 21.0+ (8ecc13f50), machine FS-A1GT (Z80 mode)

t471-openmsx-0
t471-openmsx-3
t471-openmsx-7
t471-openmsx-11
t471-openmsx-15

## Interpreting test results

Quote `t471.asm`:

```
; HOW IT WORKS
;   IE1 (R#0 bit 4) is enabled but the Z80 stays in DI, so S#1 bit 0 (FH) is a
;   latched, read-to-clear "a line match happened" flag.  Every measurement
;   starts by syncing the raster on a match of a known R#19, after which all
;   positions are reached with a 228 T-state per line delay loop.  For each
;   R#19 the flag is then read at two points:
;
;     1 : late in THIS frame, after the match would fall here
;     2 : in the top border of the NEXT frame
;
;   in three configurations:
;
;     LN=0  : R#9 bit 7 stays 0
;     LN=1  : R#9 bit 7 stays 1
;     TOGGL : R#9 bit 7 set to 1 in the top border and back to 0 after display
;             start -- what Sex Bomb Bunny does
;
; READING THE OUTPUT
;   One character per R#19 value, low to high:
;
;     1  the match landed in this frame
;     2  the match landed in the next frame's top border
;     B  BOTH: one match produced two interrupts   <-- the bug
;     .  no match at all: it would fall after the display line counter is
;        reset in the top border, so the counter never reaches that value
;     ?  a control read found the flag already set; that entry lost sync
;
;   The R#19 where '1' turns into '2' measures where the display start is, so
;   the LN=0 and LN=1 rows should differ by exactly 10 lines.  In TOGGL the
;   counter was started at the LN=1 display start, so its transition must sit
;   at the same R#19 as the LN=1 row.
;
;   Expected on hardware (and on openMSX with PR #2191), for vadj=7:
;     LN=0  111111222222222222222.....
;     LN=1  11111111111111112222222222
;     TOGGL 11111111111111112222222222   <- same transition as LN=1
;   openMSX 21.0 and older instead give:
;     TOGGL 111111BBBBBBBBBB22222.....   <- 10 values fire twice
;
;   Where the '.' run starts also measures the line at which the display line
;   counter is reset; openMSX assumes line 8+vadj and the source flags that as
;   an unverified assumption.
```

## Observations

* The test polls `FH` which is known to be problematic on real hardware. Not sure if it affected results.
* With `LN=0` the real MSX shows one less `1` than openMSX.
* With `LN=1` and `TOGGL` the real MSX never shows any `1`’s at all.
* With `TOGGLE` obviously the real MSX shows no `B` at all.
* The `LN=0` “measured display start” is off by one.
* The `LN=1` “measured display start” differs significantly, but I doubt that draws the right conclusion since the 1-2 transition is not visible at all in the sampled region.

ika-musume:
IKA9958 correction: 
```systemverilog
case({REG.Data.NT_n, REG.Data.LN}) begin
...
```
is correct. There was nothing wrong with the schematic; the error was introduced while translating the schematic into code.

sndpl:
The PR has been rewritten along the lines @m9710797 proposed, and the description above has been updated to match. Summary of what happened since the last round:

**Hardware, FS-A1ST.** T471 v2 (attached below) decouples the LN write from the R#19 arming — v1 wrote both at line `10+vadj`, i.e. two lines *after* the counter reset, which is why grauw's LN=1 and TOGGL rows showed no `1` at all: the write had simply not taken effect yet. v2 writes LN at `2+vadj`, and adds a `LATCH` row that sweeps the line at which LN=1 is written and reports whether it took effect (`E`) or came too late (`L`). Results at two adjust values:

```
                    SET ADJUST(0,0), vadj=7          SET ADJUST(0,-4), vadj=3
LN=0                1111111222222222222222.....      1111111222222222.........
LN=1                11111111111111111222222222       11111111111111111222222222
TOGGL               11111111111111111222222222       11111111111111111222222222
LATCH               EEEEEELLLLLLLLLLLLLLLLLLLL       EEEEEELLLLLLLLLLLLLLLLLLLL
display start       LN=0 line 41, LN=1 line 31       LN=0 line 37, LN=1 line 27
```

* `TOGGL` is identical to `LN=1`, no `B`: moving the display start mid-frame, the way SBB does, gives exactly one interrupt per match. That was the missing direct evidence.
* The `LATCH` edge sits at the same *index* (6) at both adjusts while the display start moved by 4 lines: LN is sampled at `8 + vadj`, not at display start. Accounting for the test's ~1 line labelling offset (the sync lands 0.95 line into its line), the edge falls between lines 14 and 15 at neutral adjust — the `vcntr + V == 15` reset from IKA9958, whose `{NT_n, LN}` correction was also confirmed in the meantime.
* The `.` run starts at the same R#19 as in openMSX: no match beyond the reset. So the comparator does not see the counter's negative region and openMSX's "never" was right; the 8-bit double-match I speculated about earlier does not exist.

**The implementation** is now the single-counter model: the reset is its own sync point at `8 + vadj` (moving with the adjust), `displayStart` is latched there from R#9 sampled at that moment, DISPLAY_START / VSCAN / HSCAN are scheduled from it, and across a frame start the values are only re-expressed relative to the new frame — so the carry-over into the next top border needs no second variable, and a match at or after the next reset simply never happens. R#9 bit 7 writes only reschedule the display end, which is what makes overscan fall out of the model instead of being a special case. `prevDisplayStart` is gone. Savestate 12 → 13, rebased over #2193.

openMSX with this PR reproduces both photos character for character (the emulated FS-A1ST differs by one frame-boundary value in the LN=1 row). SBB is steady at `SET ADJUST` 0, 5 and −4; old savestates load; PAL and NTSC BASIC boots are unchanged.

Two pre-existing corners stay approximated and are called out in the description and in the code: an adjust write inside the top border that moves the reset line behind the raster (hardware would skip the reset that frame), and a 50/60Hz switch. Neither affects known software; a T471 row for the first one is in preparation, since it is a cheap, falsifiable prediction of the model and Sandy has the machine on the desk.

Photos of the two ST runs follow in the next comment.

---

*Transparency note: this analysis, the code and this comment were produced by Claude (Opus), an AI assistant, posting under Sandy's account and supervision.*

m9710797:
Thanks! I did not yet read the patch in detail. But one thing is already strange to me:
* You introduce a new `syncLineCountReset` which resets the counter to a "fixed" value (I mean a value that does not depend on vertical adjust). The moment `syncLineCountReset` triggers does depend on vertical set adjust.
* But implementing this in hardware requires another counter. (Weird to need a 2nd counter to be able to reset a counter).

What if we reset the counter at a fixed position in the frame (and then preferably a position for which we already have a sync point, maybe frameStart?). And then reset the counter to a value based on the vertical set adjust setting.

As I said, I didn't read the code in detail yet. But at first sight it seemed rather complex (more code changes than what I was expecting). But maybe that's because you introduced a (needless) extra sync point?

sndpl:
Fair challenge — but the two-counter structure is the chip's, not mine. In IKA9958 the display line counter (`vacntr`) is reset by a comparator on the raster counter:

```systemverilog
wire vadd_eq15 = (vcntr + {{5{REG.Data.V[3]}}, REG.Data.V}) == 9'd15;
always_comb vacntr_rst = vacntr_ci & vadd_eq15;
```

to one of four constants that depend only on LN and 50/60Hz. So the adjust moves *when* the counter starts, not what it starts from — which is just what a vertical shift is. In the emulator that comparator is a sync point; there is no second counter, only a scheduled moment, the same kind of thing DISPLAY_START or VSCAN already are. It costs one sync per frame.

Resetting at frame start to an adjust-dependent value would sample R#9 at the wrong moment, and that is measurable: in the `LATCH` row on the FS-A1ST, LN=1 written at lines 8–13 (all *after* frame start) took effect, at 15 and later it did not. A frame-start latch would show `L` across that whole row. The `ADJ` rows Sandy is about to run test the same thing for R#18 itself.

The sync point is also what let me drop `prevDisplayStart`: with the origin latched at line 8+vadj there is exactly one origin alive at any time, so the carry-over into the next top border needs no second variable. Re-deriving `displayStart` at frame start puts two origins in flight between frame start and the reset — the old one for the carry-over match, the new one for everything after — which is where the original bug came from. Doing the sampling without a sync point would mean recording R#9/R#18 write history instead, so more state, not less.

On size: of +168/−91, 91 added and 57 removed lines are comments. Code is +77/−34, mostly the schedule/exec pair (~45) and serialization (~12); `scheduleHScan` shrank by a line and lost the wrap-around and `lineCountResetTicks` special cases.

---

*Transparency note: this analysis, the code and this comment were produced by Claude (Opus), an AI assistant, posting under Sandy's account and supervision.*

sndpl:
IMG_4289
IMG_4290
[t471_v2.zip](https://github.com/user-attachments/files/32239113/t471_v2.zip)

m9710797:
I've asked another AI model for a review. Here's what it came up with (I didn't have time yet to check myself):

[quote]
Negative displayStart and getLineZero()

After frameStart and before the reset, displayStart is the previous origin expressed in this frame, so it is negative. getLineZero() is displayStart / TICKS_PER_LINE. C++ integer division truncates toward zero, so this is not floor-division: a value that should be line −230 can come out as −229.

SpriteChecker and PixelRenderer call getLineZero() live. For normal frames the last top-border sprite line is after the reset, so they are fine. During SBB-style overscan, isDisplayArea stays true into the next top border, and those few lines can be drawn/checked against the wrong Y. Unlikely to reintroduce the shake, but it is a real consequence of making displayStart straddle the frame boundary.
[/quote]

If I understand correctly it says `displayStart` can be negative (in some regions), and then the division rounds to the wrong direction. But it's not clear to me if this is also an actual problem.

sndpl:
The review is right, and it is measurable inside the emulator. `machine_info VDP_msx_y_pos` goes through the same `getLineZero()` as `PixelRenderer` and `SpriteChecker`, so I sampled it every 20 µs across a frame boundary on the emulated FS-A1ST (neutral adjust, counter reset at raster line 15):

```
              raster 258..261        raster 0..14                   raster 15
before fix:   216 217 218 219   |   219 220 221 ... 233        |   -27
after fix:    216 217 218 219   |   220 221 222 ... 234        |   -27
```

Truncating division repeated line 219 at the frame boundary; with floor division the counter simply keeps counting until the reset, which is what the chip does. Same picture on the PAL NMS8250 (247 | 247 248 … before, 247 | 248 249 … after). Pushed as a second commit: `getLineZero()` now floors, and it is the only place that divides `displayStart`.

On the impact, it is worth being precise, because the same lines were much *more* wrong before this PR:

* Those raster lines (0 to 7+vadj) only carry picture when software keeps the display on across the frame boundary (the SBB overscan); otherwise they are border and the value is unused.
* openMSX renders from raster line 18 (60Hz) / 45 (50Hz), so the region is visible only when `8 + vadj > 18`: never in 50Hz, in 60Hz for vertical adjusts of 11 and up — for SBB that is `SET ADJUST` y = −6..−2, one to five lines at the very top of the picture.
* Before this PR those lines were drawn against the *new* frame's `displayStart`, i.e. off by `6 + ds_new − ds_prev` lines: 6 even when LN did not change (262 − 256), 16 for SBB's 32 → 42 change. So: was 6–16 lines off, was 1 line off with truncation, exact now.

For sprites the same value feeds `displayDelta` and `collisionY` in that region; the `l0 = getLineZero() − 1` "last top-border line" check is negative there and does not fire, which is right — the real one happens after the reset. None of this touches interrupt timing; T471 v3 and the SBB recordings are unchanged with the fix.

What is *not* verified against hardware is what the chip displays in that top border during overscan — the model says pattern row `L + 262 − ds_prev`, which is the natural reading of a counter that keeps running, but nobody has photographed it. That would be a display test rather than an FH test, and it is independent of this PR's interrupt fix.

---

*Transparency note: this analysis, the code and this comment were produced by Claude (Opus), an AI assistant, posting under Sandy's account and supervision.*

sndpl:
Follow-up: the floor fix alone did not make the carried-over rows render correctly, because `PixelRenderer` and `SpriteChecker` work lazily. Lines between their last sync and the counter reset were drawn against the *new* line zero. Third commit adds `Renderer::updateLineZero()` and a `SpriteChecker::sync()` at the reset, the same pattern as a vertical scroll write.

Checked with a screen 5 ruler (VRAM row *r* has a white pixel at x = *r*) under overscan on the emulated FS-A1ST, vertical adjust nibble 8 (reset at raster 23, display start 50):

```
raster 18..22   raster 23..49        raster 50..
before: 224..228   229..255            0 1 2 ...    (carry-over 6 rows off)
after:  230..234   229..255            0 1 2 ...    (= 262 - 50 + L, then (L - 50) & 255)
```

Two things worth knowing for the overscan side of this, both independent of the interrupt fix:

* In IKA9958 the latch that gates active video (`gc103`/`gc106` in `IKA9958_st.sv`) is set by the 192/212 match and cleared only when the counter reaches 0 again; the reset to −27/−17 does not touch it. So once the end-of-display match is missed, the picture stays on through the frame wrap and the top border until display start. openMSX's `isDisplayArea` behaves the same way.
* Which VRAM rows the chip fetches in that stretch is not in the die-shot RTL yet (address logic is still open there). openMSX shows `(L − displayStart) & 255` after the reset, i.e. rows 229..255 and then 0, which is what overscan demos have always looked like in the emulator. Sandy can photograph that region on the FS-A1ST with a ruler picture, so we can check both the post-reset rows and the carry-over rows at a large vertical adjust.

T471 v3 and the SBB recordings are unchanged.

---

*Transparency note: this analysis, the code and this comment were produced by Claude (Opus), an AI assistant, posting under Sandy's account and supervision.*

PR #2193, opened 2026-09-11
VDP: the access grid at the vertical border, and the last six command start delays

sndpl:
Three changes, on top of current master.

1. **The access grid switches one line before the display area.** The VDP fetches
   the sprite data of a display line during the line before it, and the grid follows
   the fetching: it switches to the display tables 162 cycles into the line before
   the display area and back 162 cycles into the line after it — the same point in
   the line at both ends. openMSX kept `tabScreenOff` for both of those lines, which
   gives a command **154 slots where the hardware gives it 44** on the line before
   the display area. For one line a frame a command therefore ran several times too
   fast.

2. **The command start delay of the other six commands**, from the new `*i.vcd`
   captures: POINT 45, LMCM 58, PSET/SRCH/LMMC 70 and HMMC 94 cycles from the rising
   `/CSW` edge, so 63, 76, 88 and 112 from openMSX's port-write timestamp. All twelve
   commands now start when the hardware does.

3. **doc**: the analysis behind both, plus what the new character and text captures
   say.

### How it was checked

* The two new switch moments fire at exactly cycle 162 of the line before and the
  line after the display area, verified against the emulator's own frame counters.
* The generated slot tables, dumped from the real `VDPAccessSlots.cc` and compared
  against the independent model in `doc/internal/vdp-vram-timing/`: **120384 table
  entries, 304 slot classes and 6272 padding spans agree exactly.**
* The twelve command startup values come from one shared-pin-delay fit over all 328
  launches, which leaves 4 of them unsatisfied at the best offset against 15 at the
  worst.
* C-BIOS boots unchanged, and the too-fast and dummy-read paths behave as before.

### Savestate format

`VDP` goes to version 12: `isSlotTableDisplayArea` and its two sync points. Older
savestates load with it equal to `isDisplayArea`, which differs for one line a frame.

### Not here

The left blanking of the line before the display area runs the sprites-off comb
rather than the display-off one — fifteen slots six cycles later instead of sixteen
— which would need a fourth table for one blanking region per frame. The character-
mode line padding (+2 at 1341, +1 at 1352, measured) is documented but not
implemented.

---

*Transparency note: this branch, its analysis and this description were produced by
Claude (Opus), an AI assistant, working under Sandy's supervision.*


src/video/VDPAccessSlots.cc
```diff
@@ -264,7 +264,8 @@ struct CycleTable : AccessTable
 
 		size_t out = 0;
 		for (auto idx : xrange(NUM_DELTAS)) {
-			bool cmd = (FIRST_CMD_DELTA <= idx) && (idx < LAST_CMD_DELTA);
+			bool cmd = ((FIRST_CMD_DELTA <= idx) && (idx < LAST_CMD_DELTA)) ||
+			           ((FIRST_CMD_DELTA_2 <= idx) && (idx < LAST_CMD_DELTA_2));
```
m9710797: I don't like this. Can't we make all `CMD_xxx` a single continuous chunk? Sure that means in this PR you can't just append new values at the end ... you need to insert in the middle, and/or shuffle some things around.

While you're add it, maybe also place `CPU_16_ANY` next to the other `CPU_xxx` values (because in a previous PR you already introduced another such non-contiguous check: the check for `any` in the line below, I think it would be cleaner if that were also just inside the range `[FIRST_CPU_DELTA, LAST_CPU_DELTA)`.

m9710797 (commented):
In the commit message you say `POINT` is the only command that doesn't write. That's not true: `SRCH` also doesn't write.

But it's indeed weird that `POINT` is the only command that deviates from the pattern. How confident are you in the analysis of the measurements? Should I make more measurements to increase the confidence? If so, then which commands?

doc/internal/vdp-vram-timing/2026-measurement-analysis.md
```diff
@@ -872,3 +872,124 @@ Not emulated, and why:
...
+matter: openMSX performs the transfer access at the moment the byte arrives rather than at
+the next slot, which is a TODO that predates this work.
+
+## 18. Character and text modes
+
+The `*i.vcd` set also adds character and text captures: `noCpu` with R#9 S1/S0 = 0 and = 16,
+and `wrCpu` at 72 cycles, for screens 1, 2, 3 and 4 and for both text widths, each with the
+display off, sprites off and sprites on.
+
+Text mode has no refresh chain this analysis can find -- the refresh accesses do not carry
```
m9710797: Please re-read `vdp-timing-2.html`. There I documented the refresh reads in the text modes. They are clustered together, so not very useful as anchor points to interpolate the timing.

But if you look at the RAS signal directly, there's probably still sufficient info to do a reliable timing mapping from the .vcd file to VDP clock cycles. And again look at `vdp-timing-2.html` for more hints, or at least cross-check our findings.

doc/internal/vdp-vram-timing/2026-measurement-analysis.md
```diff
@@ -872,3 +872,124 @@ Not emulated, and why:
...
+| YMMM | **82** | 100 | R source | 29/29 |
+| HMMV | **94** | 112 | W dest | 28/28 |
+| LINE | **94** | 112 | R dest | 28/28 |
+| HMMC | **94** | 112 | W dest | 28/28 |
+
+Eleven of the twelve are `46 + 12k`, so section 11's `S0 = 10 + 12N` survives with
+N = 3, 4, 5, 6, 7. **POINT is the exception**: 45, one cycle below LMMM, whose first access
+is the same source read. It is not a rounding artefact -- at the fitted `phi` the admissible
+band for POINT is the single value 45 over all 26 launches, and 46 is outside it. POINT is
+the only command that never writes anything.
```
m9710797: As mentioned in another review: `SRCH` also never writes.

doc/internal/vdp-vram-timing/2026-measurement-analysis.md
```diff
@@ -872,3 +872,124 @@ Not emulated, and why:
...
+* **TEXT1 and TEXT2 share the table**, as openMSX assumes.
+* **The character table does not change when sprites are disabled.** `vdp-timing-2.html`
+  lists this as unmeasured and guesses that the sprite bandwidth becomes available to the
+  CPU as it does in bitmap modes. It does not: sprites off scores exactly the same 100 % on
+  the unchanged table.
+* **Nor when the display is disabled**, in either character or text mode. For text that
+  answers the other open question in `vdp-timing-2.html`: screen-off in text mode is *not*
+  the common screen-off schema, it is the text schema (0 % of text writes land on the
+  display-off table). Both confirm what openMSX does after
+  [issue 1754](https://github.com/openMSX/openMSX/issues/1754).
```
m9710797, line 979: Did I understand correctly:
* Bitmap screen modes have 3 access modes: `dispOff`, `sprOff` and `sprOn`.
* But BOTH text and character modes only (each) have a single access mode, doesn't matter if sprites are on/off (for character mode) or if screen is off?

m9710797 (commented):
Just checking: does this implementation also work in "overscan mode"? (You can read https://github.com/openMSX/openMSX/pull/2191 for some background info on how overscan works .. play with the 192 vs 212 line counter).

We discussed rotating the openMSX start-of-line so that it's aligned with the VDP's start-of-line (at least for this aspect of the access slots). And then you decided against it because it would be quite invasive (I agree), and scheduling in the middle of the (openMSX) line is simple enough. But for my understanding, if we were to do this, then `slotTableEndSyncTime` and `vScanSyncTime` would coincide, right? And then we only need to add one additional `syncPoint` instead of two? Or did I miss something? Will that also change some other timing (e.g. maybe we can't change `vScanSyncTime` without influencing other stuff).

sndpl:
All four review points done and pushed; the two questions answered below, and the pointer to
`vdp-timing-2.html` paid off twice — it gave me the anchor the text captures needed, and the
result cross-checks what I had fitted.

# The review

**`SRCH` also never writes — you are right**, and that kills the observation. POINT and SRCH
are both read-only and they are 25 cycles apart, so "never writes" explains nothing.
Corrected in the commit message and in the doc, where the sentence now reads that the value
follows neither the operand, nor the accesses per unit, nor byte-versus-pixel, nor whether
the command writes at all.

**The `Delta` ranges are contiguous again.** `CPU_16_ANY` sits with `CPU_16` and `CPU_28`,
the command steps and the startup delays are one range, and `NUM_DELTAS` is still 22 because
the reshuffle only renumbers. The only thing left that singles out `CPU_16_ANY` is one named
constant for the one thing that is actually different about it:

```cpp
bool cpu = (FIRST_CPU_DELTA <= idx) && (idx < LAST_CPU_DELTA);
bool skipTight = cpu && (idx != CPU_ANY_DELTA);
```

**Overscan: it works, and testing it found one real hole, now closed.** I drove R#9 bit 7
the way Sex Bomb Bunny does — LN=1 early, LN=0 mid-frame — with the phase of the second
write swept across the whole frame, 840 frames, and logged the two switches together with
`isDisplayArea` and the frame starts. Of those frames:

| | frames |
|---|---|
| normal: grid on, grid off, VSCAN | 788 |
| **no events at all, grid still on from the previous frame** | **47** |
| grid on, never off (overscan starts) | 2 |
| grid off + VSCAN, no start (grid was already on) | 2 |
| grid on, VSCAN, but no grid off | **1** |

The 47 are the overscan frames: when the LN write moves the display end into the past,
`scheduleVScan()` removes the pending sync point and does not re-add it, my end switch
follows because it is derived from the same `vScanSyncTime` by a constant, and the grid
correctly stays on across the frame boundary — exactly as `isDisplayArea` does.

The last row is the hole. If a reschedule lands in the 40 cycles between my end switch and
VSCAN, the old end sync point is removed and the new one is already in the past, so the grid
stays on for a whole extra frame. One frame in 840 of deliberately pathological toggling,
but it is a real bug. Both handlers now carry a backstop:

```cpp
void VDP::execDisplayStart(EmuTime time) {
    setSlotTableDisplayArea(true, time);   // normally switched a line ago
    ...
void VDP::execVScan(EmuTime time) {
    setSlotTableDisplayArea(false, time);  // normally switched 40 cycles ago
```

Those are correct unconditionally — the grid is always on at display start and always off at
VSCAN — and in overscan VSCAN is not reached at all, so they do not disturb it. With them
the backstop fires exactly once in the same 840-frame test and never in a normal boot, and
the toggling of the flag (with the `cmdEngine->sync()` that goes with it) is now in one
place, `setSlotTableDisplayArea()`.

# Would one sync point be enough if we rotated the line?

No, I don't think so, and the reason is that the 40 cycles are real rather than an artefact
of the origin. `displayStartSyncTime` and `vScanSyncTime` are both at cycle **202** of their
line, where VR flips; the grid switches at cycle **162**. Rotating openMSX's line origin by
162 makes my two switches land on line boundaries, but it moves VR by the same 162, so the
two are still 40 cycles apart and I would still need my own sync point at the end.

They would coincide only if `vScanSyncTime` itself moved to 162 — and that is the "influences
other stuff" you suspected: VSCAN is the VBLANK IRQ and the moment `isDisplayArea` drops for
the renderer and the sprite checker. Moving it 40 cycles to tidy up a sync point would be a
bad trade.

The start side is worse either way: even with the line rotated, the grid start is a whole
line plus 40 before `displayStartSyncTime`, so it needs its own sync point regardless. So
the rotation buys a nicer set of constants (1368 and 0 instead of 1408 and 40) and nothing
else. Two sync points per frame is not a cost worth restructuring for.

# How confident am I in POINT = 45, and what would raise it?

Confident that it is what these captures say; less confident that it is the *rule*.

What the fit gives is sharp. At the shared pin delay that minimises the total error over all
328 launches, POINT's admissible band is the single value 45 over all 26 of its launches, and
46 lies outside it — it is not a case of 45 winning by one launch. And the fit is not fragile:
the optimum sits in a plateau where the eleven other commands keep their values, and the
total unsatisfied count is 4 of 328 there against 15 at the worst offset.

What I cannot rule out from these captures alone:

* they are all **display off, one machine, one session**. The other eleven commands were
  measured the same way and they land on the grid, which is reassuring, but a systematic
  quirk of the POINT setup code — a different register order, a different last write —
  would look exactly like this.
* 26 launches is the smallest set of the twelve.

**If you want to spend a session on it, the measurements I would ask for**, in order:

1. **POINT with sprites off and with sprites on**, `nx4-ny2`-style, a few captures each. This
   is the strongest one: display-off pins a single integer, and the other two grids have to
   agree with 45 plus the sprite addend. A setup quirk would not survive changing the grid
   under it.
2. **POINT and LMMM in the same capture**, alternated, so the 1-cycle difference is measured
   against the same register-write sequence and the same time base. That removes everything
   except the commands themselves.
3. **LMCM**, the other interesting one: 58 is the only N = 4, and it is the only command
   whose first access is a source read that does not share a value with another command. It
   was 28 of 28 launches, so I have no reason to doubt it, but it is the second thing I would
   confirm.

I would not bother re-measuring the other ten: they agree with the grid and with each other.

# Text mode: your pointer was the answer

`vdp-timing-2.html` has it — seven refresh reads per line instead of eight, clustered at
74 to 122 instead of spread over the line. That is why my chain detector found nothing: it
looks for the 128-cycle row-incrementing chain, which text simply does not have.

The cluster works as an anchor, and the captures reproduce your 2013 numbers exactly: 74, 82,
90, 98, 106, 114, 122, the two dummy reads at 230 and 238, and the rendering groups from 246
in steps of 48, each `g, g+18, g+24, g+30, g+36, g+42`. Twelve years and a different machine.

**Cross-check.** Anchoring on the cluster rather than on the mark I had fitted puts 90–95 %
of the CPU writes on a `slotsText` slot and uses 46 of the 47, against the fit's 100 % and 45.
Same answer; the difference is that a run of CPU writes in the eight-apart slots can
impersonate the cluster in some lines. And your documented 47-slot list *is* openMSX's
`slotsText`, which is the third independent route to the same table.

**And it narrows the padding.** Anchored on the cluster, every access from 74 through the
last rendering read at 1200 is at an identical position with S1/S0 = 0 and = 16. So all three
removed cycles are between cycle 1200 and the next line's cluster at 74 — the right blanking,
the same region as in the bitmap and character modes. Pinning them exactly needs an access
*inside* that region, and the CPU slots at 1206..1362 are there: **a text `wrCpu` capture
with S1/S0 = 16** would do it. The `noCpu` captures that exist cannot, because nothing is
accessed there.

# And yes, your summary is right

* bitmap: three access modes, `dispOff` / `sprOff` / `sprOn`;
* character: one, whether sprites are on or off and whether the display is enabled or not;
* text: one, likewise.

Measured that way: in character mode all three states put 100 % of their CPU writes on
`slotsChar` and use the same 30 slots, and in text mode all three put 100 % on `slotsText`
and 0 % on the display-off table. The only wrinkle is the one this PR is about — the
*vertical border* of a bitmap screen is the display-off table, and the boundary between them
is not where the display area is.

---

*Transparency note: this analysis, the code and this comment were produced by Claude (Opus),
an AI assistant, posting under Sandy's account and supervision.*

PR #2194, opened 2026-09-12
MSXPiDevice: emulate the MSXPi CPLD v1.6 at I/O-port level

costarc:
## MSXPiDevice: CPLD v1.6 /WAIT flow control, and two transfer fixes

Updates the existing MSXPi device for MSXPi interface v1.6. No other device is
touched. The new flow control is negotiated, so a ROM that does not ask for it
keeps using the polled path exactly as before.

### Hardware /WAIT flow control

MSXPi v1.6 adds a CPLD that asserts /WAIT while a byte is in flight, letting the
MSX move payloads with `INIR`/`OTIR` at 21 T-states per byte instead of polling
the status port for each one.

Wait mode is what the MSXPi ROM's Ethernet UNAPI transport (`ETH_BEGIN` …
`ETH_END`) and its disk burst transfers both use. The ROM probes for it and
falls back to the polled path when the read-back says it is unavailable, so
without this the fast path simply cannot be reached — or tested — under
emulation.

Port `$57` becomes a wait-mode register as well as the version port: writing
`$01` enables wait mode, `$00` disables it, other values are ignored (matching
the CPLD's `mode_reg`). A read returns `$FF` with wait mode on, `$FE` with it
off.

Those values matter. Existing MSXPi binaries ask "am I emulated?" on every byte
with `in a,($57)` / `cp $FE` / `jr c,physical`, so anything below `$FE` means
"real hardware" to code already in the field. Real CPLD v1.6 answers `$0E`/`$8E`,
which would therefore send stock code down the physical path under emulation.
`$FE`/`$FF` keep both states above that threshold, and real hardware can never
collide with them because the CPLD pins bit 6 low, capping `$57` at `$BF`.

In wait mode a data-port read blocks until the byte arrives, bounded at 250 ms
so a dead server cannot freeze openMSX. Two optional settings, both off by
default: `<wait_cycles>` charges the guest for the stall via
`MSXCPU::waitCyclesZ80` (as `VDP.cc` and `TurboRFDC.cc` do), and
`<rdy_fail_every>` injects a low `RPI_READY` every N reads to exercise retry
paths without hardware.

### Received bytes are no longer discarded

`readLoop()` capped the queue at 16 KB and dropped the rest, so the device lied
about what it had received — no error anywhere. A 16384-byte block with its
header and checksum is 16389, five over the cap, so large transfers lost their
tail including the checksum byte the MSX then waited on forever. It now checks
for room before reading; the socket buffer fills and the server blocks, which is
how real hardware behaves anyway. Bytes are handed over in 512-byte batches
rather than under one lock, since the emulated Z80 takes that mutex on every
status poll.

### TCP_NODELAY on the server socket

Each `OUT` to the data port is its own one-byte write. With Nagle plus the
server's delayed ACK that cost ~40 ms each, so a 512-byte burst write took
minutes and looked like a hung transfer. The protocol is request/response, so
there is nothing to coalesce. Same binary except this `setsockopt`: 50 s with
it, still stalled past 600 s without.

### Testing

Verified on real hardware (Canon V-25, MSXPi v1.6) and under openMSX: 256 KB in
51.95 s down from 1:45.00, byte-for-byte round trips, and InterNestor Lite
resolving names through the device. `share/extensions/MSXPi.xml` adds the sha1
for the v1.6 BIOS; earlier ROMs still resolve by their own hashes.


costarc:
Please hold. 
I have one more change to add.

share/extensions/MSXPi.xml
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" ?>
+ <?xml version="1.0" ?>
```
m9710797: Accidental edit? Please correct this.

From automatic SonarCloud Code Analysis:
`Remove all characters located before "<?xml".`

share/extensions/MSXPi.xml
```diff
@@ -20,6 +20,7 @@
         <ROM id="MSXPi BIOS">
           <mem base="0x4000" size="0x4000"/>
           <rom>
+            <sha1>7ab3bd277598af4db3a50357c4644dace720214e</sha1> <!-- Build 1.6.20260912.050 - BASIC BIOS, MSX-DOS1, UNAPI and Performance improvements-->
             <sha1>3cfffa59afef160439085a8a496953edcaeb0662</sha1> <!-- Build 1.5.20260808.020 - BASIC BIOS and MSX-DOS1-->
             <sha1>c429077a5f58b006125247a67783ea4ea6fb7389</sha1> <!-- Build 1.2.20250923.792 - BASIC BIOS only-->
             <filename>msxpibios.rom</filename>
```
m9710797, line 36: In `MSXPiDevice.cc` below, you added two new tags `wait_cycles` and `rdy_fail_every`. I understand these are optional. Nevertheless it can be useful to add those here (possibly with their default value). E.g. this is useful as documentation that these options exist, and then where exactly they should be placed, etc.

src/MSXPiDevice.cc
```diff
@@ -1,13 +1,20 @@
 #include "MSXPiDevice.hh"
+#include "MSXCPU.hh"
 #include "Timer.hh"
+#include <chrono>
+#ifndef _WIN32
+#include <netinet/tcp.h> // TCP_NODELAY (winsock2.h provides it on Windows)
+#endif
```
m9710797: In general it's a good idea to try to minimize the number of `#ifdef` blocks in the code. Our `Socket.hh` header in openMSX is used to abstract platform specific network related stuff. So preferable all network related `#ifdef` stuff should be located in the `Socket.{hh,cc}` files.
* So instead we should move this `#include` to `Socket.hh`.
* Or even better: we should add a `sock_setNoDelay(SOCKET sd)` helper function in `Socket.hh`, similar to how there already is a `sock_setNonBlocking(SOCKET sd)` function.
* Follow up question, maybe not for this pull request: should we then get rid of the `sock_setIntOption()` and `sock_getIntOption()` helpers, and replace the current callers with more specific helpers?

src/MSXPiDevice.cc
```diff
@@ -45,6 +53,48 @@ byte MSXPiDevice::readIO(uint16_t port, EmuTime time)
...
+			// openMSX way to model wait states (MSXCPU::waitCyclesZ80,
+			// as used by VDP.cc and TurboRFDC.cc) and it advances
+			// EMULATED time, which blocking this thread does not.
+			//
+			// It cannot replace the block below: waitCycles takes a
+			// count known in advance, whereas the byte itself arrives
+			// from a socket whenever the server gets round to it.
+			if (waitCycles > 0) {
+				time = getCPU().waitCyclesZ80(time, waitCycles);
+			}
```
m9710797: I'm not sure this is correct. But maybe I don't fully understand the intention. So let's discuss.

This is my understanding, but maybe I'm wrong, then please correct me:
* We don't exactly know when the next data will arrive. It might take up to 250ms (I understand usually it's less than that).
* There is a mode in the hardware where we block the Z80 when no data is available yet. So from the Z80 point of view, every `IN` instruction always returns a byte (no need for a retry loop in Z80 code).
* We want to emulate the same behavior in openMSX.

But there's a problem in the current implementation:
* The `rxCx.wait_for()` call below, can block anywhere from 0ms to 250ms (I'm simplifying a bit).
* BUT the call to `waitCyclesZ80()` always waits for the *same* amount of time. And it even waits if `rxQueue` is not empty.
* Instead we should measure how much *real* time the `rxCv.wait_for()` took. And then pass a similar amount of time to `waitCyclesZ80()`.
* Also `waitCyclesZ80()` is probably not the correct function. Why only on Z80 (not also on R800). I think `MSXCPU::wait()` might be more appropriate. To that function you pass a destination time (must be in the future) instead of a number of cycles, that might also be easier to integrate with the "variable" wait duration. So then it becomes something like `getCPU().wait(time + EmuDuration::usec(host-wait-time-in-micro-seconds)`.

Next there might be a practical limitation in openMSX. This `readIO()` function gets called from the main thread. This main thread does the MSX emulation, but also sound generation, video rendering, interaction with the (built-in) GUI, ... But now if `rxCv.wait_for()` indeed block up-to 250ms, then all these tasks are blocked. For sure this will cause a buffer under-run in sound generation, VDP rendering will stall for up-to 15 NTSC frames, GUI will be unresponsive, ... This IS a real problem and TBH I don't know a good solution for it (in the current openMSX architecture .. till now we could assume that *every* CPU instruction will finish in a fairly short amount of time (at most a few 100 cycles)). But if this is a very infrequent situation then maybe we can live with it (and have pops in the audio etc).

src/MSXPiDevice.cc
```diff
@@ -70,8 +120,23 @@ byte MSXPiDevice::peekIO(uint16_t port, EmuTime /*time*/) const
 			if (!rxQueue.empty()) return 0x02; // byte available
 		}
 		return 0x00;
-	case 0x57: // Version
-		return 0xFE; // Used to know when its the extension and not the actual physical interface
+	case 0x57: // Version + wait-mode read-back
+		// This port has to answer TWO questions at once: "is wait mode on?"
+		// and "am I openMSX or real hardware?".  msxpi_bios.asm:97 asks the
+		// second one on EVERY BYTE:
```
m9710797: Can you explain this is more detail. In general in openMSX we do NOT want to emulate features that let the MSX software test if they are running an a real MSX or on an emulator. Why should we deviate from this rule here? Is there any reason why the `msxpi_bios` MUST know this information?

src/MSXPiDevice.cc
```diff
@@ -110,6 +191,19 @@ void MSXPiDevice::writeIO(uint16_t port, byte value, EmuTime time)
...
+	// and on the heap rather than the stack - a buffer this size is not
+	// something to put on a thread stack.
+	//
+	// MAX_QUEUE_SIZE is deliberately far larger than any single block: the
+	// queue only ever grows to what is actually used (cb_queue starts at zero
+	// capacity and doubles), so the ceiling costs nothing until a transfer
+	// needs it.  Now that nothing is discarded it is a throughput knob rather
+	// than a correctness limit: too small merely stalls, it cannot lose data.
+	static constexpr size_t MAX_QUEUE_SIZE = 64 * 1024;
+	std::vector<char> buf(64 * 1024);
```
m9710797: Detail: use `MAX_QUEUE_SIZE` instead of repeating `64 * 1024`?

Any reason why you switched from a `std::array` to `std::vector`? A valid reason could be to not put too large arrays on the stack? But 64kB is probably still OK. Especially because this code runs on a separate thread which doesn't call too many other functions.

There is one difference:
* The `array` was left uninitialized.
* The `vector` is initialized with zeros.  (And has some overhead of heap-allocation and indirection.)

But that initialization is needless work (`sock_recv()` anyway overwrites the content of `buf` before we use it). OTOH this is done just once when the helper thread starts, so it doesn't really matter.

If you don't want to use `std::array` for some reason, and also don't want the cost of zero-initialization, then maybe our `MemBuffer` class is a good alternative (read the documentation on lines 16-30 in `src/utils/MemBuffer.hh`).

src/MSXPiDevice.cc
```diff
@@ -123,6 +217,18 @@ void MSXPiDevice::readLoop()
...
+			// Nagle enabled those are held back until the previous one is
+			// acknowledged, and the server's delayed ACK makes that ~40 ms
+			// EACH: a 512-byte /WAIT burst write then takes minutes instead
+			// of milliseconds, which looks exactly like a hung transfer.
+			// The protocol is request/response, so there is nothing to
+			// coalesce anyway.
+			{
+				int one = 1;
+				setsockopt(sock, IPPROTO_TCP, TCP_NODELAY,
+				           reinterpret_cast<const char*>(&one), sizeof(one));
```
m9710797: See my comment above, maybe this should become a helper function in `Socket.hh`?

m9710797 (commented):
Hi. I don't know if the code was ready for review already. Maybe in the future you could use the github feature to mark your PR as draft? That's easier (more clear) than using plain comments.

Anyway, here's some (early?) feedback.

src/events/Socket.hh
```diff
@@ -48,9 +48,10 @@ void sock_setNonBlocking(SOCKET sd);
 void sock_setIntOption(SOCKET sd, int level, int optName, int value = 1);
 // Get an integer socket option (e.g. SO_ERROR), 0 on failure.
 [[nodiscard]] int sock_getIntOption(SOCKET sd, int level, int optName);
+// Disable Nagle's algorithm (TCP_NODELAY): send small writes immediately.
+void sock_setNoDelay(SOCKET sd);
 // Non-blocking readiness poll (zero timeout): data ready or pending connection.
-[[nodiscard]] bool sock_readable(SOCKET sd);
-// Build an IPv4 sockaddr_in (network order); hostIp==0 -> INADDR_ANY.
+[[nodiscard]] bool sock_readable(SOCKET sd);// Build an IPv4 sockaddr_in (network order); hostIp==0 -> INADDR_ANY.
```
m9710797, line 54: Accidental wrong edit. Wrongly removed newline?

costarc:
@m9710797

Thanks for all the feedback, review and suggestions.
After reading this question from you: 
"[m9710797](https://github.com/m9710797)
[8 hours ago](https://github.com/openMSX/openMSX/pull/2194#discussion_r3999116190)
Contributor
Can you explain this is more detail. In general in openMSX we do NOT want to emulate features that let the MSX software test if they are running an a real MSX or on an emulator. Why should we deviate from this rule here? Is there any reason why the msxpi_bios MUST know this information?"

I realized my aproach was not ideal. The emulated device has indeed differences on its behaviour compared to the actual hardware. I know some compromise must be done, for example, as far as I know, /wait states cannot be implemented in openMSX devices to behave exactly like the actual hardware signal, therefore some behavioural implementation needs to be implemented.
Despite potential contstraints, such as my lack of deeper knowledge in both openMSX and C++, I commited to make a better device emulation, and with assistance of an AI tool, redesigned and re-wrote the MSXPiDevice to better mirror my hardware interface.
In this design, there is no need for any client to test if its running in a actual hardware on in the emulator. 
The application cannot also findout if the MSX is emulated, unless it tries by analyzing timings in operations.

Then here it is, I am changing for draft for now to receive the inputs and apply further changes as needed.

Addressing your previous comments individually - although many are now no longer relevant, since most of the code are now gone or your suggestions applied.

Looking at the CPLD source again, the hardware never stalls while waiting for data: /WAIT is wait_mode and SPI_RDY and ..., and the Pi only raises READY once it has the byte and is about to clock it. With READY low, an IN in wait mode returns immediately with the old byte, and the driver checks $56 before a burst.

So I rewrote the device to model the CPLD instead of a byte queue. The server now tells the device when "READY" is up and with which byte, over the socket, before the MSX asks for it. readIO() never waits on the socket. A /WAIT stall is exactly one emulated SPI transfer, done with getCPU().wait(doneTime) as you suggested, so it also works for the R800. wait_cycles and the 250 ms wait_for are gone.

The only remaining wait is when the MSX polls $56 for a byte the server hasn't offered yet. There the device waits only while emulation is ahead of real time (RealTime::timeLeft()), time that would otherwise be slept, so it can't cause underruns or VDP stalls. It brought a 256 KB copy from 98 s to 40 s, against 35 s on real hardware.

I am also applying the process we agreed previously, which is keep the sha1 for the previous released roms in the MSXPi.xml, and making the previous roms available for download in the MSXPi repository.

I will continue to run test for some more days, to assure there is not any change needed to the ROM, which is where the bigger impacts are. Once I am happy with the new implementation, I will drop a message requesting the merge.
In the mean time, fell free to review and contribute.
Thanks.

m9710797:
Thanks a lot for your changes (and making them so quickly).
I only had time to skim your latest version. At first sight I like the changes. I'll try to make time the following days to look at them in more detail.

src/MSXPiDevice.cc
```diff
@@ -1,152 +1,395 @@
...
+	bound.reset();
+	srValue = 0x00;
+	waitMode = false;
+}
+
 void MSXPiDevice::reset(EmuTime /*time*/)
 {
+	// The CPLD does not see the MSX reset line, but the device registers
+	// clear at power-up, and a machine reset in openMSX must not leave a
+	// stretched transfer behind.
```
m9710797, line 71: Question for clarification:
I'm confused by this comment, you say in real hardware the CPLD doesn't see the MSX reset, but then in the emulation you anyway call `cpldReset()`?

You say "... reset must not leave a stretched transfer behind ...". Can't the same happen on real hardware? If so, then openMSX should be able to reproduce that situation (even if it is undesirable).

If you need different behavior on reset and power-on, you can override the `MSXDevice::powerUp()` method (in addition to `MSXDevice::reset()`, and if you need there;s also `MSXDevice::powerDown()`). By default `powerUp()` does the same as `reset()`, but a few devices (and maybe `MSXPiDevice` in one of them) want a different behavior for both.

src/MSXPiDevice.cc
```diff
@@ -1,152 +1,395 @@
...
+
+void MSXPiDevice::waitForPeer(EmuTime time)
+{
+	// Emulation runs ahead of real time in slices and then sleeps to throttle.
+	// A byte the server offers during that sleep would only be seen at the
+	// start of the next slice, so every request/response exchange with the
+	// server would cost a whole slice of emulated time - far more than the
+	// real Pi needs. Instead, spend the time emulation is ahead of real time
+	// here, waiting for the offer. That is time RealTime would have slept
+	// anyway, so this never makes emulation fall behind real time.
```
m9710797, line 169: Just a remark, no need to change anything (because I don't know how to do better).

This is an interesting approach: making use of the time emulation runs ahead of real-time.

However keep in mind that next to emulation of the Z80 and other MSX devices, openMSX also has to generate the video output (+ scaling), do audio generation, render the GUI, execute Tcl scripts, .... Normally these tasks run in the rest of the real-time-slice. But now if we eat up that slice waiting for network data, there might not be enough time left for those other tasks.

In other words: there's a good reason why openMSX emulates in such time-slices. Running emulation ahead of real-time, hopefully building a sufficient lead, so that we don't fall behind when those other tasks run. IIRC a time-slice is typically the duration of one (MSX) frame.

I don't know if this is a problem, maybe it's not. Or maybe it only becomes a problem on not so powerful host machines (or when the host machine is heavily loaded) when it's only just fast enough to run openMSX smoothly, and now because of the extra wait (in the main thread) it's no longer running smooth.

That being said, it is an interesting approach and I can see how it improves lag in network transfer. But I don't have a good solution for the above problem.

src/MSXPiDevice.cc
```diff
@@ -1,152 +1,395 @@
...
-		if (sock != OPENMSX_INVALID_SOCKET) {
-			readRequested = true;
-		}
+		[[fallthrough]];
+	case 0x5A:
+		update(time);
+		latch = value; // D_buff_msx follows D on every $56/$5A write
+		// spi_en write term: a write starts a transfer, READY or not
+		if (!busy) startTransfer(time);
+		(void)stall(time);
```
m9710797, line 257: Maybe drop the `(void)` and the `[[nodiscard]]` annotation on `stall()`? Because apparently it is valid to call `stall()` and ignore the return value.

Or if you think it's very exception to ignore the return value, and want to make such exceptions explicitly visible in the code, then the current approach is fine.

---

TBH, I'm suprised to see a stall (wait cycles) on the write-side. Why does the Z80 need to be stalled, it doesn't not need to wait for a result to become available. But if the real hardware works like that, then we should also emulate it.

src/MSXPiDevice.cc
```diff
@@ -1,152 +1,395 @@
...
-		if (readRequested) {
-			std::lock_guard lock(mtx);
-			if (!rxQueue.empty()) {
-				return rxQueue.front();
-			}
-		}
-		return 0xff;
+	case 0x56: // "0000000" & (SPI_en_s or not SPI_RDY)
+		return (busy || !ready(time)) ? 0x01 : 0x00;
+	case 0x57: // wait_mode & '0' & "00" & MSXPIVer
```
m9710797, line 233: `wait_mode & '0' & "00" & MSXPIVer`: I think this is verilog (or VHDL?) notation for concatenate bits? In the context of c++ code that notation might be confusing (because `&` has a very different meaning). Actually this comment doesn't say much more than the implementation `(waitMode ? 0x80 : 0x00) | MSXPI_VERSION`, so maybe we don't need this comment at all?

Similar comment above for `"0000000" & (SPI_en_s or not SPI_RDY)`, with the addition that names like `SPI_en_s` and `SPI_RDY` need more explanation (better stick to the c++ variable names, using two names for the same thing only adds confusion).

src/MSXPiDevice.cc
```diff
@@ -1,152 +1,395 @@
...
+			if (!busy && ready(time)) startTransfer(time);
+			return shiftRegister(stall(time));
 		}
-		return 0xff; // No data ready
+		return shiftRegister(time);
 	default:
-		return 0xff;
+		return 0xFF;
 	}
 }
```
m9710797, line 226: May I ask to document to *exact* behavior of these 3 IO ports 0x56, 0x57, 0x5A better. I mean I can read the c++ implementation, and that seems to make sense to me. But I don't know what the behavior of the actual hardware is. So I don't know if both match.

I must say I find the interface a bit unusual (at least the way I understand it now):
* On the write-side it's possible to transfer full bytes (ports 0x56 and 0x5A).
* On the read side (port 0x5A) the Z80 reads a byte as 9(?) individual bits? So the MSX software needs to reassemble those bits into a byte?
* In non-wait-mode these bits must be read at precise moments in time (needs specially timed Z80 code, will break on some machines in 6MHz or 7MHz mode, or in a turbor machine in R800 mode).
* In wait-mode, you can just read all bits.

So there is extra hardware (the wait mode) to make reading bits easier (no need for a polling loop, or no need for precisely timed code). But at least from a software programmer point of view, I think, I would be happier if the hardware would assemble the byte for me which I then can read (possibly with a polling loop).

But hey, if the hardware works like that, then we should also emulate it like that. I just want to make sure that my understanding of the hardware is correct.

---

UPDATE:
Maybe I misunderstood. Does WAIT stay active until the full byte is received? And then the full byte can be read as a whole? And these 9 bit intermediate results can only be observed in non-wait mode?

Anyway, to me this code isn't obvious, so maybe it is a good idea to better document it.

src/MSXPiDevice.cc
```diff
@@ -1,152 +1,395 @@
...
+static constexpr byte PROTOCOL_VERSION = 1;
+
+static constexpr byte MSXPI_VERSION = 0x0E; // MSXPIVer "1110"
+static constexpr uint16_t SERVER_PORT = 5000;
+
 MSXPiDevice::MSXPiDevice(const DeviceConfig& config)
 	: MSXDevice(config)
+	, transferTime(EmuDuration::usec(config.getChildDataAsInt("transfer_time", 20)))
+	, readyTail(EmuDuration::usec(config.getChildDataAsInt("ready_tail", 0)))
+	, readyGap(EmuDuration::usec(config.getChildDataAsInt("ready_gap", 0)))
```
m9710797, line 29: Can you explain why, as a normal openMSX user (not for you as a developer), I would ever want to change these 3 parameters? Are there variations of `MSXPiDevice` (I mean real hardware) that do have other values, or can a user reasonably create such variations?

I'm asking if we actually need to make these 3 configurable. Can't those be just constants in the code?

src/MSXPiDevice.hh
```diff
@@ -2,48 +2,121 @@
...
+	// timing of the emulated Pi, in emulated time
+	EmuDuration transferTime;   // E1 -> E10
+	EmuDuration readyTail;      // E10 -> READY low, for a releasing offer
+	EmuDuration readyGap;       // READY low before the Pi offers again
+
+	// --- shared with the reader thread, under mtx ---
+	mutable std::mutex mtx;
+	std::deque<Offer> offers;   // visible offers
+	std::condition_variable offerCv; // signalled when offers become visible
+	std::atomic<size_t> offerCount = 0;
```
m9710797, line 106: I checked the code, I think `offerCount` is *always* equal to `offers.size()`, correct?

So then my question is: why store this redundant information? Maybe the reason is: because then we can read `offerCount` without the need for taking the lock `mtx`?

Is that worth it? Do we access `offerCount` on such a hot code path that we cannot afford taking a lock? It's a tradeoff of course simpler code vs faster code.

Would be nice to add this is a comment:
* (if my understanding is correct) document that `offerCount == offers.size()` and both must always be updated together.
* Document exactly which variables can only be read/written while holding `mtx`.  (E.g. `offerCount` is NOT part of that set (but the current code layout/comments suggest that it is).

src/MSXPiDevice.cc
```diff
@@ -1,152 +1,395 @@
...
+{
+	switch (op) {
+	case OP_HOLD:
+		pendingHold.push_back({arg, true});
+		break;
+	case OP_OFFER: {
+		// a run of holds becomes visible together with its closing offer
+		std::lock_guard lock(mtx);
+		for (const auto& o : pendingHold) offers.push_back(o);
+		offers.push_back({arg, false});
```
m9710797, line 294: Detail, tiny code tweak:
* `offers.push_back({arg, false});` This constructs a temporary `Offer` object and then puts a copy of that object in the `offers` `deque`.
* A tiny bit better is: `offers.emplace_back(arg, false);` This directly constructs the `Offers` object inside the `offers` `deque` (no temporary object, no copy).

(At least in theory, in practice the compilers optimizer might optimize the first version to the same code as the second.)

Caveat: untested. Older clang versions didn't support this yet (for simple structs like `Offer` without an explicit constructor), and some versions of macos shipped with that clang version. I *think* we no longer have to worry about this, but I'm not 100% sure.

src/MSXPiDevice.cc
```diff
@@ -1,152 +1,395 @@
...
+void MSXPiDevice::handleFrame(byte op, byte arg)
+{
+	switch (op) {
+	case OP_HOLD:
+		pendingHold.push_back({arg, true});
+		break;
+	case OP_OFFER: {
+		// a run of holds becomes visible together with its closing offer
+		std::lock_guard lock(mtx);
+		for (const auto& o : pendingHold) offers.push_back(o);
```
m9710797, line 293: Replace this loop with:
`append(offers, pendingHold)`
Caveat: untested, `append()` is our own utility function. I know that for `std::vector` this will work, but I'm not sure about `std::deque`. If it doesn't work then you can still use:
`offers.insert(offers.end(), pendingHold.begin(), pendingHold.end())`

m9710797 (commented):
Hi,

Code seems plausible. But my biggest question is: I don't know exactly how the real hardware behaves, so I can't really tell if the emulation does the same as the hardware.

During the review my understanding of the code grew. So it's possible some of my earlier review comments don't make sense. I did went back and updated/removed some of my earlier comments, but I may have missed some.

So I'd like to get a better understand of the real hardware before I can *really* review and merge this PR.

src/MSXPiDevice.cc
costarc: Thanks for the thorough review. Here is how the real interface works; the source of truth is the CPLD code in the MSXPi repo, hardware/CPLD_Project/MSXPi.vhd (branch release/v1.6).

The interface's heart is a small CPLD, that implements a shift register to convert MSX bytes to serial bits for Raspberry Pi. MSX read and write bytes from/to the Interface using I/O port $5A, using ports $56 and $57 for control.
The interface has s limited number of I/O pins available, reason why the communication with RPi could not be implemented using full bytes (paralell).

The revision 1.6 of the CPLD (and corresponding BIOS routines) implements the following logic:

Start a transfer: OUT ($56),x (x ≠ $FF) or OUT ($5A),x start one. The written value is latched and sent to the Pi. In wait mode, IN ($5A) also starts one, but only if the Pi is ready.
READY: a signal the Pi drives high when it has a byte ready and is about to clock. A 10K pull-down keeps it low when no Pi is there.
$56 read: bit 0 = 1 while a transfer is still running or the Pi is not ready, 0 when the last transfer has completed and the Pi is ready. Other bits are 0.
$5A read: the byte received from the Pi, once the transfer has completed.
$57 read: bits 3–0 are the CPLD version (1110), bit 7 is the wait-mode flag, so $0E or $8E.
$57 write: $01 turns wait mode on, $00 turns it off, other values are ignored.
OUT ($56),$FF: resets the transfer logic ($5A then reads $00) and turns wait mode off.

Normal (polled) software does: poll $56 until 0 → OUT ($56),0 → poll $56 until 0 → IN ($5A). Writing a byte: poll $56 until 0 → OUT ($5A),byte.

In wait mode the CPLD holds /WAIT for the duration of one transfer, so INIR/OTIR can move blocks without polling. It only does so while READY is high: if the Pi is not ready, the I/O cycle is not stretched at all. That is why drivers poll $56 once before a burst.

The use of INIR/OTI and use using /wait states per I/O block instead of bytes in this revison was a breakthrough in terms of transfer speed, and load times were improved by over 3 times, making loading a file from RPi or over the internet comparable to load times form a MFR SD Card.

costarc: > Question for clarification: I'm confused by this comment, you say in real hardware the CPLD doesn't see the MSX reset, but then in the emulation you anyway call `cpldReset()`?
> 
> You say "... reset must not leave a stretched transfer behind ...". Can't the same happen on real hardware? If so, then openMSX should be able to reproduce that situation (even if it is undesirable).
> 
> If you need different behavior on reset and power-on, you can override the `MSXDevice::powerUp()` method (in addition to `MSXDevice::reset()`, and if you need there;s also `MSXDevice::powerDown()`). By default `powerUp()` does the same as `reset()`, but a few devices (and maybe `MSXPiDevice` in one of them) want a different behavior for both.

You're right. The CPLD has no reset input: only power-up and OUT ($56),$FF clear it. An MSX reset leaves it as it was, including an armed transfer. I'll move the state initialisation to powerUp() and make reset() leave the CPLD alone, so openMSX behaves like the hardware there. (A stall cannot span a reset in openMSX, since the stall is just a wait() to a future time, so nothing is lost.)

src/MSXPiDevice.cc
costarc: I'll drop [[nodiscard]] and the (void).

The write-side stall is real hardware behaviour. The CPLD's write latch is transparent, and the Pi only copies it into its shift register when it starts clocking. Without /WAIT on writes, an OTIR would put the next byte on the bus before the Pi has taken the previous one, and bytes would be lost. So the CPLD asserts /WAIT for OUT transfers as well, in MSXPi.vhd: spi_en includes writes, and wait_assert uses spi_en.

src/MSXPiDevice.cc
costarc: Agreed, those comments came straight from the VHDL and don't help in C++. I'll describe the ports in terms of the C++ members instead, and document them once in the header.

src/MSXPiDevice.cc
costarc: Your update is correct. Software always reads a whole byte: in polled mode it waits until $56 reads 0 before IN ($5A), and in wait mode the IN is stretched until the byte is complete. The Z80 never has to assemble bits.

shiftRegister() also returns the half-shifted value because that is what the real register holds mid-transfer. It's only visible to software that reads $5A too early, which correct software never does. I'll add a block comment in MSXPiDevice.hh describing the three ports as in my summary above, and a note that the in-between values only exist for faithfulness.

src/MSXPiDevice.cc
costarc: Yep, there are no hardware variants with other values. transfer_time is roughly how long the Pi takes to clock a byte. ready_tail and ready_gap were knobs I added to try to reproduce a timing race seen on hardware. They didn't reproduce it, so they have no value for users. I'll make them constants: transfer time 20 µs, and no tail or gap. They'll be removed from MSXPi.xml.

src/MSXPiDevice.hh
costarc: Yes, it's an atomic copy of offers.size(), so the $56 status poll (tens of thousands of reads per second during a transfer) doesn't have to take the mutex. I'll add comments stating that the two must always be updated together under mtx, and listing exactly which members are protected by mtx, which are atomic, and which belong to only one thread.

src/MSXPiDevice.cc
costarc: Will do: offers.emplace_back(arg, false). Offer is a plain aggregate, so this relies on C++20 parenthesised aggregate initialisation; I'll build it on the CI platforms to confirm older clang accepts it. If not, I'll give Offer a constructor.

src/MSXPiDevice.cc
costarc: will do.

costarc:
Hi @m9710797.

Just updated the PR with the changes.
I am very happy how it went, hopefully there won't be too many additional changes to make. :)

Edit: I reverted this change, so please disconsider -> **Please note I made a change to build/msvc/genconfig.py - the build under msvc was being tagged as mingw - I thought that was incorrect, and made the change, but let me know if there is more to that than I know, and I revert in my PR.**

How your comments shaped it:
No more blocking. The server now tells openMSX "READY is up, here's the byte" before the MSX asks, so readIO() never waits on the network. A /WAIT stall is exactly one emulated transfer, done with MSXCPU::wait() as you suggested, which also covers the R800.

Reset behaves like the hardware. The interface has no reset input, so the state is set up in powerUp() and an MSX reset leaves it alone.

Clearer code. The VHDL-style comments are gone, the ports are documented in plain C++ terms, members are grouped by the thread that owns them, and offerCount is explained.

Smaller details you pointed out:

sock_setNoDelay() now lives in Socket.hh/.cc.
emplace_back() is used for the offers.
The newline lost in Socket.hh is back.
The timing is now a constant instead of a setting.

About your remark on using the time emulation runs ahead: I agree it's a trade-off. It only kicks in while the MSX is polling for a byte from the Pi, and only while emulation is ahead of real time, keeping openMSX's own slack. In practice it made a big difference: a 256 KB disk copy went from 98 s to 40 s, close to the 35 s on a real Canon V-25.

One small addition

A new msxpiserver_port setting (default 5000) lets you point openMSX at a server on another port (replacing the previous implementation where it was hardcoded).

Lookgin forward for you review and comments.
Thanks
Roni

m9710797:
Thanks! I'll review, but it might take a few days.

MBilderbeek:
If you make a fix commit on a commit in a series inside one PR, I strongly prefer to fold that fix into the original commit, so the fault doesn't end up on the master tree (and may break a build when bisecting).

costarc:
> If you make a fix commit on a commit in a series inside one PR, I strongly prefer to fold that fix into the original commit, so the fault doesn't end up on the master tree (and may break a build when bisecting).

Hi @MBilderbeek.

Thanks, that makes sense: I've rewritten the branch accordingly.

The series is now three commits on top of current master, with the merge commits gone:

MSXPiDevice: emulate the MSXPi CPLD v1.6 at port level — the original /WAIT commit, the CPLD rewrite and the fixes from your review, folded into one.
MSXPi.xml: update the supported MSXPi ROMs — also drops the three timing settings the device no longer reads.
MSXPiDevice: configurable server port — with the build fix folded in, so the commit that broke the Windows-VC build never exists.

I built each of the three commits on its own (MSVC x64) to make sure a bisect can't land on a broken tree, and the final state also builds with MinGW and on Linux.

Thanks for the feedback.

m9710797:
About @MBilderbeek's remark. In general I agree, a clean commit history is nice, and it makes future bisect easier. HOWEVER, now for me it make doing the incremental review more difficult. I can't see anymore what you changed compared to my last review.

So what I did was: I went back to my review remarks, look those up in the final code, and check if they are properly addressed. And then I trust you didn't make too many other changes. I did see the change about a configurable msxpiserver_port.

So for the future (also @MBilderbeek ): a clean commit history is nice, but maybe wait with that until the review is done.

Anyway, code looks fine now. I'll merge it. Thanks!

MBilderbeek:
@m9710797 I think you can still see exactly the diff by clicking on the links under 'force-pushed', like this one: https://github.com/openMSX/openMSX/compare/a9a599ef63ef321413194d0a9847f87008fe1c28..1d4b8aeeeefe82c03dd5cf0f12eac3613d91ba99

m9710797:
@MBilderbeek unfortunately no (try that link yourself). Because the PR was rebased on last mater, it also shows all changes in all other files. But you're also right that maybe I could have used this for the changes in MSXPiDevice itself.

MBilderbeek:
> @MBilderbeek unfortunately no (try that link yourself). Because the PR was rebased on last mater, it also shows all changes in all other files. But you're also right that maybe I could have used this for the changes in MSXPiDevice itself.

Ah, right, if it also does a rebase on master (which is usually not needed, unless there are conflicts), it becomes hard to see. So, maybe a good way to handle this is: fine to force-push, but a force-push with a rebase on master must be in a separate force-push, so that it doesn't get mixed in the changes that are relevant for the PR.

costarc:
@m9710797 Thank you for all the reviews, suggestions and for merging the code.
Thanks @MBilderbeek for all costructive feedback and suggestions.

PR #2197, opened 2026-09-17
DO NOT MERGE: AI generated fixes for the SCC

seanyoung:
I cannot take credit for any of this - all AI generated.

I asked Claude to compare the SCC emulation in openMSX with https://github.com/ika-musume/IKASCC/

It generated this code. It sounds much better to me.

Not sure where to go with this next. Each commit needs careful reviewing.

I thought I'd share this in case it is useful. Feel free to close this PR.


m9710797:
Improving SCC emulation based on `IKASCC` has long been on my wish list (but never near the top of my TODO list). So thanks a lot for tackling this!

I will review these changes. But I may only have time for it next week.
If I leave review comments here, will you then pass them to Claude?

---

Some initial comments/questions without having looked at the code yet:
* How was this tested? Some (vague) idea I had was:
   *  Do a very mechanical transformation of the verilog code to some executable format (c++ or something else). The transformation should be very simple so that we're confident it's correct.
   * Create a set of test cases. I'm not sure what the best approach is here (recorded from MSX software, or create some manual test cases to trigger corner cases?)
   * Then run the test cases on both the (transformed) `IKASCC` code and openMSX. Both should produce identical output.
* OpenMSX generates SCC sound with a sample frequency of 3.57MHz / 32 = 112kHz. Does the real hardware do the same? I can imagine the digital sound output could change at any clock cycle.
  *  That would require a sample rate of 3.57MHz, but that would make the resampler step very slow (to bring the sample rate down to 44.1kHz or 48kHz).

seanyoung:
The way I've been thinking about this is to create a set of test cases and run them against an SCC on a breadboard and our SCC emulation. That way we can run the same tests against an SCC+ too, for which we have no verilog. It's also a just a better test than simulated/translated verilog: testing the actual hardware.
The nice thing is we can have tons of tests including all known SCC music.
Happy to work on this, will take some time to setup and run.

m9710797:
Interesting idea that breadboard. There are some tradeoffs:
* On the breadboard it might be hard to 100% control the initial state (e.g. the exact phase of all the internal counters). In a software emulation (I mean both openMSX-like and verilog-like) you have perfect control over the experiment.
* Running hundreds of automatic tests might be easier in software compared to capturing data in the real world (and then transfer it to software to be able to compare).
* A very important point: only measurements on real hardware can be 100% trusted. And there are some hints that the `IKASCC` schema might not be 100% correct (*1).

So maybe we can do both? Have a software model for quick (extensive) experimentation, and then also (sparse) real hardware measurements for final verification?

I started reading the commits in this PR, I did a quick first pass. From the commit messages I have the impression that Claude already did make an executable model for the verilog code? So maybe we're not too far off the software experimentation setup?

So my plan is to start reviewing in more detail. I'll leave review comments on the individual commits.

---

(*1) I did some some experiments with a (weaker) AI agent myself. The top 30 lines in `openmsx/src/sound/SCC.cc` contain the formula obtained by Jon De Schrijder. The detail that's important here is that the final sum for the sound-output (11-bits) contains a term `640` (and I've just now confirmed it with a volt-meter: silence outputs the value `640`, not `0`). I've asked AI to locate that term in the `IKASCC` schema. It hypothesized that `640 = 5 * 128`, and that we can get this my flipping the MSB bit of each of the 5 8-bit (signed) channel outputs (flipping that bit is equivalent to adding 128, and then interpreting the result as unsigned). But the AI model could absolutely not find any evidence for that bit flip, nor in the multiplier, nor the adder, nor anything after that. So maybe the AI model is wrong, or maybe the verilog code does not match the hardware (and that difference might be intentionally or not).

src/sound/SCC.cc
```diff
@@ -390,7 +390,11 @@ void SCC::setFreqVol(unsigned address, uint8_t value, EmuTime time)
 			per >>= 8;
 		}
 		period[channel] = per;
-		incr[channel] = (per <= 8) ? 0 : 32;
+		// The SCC needs 9 clock cycles to compute an output sample; when
+		// the period is shorter than that the multiplier is restarted
+		// before it ever finishes and the output latch keeps its value.
+		// A period of exactly 8 (9 cycles) still works.
+		incr[channel] = (per < 8) ? 0 : 32;
```
m9710797, line 397: Because, as you say in the commit message, this contradicts NYYRIKKI's observation, we should verify this on real hardware ourselves.

I'm leaning towards believing `IKASCC`, but we should measure to be sure. E.g. NYYRIIKI could have made a typo in his post. Or period 8 may have resulted in such a high frequency (13-14kHz ?) that it became inaudible (maybe such frequencies were attenuated too much in NYRRIKKI's setup?).

src/sound/SCC.cc
```diff
@@ -371,6 +372,37 @@ void SCC::writeWave(unsigned channel, unsigned address, uint8_t value)
...
+		// 4 bit frequency
+		per >>= 8;
+	}
+	period[channel] = per;
+	// The SCC computes each sample with a serial multiplier that needs 9
+	// master clock cycles, one more than the period+1 cycles available
+	// below period 8. Under that the multiplier is restarted before it
+	// ever finishes and the output latch keeps its old value. The wave
+	// pointer is driven by the same frequency counter and keeps running.
+	latchOutput[channel] = per >= 8;
```
m9710797, line 393: If I read the code correctly, then at all times `latchOutput[i] == orgPeriod[i] >= 8`. So `latchOutput[]` is redundant info, and it's cheap to recompute. Then maybe we don't need the `latchOutput[]` array at all?

I do understand that the same could be said about the old `incr[]` array. And I do see that later revisions in this PR still change `latchOutput[]`. But in general: when we're changing the code anyway, maybe we can eliminate some redundant (and cheap) state.

src/sound/SCC.cc
```diff
@@ -371,6 +372,37 @@ void SCC::writeWave(unsigned channel, unsigned address, uint8_t value)
...
+}
+
+// Advance the wave pointer over 'num' output samples (32 master clock cycles
+// each), without touching the output latch.
+void SCC::advancePos(unsigned channel, unsigned num)
+{
+	unsigned period2 = period[channel] + 1;
+	unsigned newCount = count[channel] + num * 32;
+	count[channel] = newCount % period2;
+	pos[channel] = (pos[channel] + newCount / period2) % 32;
```
m9710797, line 403: This comment is a more general remark (not strictly only this commit).

In this function the constant `32` appears twice:
* Once as the input-rate reduction factor (see the calculation of `INPUT_RATE` on line 116).
* Once as the length of the SCC waveform.

But these don't need to be the same. In principle we could reduce the input rate reduction (to 16, 8 or even 1) to get a tradeoff between better sound quality and computation cost (I see that later commits in this PR improve sound quality by applying a box-filter over these 32 samples). OTOH the length=32 of the SCC waveform is fixed.

Maybe it would be a good idea to introduce two different named constants for these two values `32`.

m9710797 (commented):
No remarks on commit 3 (d85c579) "SCC: apply deformation register bits 0 and 1 to the period immediately"

m9710797 (commented):
No remark on commit 4 (a13b989) "SCC: update the sound stream before changing the deformation register"

m9710797 (commented):
No remarks on commit 5 (bc42a92) "SCC: don't refresh the output on a frequency write while stalled"

src/sound/SCC.cc
```diff
@@ -507,8 +559,8 @@ void SCC::generateChannels(std::span<float*> bufs, unsigned num)
 	for (unsigned i = 0; i < 5; ++i, enable >>= 1) {
 		if (!(enable & 1) || (!volume[i] && (out[i] == 0.0f))) {
 			bufs[i] = nullptr; // channel muted
-			// Update phase counter.
-			advancePos(i, num);
+			// The wave pointer keeps running.
+			pos[i] = (pos[i] + advanceCounter(i, num * 32)) % 32;
```
m9710797, line 563: See my remark on an earlier commit. The 2 values `32` have a different meaning.
The code would be easier to understand if we would replace these with the different named constants.

src/sound/SCC.cc
```diff
@@ -397,14 +397,66 @@ void SCC::updatePeriod(unsigned channel)
...
+			clocks -= cnt + 1;
+			unsigned p = org + 1;
+			steps = 1 + clocks / p;
+			clocks %= p;
+			cnt = org;
+		}
+		cnt -= clocks;
+	}
+	counter[channel] = cnt;
+	return steps;
```
m9710797, line 459: This code looks fine to me. And it's very good to get these corner cases correct.
But do keep in mind that `deformValue != 0` is the exceptional case. So we should tune performance for `deformValue == 0`. When looking at this function in isolation, maybe we should have the common path `(deformValue & 3) == 0` first, and annotate it with `[[likely]]`? But see also my next remark.

src/sound/SCC.hh
```diff
@@ -76,6 +76,8 @@ private:
 	std::array<bool, 5> latchOutput;
 };
 
+SERIALIZE_CLASS_VERSION(SCC, 2);
```
m9710797, line 79: Very good that you thought about a version bump for the serialization format.

I did see that the later commits in this PR, bump the version again (2 times I think to version 4). In general that's the correct thing to do. But because we will merge this full PR in one go, we can bump the version just once. So this remark actually applies to those later commits.

src/sound/SCC.cc
```diff
@@ -519,25 +571,20 @@ void SCC::generateChannels(std::span<float*> bufs, unsigned num)
...
-				//       this will take more than 1 iteration
-				while (count2 >= period2) [[unlikely]] {
-					count2 -= period2;
-					pos2 = (pos2 + 1) % 32;
+				//       this steps more than once per sample
+				if (unsigned steps = advanceCounter(i, 32); steps != 0) [[unlikely]] {
+					pos2 = (pos2 + steps) % 32;
 					out2 = volAdjustedWave[i][pos2];
 				}
 			}
```
m9710797, line 586: This inner loop (lines 578-586) is (probably) critical for performance. This commit (inside the call to `advanceCounter()`) introduces an extra check (on `deformValue`) in every iteration. But that check is loop-invariant.

Maybe it makes sense to move that check before the `for` loop on line 578. In other words have a fast path for `deformValue == 0` and a slow path for the other (unlikely) case. And then to avoid code duplication we could maybe give the `advanceCounter()` function a template parameter `bool NON_ZERO_DEFORM`? And then in the function have some `if constexpr` checks. Or maybe you have a better idea for a a good tradeoff between source code duplication (undesired) and separate fast/slow execution path (desired).

m9710797 (commented):
No remarks on commit 7 (d129e9f) "SCC: an MSX reset rewinds the wave pointers and reloads the counters".

src/sound/SCC.hh
```diff
@@ -75,9 +79,10 @@ private:
 	std::array<bool, 5> rotate;
 	std::array<bool, 5> readOnly;
 	std::array<bool, 5> latchOutput;
+	std::array<int8_t, 2> waveLatch; // channel 4 and 5, see stepLatched()
 };
 
-SERIALIZE_CLASS_VERSION(SCC, 2);
+SERIALIZE_CLASS_VERSION(SCC, 3);
```
m9710797, line 85: As I mentioned before. In general this is the right thing to do.
But because we will push this PR as a whole, it's maybe better to bump the serialization version just once.

src/sound/SCC.cc
```diff
@@ -470,6 +471,89 @@ unsigned SCC::advanceCounter(unsigned channel, unsigned clocks)
...
+			b = narrow_cast<int8_t>((uint8_t(ram[pos2]) & ~mask) |
+			                        (uint8_t(latch) & mask));
+		}
+		current = adjust(b, volume[channel]);
+		s += p;
+	});
+	if (!captured) latch = ram[pos2];
+	pos[channel] = pos2;
+	return current;
+}
```
m9710797, line 541: This gets called for channel 4 and 5 from the inner loop of `generateChannels()`, so performance is important here.

I do understand that accurate emulation will be a bit slower than what we had before. But I also think (this version of) this function can be tweaked a bit. For example:
* (I think) `capture`, `s`, and `p` are only needed when `steps != 0` (and `step == 0` is the more common case I believe). So maybe those variables (and the calculation to initialize them) can be moved inside a test for `step != 0`?
* And then updating `pos[channel]` is not needed when `step == 0`.

Small stuff like that, optimizing the more frequent case.

src/sound/SCC.cc
```diff
@@ -582,7 +668,14 @@ void SCC::generateChannels(std::span<float*> bufs, unsigned num)
...
-			pos[i] = (pos[i] + advanceCounter(i, num * 32)) % 32;
+			advanceBlock(i, num);
+		} else if (isLatched(i)) {
+			auto out2 = out[i];
+			for (auto j : xrange(num)) {
+				bufs[i][j] += out2;
+				out2 = stepLatched(i, out2);
+			}
+			out[i] = out2;
 		} else {
```
m9710797, line 679: I like that you check for `isLatched(i)` outside the inner loop (don't repeat this loop-invariant check for every sample).

The structure of the code for latched (channel 4 and 5 (in certain modes)) and non-latched, looks very different: a call to `stepLatched()` vs inlined non-latched calculations. Because of this visual difference it took me a few moments to recognize the structure. It surprised me that the simpler channels 1-3 actually take more code than the more difficult channels 4-5 (of course it only looks that way and the simpler channels actually do execute simpler code).

So no problem, just confusion from my side. But maybe there is some way to make both parts more visually similar? Or maybe add 1 line of comment in each block (something like: `// channels 4-5`, `//channels 1-3`).

src/sound/SCC.cc
```diff
@@ -725,14 +734,18 @@ void SCC::Debuggable::write(unsigned address, uint8_t value, EmuTime time)
 	auto& scc = OUTER(SCC, debuggable);
 	scc.updateStream(time);
 	if (address < 0xA0) {
-		// read wave form 1..5
-		scc.writeWave(address >> 5, address, value);
+		// write wave form 1..5
+		unsigned channel = address >> 5;
+		if ((channel == 4) && (scc.currentMode == Mode::Real)) {
+			channel = 3; // one RAM for channels 4 and 5
+		}
```
m9710797, line 741: This is part of the debug-interface (cannot be triggered via a MSX program, only via actions in the debugger). I was wondering if for the debugger we should also apply this shared channel 4/5 stuff. Probably we should. And then this code change is correct.

But then shouldn't we also do that in `SCC::Debuggable::read()`?

src/sound/SCC.hh
```diff
@@ -82,7 +80,7 @@ private:
 	std::array<int8_t, 2> waveLatch; // channel 4 and 5, see stepLatched()
 };
 
-SERIALIZE_CLASS_VERSION(SCC, 3);
+SERIALIZE_CLASS_VERSION(SCC, 4);
```
m9710797, line 83: As already mentioned before, because we'll merge this PR as one block, maybe it's sufficient to bump the serialization version only once?

src/sound/SCC.cc
```diff
@@ -218,18 +218,23 @@ uint8_t SCC::readMem(uint8_t addr, EmuTime time)
 	if (((currentMode == Mode::Real) && (addr >= 0xE0)) ||
 	    ((currentMode != Mode::Real) && (0xC0 <= addr) && (addr < 0xE0))) {
 		updateStream(time);
-		setDeformReg(0xFF, time);
+		setDeformReg(0xFF);
+	}
+	// A read in 'rotation' mode returns the byte being played, so bring
+	// the wave pointers up to date first.
+	if (std::ranges::any_of(rotate, [](bool r) { return r; })) {
```
m9710797, line 225: This test with `any_of()` (a loop over `std::array<bool, 5> rotate`) made me wonder:
* Should we instead store the 5 `bool`s as 5 bits in a single `uint8_t`? Then this loop becomes a single `!= 0` test.
* In `readMem()` we need (simple) bit-manipulation instead of testing a single `bool`. But that function is not called from the performance critical `generateSound()` function.
* Initializing a single `uint8_t rotate` is simpler than initializing `std::array<bool, 5> rotate`.

And maybe the same for `std::array<bool, 5> readOnly`?

src/sound/SCC.cc
```diff
@@ -753,15 +766,19 @@ SERIALIZE_ENUM(SCC::Mode, chipModeInfo);
...
 	             "period",      orgPeriod,
 	             "volume",      volume,
-	             "ch_enable",   ch_enable,
-	             "deformTimer", deformTimer,
-	             "deform",      deformValue);
+	             "ch_enable",   ch_enable);
+	if (ar.versionBelow(version, 4)) {
+		Clock<CLOCK_FREQ> deformTimer(EmuTime::zero()); // no longer used
+		ar.serialize("deformTimer", deformTimer);
+	}
```
m9710797, line 780: There's no need to load fields that are no longer needed.

More in detail:
* Saving is always done using the latest version (so we'll never save the `deformTimer` anymore).
* For loading (not saving) on-disk savestates we can encounter an older savestate version. But on-disk savestates are in XML format, and for those it's fine to skip loading certain fields.
* For the in-memory savestates we use a binary format and there we can't easily skip loading fields. But for in-memory savestate the version is always the latest version (so we don't save and then also don't load `deformTimer`).

So I think lines 777-780 can just be dropped. However we do still need the version bump. Because older openMSX executables will still try to load the `deformTimer` field. Instead we want to give an error message that the savestate version is too new (the serialization framework itself already takes care of that).

But then see also my remark about only bumping the version number once for this PR. We should keep the comment about "removed 'deformTimer'".

src/sound/SCC.cc
```diff
@@ -515,8 +515,9 @@ bool SCC::isLatched(unsigned channel) const
 
 // One 32-cycle sample of a latched channel: advance the counter, the wave
 // pointer and the latch. Returns the output at the end of the sample, or
-// 'current' when there was no step.
-float SCC::stepLatched(unsigned channel, float current)
+// 'current' when there was no step, and in 'average' the mean output over
+// the sample (see generateChannels()).
+float SCC::stepLatched(unsigned channel, float current, float& average)
```
m9710797, line 520: I don't like output parameters (like `average`). Instead return multiple values, e.g. via a small `struct`:
```
struct StepLatchedResult {
    float current;
    float average;
};
StepLatchedResult SCC::stepLatched(...) {
    ...
    return {.current = ..., .average = ...};
}
...
auto [current, average] = stepLatched(...);
```
Often this makes the code easier to understand (replaces reference-semantics with value-semantics). 

But also for the compiler: with a `float` reference (or pointer) parameter, the compiler has to make worst-case assumptions that this parameter may alias any other `float` variable. Meaning any write to a `float` variable may change the value of `average`. Or a write to `average` may change the value of any other `float` variable. When you use the return path, this aliasing problem disappears.

src/sound/SCC.cc
```diff
@@ -681,21 +695,33 @@ void SCC::generateChannels(std::span<float*> bufs, unsigned num)
...
+				unsigned last = 0;
+				repeat(steps, [&] {
+					acc += out2 * float(s - last);
+					last = s;
+					pos2 = (pos2 + 1) % 32;
+					out2 = volAdjustedWave[i][pos2];
+					s += p;
+				});
+				acc += out2 * float(32 - last);
+				bufs[i][j] += acc * (1.0f / 32.0f);
```
m9710797, line 724: I haven't fully worked this out yet, but here's an idea we can explore, the goal is to improve performance on the hot code path (`generateChannels()`).

I'm assuming we've already split this path for the cases `deformValue == 0` (very likely) and `!= 0` (unlikely). I'm only talking about tweaking that fast path.

In `advanceCounter()` we calculate the number of steps. This has a fast path for no steps, and a division on the slow path. Then here we replicate the fast (`steps == 0`) and slow path. And then in the slow path loop over the number of steps. What if we inline `remaining()` and `advanceCounter()` in here (*1). And then in the slow path, instead of calculating the number of steps (via a division) outside the loop, change the loop to iterate until `s > 32` (and then simplify all the bookkeeping).

m9710797 (commented):
In the PR description you say: "It sounds much better to me".
I assume that's mostly because of this commit?

m9710797 (commented):
No remarks on commit 11 (71c93ef): "MSXMixer: don't drop the pending stream when the mixer parameters change".

src/sound/SCC.cc
```diff
@@ -513,104 +523,250 @@ bool SCC::isLatched(unsigned channel) const
...
+				stealUntil[channel] = e.cycle + ACCESS_CYCLES;
+				break;
+			case Event::RELOAD:
+				doReload = true;
+				restart  = e.data & 1;
+				break;
+			default:
+				UNREACHABLE;
+			}
+			ev.erase(ev.begin());
```
m9710797, line 669: I don't like this. Erasing the first element of a `std::vector` in a loop becomes `O(n^2)`.
Instead we can maybe replace `std::vector<Event>` with our `CircularBuffer<Event, 64>` class. (Because also other parts of the code never put more than 64 elements `events`). And with `CircularBuffer` you have `O(1)` push/pop front/back.

src/sound/SCC.cc
```diff
@@ -513,104 +523,250 @@ bool SCC::isLatched(unsigned channel) const
...
-	const auto& ram = wave[channel];
-	unsigned pos2 = pos[channel];
-	bool captured = false;
+// The cycle of the stream at which a CPU access at 'time' lands. The stream
+// has been generated up to some sample at or before 'time' (updateStream()
+// is called first); the access is that many cycles into what follows.
+uint64_t SCC::accessCycle(EmuTime time)
+{
+	auto [n, frac] = getEmuClock().getTicksTillAsIntFloat(time);
+	return streamCycle + 32 * uint64_t(n) + unsigned(32.0f * frac) + 1;
```
m9710797, line 532: I understand that with the current available infrastructure this is the best we can do. But:
* We calculate the distance as an integral and fractional part.
* Multiply both parts by 32, truncating the (float) fractional part to an integer.
* Then return the sum.

So basically what we wanted to know was: "what's the distance measured using a clock that ticks 32 times faster".

I don't know how yet, but it seems to me that it should be possible to calculate this in a more direct way (though not by constructing a new `DynamicClock` because initializing the internal `DivModBySame` object is expensive). ... Maybe via:
* Query `getEmuClock().getPeriod()` (that should be cheap? please check)
* Divide that by 32.
* Then use `operator/` or `div()` or `divUp()` on two `EmuDuration` objects.

Hmm, but is this cheaper/simpler than the current implementation?

src/sound/SCC.cc
```diff
@@ -209,6 +223,28 @@ void SCC::setMode(Mode newMode)
 	currentMode = newMode;
 }
 
+// Which channel's wave RAM window an address falls in, for reading or for
+// writing, or -1.
+int SCC::waveChannel(uint8_t address, bool write) const
```
m9710797, line 228: Instead of using `-1` as a special value, we could return `std::optional<unsigned>`.
That might make the callers easier (I haven't checked in detail). And this function isn't used in any performance critical code path.

src/sound/SCC.cc
```diff
@@ -220,61 +256,52 @@ uint8_t SCC::readMem(uint8_t addr, EmuTime time)
...
 	}
-	// A read in 'rotation' mode returns the byte being played, so bring
-	// the wave pointers up to date first.
-	if (std::ranges::any_of(rotate, [](bool r) { return r; })) {
+	if (int channel = waveChannel(addr, false); channel >= 0) {
+		// The access takes the RAM's address away from the channel for
+		// the bus cycle, see runSample(). And in rotation mode the byte
+		// returned depends on the wave pointer, so bring it up to date.
 		updateStream(time);
+		queueEvent(unsigned(channel), {accessCycle(time), Event::READ, uint8_t(addr & 0x1F), 0});
```
m9710797, line 264: Do we have any idea if `SCC+` (thus `Mode::Compatible` and `Mode::Plus`) also suffer from this RAM-shared-with-CPU-problem? I'm not sure at all, but I seem to remember that this "crackling" problem didn't occur on `SCC+`, is that correct???

If so, then maybe for `SCC+` we shouldn't do this expensive(?) event queuing stuff? OTOH more uniform code (fewer special cases) is also a nice property.

src/sound/SCC.cc
```diff
@@ -513,104 +523,250 @@ bool SCC::isLatched(unsigned channel) const
...
+	return streamCycle + 32 * uint64_t(n) + unsigned(32.0f * frac) + 1;
+}
+
+void SCC::queueEvent(unsigned channel, const Event& event)
+{
+	auto& ev = events[channel];
+	// Accesses arrive in order of time and the stream is never more than
+	// a few samples behind, so this stays short. Should it not, apply the
+	// oldest right away rather than lose it.
+	if (ev.size() >= 64) {
```
m9710797, line 541: Can you explain how you got to this value `64` for the maximum queue size?

I assume it depends on the settings for `samples` and `frequency`? And then how the CPU is accessing the SCC registers. Let's assume the absolute worst case:
* `samples=2048`, `frequency=44100`, that gives a buffer of `46ms`.
* That's 116k clock cycles @ 3.57MHz.
* Assuming the Z80 is writing to the waveform registers using an unrolled sequence of `LDI` instructions, that's 18 cycles per instruction.
* So together there could be 9k events queued between two invocations of `generateChannels()`.

(And we could even set `samples=4096`, `frequency=22050`, overclock the Z80, or use R800). So in theory it's possibly to have more than 64 queued events. In practice this will almost never happen. Maybe worst case the CPU does a `LDIR` to all 5 x 32 waveform registers?

Hmm, as a solution I was going to propose to do a `updateStream()` if we detect the queue is full (more than 64 elements). But then if I check the code: `queueEvent()` is called from 3 locations, and for all these we do already call `updateStream()` before calling `queueEvent()`.

Do I understand correctly that after a call to `updateStream()` the queue should be drained? If that's the case (and we always call `updateStream()` before `queueEvent()). Doesn't that mean a queue-size of 1 is all we need? And if that's the case then a lot of the code can be simplified I think. Did I miss something?

src/sound/SCC.cc
```diff
@@ -513,104 +523,250 @@ bool SCC::isLatched(unsigned channel) const
...
+		applyEvent(channel, ev.front());
+		ev.erase(ev.begin());
+	}
+	ev.push_back(event);
+}
+
+// Apply an event without regard for its cycle.
+void SCC::applyEvent(unsigned channel, const Event& event)
+{
+	switch (event.kind) {
```
m9710797, line 551: This switch is similar to the one on line 652 (inside `runSample()`). The difference is this version only gets called when the channel is silent (from `skipBlock()`) while the other is the more general one. Is that correct?

Even though both are semantically similar, the implementation is very different. And it wasn't obvious to me if this version is correct or not. E.g. why don't we need to update `stealByte[]` and `steelUntil[]` here? (I think it's because `skipBlock()` unconditionally sets `steelUntil = 0` (and in that case `steelByte` doesn't matter). But this took me quite a while to figure out. (And did I even understand correctly?)

So maybe we should document this better. Would it be a good idea to add a `bool SILENT` template parameter to this function. So that we only have a single function (in source code) anymore that handles events. Then in the implementation we can have `if constexpr` to deal with the differences. And then we also have a good location in the source code to explain those differences (e.g. why updating certain things isn't needed).

src/sound/SCC.cc
```diff
@@ -513,104 +523,250 @@ bool SCC::isLatched(unsigned channel) const
...
+	case Event::READ:
+		break;
+	case Event::WRITE:
+		writeWave(channel, event.address, event.data);
+		break;
+	case Event::RELOAD:
+		reload(channel, event.data & 1);
```
m9710797, line 558: When queuing a `RELOAD` event we already mask data to a single bit. No need to mask again here (and same on line 664).

src/sound/SCC.cc
```diff
@@ -513,104 +523,250 @@ bool SCC::isLatched(unsigned channel) const
...
+	multiplier[channel] = {.active = true, .bit = 0, .byte = 0, .vol = volume[channel]};
+}
+
+// One sample of a channel for the samples in which something happens,
+// exact to the cycle: the frequency counter, the serial multiplier reading
+// its 8 bits and writing the output latch, the channel 4/5 capture, and
+// whatever the CPU does to the RAM port meanwhile. 'base' is the stream
+// cycle before the sample's first edge. Returns the mean output over the
+// sample, and leaves the output latch in out[].
+float SCC::runSample(unsigned channel, uint64_t base)
```
m9710797, line 581: I haven't reviewed this function in detail yet (I'm running out of time for today, I'll continue later).

I believe this function might be important for performance, it's called for every sample in the inner loop. It does perform many checks, but several of these checks are loop invariant. So maybe we can explore giving this function some template parameters:
* `bool DEFORM_VALUE_ZERO`: deformValue == 0
* `bool SHARED_CH45`: isLateched(channel)
* `bool PENDING_EVENT`: there is some pending event at the start of the loop, during the loop it may get handled, that's fine (no need to exit this loop and enter another more efficient loop).

Then outside the loop over all samples, we test these conditions just once, and then inside the loop (actually there multiple loops now) call the appropriate specialized version of this function.

m9710797 (changes requested):
I ran out of time, this last commit is only partially reviewed. But I already wanted to submit my review now. If you want you can already start processing the review comments.

I must say, also for all the previous commits, I mostly trusted the implementation was correct, or at least the reverse engineering was correct. I did always check that the comments or commit message matched with what I saw in the implementation.

In the long term it would be nice if we can better verify these finding against measurements on real hardware. But all seems plausible to me. So we don't necessarily have to wait with merging this PR until all verification is done.

But I would like to see my review comments addressed before we merge. Some of my comments are only questions, or some might be plain wrong. So please discuss, don't just change because I made a remark.

ika-musume:
IKASCC author here. The signed output is an intentional normalization for SoC integration. The missing `640` is explained by the difference between signed audio samples and the original SCC's DAC-oriented representation.

1. **How the original SCC represents and sums channel outputs**

   Page 10 of the official SCC datasheet, which is not publicly available, specifies that waveform samples are 8-bit signed values. For the R2R DAC's internal circuit diagram, see [this tweet](https://x.com/v9938/status/1476135157820432385).

   After volume multiplication, the original circuit converts each channel's signed result to offset binary by inverting its MSB. This is visible in [the reconstructed schematic](https://github.com/ika-musume/IKASCC/blob/main/docs/K051649.pdf): on page 2, channel 1's output latch uses `Q` for bits 0–6, but **`Q̅` for bit 7** at B6-F1. Channels 2–3 repeat this design, and channels 4–5 have the same MSB inversion.

   Interpreted as unsigned, this transformed byte equals `signed_sample + 128`. The summer adds these unsigned channel values, giving:

   ```text
   sum(channel_sample + 128)
       = sum(channel_sample) + 5 × 128
       = sum(channel_sample) + 640
   ```

   Thus, the summer's silence value is `640`. The bias comes from the channel output representation, not from the external R2R resistor network.

2. **Why IKASCC outputs signed audio**

   IKASCC deliberately retains signed channel results and sums them into a signed 11-bit output. This makes the core easier to integrate into an SoC's digital audio mixer without requiring the surrounding system to remove a DAC-specific bias.

   The original hardware also has an AC-coupling capacitor after the DAC, which removes the steady DC component from the audio path. Reproducing that DC bias is therefore unnecessary for the intended signed audio interface. This is a representation choice, rather than an attempt to reproduce the physical DAC-driving bits directly.

3. **What this means for the suspected missing offset**

   The hypothesis `640 = 5 × 128` was correct. The evidence for the MSB inversion is the selection of the flip-flop’s complementary output, so there is no separate NOT gate to find.

   IKASCC's Verilog omits this conversion intentionally. Comparing its signed output directly with the original circuit's biased sum will therefore show a constant difference of `640`. This particular difference does not indicate an error in the waveform or volume calculation.

4. **How to reproduce the original summer's representation in Verilog**

   Invert **each channel’s MSB before unsigned addition**, explicitly extending each operand to 11 bits:

   ```verilog
   wire [10:0] dac_sum =
       {3'b000, ~ch1_sound[7], ch1_sound[6:0]} +
       {3'b000, ~ch2_sound[7], ch2_sound[6:0]} +
       {3'b000, ~ch3_sound[7], ch3_sound[6:0]} +
       {3'b000, ~ch4_sound[7], ch4_sound[6:0]} +
       {3'b000, ~ch5_sound[7], ch5_sound[6:0]};
   ```

   Register this result with the appropriate output timing. Numerically, it is equivalent to adding `640` to the signed sum.

   Simply casting the unchanged channel bytes to unsigned is **not** equivalent, nor is inverting only the final sum's MSB. For exact physical pin compatibility, the subsequent output-latch and pad-driver polarities must also be reproduced; `dac_sum` describes the internal summer value.

5. **Impact on the current C++ emulation**

   If the C++ audio path already sums signed, volume-scaled channel samples, it does not need an additional `640` for this reason. Adding it would introduce a DC bias into that audio representation.

   For validation, compare like with like: compare the C++ signed output directly against IKASCC's signed output, or subtract `640` from the original circuit’s biased sum before comparing. If testing physical output pins or analog voltages, account for the output polarity and DAC conversion as well.

   This explains the offset discrepancy specifically; timing, truncation, RAM-access interactions, and other hardware behavior still need their own verification. I've recently been preparing a "hat" for an FPGA dev board to analyze SH3 and Ensoniq sound chips, as well as a longstanding unresolved issue in IKASCC: the access timing of the SCC's internal RAM. This timing is important for accurately emulating the characteristic noise heard when the Z80 accesses that RAM while the SCC channels are playing.

**If anyone involved is interested in the internal architecture of the SCC+ and would like an accurate Verilog model and analysis, you are welcome to send me a chip. I would be happy to reverse engineer it free of charge and publish the results.**

m9710797:
@ika-musume Thanks for the clarification.

I didn't mean to imply there was a mistake in the verilog code. I only wanted to point out a difference compared to real hardware (for when later someone tries to compare with real hardware).

And indeed dealing with signed values is much easier when integrating into a larger structure. OpenMSX also fully omits this 640 bias.

m9710797:
Hi,

I finished the review (2nd part of last commit), no additional remarks. I'm still assuming the reverse engineering is all correct. I only checked that the description in the commit messages matches the implementation in the PR.

The most difficult part of the review was verifying that all code paths are consistent. I mean for example the `skipBlock()` function or the "events-is-empty" path in `sample()`. But I think it's all correct. And I do believe those optimizations are important. More on optimizations in the 2nd part of my comment.

@seanyoung How do you propose we continue? Can you (or Claude) process my review comments?

---

More about optimizations (in the full `generateChannels()` code path).

The SCC runs at 3.57MHz, and the digital sound output can change at any clock. So in principle we should emulate a sound stream with a sample rate of 3.57MHz. Emulating clock-per-clock would (probably) also result in the simplest c++ emulation code. Unfortunately that will also be slow. Instead we emulate at a sample rate that's 32x lower (112kHz). 3/57MHz would be slow both because the emulation loop has to iterate 32x more times. But also because resampling the audio signal to 44.1kHz or 48kHz is much more expensive if we start from 3.57MHz than from 112kHz.

A second optimization is emulating "silence". In openMSX we have a few "Boosted" machine configurations (and these are quite popular). They do contain a SCC sound chip, but most of the time the user won't be running any SCC code. And then it's important that emulating an idle SCC chip doesn't take too much time.

Other optimizations are:
* The deform-register. This is actually a kind of test register, but some software "abuses" it to get various effects. I do think accurate emulation is a must. However using this register is the exception, most software leaves this register at zero. So it's a bonus if we can exploit that and gain some emulation speed for the common case.
* Similar for the case of "write-to-waveform-during-playback". This results in some "crackling" noise. I do agree we should emulate this accurately (now that we know how to do that). But because it causes noise, most real software tends to avoid this scenario, and then maybe we could also optimize for this more common case (also because it requires emulating the multipliers cycle-per-cycle, that's expensive).

What I'd propose is to merge the PR (close to) how it is now.
* Do address my review remarks.
* But focus less on my remarks about optimizations.

Then once it's merged we can focus on optimizations:
* I'd propose then to maybe first undo all optimizations.
* And then re-apply them, but maybe in such a way that it's easier to see that the optimized code path is fully equivalent to the previous unoptimized path.
* And my hope is that we can do that without too much source code duplication. My current idea for that is to use c++ templates to create specialized code paths from a single source code function.

@seanyoung What do you propose?

PR #2199, opened 2026-09-21
Fix ImGui Cheat Finder: restore lazy loading of _cheat.tcl

sndpl:
Commit e2cde0cb0 removed 'start' and 'search' from the _cheat.tcl lazy registration, because those bare names were announced as global commands (tab-completion, syntax highlighting) while no such commands exist. But they were also the trigger that made lazy_handler() source _cheat.tcl: the ImGui Cheat Finder calls cheat_finder::start and cheat_finder::search, and lazy_handler() matches on [namespace tail], so those calls resolved via 'start'/'search'.

Without them the script is never loaded, both calls end up in 'unknown' and throw "invalid command name". ImGuiManager::execute() silently ignores CommandException, so pressing "Start a new search" (or any of the compare buttons) does nothing at all. Broken since e2cde0cb0, i.e. after the 21.0 release.

Register the fully qualified names instead, and let lazy_handler() match on those as well as on the bare proc name. This keeps _cheat.tcl loading on demand without announcing commands that don't exist.


MBilderbeek:
Oops, nice find. @m9710797 what do you think about this solution?

m9710797:
Merged.
I haven't tested this, i trust you did, and it works.
But I did read the code, and the approach looks fine to me.

PR #2200, opened 2026-09-22
Fix pkg-config 0.29.2 glib compilation with C23 bool keyword

meesokim:
Fix compilation error in glib/glib/goption.c when building internal pkg-config 0.29.2 under C23 (GCC 15) standard where 'bool' is a reserved keyword.


MBilderbeek:
@meesokim thanks, but isn't this patch working as well: 8b59053d1294d7c623c108d63e784f749b95d215 ?

